PR #147, opened 2014-12-25
	fix haselcast plugin issue

wmz7year:
issue info https://community.igniterealtime.org/thread/54122

	There are diffent classloader load haselcast classes will have this problems


tevans:
We have not been able to reproduce this issue in our cluster plugin testing. Closing the PR for now, pending further research.

PR #151, opened 2015-01-09
OF-857: Synchronize I/O for C2S connection

tevans:
Use a lock to ensure the integrity of IoSession (MINA), especially when
using compression or TLS filters

### DIFF
--- a/src/java/org/jivesoftware/openfire/nio/NIOConnection.java
+++ b/src/java/org/jivesoftware/openfire/nio/NIOConnection.java
@@ -19,9 +19,9 @@
 
 package org.jivesoftware.openfire.nio;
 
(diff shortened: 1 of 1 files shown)


Flowdalic:
Another thought: If we need to synchronize the `ioSession.write()`, then how did we previously guarantee the stability of the stanza order?

tevans:
Based on [this issue] (https://issues.apache.org/jira/browse/DIRMINA-653), I believe the problem is actually in the MINA library - async handling for longer-running I/O filter chains (e.g. with compression, TLS, etc) appears to be buggy. The recommended approach (in that post) was to use a Semaphore to lock the IoSession during I/O operations. My goal with the PR was to provide a testable quick fix, pending additional research into the MINA codebase.

Regardless, I believe a ReentrantLock would accomplish essentially the same result if that is preferable.

PR #152, opened 2015-01-09
OF-855: Ensure valid connection for session packets

tevans:
Check the underlying connection for a given session before delivering a
packet; if the connection is missing, clean up session and resend packet


tevans:
The isClosed check was redundant (also checked in the conn.deliver() method). Based on the submitted bug report, this was creating a trap door that would allow packets to be silently dropped in certain exception scenarios.

PR #153, opened 2015-01-10
OF-863: Fix multiple NPEs in cluster mode

tevans:
This patch fixes several NullPointerExceptions that were observed in a
cluster when running under high load and latency, including related
issues in the BOSH connector and hazelcast plugin. Some related logging
improvements are also included.

Some of the NPEs were generated during the startup phase before the
server was fully initialized. These have been corrected by checking
appropriate "ready state" indicators.

In general, sessions hosted in remote cluster nodes that cannot be
addressed due to timeouts during remote task execution should be
considered temporarily "unavailable" to the cluster member that
initiated the call.

### DIFF
--- a/src/java/org/jivesoftware/openfire/IQRouter.java
+++ b/src/java/org/jivesoftware/openfire/IQRouter.java
@@ -97,6 +97,7 @@ public class IQRouter extends BasicModule {
         }
         JID sender = packet.getFrom();
         ClientSession session = sessionManager.getSession(sender);
+        Element childElement = packet.getChildElement(); // may be null
         try {
             // Invoke the interceptors before we process the read packet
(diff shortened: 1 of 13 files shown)


tevans:
Per the cluster SPI change, our testing showed that under high load the cluster members may time out for synchronous task operations, leaving the corresponding remote session proxy in an uninitialized state. Merely changing from NPE to ISE does not actually fix this problem, but it does allow a more consistent treatment throughout the codebase as we begin applying the required exception handling logic.

Specifically as it relates to RemoteSession, the clustering code needs to be updated anyway to better utilize caching rather than remote task execution to access a session's properties. More to come on this issue ... stay tuned.

PR #155, opened 2015-01-11
OF-855/OF-857 improvements

Flowdalic:
I'm pretty sure that 4158dd77e4776fa263373e7cb83467ebd5cd3e6d is sound. 7bbfb9dbc49600ea8ffd02129909c152a2f73a72 should be too since I doubt that there is a case where Connection in the constructor is null. But since I don't longer have a Openfire instance to test I can't confirm it.


tevans:
I'm ok with these recommendations, agree it's better to ensure a valid connection rather than trying to chase various exception conditions around elsewhere. Note that the real issue seems to be related to MINA's async handling for long-running I/O filters. More to come on that ... under separate cover.

Per Flow's comment, we will need to run this through some regression cycles before releasing.

tevans:
I've got additional changes to make here, so will merge and keep moving forward.

PR #156, opened 2015-01-12
OF-857: Synchronize connection/session cleanup

tevans:
This patch is to prevent a stack overflow in certain exception cases
where a connection has been closed, but its corresponding session is
still present in the routing table. The patch synchronizes the internal
connection close logic before handing the connection back to mina to 
close the IoSession asynchronously.

### DIFF
--- a/src/java/org/jivesoftware/openfire/SessionManager.java
+++ b/src/java/org/jivesoftware/openfire/SessionManager.java
@@ -641,7 +641,7 @@ public class SessionManager extends BasicModule implements ClusterEventListener 
         JID searchJID = new JID(originatingResource.getNode(), originatingResource.getDomain(), null);
         List<JID> addresses = routingTable.getRoutes(searchJID, null);
         for (JID address : addresses) {
-        	if (originatingResource != address) {
+        	if (!originatingResource.equals(address)) {
 	            // Send the presence of the session whose presence has changed to
 	            // this user's other session(s)
 	            presence.setTo(address);
--- a/src/java/org/jivesoftware/openfire/SessionManager.java
+++ b/src/java/org/jivesoftware/openfire/SessionManager.java
@@ -641,10 +641,12 @@ public class SessionManager extends BasicModule implements ClusterEventListener 
         JID searchJID = new JID(originatingResource.getNode(), originatingResource.getDomain(), null);
         List<JID> addresses = routingTable.getRoutes(searchJID, null);
         for (JID address : addresses) {
-            // Send the presence of the session whose presence has changed to
-            // this other user's session
-            presence.setTo(address);
-            routingTable.routePacket(address, presence, false);
+        	if (originatingResource != address) {
+	            // Send the presence of the session whose presence has changed to
+	            // this user's other session(s)
+	            presence.setTo(address);
+	            routingTable.routePacket(address, presence, false);
+        	}
         }
     }
 
--- a/src/java/org/jivesoftware/openfire/nio/NIOConnection.java
+++ b/src/java/org/jivesoftware/openfire/nio/NIOConnection.java
@@ -255,7 +255,15 @@ public class NIOConnection implements Connection {
 
     public void deliver(Packet packet) throws UnauthorizedException {
         if (isClosed()) {
-            backupDeliverer.deliver(packet);
+        	// OF-857: Do not allow the backup deliverer to recurse
+        	if (backupDeliverer == null) {
+        		Log.error("Failed to deliver packet: " + packet.toXML());
+        		throw new IllegalStateException("Connection closed");
+        	}
+        	// attempt to deliver via backup only once
+        	PacketDeliverer backup = backupDeliverer;
+            backupDeliverer = null;
+            backup.deliver(packet);
         }
         else {
             boolean errorDelivering = false;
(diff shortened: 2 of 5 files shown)


tevans:
FYI - looks like this still needs some more work. Stay tuned ... please don't merge this PR (yet).

tevans:
Ok, I was able to reproduce (and fix) the stack overflow locally. We now have a backup deliverer that will attempt to write message packets into the offline store, and will drop IQ and Presence packets.

This PR is now ready for review / merge.

src/java/org/jivesoftware/openfire/nio/NIOConnection.java
```diff
@@ -261,7 +255,15 @@ public boolean isSecure() {
...
     public void deliver(Packet packet) throws UnauthorizedException {
         if (isClosed()) {
-            backupDeliverer.deliver(packet);
+        	// OF-857: Do not allow the backup deliverer to recurse
+        	if (backupDeliverer == null) {
+        		Log.error("Failed to deliver packet: " + packet.toXML());
+        		throw new IllegalStateException("Connection closed");
+        	}
+        	// attempt to deliver via backup only once
+        	PacketDeliverer backup = backupDeliverer;
```
tevans: Note: I added this "belt and suspenders" code to prevent recursion because the ComponentConnectionHandler still uses the standard PacketDeliverer as a backup.

src/java/org/jivesoftware/openfire/SessionManager.java
```diff
@@ -641,10 +641,12 @@ public void broadcastPresenceToOtherResources(JID originatingResource, Presence
         JID searchJID = new JID(originatingResource.getNode(), originatingResource.getDomain(), null);
         List<JID> addresses = routingTable.getRoutes(searchJID, null);
         for (JID address : addresses) {
-            // Send the presence of the session whose presence has changed to
-            // this other user's session
-            presence.setTo(address);
-            routingTable.routePacket(address, presence, false);
+        	if (originatingResource != address) {
```
wmz7year: Here you should use equals?

src/java/org/jivesoftware/openfire/nio/OfflinePacketDeliverer.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public OfflinePacketDeliverer() {
+        this.messageStrategy = XMPPServer.getInstance().getOfflineMessageStrategy();
+    }
+
+    @Override
+	public void deliver(Packet packet) throws UnauthorizedException, PacketException {
+    	
+        if (packet instanceof Message) {
+            messageStrategy.storeOffline((Message) packet);
```
Flowdalic: Shouldn't we also take carbons into consideration? I think we should check here if it's an carbon and *not* store the message then.

src/java/org/jivesoftware/openfire/SessionManager.java
dwd: "If the current address is not that of the originator, send the presence." - this matches the comment, in as much as it's sending the presence of the session to this user's *other* sessions. The interesting question is whether the presence is sent to the session that generated it somewhere else, since (I think) it's meant to be.

src/java/org/jivesoftware/openfire/nio/OfflinePacketDeliverer.java
dwd: You're right in that it needs checking for; but I think we can best check in OfflineMessageStrategy.storeOffline can't we?

src/java/org/jivesoftware/openfire/SessionManager.java
Flowdalic: I guess @wmz7year meant that `!equals()` should be used instead of a comparison of the object references.

src/java/org/jivesoftware/openfire/nio/OfflinePacketDeliverer.java
Flowdalic: Good point, checking if the message is a carbon should go into https://github.com/igniterealtime/Openfire/blob/877c20eb2e3b54a702307190ac2bb99bc30c3013/src/java/org/jivesoftware/openfire/OfflineMessageStrategy.java#L82

src/java/org/jivesoftware/openfire/SessionManager.java
dwd: Oh, see, this is why I shouldn't be allowed to get near Java. :-)

wmz7year: @Flowdalic Year, That's my mean.

tevans: Yes, good catch. I will update the PR for this.

src/java/org/jivesoftware/openfire/nio/OfflinePacketDeliverer.java
tevans: Makes sense ... I will add.

src/java/org/jivesoftware/openfire/nio/NIOConnection.java
wmz7year: I'm not sure if throws IllegalStateException who catch it? MINA?

tevans:
FYI - I believe this one is ready to go now. Further comments/concerns?

Flowdalic:
If I'm not mistaken, `shouldStoreMessage()` is only invoked when the the offline message strategy type is 'store'. If type is e.g. 'bounce' then we would send a error message response back. And while I don't have the spec in front of me, I don't think that should be the case for carbons. So IMHO it makes sense to have two phases to decide what to store.

PR #157, opened 2015-01-13
Fix non working STUN server address in JingleNodesPlugin.java

J0s3f:
Multiple STUN servers are tried, in case one fails.

### DIFF
--- a/src/plugins/jingleNodes/src/java/org/jinglenodes/JingleNodesPlugin.java
+++ b/src/plugins/jingleNodes/src/java/org/jinglenodes/JingleNodesPlugin.java
@@ -40,6 +40,9 @@ import org.xmpp.jnodes.nio.LocalIPResolver;
 import org.xmpp.jnodes.nio.PublicIPResolver;
 
 public class JingleNodesPlugin implements Plugin {
+	
+	private static final String[] 	stun_host = 	{"stun.l.google.com",	"stun.schlund.de",	"stun.iptel.org ",	"stun4.l.google.com"};
+	private static final int[] 		stun_port =		{19302,					3478,				3478,				19302};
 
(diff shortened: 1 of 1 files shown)


src/plugins/jingleNodes/src/java/org/jinglenodes/JingleNodesPlugin.java
```diff
@@ -40,6 +40,9 @@
 import org.xmpp.jnodes.nio.PublicIPResolver;
 
 public class JingleNodesPlugin implements Plugin {
+	
+	private static final String[] 	stun_host = 	{"stun.l.google.com",	"stun.schlund.de",	"stun.iptel.org ",	"stun4.l.google.com"};
```
tevans: I would prefer to see configuration properties for things like public network addresses.

J0s3f: I think there should be a method in the core to get the external IP address and information about the reachability. Plugins could then query this information. 
Sadly,  I am not familiar enough with Openfire to implement it easily. 
I just needed a fix for the Jingle plugin.

guusdk:
Using org.jivesoftware.util.JiveGlobals#getProperties it is easy to get a list of property values. I suggest you use that to replace the hard coded properties. To simplify things, I suggest you put one host and one port in one property, separated by a colon: `example.org:1234`

guusdk:
https://github.com/igniterealtime/Openfire/pull/161 is pretty similar. Perhaps efforts can be combined?

deleolajide:
This has been covered by https://github.com/igniterealtime/Openfire/commit/9cc37be2783a213995e54319f176d9b5be2afd7e

PR #161, opened 2015-01-19
change the default STUN server and make the UDP port range configurable

linuschien:
1. stun.xten.com works no more, the STUN server should be configurable.
2. The UDP port range should be configurable because of the enterprise firewall policy.
3. The result date should be the startDate for the remain cases.
4. https://igniterealtime.org/issues/browse/OF-856

### DIFF
--- a/src/plugins/jingleNodes/src/java/org/jinglenodes/JingleNodesComponent.java
+++ b/src/plugins/jingleNodes/src/java/org/jinglenodes/JingleNodesComponent.java
@@ -37,104 +37,104 @@ import org.xmpp.packet.IQ;
 import org.xmpp.packet.PacketError;
 
 class JingleNodesComponent extends AbstractComponent {
-	private static final Logger Log = LoggerFactory.getLogger(JingleNodesComponent.class);
+    private static final Logger Log = LoggerFactory.getLogger(JingleNodesComponent.class);
 
-	private static final String UDP = "udp";
-	private static final String PROTOCOL = "protocol";
-	private static final String HOST = "host";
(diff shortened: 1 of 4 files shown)


guusdk:
The commits in this pull request appear to have applied an automatic formatter. As a result, pretty much the entire plugin has been modified. This obscures your changes. Can you please rectify this?

Also, are you aware that a similar change is proposed in https://github.com/igniterealtime/Openfire/pull/157 ?

linuschien:
1. Sorry about auto-format. After I fixed, it could diff more clearly.
2. Sorry, I don't think hard code STUN server address and port would be a good idea. Giving a default value and open configuration would be mush better.

guusdk:
(I never argued we should have hardcoded addresses?)

deleolajide:
I think this discussion may have been overtaken by events. I just need to check and confirm. I recently accepted a PR for jinglenodes plugin that has implemented this and other things. I also replaced the outdated library jar file with the most current source code and tested it ok from the openfire meetings plugin.

guusdk:
Oh, that'd be good. That same merge would then also obsolete #157, I think? it would be great if you can verify and close both pull requests!

deleolajide:
The pull request is https://github.com/igniterealtime/Openfire/commit/9cc37be2783a213995e54319f176d9b5be2afd7e

and only deals with items 1 & 2 of this pull request. I think items 3 and 4 are for OF-856 and are probably outstanding

guusdk:
I have cherry-picked and merged @linuschien fixes for 3 and 4, and created a new PR https://github.com/igniterealtime/Openfire/pull/458

PR #164, opened 2015-01-28
OF-626: Incorporate review feedback

tevans:
- cleanup lib versions to remove obsolete MINA notes
- use C2S compression policy rather than S2S compression policy for BOSH
connector


tevans:
It would be good to merge this change in prior to releasing 3.10.0. The versions.txt is just documentation; the other change ensures that we use the C2S compression setting for BOSH rather than using the S2S compression setting.

PR #168, opened 2015-02-02
OF-873 Openfire MUST return an empty query if no items are found (disco#...

sco0ter:
...items) instead of an error.


dwd:
Right, I'll close this one and we can resurrect it if needs be later.

PR #169, opened 2015-02-02
OF-874 disco#items request SHOULD return connected or available resources

sco0ter:


dwd:
Shouldn't this only list resources for which the requester already has presence? Also, I'm interested as to what needs this, and which other servers implement it.

sco0ter:
Not sure, what you mean with presence. It should behave like Example 15. in XEP-0030: Return all resources of the requesting user.
I saw this in ejabberd, too.

sco0ter:
I now see what you mean. You are right, this probably needs some security check.

Flowdalic:
I wonder if Openfire has a `boolean isSubscribedTo(Jid myJid, Jid otherJid)`. If not, then I think we should add such an method, because in XMPP there are just so many places where need to make decisions based on if an entity is subscribed to anthers entity presence.

dwd:
It's more complicated than that. If I send directed presence to you, you should probably see that one even if you're not subscribed. It's basically the same logic, and function, as a probe, and probably dates from when probes weren't sent from end clients at all.

So there's a "canSeeSession" or something needed.

sco0ter:
I've found the canProbePresence method for that. Unfortunately I don't know where and if Openfire has a "canSeeSession" for directed presence. Anybody an idea?

dwd:
canProbePresence is enough, though it *should* also handle directed presence.

PR #170, opened 2015-02-03
OF-875 Roster requests to bare JID of the user are not responded

sco0ter:
There were some issue with this code:

1. Before this change the else branch (removeRosterItem(roster)) did return null and therefore violated IQ semantics.
2. IQs targeted at the user's own bare JID were routed to the else branch, which was wrong.
3. IQs targeted directly to the server were responded to. Prosody returns an error in this case and I think it's correct.

It now checks if the sender equals the recipient or the recipient is null. Otherwise return service-unavailable.

I don't understand the comment (// The packet must be a roster removal from a foreign domain user.)
AFAIK IQHandlers only deal with IQs which were routed to the own server. Foreigen domains are already handled by IQRouter (I think).
I also don't understand why it must be a roster removal then. Anybody?
There's also a similar "removeItem" method which already deals with roster item deletion.

Please review, but I guess the else branch is safe to remove.

### DIFF
--- a/src/java/org/jivesoftware/openfire/handler/IQRosterHandler.java
+++ b/src/java/org/jivesoftware/openfire/handler/IQRosterHandler.java
@@ -110,19 +110,18 @@ public class IQRosterHandler extends IQHandler implements ServerFeaturesProvider
     @Override
 	public IQ handleIQ(IQ packet) throws UnauthorizedException, PacketException {
         try {
-            IQ returnPacket = null;
+            IQ returnPacket;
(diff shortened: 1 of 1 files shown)


dwd:
I *think* the right fix might involve stamping missing "to" with the user's bare jid... But I suspect doing so might have knock-on effects.

sco0ter:
I did exactly this (to fix OF-872 and OF-462) and then I encounterd the misbehavior in IQRosterHandler. So this should be fixed before.

PR #176, opened 2015-02-06
Mark Openfire 3.10.0 Release Candidate

akrherz:


akrherz:
went against 3.10 instead

PR #177, opened 2015-02-06
OF-874: Fix NPE if from/to is missing

tevans:
Returns an empty iterator if the inbound packet is incomplete.


akrherz:
@sco0ter could you kindly review this, so we can merge it and get the RC out the door, thanks

PR #178, opened 2015-02-07
OF-869: Upgrade to latest Jetty (9.2.7)

tevans:
* Updated JAR file dependencies for Jetty 9.2.7.v20150116
* Extended max HTTP header size to 32k (jetty default is 8k), and added a property to configure it as needed.


tevans:
FYI - we have been running this update for a while now under significant load and have observed noticeably fewer Jetty errors in the BOSH endpoint (particularly related to https://bugs.eclipse.org/bugs/show_bug.cgi?id=425422).

Any chance we could get this pulled in?

PR #185, opened 2015-02-19
OF-877 breaks plugins webservices fix

deleolajide:
This fix reverts behavour to reset the web contexts handler collection as it removes the webservices added by plugins at class construction time.

### DIFF
--- a/src/java/org/jivesoftware/openfire/http/HttpBindManager.java
+++ b/src/java/org/jivesoftware/openfire/http/HttpBindManager.java
@@ -123,9 +123,9 @@ public final class HttpBindManager {
     public static Map<String, Boolean> HTTP_BIND_ALLOWED_ORIGINS = new HashMap<String, Boolean>();
 
     private static HttpBindManager instance = new HttpBindManager();
-    
+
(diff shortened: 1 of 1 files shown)


tevans:
Hi Dele - 

With this proposed change, will we still able to use the admin console to stop/start and reconfigure the HTTP connector? For example, will I be able to change the HTTP port from 7070 to 5280 without restarting Openfire? That's what the initial patch was intended to fix.

Thanks,
Tom

deleolajide:
Hi Tom :-)

It is interesting you ask that question, cos I don't recollect ever having a problem changing HTTP-BIND ports to restart jetty web service on the new ports  without restarting openfire. That works fine for me. The only exception is when ports under 1024 are specified on linux and then it fails, disabling http-bind service. 

Just moved my ports from 7070/7443 to 8080/8443 and ofmeet and hawtio worked fine without a restart with this fix.

The only restart issue I experience and may require some effort to fix is the situation when installing a plugin that creates a webservice without restarting openfire. Not sure if it is the responsibility of the plugin or the HttpBindManager. Probably both.

tevans:
Ok - thanks Dele.

PR #191, opened 2015-03-06
OF-888: Fix outgoing S2S packet delivery failure

tevans:
Prevent recursion when attempting to send an error response packet via S2S to a non-responsive remote domain.


tevans:
Strange build failure (was working fine locally) ... but I am looking into it. The offending class (org.xmpp.packet.PacketExtension) is located in the lib/merge/tinder.jar file, which is presumably in the build classpath.

PR #192, opened 2015-03-06
OF-858: Auto create new nodes in PEP

nobelium:
This pull request is for https://igniterealtime.org/issues/browse/OF-858

### DIFF
--- a/src/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
+++ b/src/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
@@ -21,6 +21,7 @@
 package org.jivesoftware.openfire.pep;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
@@ -278,7 +279,14 @@ public class IQPEPHandler extends IQHandler implements ServerIdentitiesProvider,
      * in the server's disco#info result (as per section 4 of XEP-0163).
      */
     public Iterator<String> getFeatures() {
-        return XMPPServer.getInstance().getPubSubModule().getFeatures(null, null, null);
+        Iterator<String> it = XMPPServer.getInstance().getPubSubModule().getFeatures(null, null, null);
+        ArrayList<String> features = new ArrayList<String>();
+        while (it.hasNext()) {
+            features.add(it.next());
+        }
+        // Auto Creation of nodes is supported in PEP
+        features.add("http://jabber.org/protocol/pubsub#auto-create");
+        return features.iterator();
     }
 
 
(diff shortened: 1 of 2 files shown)


nobelium:
Hi, I have updated the PR. Please take a look.

nobelium:
I have updated the PR with new method for creating nodes. Please take a look.

src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
```diff
@@ -1084,13 +1105,83 @@ private void getPublishedItems(PubSubService service, IQ iq, Element itemsElemen
...
+        }
+    }
+
+    /**
+     * Response Object returned by createNodeHelper method
+     */
+    private class CreateNodeResponse {
+        private PacketError.Condition creationStatus;
+        private Node newNode;
+        private Element pubsubError;
```
dwd: I think these should be set within a constructor, and possibly should be final. I don't think these should have setter methods.

src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
```diff
@@ -1084,13 +1105,83 @@ private void getPublishedItems(PubSubService service, IQ iq, Element itemsElemen
...
         // Get sender of the IQ packet
         JID from = iq.getFrom();
+        CreateNodeResponse response = new CreateNodeResponse();
         // Verify that sender has permissions to create nodes
         if (!service.canCreateNode(from) || (!UserManager.getInstance().isRegisteredUser(from) && !isComponent(from)) ) {
             // The user is not allowed to create nodes so return an error
-            sendErrorPacket(iq, PacketError.Condition.forbidden, null);
-            return;
+            response.setStatus(PacketError.Condition.forbidden);
+            return response;
```
dwd: My previous comment would mean this would be:

return new CreateNodeResponse(PacketError.Condition.forbidden);

src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
sco0ter: I agree.

src/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
```diff
@@ -278,7 +279,14 @@ public IQHandlerInfo getInfo() {
      * in the server's disco#info result (as per section 4 of XEP-0163).
      */
     public Iterator<String> getFeatures() {
-        return XMPPServer.getInstance().getPubSubModule().getFeatures(null, null, null);
+        Iterator<String> it = XMPPServer.getInstance().getPubSubModule().getFeatures(null, null, null);
+        ArrayList<String> features = new ArrayList<String>();
+        for (;it.hasNext();) {
```
sco0ter: Just a very minor remark: Personally I like a while loop here better for readability. But eventually it doesn't really matter.

Flowdalic:
Should make CreateNodeResponse immutable, s/for/while/, and maybe a space between `if` and `(` (like it's done everywhere else in the codebase). Other then that UACK from me. +1

nobelium:
Thanks for reviewing. Working on it!!

nobelium:
I have updated the PR. Please take a look!!

PR #194, opened 2015-03-06
Disable inband account registration by default

nobelium:


nobelium:
Closing this request as discussed on the thread.

PR #204, opened 2015-04-22
Updated jingleNodes plugin to work with Public/Private IP and Configurable Ports/Servers:

Guenther93:
* Made ports configurable via the admin interface
* Made STUN test server default to google and configurable via the admin interface (previous hardcoded server was dead)
* Made Local/Public IP logic work for a server behind a firewall (previous code was mixing local and remote IP concepts and would only work outside of a firewall)
* Added some logging to the verify code to make it easier to find out why verification failed - previously it was a blackhole that didn't distinguish between the stun server being unreachable or the ip's mismatching.


akrherz:
@Guenther93 great job with submitting your PR :)  One thing you'll want to consider in the future is to commit your changes against a local branch and not master, this will make things easier for you to maintain as master may not immediately take this PR.

deleolajide:
I need this PR.

PR #205, opened 2015-04-23
OF-115 Remove race during outgoing session creation

dwd:
This one is hard to duplicate in the lab, so this may not actually cure the
issue, however I suspect it does.

The existing code is based around a route of last resort, the
OutgoingSessionPromise, which builds a helper - the PacketsProcessor - and
creates the new session. The PacketsProcessors operate from within a thread
pool, and thus have contended execution time.

As new packets arrives from onward transmission during the session setup, the
PacketsProcessor queues them for retransmission. However, the act of session
creation also creates the outgoing route, meaning that later packets will
bypass the queue, and be transmitted out of order.

This should give the impression of a randomly reordered set of stanzas, but
would in fact be two consecutive sequences of ordered stanzas which are
interleaved, hence if the two sequences are A-E, F-J, one might see:

A, B, F, G, C, H, I, D, J, E

The fix here is two ensure that all packets are queued until the queue itself
is empty. In the solution provided, an additional helper class acts as the
route's RoutableChannelHandler, and stores packets in a queue until the real
ServerSession can catch up with the pending traffic. After that it simply
passes the process() call argument through.


src/java/org/jivesoftware/openfire/server/LocalOutgoingServerProxy.java
```diff
@@ -0,0 +1,176 @@
+/**
+ * 
+ */
+package org.jivesoftware.openfire.server;
```
Flowdalic: I just want to point out that if we ever want openfire to get into debian or redhat/fedora, then we need to have a license header for every source file.

src/java/org/jivesoftware/openfire/server/LocalOutgoingServerProxy.java
```diff
@@ -0,0 +1,176 @@
...
+    public LocalOutgoingServerProxy(final String domain, ServerSession session) {
+        this.domain = new JID(domain);
+        this.session = null;
+        this.packets = null;
+    }
+    
+    private static int active = 0;
+    
+    protected synchronized void started() {
+        active = active + 1;
```
Flowdalic: I guess it's intentional that you spelled out the increment instead of `active++`?

src/java/org/jivesoftware/openfire/server/LocalOutgoingServerProxy.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /* (non-Javadoc)
+     * @see org.jivesoftware.openfire.ChannelHandler#process(org.xmpp.packet.Packet)
+     */
+    @Override
+    public synchronized void process(final Packet packet) throws UnauthorizedException,
+            PacketException {
+        if (this.session != null) {
+            this.session.process(packet);
```
Flowdalic: I would remove the following indentation level by adding a `return` at the end of the if's body, so you could remove the else clause.

src/java/org/jivesoftware/openfire/server/LocalOutgoingServerProxy.java
```diff
@@ -0,0 +1,176 @@
...
+        } else {
+            if (packets == null) {
+                packets = new LinkedBlockingQueue<Packet>();
+                log.info("Queued packet for {}.", domain.toString());
+            }
+            packets.add(packet.createCopy());
+            if (isTrying == false) {
+                final String fromDomain = packet.getFrom().getDomain().toString();
```
Flowdalic: I guess it's guaranteed that the stanza has a from and to attribute set, i.e. it's not null.

src/java/org/jivesoftware/openfire/server/LocalOutgoingServerProxy.java
```diff
@@ -0,0 +1,176 @@
...
+                            started();
+                            try {
+                                ServerSession s = LocalOutgoingServerSession.authenticateDomain(fromDomain, toDomain); // Long-running.
+                                if (s != null) {
+                                    sessionReady(s);
+                                } else {
+                                    sessionFailed();
+                                }
+                            } catch(Exception e) {
+                                sessionFailed();
```
Flowdalic: Don't we want to log `e`, at least with a debug log level?

src/java/org/jivesoftware/openfire/server/LocalOutgoingServerProxy.java
```diff
@@ -0,0 +1,176 @@
...
+                final String toDomain = packet.getTo().getDomain().toString();
+                if ((failureTimestamp == -1) || ((System.currentTimeMillis() - failureTimestamp) >= 5000)) {
+                    isTrying = true;
+                    log.debug("Spinning up new session to {}", domain.toString());
+                    pool.execute(new Runnable() {
+                        public void run() {
+                            log.debug("Initiating connection thread for {} -> {} ({})", fromDomain, toDomain, domain.toString());
+                            started();
+                            try {
+                                ServerSession s = LocalOutgoingServerSession.authenticateDomain(fromDomain, toDomain); // Long-running.
```
Flowdalic: Hmm, Code and comment in the same line is at least controversial in the Java community.

src/java/org/jivesoftware/openfire/server/LocalOutgoingServerProxy.java
```diff
@@ -0,0 +1,176 @@
...
+        } else {
+            if (packets == null) {
+                packets = new LinkedBlockingQueue<Packet>();
+                log.info("Queued packet for {}.", domain.toString());
+            }
+            packets.add(packet.createCopy());
+            if (isTrying == false) {
+                final String fromDomain = packet.getFrom().getDomain().toString();
+                final String toDomain = packet.getTo().getDomain().toString();
+                if ((failureTimestamp == -1) || ((System.currentTimeMillis() - failureTimestamp) >= 5000)) {
```
Flowdalic: So the intention is to try at most every 5 seconds? This value may be a good candidate for a openfire property setting.

src/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -413,4 +414,57 @@ public String getCipherSuiteName() {
...
+                }
+                Message reply = new Message();
+                reply.setID(packet.getID());
+                reply.setTo(packet.getFrom());
+                reply.setFrom(packet.getTo());
+                reply.setType(Message.Type.error);
+                reply.setThread(((Message)packet).getThread());
+                reply.setError(PacketError.Condition.remote_server_not_found);
+                routingTable.routePacket(reply.getTo(), reply, true);
+            }
```
Flowdalic: I'm a fan of defensive programming, and would put an `else` clause throwing an AssertionError here.

tevans:
This looks good to me. If we can get the merge conflicts cleaned up, let's merge this one into the pending 3.11 master.

dwd:
Rebased.

PR #207, opened 2015-04-24
OF-907 - Accept SSLv2 Hello

dwd:


godly:
Is this the right fix? If I got it right some Psi _clients_ have problems to connect while the server connections are fine. So this fix wouldn't help as "clientmode" still does not support SSLv2.

I'd prefer system properties to modify the supported protocols if needed and keep the current default as is ("TLSv1", "TLSv1.1", "TLSv1.2") as it makes no sense to re-enable support for Poodle. Anyhow if one wants to support old clients setting a system property is usually fine.

Sorry for the late update and reopening the issue.

dwd:
I merged this manually, actually, after clarifying the comment, but "server" in this case is referring to the passive connection side, not the question of whether it's a C2S or S2S session - anything connecting to Openfire can (with this patch) use an SSLv2 Hello message to initiate TLS.

Secondly, this doesn't re-enable support for Poodle, or indeed SSLv2 - it just allows the first message of the handshake, the Hello message, to use the SSLv2 syntax. The server will respond (using SSLv2's protocol) that it supports SSLv3.1, v3.2, and v3.3 (which are TLSv1, v1.1, v1.2 in fact).

PR #209, opened 2015-04-28
OF-909: Include ack in BOSH response body element

tevans:
Per spec, the ack attribute in the response <body/> allows a BOSH client
to verify that sent stanzas with the corresponding rid(s) have been
successfully received by the server.

This PR will merge to master. See #210 for corresponding change for the 3.10 branch.

### DIFF
--- a/src/java/org/jivesoftware/openfire/http/HttpBindServlet.java
+++ b/src/java/org/jivesoftware/openfire/http/HttpBindServlet.java
@@ -348,14 +348,6 @@ public class HttpBindServlet extends HttpServlet {
         sendLegacyError(context, error, null);
     }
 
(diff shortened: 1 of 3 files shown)


tevans:
This patch approximates the behavior that was observed in the pre-3.10 builds (before the async request/response was introduced). Although I agree that the spec recommends leaving the 'ack' attribute out of the response body when it is equal with the 'rid' from the corresponding request (an implied acknowledgement), I took the liberty of overriding this recommendation (in the spirit of the [RFC-2119] (http://www.ietf.org/rfc/rfc2119.txt) definition of 'SHOULD NOT') in retaining the pre-existing behavior of returning an explicit ack in all cases.

Happy to discuss further as needed.

PR #211, opened 2015-04-30
Plugin: userservice. Get REST responses in JSON too

alexproca:
Return responses from **userservice** REST plugin in json format too for easy consumption from javascript.
In order to get responses in JSON add header

**Accept: application/json**

### DIFF
--- a/src/plugins/userservice/src/java/org/jivesoftware/openfire/plugin/service/UserGroupService.java
+++ b/src/plugins/userservice/src/java/org/jivesoftware/openfire/plugin/service/UserGroupService.java
@@ -25,7 +25,7 @@ public class UserGroupService {
 	}
 
 	@GET
-	@Produces(MediaType.APPLICATION_XML)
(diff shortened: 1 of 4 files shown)


Redor:
Hi, thy for your contribution. Your PR looks good and I will merge it, but I would recommend you to take the REST API plugin, which contains all the userservice features. The only thing you need to change is the path of the URL (from userservice to restapi). The REST API plugin contains already the support of JSON (consuming and response). 

REST API documentation: https://www.igniterealtime.org/projects/openfire/plugins/restapi/readme.html

alexproca:
Hello,

I know there is another api for all services but right now I need only acces to users. Maybe I’il look into REST API plugin to enable json too with another PR.

Best regards,
Alex Proca

Redor:
REST API support already fully JSON.

alexproca:
I did not look into it at all. Thanks for pointing that.

PR #215, opened 2015-05-12
Order Fastpath Workgroups by Display Name

drowe:
Following suit with Users/Groups, Workgroups should order in alpha, shouldn't they?


Redor:
+1

guusdk:
Thanks for your contribution!

PR #217, opened 2015-05-15
Make module loading driven by a file

randylayman:
This change moves the list of modules to be loaded from hard coded in XMPPServer to defined in conf/modules.xml.  My motivation is to allow me to replace Openfire behavior defined in modules that can't be overridden or changed in a plugin.

There are two basic elements of the change.  First, the loadModules method has been updated to open the modules.xml file and pass the contents of the file to loadModule.  

The second part of the change is to modify the modules map to use a String as the key.  This allows for the modules to be looked up by the name they implement instead of the class that is actually implementing.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -1176,6 +1176,7 @@
 
         <copy tofile="${target.solaris.root}/pkginfo" file="${basedir}/build/pkg/pkginfo" />
         <copy tofile="${target.solaris.root}/opt/openfire/resources/solaris/openfire.xml" file="${basedir}/build/solaris/openfire.xml" />
+        <copy tofile="${target.solaris.root}/opt/openfire/resources/solaris/modules.xml" file="${basedir}/build/solaris/modules.xml" />
         <copy tofile="${target.solaris.root}/checkinstall" file="${basedir}/build/solaris/checkinstall" />
         <copy tofile="${target.solaris.root}/postinstall" file="${basedir}/build/solaris/postinstall" />
         <copy tofile="${target.solaris.root}/preremove" file="${basedir}/build/solaris/preremove" />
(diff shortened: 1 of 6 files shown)


akrherz:
Thanks @randylayman , this looks interesting.  We would need additional changes to make sure this new conf file is bundled with the various install artifacts generated by the build system.

randylayman:
Can you provide some guidance on where I'd need to make changes?

akrherz:
@randylayman , check out the build/ folder.  In general, it may be best to wait for further feedback from folks like @dwd to make sure this type of change would be accepted so you don't waste any time with further improvements.

randylayman:
I'm not sure how to test the installer changes but made the changes that I believe are necessary.

deleolajide:
This is a good change. I am curious about why a new xml file instead of a module node in openfire.xml

randylayman:
Any change in the modules file will affect the inner workings of Openfire and will possibly (probably?) lead to things not working as expected.  To me, changing a module is something that should be done with care and very intentional forethought.  Putting it in the openfire.xml file seems to make it too easy for installations to make changes without understanding the ramifications.

With that said, if you think its better in the openfire.xml file I will gladly move it.

deleolajide:
Thanks for the reply. I am just curious and wanted to appreciate your decision. It is for @dwd to decide

randylayman:
@dwd Is there anything else I need to do to have this PR reviewed and given a thumbs up/thumbs down/needs changes decision?

dwd:
Sorry, I'm being rubbish and distracted. I have reviewed, agree that the
driver config file should be separated. I'll merge when I'm at a desktop.

randylayman:
Not a rush, just wanted to make sure I hadn't missed a step in a process that was keeping this from being seen.

guusdk:
Note that there's a small issue being introduced here. See https://igniterealtime.org/issues/browse/OF-957

Fix: https://github.com/igniterealtime/Openfire/pull/322

guusdk:
We've been using the new modules.xml file for a while now, and to be honest, it's not working out for me. The implementation itself appears to be fine - things work as intended. I am, however, repeatedly running into issues that relate to builds / instances that have 'out-of-date' versions of the modules.xml. That's likely caused by one or more issues in the build or upgrade process, but when I was thinking about it - what does modules.xml solve what cannot be done with an overlay? Yes, the solution presented here is a cleaner solution compared to overriding XMPPServer in an overlay, but perhaps I'm starting to think that the added complexity outweighs the benefits of this change. By far most user (or even developers) won't ever need this functionality - for those that do need, I think we have a fitting workaround: https://www.igniterealtime.org/builds/openfire/docs/latest/documentation/overlay.html

guusdk:
https://igniterealtime.org/issues/browse/OF-1005

guusdk:
PR for rollback: https://github.com/igniterealtime/Openfire/pull/455
I'd welcome some feedback!

PR #219, opened 2015-05-28
OF-883: Compensate for new half-duplex close of TCP channel in MINA

guusdk:
MINA 2.0.8 introduces support for half-duplex close of TCP channels (see DIRMINA-785).
As part of this change, IoHandlers got a new method. The default behavior of the
IoHandlerAdapter closes the MINA session when this handler is triggered. However,
without an Openfire-specific extension (in ConnectionHandler), this fails to close
Openfire-maintained resources. As a result, CPUs could start to spin in NIO code,
resulting in 100% CPU cycles.

To fix this issue, ConnectionHandler now overrides the default functionality and
triggers a normal closure of the session (both in Openfire as well as MINA context).

Additionally, to prevent the CPU spin, MINA sessions need to be closed synchronously
(as opposed to the async closure Openfire had up until now). I cannot quite explain
the need for this change, other than a suspision of race conditions in the async
setup.

### DIFF
--- a/src/java/org/jivesoftware/openfire/nio/ConnectionHandler.java
+++ b/src/java/org/jivesoftware/openfire/nio/ConnectionHandler.java
@@ -108,6 +108,14 @@ public abstract class ConnectionHandler extends IoHandlerAdapter {
     }
 
     @Override
+    public void inputClosed( IoSession session ) throws Exception {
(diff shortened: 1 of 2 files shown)


tevans:
Looks good to me. Thanks for the helpful summary of your analysis.

PR #220, opened 2015-05-29
Two additional fixes that relate to OF-883

guusdk:
The commits in this pull request address:
- an issue where only the MINA session, not the entire Openfire connection, is closed upon an error
- re-closing a connection that's already being closed (potentially looping)

More details can be found in the commit messages.


guusdk:
Ok, the commit above should address those concerns.

PR #225, opened 2015-06-01
Allow users to specifiy custom classes to map credentials from certificates

victorqhong-bah:
This change allows users to specify classes in the Openfire properties (similar to AuthorizationManager) to map different identity credentials from certificates. This change is useful during client authentication with certificates (mutual authentication) in order to provide different credentials to the AuthorizationMapping (e.g. full subject distinguished name instead of only common name).

This change is taking code out of CertificateManager.java and moving it into CNCertificateIdentityMapping and SANCertificateIdentityMapping which implement a new CertificateIdentityMapping interface. In CertificateManager it fell back to the CN if the SAN did not have any XMPP OIDs present. In this pull request it is currently adding CN to any list of identities that might have come from the SAN field. Most code appears to only use the first identity but if there is a concern about this change then the pull request can be updated to break out of the identity mapper loop when the first identity mapper returned a non-empty list of identities.


dwd:
Sorry for not having reviewed this sooner. I think this is essentially correct, but:

 -  it really requires there to be two lists of user-defined classes, one for peer server authentication, and one for client authentication. Clients might reasonably be authenticated by treating the mailbox as a Jid, for example, whereas servers need sRVName considering, etc.

- I need to test this considerably, since it's security code, and I'm short on time currently.

victorqhong-bah:
Thanks for reviewing. Since you're short on time, let me know what I can do to help. I can add another property for server mappings, is there anything I can do to help the testing process?

PR #235, opened 2015-06-11
Prevent sessions in admin console from showing UnknownHostException

AndiHeusser:
Updated JSP pages around sessions to prevent them from blowing up when an UnknownHostException is thrown and show 'Invalid session/connection' instead. The user can then manually remove those sessions.

### DIFF
--- a/src/web/component-session-details.jsp
+++ b/src/web/component-session-details.jsp
@@ -151,13 +151,13 @@
             <fmt:message key="session.details.hostname" />
         </td>
         <td>
-        	<% try { %>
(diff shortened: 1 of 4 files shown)


akrherz:
Would be nice if this commit mentioned OF-889

AndiHeusser:
Yes, ah I see. I thought that because my branch is named of889, that'd be good enough but it should be in the commit comment too. Let me see if I can amend the commit comment.

AndiHeusser:
Ah, didn't do as nicely as I wanted. I'll recreate the branch and submit a new PR with a cleaner commit history. Sorry.

PR #239, opened 2015-06-17
Enabled SSL Connection Pooling

speedy01:
per http://docs.oracle.com/javase/8/docs/technotes/guides/jndi/jndi-ldap.html#pooling  SSL Connection Pooling is disabled when using a custom ssl socket. The custom socket was used to accept unsigned/untrusted/expired ssl certs. I removed the use of the custom socket, and also added the required system property to enable SSL pooling, which was missing.


dwd:
Pushed independently after fixing the description a bit.

PR #240, opened 2015-06-17
OF-919 Update bundled jDTS driver to 1.3.1

akrherz:


akrherz:
Speedy verified that the md5sum of the jtds.jar file included in this PR works for him, merging.

PR #241, opened 2015-06-17
OF-922: Revert MINA to 2.0.7

tevans:
This is in response to several performance problems discovered with connection management in the 2.0.9 MINA release. There are two commits here: one to revert the original upgrade, and another to remove an extraneous override that was added later. 

I am submitting this PR against the master branch, but we should be able to cherry-pick and/or merge into the 3.10 branch if we want to ship a 3.10.2 patch release. Refer to ticket for details.


akrherz:
@deleolajide Do you have any comments on doing this?  Is there anything specific your plugins needed with MINA 2.0.9 that does not properly work in 2.0.7?

deleolajide:
Thanks for the heads up. I am immune to Mina. I am mostly web-based using Websockets from Jetty

PR #244, opened 2015-06-28
Of 926 clients can't authenticate using ldap ssl

speedy01:


dwd:
Needs the merges removed. Try git pull --rebase (or a proper rebase). If this seems like it'll be too tricky, I'll do it later.

speedy01:
I'm unsure how to do this, and I'm afraid I mess things up even more.

speedy01:
ok. I think I figured it out!

dwd:
I don't understand how this is only affecting LDAP connections, and not any certificate validation. Obviously if it's affecting XMPP connections, this is a no-go.

speedy01:
this is only being called by LdapManager.java 
https://github.com/speedy01/Openfire/commit/3fd44c51d6328f649e4fa9b1ec6801c50f5c2cfa

This doesn't affect xmpp.  It returns the behavior that was used prior to PR #239, without using SimpleSSLSocketFactory (custom ssl factory). Since a custom ssl factory is not being used, connection pooling with ssl still works.   Ive also added a system properties which lets the user disable this, allowing for only valid ssl certs when connecting to ldap.

speedy01:
submitted pr 364 to replace this one.

PR #247, opened 2015-07-02
Fix collection requests XEP-0059

ishmakov:
https://community.igniterealtime.org/thread/56171
https://community.igniterealtime.org/thread/56173


Redor:
+1

PR #248, opened 2015-07-03
FIX: Monitoring plugin list request with max tag does not work correctly

ishmakov:
FIX: Monitoring plugin list request with max tag does not work correctly
https://community.igniterealtime.org/thread/56168

### DIFF
--- a/src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/JdbcPersistenceManager.java
+++ b/src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/JdbcPersistenceManager.java
@@ -629,6 +629,7 @@ public class JdbcPersistenceManager implements PersistenceManager {
 			}
 		}
 		querySB.append(" )");
+		querySB.append(SELECT_CONVERSATIONS_GROUP_BY);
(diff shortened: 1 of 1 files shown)


ishmakov:
It's need to review by author or somebody, that it does not break other feature in Monitoring plugin.
At my first rewiew all are ok, but to be shure, better check it.

leonroy:
I'll take a look.

leonroy:
The SQL fu is strong with this one...that's one heckuva select and nice usage of GROUP BY. I don't think anyone's ever used GROUP BY in a single SELECT in the Openfire code base.

Going to fire up my test data and see if all the other monitoring queries still work ok fine,

thanks!

leonroy:
Just looking at writing some integration tests to make verification of the monitoring plugin MUCH easier.

Do you know if there's a decent XMPP client which supports XEP-0136?

ishmakov:
@leonroy Unfortunately I do not have a client, I write an extension gloox library.
But I spend testing via QIP client allows you to connect via the XMPP protocol and having a console, where you can send your XML request and read the response.

dwd:
Sorry, this took me ages to get around to testing.

PR #249, opened 2015-07-07
Allow users to specifiy custom classes to map credentials from certificates

victorqhong-bah:
Resubmitting this pull request from a different branch. See [here](https://github.com/igniterealtime/Openfire/pull/225).

"This change allows users to specify classes in the Openfire properties (similar to AuthorizationManager) to map different identity credentials from certificates. This change is useful during client authentication with certificates (mutual authentication) in order to provide different credentials to the AuthorizationMapping (e.g. full subject distinguished name instead of only common name).

This change is taking code out of CertificateManager.java and moving it into CNCertificateIdentityMapping and SANCertificateIdentityMapping which implement a new CertificateIdentityMapping interface. In CertificateManager it fell back to the CN if the SAN did not have any XMPP OIDs present. In this pull request it is currently adding CN to any list of identities that might have come from the SAN field. Most code appears to only use the first identity but if there is a concern about this change then the pull request can be updated to break out of the identity mapper loop when the first identity mapper returned a non-empty list of identities."

### DIFF
--- a/src/java/org/jivesoftware/openfire/clearspace/ClearspaceX509TrustManager.java
+++ b/src/java/org/jivesoftware/openfire/clearspace/ClearspaceX509TrustManager.java
@@ -77,7 +77,7 @@ public class ClearspaceX509TrustManager implements X509TrustManager {
         if (verify) {
             int nSize = x509Certificates.length;
 
-            List<String> peerIdentities = CertificateManager.getPeerIdentities(x509Certificates[0]);
+            List<String> peerIdentities = CertificateManager.getServerPeerIdentities(x509Certificates[0]);
 
             if (getBooleanProperty("clearspace.certificate.verify.chain", true)) {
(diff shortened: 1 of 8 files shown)


dwd:
I've finally given this the review it deserves, and it's mostly positive. However I notice that this code is not conformant to RFC 6125, in particular, CommonName mapping are used even if there are existing SANs present. The old code used CommonNames as a fallback in the absence of matching SANs (still not right), but this uses both at all times.

This is problematic, because the presence of any SAN in the certificate implies the CommonName matching rules should not be used.

We could have the CommonName mapping return nothing if it finds matching SAN types?

victorqhong-bah:
Thanks for the feedback. I knew about the difference in functionality but didn't know the original behavior was from RFC 6125. I'll make a change that addresses this.

dwd:
Thanks!

PR #250, opened 2015-07-09
Of 929

leonroy:
As per here:
https://igniterealtime.org/issues/browse/OF-929

I was looking for a way to quickly generate a test Openfire environment populated with users, roster entries and messages.

Jive had a plugin already which did the first two. It now does number three as well, generating plenty of test data for the monitoring plugin..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 target/
 work/
+*.class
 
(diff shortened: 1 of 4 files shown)


leonroy:
Appreciate it if someone could peer review this before I merge it!

PR #251, opened 2015-07-10
Restored and separated Database-Query for MYSQL

tomatis:
This patch repairs an incompatibility of the current default PURGE_FOR_SIZE - SQL-query by adding a separated variant for the database-type mySQL.
Previously this incompatibility was provoked by "fixing" the default-query for hsqldb - see OF 789 [https://igniterealtime.org/issues/browse/OF-789] and thread on https://community.igniterealtime.org/message/250575 )


Redor:
@sco0ter Could you please review/comment that?

PR #259, opened 2015-07-24
Initial xep-0198 message ack implementation

dwd:
This patch adds a basic, non-resumable, XEP-0198 acking solution. It will
redirect "lost" messages to offline storage, but does not attempt to suspend
the session, so will not allow any resumptions.

### DIFF
--- a/src/java/org/jivesoftware/openfire/SessionManager.java
+++ b/src/java/org/jivesoftware/openfire/SessionManager.java
@@ -25,6 +25,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
+import java.util.Deque;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -67,11 +69,13 @@ import org.jivesoftware.openfire.session.OutgoingServerSession;
 import org.jivesoftware.openfire.session.RemoteSessionLocator;
 import org.jivesoftware.openfire.session.Session;
 import org.jivesoftware.openfire.spi.BasicStreamIDFactory;
+import org.jivesoftware.openfire.streammanagement.StreamManager;
 import org.jivesoftware.openfire.user.UserManager;
 import org.jivesoftware.openfire.user.UserNotFoundException;
 import org.jivesoftware.util.JiveGlobals;
 import org.jivesoftware.util.LocaleUtils;
 import org.jivesoftware.util.Log;
+import org.jivesoftware.util.XMPPDateTimeFormat;
 import org.jivesoftware.util.cache.Cache;
 import org.jivesoftware.util.cache.CacheFactory;
 import org.slf4j.Logger;
@@ -1245,10 +1249,20 @@ public class SessionManager extends BasicModule implements ClusterEventListener/
 
                     // Re-deliver unacknowledged stanzas from broken stream (XEP-0198)
                     if(session.getStreamManager().isEnabled()) {
-	                    Map<Long,Packet> unacknowledgedStanzas = session.getStreamManager().getUnacknowledgedServerStanzas();
+                            session.getStreamManager().setEnabled(false); // Avoid concurrent usage.
+	                    Deque<StreamManager.UnackedPacket> unacknowledgedStanzas = session.getStreamManager().getUnacknowledgedServerStanzas();
 	                    if(!unacknowledgedStanzas.isEmpty()) {
-	                    	for(Entry<Long,Packet> unacknowledgedStanza : unacknowledgedStanzas.entrySet()) {
-	                    		router.route(unacknowledgedStanza.getValue());
+	                    	for(StreamManager.UnackedPacket unacked : unacknowledgedStanzas) {
+	                    	    if (unacked.packet instanceof Message) {
+	                    	        Message m = (Message)unacked.packet;
+        	                        Element delayInformation = m.addChildElement("delay", "urn:xmpp:delay");
+        	                        Element delayInformationOld = m.addChildElement("x", "jabber:x:delay");
+                                        delayInformation.addAttribute("stamp", XMPPDateTimeFormat.format(unacked.timestamp));
+                                        delayInformationOld.addAttribute("stamp", XMPPDateTimeFormat.formatOld(unacked.timestamp));
+                                        delayInformation.addAttribute("from", serverAddress.toBareJID());
+                                        delayInformationOld.addAttribute("from", serverAddress.toBareJID());
+	                    	    }
+    	                            router.route(unacked.packet);
 	                    	}
 	                    }
                     }
--- a/src/java/org/jivesoftware/openfire/SessionManager.java
+++ b/src/java/org/jivesoftware/openfire/SessionManager.java
@@ -29,6 +29,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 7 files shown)


src/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -413,4 +429,27 @@ public String getCipherSuiteName() {
...
+
+    /**
+     * Enables stream management for session
+     * @param enable XEP-0198 <enable/> stanza
+     */
+    public void enableStreamMangement(Element enable) {
+
+    	// Do nothing if already enabled
+    	if(streamManager.isEnabled()) {
+    		return;
```
tevans: Should we reset the counter(s) on subsequent <enable />? Spec seems a little ambiguous on this point.

src/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	/**
+	 * Send an error if a XEP-0198 stanza is received at an unexpected time.
+	 * e.g. before resource-binding has completed.
+	 */
+	public void sendUnexpectedError() {
+		StringBuilder sb = new StringBuilder(340);
+		sb.append(String.format("<failed xmlns='%s'>", getNamespace()));
+		sb.append("<unexpected-request xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/>");
```
tevans: Alternatively could use:
`new PacketError(PacketError.Condition.unexpected_request).toXML()`

src/java/org/jivesoftware/openfire/SessionManager.java
```diff
@@ -1241,6 +1242,16 @@ public void onConnectionClose(Object handback) {
                         presence.setFrom(session.getAddress());
                         router.route(presence);
                     }
+
+                    // Re-deliver unacknowledged stanzas from broken stream (XEP-0198)
+                    if(session.getStreamManager().isEnabled()) {
+	                    Map<Long,Packet> unacknowledgedStanzas = session.getStreamManager().getUnacknowledgedServerStanzas();
+	                    if(!unacknowledgedStanzas.isEmpty()) {
+	                    	for(Entry<Long,Packet> unacknowledgedStanza : unacknowledgedStanzas.entrySet()) {
+	                    		router.route(unacknowledgedStanza.getValue());
```
tevans: Are there any XEP-0203 (delayed delivery) implications here? There's no way (currently) to know how long the unacknowledged stanza(s) have been retained. Might we also provide configurable support for a time-based acknowledgement request (`<r />`)?

tevans:
Good stuff!

tevans:
Additional random thoughts:

I've wondered about the merits of providing a configurable means for sending unsolicited acks from the server to the client (S2C) to save the client from sending `<r />` potentially for every request, depending on the client's tolerance for unacknowledged stanzas. For example:
* We could add a server-side setting (global) to indicate the frequency of unsolicited S2C acks, default zero/disabled
* Perhaps "bend" the spec to borrow/reuse the `max='n'` on the `<enable />` element attribute to negotiate a count/time limit for unsolicited S2C acks on a per-stream basis (since we do not support the resume capabilities)

Note that none of this is intended to hold up or presume upon this PR, just providing some input in context.

src/java/org/jivesoftware/openfire/net/StanzaHandler.java
```diff
@@ -182,13 +183,28 @@ else if ("compress".equals(tag)) {
...
+        		case "enable":
+        			session.enableStreamMangement(doc);
+        			break;
+        		case "r":
+        			session.getStreamManager().sendServerAcknowledgement();
+        			break;
+        		case "a":
+        			session.getStreamManager().processClientAcknowledgement(doc);
+        			break;
+        		default:
```
tevans: Should probably also return a `<failed><not-implemented/></failed>` (sic) response here for the `<resume />` tag, although admittedly a properly implemented client should not send it since we haven't advertised the capability.

src/java/org/jivesoftware/openfire/session/LocalSession.java
Flowdalic: Appears that <enable> while already enabled doesn't seem to be covered in xep198. I think the spec should state that sending a <enable> after the SM has already been enabled is either a no-op (like it's implemented here), or result in a stream-error and subsequent stream termination (which I would prefer).

tevans:
Nice improvements ... I think this PR is fine as-is; additional incremental work can follow as needed. 

For example, I would suggest that we implement an optional/configurable timer task to send ack requests for sessions that have gone idle. This would minimize the number of duplicate stanzas that will accumulate (as many as _(requestFrequency-1)_ per session) using only the counter-based ack approach. Perhaps three minutes would be an appropriate default?

Also, I recommend that the admin console be extended to provide access to the stream management configuration options (enable/disable, stanza count, timer duration, etc.).

If these suggestions are acceptable, I would be happy to add these capabilities after the PR is merged.

dwd:
Sounds great! The other big task is to make the stream manager objects have a longer lifetime, and move the session state there, so we can do full resumption.

tevans:
More FYI - to enable stream management via BOSH we will also need to update the following classes:

- HttpSession (to advertise the `<sm/>` feature)
- SessionPacketRouter (to handle the sm stanzas)

I will add this to my list of follow-on work after the PR is merged.

dwd:
I don't think SM over BOSH makes sense - BOSH does much of the same thing anyway.

tevans:
Does it make sense to have `StreamManager` be an interface, with a configurable `DefaultStreamManager` behind it? This would be more in keeping with the other pluggable _managers_ within Openfire.

For example, this would allow an installation to configure Openfire for ack only vs ack+resume stream managers. It would also enable plugin developers to provide their own stream manager implementations as appropriate.

tevans:
WRT the BOSH handler, I was more inclined to use stream management for the websocket upgrade stream (which will reuse `SessionPacketRouter`) rather than BOSH proper, but the capability would be configurable in either case.

yigubigu:
HI, I download package night build of openfire_2015-12-06, and verify whether message will be lost during receiver offline short period of time, the message will be lost, same as before. 

Also I do some code comparison, seems SessionManager is not merged correctly. In the pull request, there are following code sniff, but they are not in current trunk, 


// Re-deliver unacknowledged stanzas from broken stream (XEP-0198)
+                    if(session.getStreamManager().isEnabled()) {
+                            session.getStreamManager().setEnabled(false); // Avoid concurrent usage.
+	                    Deque<StreamManager.UnackedPacket> unacknowledgedStanzas = session.getStreamManager().getUnacknowledgedServerStanzas();
+	                    if(!unacknowledgedStanzas.isEmpty()) {
+	                    	for(StreamManager.UnackedPacket unacked : unacknowledgedStanzas) {
+	                    	    if (unacked.packet instanceof Message) {
+	                    	        Message m = (Message)unacked.packet;
+        	                        Element delayInformation = m.addChildElement("delay", "urn:xmpp:delay");
+        	                        Element delayInformationOld = m.addChildElement("x", "jabber:x:delay");
+                                        delayInformation.addAttribute("stamp", XMPPDateTimeFormat.format(unacked.timestamp));
+                                        delayInformationOld.addAttribute("stamp", XMPPDateTimeFormat.formatOld(unacked.timestamp));
+                                        delayInformation.addAttribute("from", serverAddress.toBareJID());
+                                        delayInformationOld.addAttribute("from", serverAddress.toBareJID());
+	                    	    }
+    	                            router.route(unacked.packet);
+	                    	}
+	                    }
+                    }
                 }

dwd:
That's in the current trunk; line 1272 of SessionManager.java

yigubigu:
oh, sorry, I read the code from compiled, so didn't find this. After I review from eclipse, I found it. 
Still there are 1 or 2 messages lost during receiver is offline, but someone is sending message to him/her at same time. The offline message store has 4 records (first one is active, second one is in-active, third one is active, fourth one is inactive) for the missed messages. 
For normal offline message, there are 5 records, which has composite message and request body message.

dwd:
OK; we may have addressed that within the fixes for OF-1002 and OF-1003; you might want to check. Otherwise, test cases and patches are always welcome, but we'll see if we can add this to a testing cycle.

yigubigu:
thanks, I will take a try and update you again.

yigubigu:
the test is failed. System has detected that client is offline, but sessionManager/StreamManager could not add the missed message back to offline message store.

PR #261, opened 2015-07-29
OF-933: Initial WebSocket implementation

tevans:
Adds websocket capabilities compliant with the latest specifications (RFC 7395).

Note that the new component is implemented as a plugin, but I have included a few small modifications to core classes to use the new namespace constants for bind and session features.

I also held back a few additional changes that are dependent upon a merge of the pending stream management feature set (PR #259). The hooks for these items are noted in code comments as appropriate.

I have tested these changes using the stanza.io client library on both Chrome and Firefox.


tevans:
Looks like I need to rebase/squash this PR ... stay tuned.

PR #263, opened 2015-07-29
OF-933: Initial WebSocket implementation

tevans:
Adds WebSocket support compliant with the latest specifications (RFC 7395). This implementation also supports Openfire's new stream management ack capabilities. I have tested this plugin using the stanza.io client library on both Chrome and Firefox.


akrherz:
@deleolajide Do you have any comments on this?

@tevans I assume you mean by 'supports Openfire's new stream management' that it does not require it, so it can run cleanly on 3.10 ?

tevans:
@akrherz Excellent question ... indeed I will need to do a bit more work and testing to isolate the optional SM feature so we can enable this plugin for 3.10.x. Thanks for the feedback, update to follow shortly.

dwd:
It's not totally clear to me where it ensures the default namespace is applied to stanzas, but otherwise it looks great. Obviously if it works with stanza.io it *should* be OK, and I might just merge this in a bit anyway.

dwd:
Checked the stanza.io code, and I'm pretty sure it requires stanzas to be explicitly namespaced (as the spec does) so I'm merging this.

deleolajide:
@akrherz , thanks for the heads up. I was going to do a quick check to make sure I could replace the non standard "Openfire.Connection" in Strophe with the original "Strophe.Connection" in most of my web applications, but that can now happen later when I am free.

In any case, @tevans, I am expecting this to be up your usual high standard of excellent work and thanks for this long overdue contribution.

tevans:
@deleolajide Thanks for the feedback. Note that you will want to pull the latest websocket.js from the Strophe project (pre-release) to get a RFC-compliant implementation.

deleolajide:
@tevans, thanks for the tip. Cheers!!

PR #275, opened 2015-08-06
HazelcastClusterNodeInfo communication has big delay.

caojianlin:
Using getHostString() to benefit from not attempting a reverse lookup.
Or will cause very big delay if the ip address doese not have configured
host name.


tevans:
Makes sense to me, thanks for the contribution. We will take a look at updating the plugin to the latest upstream release (HZ 3.5.1), also including this patch.

PR #277, opened 2015-08-11
fixed MySQL Error 1093 in monitoring plugin ConversationManager & added description

PGoski:
1) add new monitoring plugin description
2) fix MySQL Error 1093(can't modify the same table which is used in the SELECT part) in ConversationManager


src/plugins/monitoring/src/java/org/jivesoftware/openfire/archive/ConversationManager.java
```diff
@@ -988,14 +988,15 @@ public void run() {
...
 					while ((message = messageQueue.poll()) != null) {
-						pstmt.setLong(1, message.getConversationID());
-						pstmt.setString(2, message.getFromJID().toBareJID());
-						pstmt.setString(3, message.getFromJID().getResource());
-						pstmt.setString(4, message.getToJID().toBareJID());
-						pstmt.setString(5, message.getToJID().getResource());
-						pstmt.setLong(6, message.getSentDate().getTime());
-						DbConnectionManager.setLargeTextField(pstmt, 7, message.getBody());
-						DbConnectionManager.setLargeTextField(pstmt, 8, message.getStanza());
+						pstmt.setInt(1, getArchivedMessageCount());
```
dwd: Wouldn't this need to be done in the same transaction?

PGoski:
I've fixed message id generation in 1f68e1b. Should work correctly now, because ArchivingTask is a singleton and run() method is synchronized -> can be executed once a time

PR #300, opened 2015-10-08
Add PluginClassLoader plugin loader isolation mechanism

wmz7year:
Different plug-ins should not be able to obtain other plug-ins through reflection resource.

	PluginClassloader for each plug-in classes and resources provides a local isolation mechanism
	that will safeguard their load classes and resources, when loading plugin all the resources
	and classes only in PluginClassloader , and does not exist in the parent Classloader.

	When the plugin needs to load a resource, it will first load plugins own resources in PluginClassloader,
	and secondly then required to load in the parent Classloader

### DIFF
--- a/src/java/org/jivesoftware/openfire/container/PluginClassLoader.java
+++ b/src/java/org/jivesoftware/openfire/container/PluginClassLoader.java
@@ -20,24 +20,38 @@
 
 package org.jivesoftware.openfire.container;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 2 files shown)


tevans:
I know it's a little late in the process (since this one is already merged), but I have some concerns with this PR (and others related to the PluginclassLoader). There are many existing plugins that rely upon certain dependencies between the core Openfire class loader and the plugin class loader(s). I would not be comfortable making this type of significant change without extensive testing of existing plugins to ensure they continue to operate correctly.

I will mention in the group chat as well.

PR #304, opened 2015-10-11
LDAP-Setup Starttls connection fix

sourceindex:
moin,

i reconfigured my openldap server to use state of the art cipher suites, protocols and also starttls insted of ssl connections.

after finishing that and changing all my applications i sadly found out that openfire was the last application in my environment that was missing the support for starttls. i found this thread and the following issue, looked at the codebase and got that daniel already submitted a patch to support starttls in the backend.

https://igniterealtime.org/issues/browse/OF-66

the missing part was just the frondend integration in setup routine. so i appended the properties with the missing parts, rebuild, setup ldap with starttls and found out that everything works as expected. so i did more testing and validated my connection setting by reviewing a tcpdump and tested more ldap queries and it still looks good to me.

it will be great if someone else can review and check if everything works for him as well, to see that feature to be fully integrate in the upcoming release. since i'm just able to jabber in english and german i filled up the other translation files with the english snippets and hope for the help of the community.

thanks for this great open-source project!

greetings,
Sebastian


wrooot:
I believe you don't have to add new i18n strings to every file (in English). If there is no such string in the corresponding language i18n file, it will use the English one for those strings automatically. At least Spark works that way.

speedy01:
It looks like it adds the option to the advance menu in the setup wizard.  PR is 3,241 additions and 3,215 deletions? might want to resubmit  openfire_i18n_en.properties to check the changes.

dwd:
This is a mess; I've rebased and created a new PR #414

PR #306, opened 2015-10-12
Made connection retry parameters configurable

AlphaGit:
Some cloud environments will enforce throttling parameters on database access (Microsoft Azure is a great example of that). As such, it is important that administrators are able to configure OpenFire's retry policy on the database login in order to prevent DoS'ing themselves from the database.

This small change does the following:

- introduces `database.defaultProvider.maxRetries` as the server property that will indicate how many retries can be used before giving up (default: 10 retries)
- introduces `database.defaultProvider.retryDelay` as the server property that will indicate how much OpenFire will wait until retrying the connection (default: 250 ms). (Note: should it be called "retryDelayMs" for better clarity? Seems weird. "retryDelayMsec"?)
- refactored some `if (condition) { statements }` into `if (!condition) return; statements;` to avoid nesting
- refactored out `ensureConnectionProvider()`, the first part that `getConnection()` did, just for clarity
- other minor adjustments (variable names, comments, indentation)

### DIFF
--- a/src/java/org/jivesoftware/database/DbConnectionManager.java
+++ b/src/java/org/jivesoftware/database/DbConnectionManager.java
@@ -49,7 +49,7 @@ import org.slf4j.LoggerFactory;
  */
 public class DbConnectionManager {
 
-	private static final Logger Log = LoggerFactory.getLogger(DbConnectionManager.class);
+    private static final Logger Log = LoggerFactory.getLogger(DbConnectionManager.class);
 
     private static ConnectionProvider connectionProvider;
     private static final Object providerLock = new Object();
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks for your contribution!

AlphaGit:
My pleasure. :)

guusdk:
Auch! There's an issue with this commit. Setting up a fresh Openfire instance now throws a StackOverFlow exception while setting up the database. This is called recursively:

        at org.jivesoftware.util.XMLProperties.parsePropertyName(XMLProperties.java:805)
        at org.jivesoftware.util.XMLProperties.getProperty(XMLProperties.java:185)
        at org.jivesoftware.util.XMLProperties.getProperty(XMLProperties.java:169)
        at org.jivesoftware.util.JiveGlobals.getXMLProperty(JiveGlobals.java:321)
        at org.jivesoftware.util.JiveGlobals.isSetupMode(JiveGlobals.java:970)
        at org.jivesoftware.util.JiveGlobals.getProperty(JiveGlobals.java:536)
        at org.jivesoftware.util.JiveGlobals.getIntProperty(JiveGlobals.java:578)
        at org.jivesoftware.database.DbConnectionManager.getConnection(DbConnectionManager.java:124)
        at org.jivesoftware.util.JiveGlobals.isSetupMode(JiveGlobals.java:981)
        at org.jivesoftware.util.JiveGlobals.getProperty(JiveGlobals.java:536)
        at org.jivesoftware.util.JiveGlobals.getIntProperty(JiveGlobals.java:578)
        at org.jivesoftware.database.DbConnectionManager.getConnection(DbConnectionManager.java:124)

The issue at hand: to set up the database, we (now) use a property. To read a property, we determine if we're in setup mode. To determine if we're in setup mode, we see if there's a database configured. Lather, rinse, repeat.

guusdk:
I've raised https://igniterealtime.org/issues/browse/OF-958 to track this issue.

AlphaGit:
Hi @guusdk! I just understood the problem. Yes, I see the problem introduced: we cannot use the database to read these configuration values until we actually have the database connections setup.

Would it be acceptable to move these configuration values to `JiveGlobals.getXMLProperty`? I wonder if there's a place where we can have a configuration that's available without DB or with DB if it happens to be open.

guusdk:
Retrieving configuration from a database that is used to connect to a database does not make sense. The XML route sounds more sensible.

PR #310, opened 2015-10-13
Handle group addition/removal per MUCRoom

sengels:
Since 3.10.0 OF-179 (https://igniterealtime.org/issues/browse/OF-179) has
been implemented. This patch makes it possible to use the restAPI for
that.


akrherz:
@sengels thanks for the PR.  Would this change also need to set plugin.xml `minServerVersion` to 3.10.0 as well?

sengels:
I don't really know enough about that code, so I would certainly do that... I'll update the PR in a minute...

Redor:
The PR looks good for me.

PR #311, opened 2015-10-16
OF-921: Reset MUC group cache for group changes

tevans:
Fixes caching issue which prevented users from joining MUC rooms in
certain cases.

### DIFF
--- a/src/java/org/jivesoftware/openfire/group/ConcurrentGroupList.java
+++ b/src/java/org/jivesoftware/openfire/group/ConcurrentGroupList.java
@@ -16,13 +16,14 @@ import org.xmpp.packet.JID;
  */
 public class ConcurrentGroupList<T> extends CopyOnWriteArrayList<T> implements GroupAwareList<T> {
 
-	private static final long serialVersionUID = -8884698048047935327L;
-	
+	private static final long serialVersionUID = 7735849143650412115L;
+
 	// This set is used to optimize group operations within this list.
 	// We only populate this set when it's needed to dereference the
 	// groups in the base list, but once it exists we keep it in sync
 	// via the various add/remove operations.
-	private transient Set<Group> groupsInList;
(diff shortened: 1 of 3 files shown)


sengels:
wait, please also see https://github.com/igniterealtime/Openfire/pull/315
Why do you add a second non-modifiable group cache?

tevans:
The original intent of the secondary cache was to reduce the overhead associated with resolving groups from the list/map contents. However, it is a fair question as to whether the associated maintenance of this additional cache (via GroupEventListener) is really worth this perceived benefit.

I will take a closer look at PR #315 before proceeding.

sengels:
First of all sorry, if my reply sounded a bit harsh; our setup currently uses LDAP to provide user and group information to openfire - this means you won't get any change trigger which could trigger your listeners... So the other solution is using a self-deprecating cache, which basically means you could use the group cache anyway... I especially like the idea about group-aware lists/maps though, so I tried to keep it as simple as possible.

tevans:
[FYI only - copied from PR #315 for context]

_I am comparing this approach with PR #311 and believe there might be a middle ground._

_In this case, every item in the list (or map) must be evaluated with every pass (or call to "includes") to determine whether it is a group. This can get pretty expensive for large groups consisting primarily of user JIDs. Unfortunately as we have seen, caching the groups themselves creates a secondary group cache that cannot be easily maintained._

_I propose that instead we continue to use the secondary cache to track the "known groups" within the list/map, reducing the number of times the non-group (or ambiguous) JIDs must be evaluated and de-referenced. However, the cache will retain only the group name corresponding to the resolved group, not the group instance itself. In this way the group can be fetched from the primary cache when needed to search for implicit JIDs._

I will submit the proposed additional changes here shortly.

sengels:
works as expected, thanks for your work.
Please, if possible, also let this patch go into the 3.10 branch.

PR #315, opened 2015-10-17
OF-921: Do not require restart to reload group memberships

sengels:
This commit removes an additional group cache that could not be updated
anymore at runtime. It lets ConcurrentGroup(List|Map) use the ordinary
cache from GroupManager instead.
This way LDAP Groups can be updated after startup by simply removing the
cache.

### DIFF
--- a/src/java/org/jivesoftware/openfire/group/ConcurrentGroupList.java
+++ b/src/java/org/jivesoftware/openfire/group/ConcurrentGroupList.java
@@ -18,12 +18,6 @@ public class ConcurrentGroupList<T> extends CopyOnWriteArrayList<T> implements G
 
 	private static final long serialVersionUID = -8884698048047935327L;
 	
-	// This set is used to optimize group operations within this list.
-	// We only populate this set when it's needed to dereference the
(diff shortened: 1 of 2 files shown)


sengels:
If you accept this patch, would it be possible to merge this patch into the 3.10 branch too?

tevans:
I am comparing this approach with PR #311 and believe there might be a middle ground. 

In this case, every item in the list (or map) must be evaluated with every pass (or call to "includes") to determine whether it is a group. This can get pretty expensive for large groups consisting primarily of user JIDs. Unfortunately as we have seen, caching the groups themselves creates a secondary group cache that cannot be easily maintained.

I propose that instead we continue to use the secondary cache to track the "known groups" within the list/map, reducing the number of times the non-group (or ambiguous) JIDs must be evaluated and de-referenced. However, the cache will retain only the group name corresponding to the resolved group, not the group instance itself. In this way the group can be fetched from the primary cache when needed to search for implicit JIDs.

I will make these changes under PR #311 and welcome further review and comments there.

sengels:
sounds like a plan, thanks...

PR #316, opened 2015-10-19
Adjust texts for Openfire Meetings vs Jitsi Meet

marclaporte:


guusdk:
Installing our own instead of Jitsi's makes sense. Thanks!

PR #317, opened 2015-10-19
Adding some inline user documentation

marclaporte:
Todo: user should be able to upload PDF presentation


guusdk:
That looks innocent enough. Thanks for the contribution!

PR #321, opened 2015-10-22
[OF-958] Avoiding DB properties when reading DB connection configuration

AlphaGit:
This should fix the problem introduced in #306.

### DIFF
--- a/src/java/org/jivesoftware/database/DbConnectionManager.java
+++ b/src/java/org/jivesoftware/database/DbConnectionManager.java
@@ -121,8 +121,8 @@ public class DbConnectionManager {
         ensureConnectionProvider();
 
         Integer currentRetryCount = 0;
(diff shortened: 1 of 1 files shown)


AlphaGit:
That's a very good point. Would it make sense for it to be at the `database` level?

```xml
<database>
  <maxRetries>2</maxRetries>
  <retryDelay>200</retryDelay>
  <defaultProvider>
    <!-- more stuff in here -->  
  </defaultProvider>
</database>
```

Would you prefer to have another particular section created for it? I'm open to suggestions.

guusdk:
pushing it one level down should work. Not sure if there's a "better" place, but this sounds good enough. If you push the change (and do a quick test, you never know), I'll merge it.

AlphaGit:
I hope it's not too much to ask, but would you mind to test it out for me? I happen to have a messed up environment right now and I won't be able to test in a while.

guusdk:
No problem. I just did the same test, everything appears to be fine. I'll merge the PR.

PR #333, opened 2015-10-28
Prefer StringBuilder.append(char) over append(String).

sco0ter:
This is the more correct API and also has very minor performance benefits.


Redor:
+1

PR #337, opened 2015-10-29
3.10

sujithmenonp:


Redor:
What should that be?

sco0ter:
I don't know either...

lkraav:
It looks like a bad merge result.

dwd:
I think this is in error.

PR #339, opened 2015-10-29
Clustering setup admin page support select cluster implement plugin

wmz7year:
Server Clustering setup support select the cluster support plugin option.
 If plugin has CacheFactory.clusteredCacheFactoryClass implement then plugin will list in the option.


wmz7year:
demo 
Like this

src/java/org/jivesoftware/util/cache/CacheFactory.java
```diff
@@ -623,38 +611,72 @@ public static synchronized void initialize() throws InitializationException {
...
+				// ignore
+			}
+		}
+    	return result;
+    }
+    
+    private static boolean createClusteredCacheFactoryStrategy() {
+    	try {
+    		PluginManager pluginManager = XMPPServer.getInstance().getPluginManager();
+    		String clusterPluginName = JiveGlobals.getProperty("cluster.plugin.name");
```
tevans: Recommend providing a default here ("Hazelcast Clustering Plugin"), but note that plugin names may also be localized.

src/java/org/jivesoftware/util/cache/CacheFactory.java
```diff
@@ -623,38 +611,72 @@ public static synchronized void initialize() throws InitializationException {
...
+			}
+    		
+    		if (plugin == null) {
+    			return false;
+    		}
+    		
+    		PluginClassLoader pluginClassloader = pluginManager.getPluginClassloader(plugin);
+    		
+    		clusteredCacheFactoryStrategy = (CacheFactoryStrategy) pluginClassloader
+    				.loadClass(clusteredCacheFactoryClass).newInstance();
```
tevans: I would also recommend that the clusteredCacheFactoryClass be resolved via local scope at runtime, rather than defined as a static variable. The following can be removed from static scope and used here instead:
```
            String clusteredCacheFactoryClass = JiveGlobals.getProperty(CLUSTERED_CACHE_PROPERTY_NAME, 
                    "org.jivesoftware.openfire.plugin.util.cache.ClusteredCacheFactory");
```
A corresponding change can be made for the localCacheFactoryClass in the initialize() method.

wmz7year: I'm create getClusteredCacheFactoryClass and getLocalCacheFactoryClass method to dynamic get clusteredCacheFactoryClass  and localCacheFactoryClass.

src/i18n/openfire_i18n_en.properties
```diff
@@ -3153,6 +3153,7 @@ system.clustering.enabled=Clustering was enabled successfully.
 system.clustering.disabled=Clustering was disabled successfully.
 system.clustering.failed-start=Failed to start or join an existing cluster. Check the error log for more information.
 system.clustering.enabled.legend=Clustering Enabled
+system.clustering.select-support-plugin=Select the cluster supports plugin,currently installed plugin will appear in the cluster support in selection.
```
tevans: For an English locale, perhaps something like this:
```
system.clustering.select-support-plugin=Select an item from the list of installed plugins that are able to provide clustering support.
```

wmz7year:
Hi @tevans , if the PluginClassloader not guaranteed isolation between plugins,this change will not work.

Cause if one plugin has ClusteredCacheFactory then all plugin's PluginClassloader can  be load ClusteredCacheFactory .

PR #341, opened 2015-10-29
Remove support for Legacy date/time (XEP-0090 and XEP-0091).

sco0ter:
These specifications are deprecated long enough (since 2007) in favor of XEP-202/203.


dwd:
Not sure if we can remove these in all cases, yet. I've asked on jdev@ for opinions.

PR #344, opened 2015-10-30
Replace deprecated org.jivesoftware.util.Log with org.slf4j.Logger

sco0ter:
Most methods in org.jivesoftware.util.Log are deprecated, stating that org.slf4j.Logger should be used instead.

This PR uses org.slf4j.Logger in Openfire core now.


Redor:
+1

PR #346, opened 2015-10-30
Use StandardCharsets.UTF_8 with Reader/Writer classes.

sco0ter:
This is a followup for fcf665a.

### DIFF
--- a/src/java/org/jivesoftware/openfire/net/SocketConnection.java
+++ b/src/java/org/jivesoftware/openfire/net/SocketConnection.java
@@ -27,6 +27,7 @@ import java.io.Writer;
 import java.net.Socket;
 import java.net.UnknownHostException;
 import java.nio.channels.Channels;
+import java.nio.charset.StandardCharsets;
 import java.security.cert.Certificate;
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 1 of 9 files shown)


src/java/org/jivesoftware/openfire/net/SocketConnection.java
```diff
@@ -145,11 +141,11 @@ public SocketConnection(PacketDeliverer backupDeliverer, Socket socket, boolean
         // DANIELE: Modify socket to use channel
         if (socket.getChannel() != null) {
             writer = Channels.newWriter(
-                    ServerTrafficCounter.wrapWritableChannel(socket.getChannel()), CHARSET);
+                    ServerTrafficCounter.wrapWritableChannel(socket.getChannel()), StandardCharsets.UTF_8.newEncoder(), -1);
```
Redor: Is that right?
Should it not be just: StandardCharsets.UTF_8

sco0ter: Yes, it is. There's no Channels.newWriter(Channel, Charset). The -1 is correct as well (taken from Channels.newWriter(Channel, String).
Thanks for reviewing!

PR #349, opened 2015-10-30
Modernize PluginManager internals with Java NIO.2 File API.

sco0ter:
NIO.2 is the more modern and faster API and smooths the way for further improvements, e.g. using java.nio.file.WatchService to watch the plugin directory for changes.

This is no critical PR, but nonetheless I think it's a good idea to modernize some code.

Please review carefully, and preferably do some tests before merging, because PluginManager is pretty important!


tevans:
Looks good to me. I've got a few custom plugins I can test this with, will merge and verify.

PR #351, opened 2015-10-30
Revert changes introduced via PR #300

tevans:
Restores earlier revision of the PluginClassLoader. 
Also, a minor fix to avoid using classes that are not part of the official JRE. This additional change was required to ensure a clean build.


wmz7year:
I'm sorry for this.But my PluginClassloader idea is right? I'm testing all plugin and find issue.Then I'll create new pr.

wmz7year:
I know the why some plugins not work. 
My PR put up isolation between plugins, including openfire.
When the plugin active use Openfire plugin resource,When you need to call Openfire not visible resulting in can not find, I will solve this problem in a new PR

deleolajide:
This PR is responsible for multiple

java.lang.LinkageError: loader constraint violation: when resolving method "xxxxxxxxxxxx"

in the ofmeet plugin

tevans:
Looks like I hosed this one up attempting to rebase. Will retry under separate cover.

PR #352, opened 2015-10-30
OF-938: Apply packet namespace via string insert

tevans:
This patch fixes a problem introduced by an earlier patch whereby a
change to the Packet namespace inadvertently cascaded to child XML
elements.


tevans:
@sco0ter Thanks for the feedback ... in this case, the desired result of the operation is actually a string rather than an element. It ought to be more efficient to manipulate the string directly.

akrherz:
This patch was tested on Ignite and appears to be working, thanks @tevans

PR #355, opened 2015-11-02
	Fix PluginClassloader not implement getResources Method

wmz7year:
Now Hazelcast plugin work fine.


wmz7year:
@akrherz @tevans  I have tested all plugins.
Just justmarried and ofmeet plugin not fine.
1. justmarried has own log4j.jar.
2. ofmeet has own slf4j.jar.
3. ofmeet System.getProperty("whack.componentManagerClass"); return null  throw NullPointerException.

Other plugins are work fine!

deleolajide:
ComponentManager componentManager = ComponentManagerFactory.getComponentManager();

is no longer working in any of my plugins even after applying this PR

PR #356, opened 2015-11-02
Update MessageRouter.java

ishmakov:
Route message to multiple target users resource if not set resource.
This fix problem with invite message to MUC room for multiple target resource


tevans:
Agree this type of global routing change is not desirable. Perhaps this would be better solved via an explicit list of invitation JIDs generated in the MUC component.

Feel free to open a thread in the community site for further discussion and feedback.

PR #357, opened 2015-11-02
	Fix hazelcast plugin xalan NoClassDefFoundError

wmz7year:
issue:https://github.com/hazelcast/hazelcast/issues/6614


wmz7year:
@akrherz I have already fix code style.

akrherz:
Great, thank you.

tevans:
Thanks @wmz7year for your ongoing contributions! Definitely want to encourage you to continue supporting Openfire.

This proposed patch appears to be a temporary workaround that has been coded into the plugin. The HZ team indicated that this issue will be resolved in their next release (3.6). I would prefer to wait for that fix rather than apply this patch. My concerns are twofold:

1) The package for the TransformerFactory is in a private namespace (com.sun.*)
2) There may be other components within Openfire that depend upon the default setting

Per the HZ team, for a temporary fix you can set the corresponding system property via the command line (using the -D option): 
```
System.setProperty("javax.xml.transform.TransformerFactory",
                "com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl");
```
When HZ 3.6 goes final (currently EA) we will update the Hazelcast plugin to pick up the upstream fix.

PR #358, opened 2015-11-02
OF-868 update roster cache on userModified

akrherz:
patch from Ignite Community user andrelab

### DIFF
--- a/src/java/org/jivesoftware/openfire/roster/RosterManager.java
+++ b/src/java/org/jivesoftware/openfire/roster/RosterManager.java
@@ -592,7 +592,18 @@ public class RosterManager extends BasicModule implements GroupEventListener, Us
     }
 
     public void userModified(User user, Map<String,Object> params) {
-        //Do nothing
+        if ("nameModified".equals(params.get("type"))) {
+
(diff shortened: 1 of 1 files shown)


src/java/org/jivesoftware/openfire/roster/RosterManager.java
```diff
@@ -592,7 +592,21 @@ public void userDeleting(User user, Map<String,Object> params) {
...
+        if ("nameModified".equals(params.get("type"))) {
+
+            for (Group group : getSharedGroups(user.getUsername())) {
+                ArrayList<JID> groupUsers = new ArrayList<>();
+                groupUsers.addAll(group.getAdmins());
+                groupUsers.addAll(group.getMembers());
+
+                for (JID groupUser : groupUsers) {
+                    String groupUsername = groupUser.getNode();
+                    synchronized ((groupUsername + " ro").intern()) {
```
sco0ter: I don't think this is required. Cache operations are thread-safe (as per JavaDoc).

akrherz: OK thanks, I have force pushed an updated changed with that removed.

PR #361, opened 2015-11-03
Add Legacy Jive build code for eventual removal

akrherz:
The intention here is to eventually remove this code with updated code that does not depend on qdox.


tevans:
Note that the binary versions of these sources are in this jar file:
https://github.com/igniterealtime/Openfire/blob/master/build/lib/ant/ant-jive-edition.jar
Presumably there are no licensing issues with these com.jivesoftware.* sources? 
Do they need any sort of preamble before committing them to Github?

akrherz:
I have email correspondence that I think covers our usage, we'd want to delete these after the maven migration anyway

PR #363, opened 2015-11-06
RestAPI Plugin - Provide the possibility to use Custom Auth Filter

totalscoccia:
Right now, RestAPI plugin provides two option for accessing its
endpoints:
1) Basic authentication: uses username and password of an existing
Openfire admin accoount
2) Secret key: uses for authentication a key defined in the Openfire
Admin console.

It may be really useful to provide the option to define another Filter
as the authentication mechanism to use for accessing the plugin' Rest
service given its fully qualified class name.


src/plugins/restAPI/src/java/org/jivesoftware/openfire/plugin/rest/service/JerseyWrapper.java
```diff
@@ -1,120 +1,141 @@
...
-	private static final String SERVLET_URL = "restapi/*";
-
-	/** The config. */
-	private static Map<String, Object> config;
-
-	/** The prc. */
-	private static PackagesResourceConfig prc;
-
-	/** The Constant JERSEY_LOGGER. */
-	private final static Logger JERSEY_LOGGER = Logger.getLogger("com.sun.jersey");
```
Redor: It is required to avoid Logging on startup

src/plugins/restAPI/src/java/org/jivesoftware/openfire/plugin/rest/service/JerseyWrapper.java
```diff
@@ -1,120 +1,141 @@
...
-
-	static {
-		JERSEY_LOGGER.setLevel(Level.SEVERE);
```
Redor: It is required to avoid logging on startup

Redor:
Could you please add the JERSEY_LOGGER back?

totalscoccia:
I'll do it as soon as I have a spare minute on Monday or Tuesday

totalscoccia:
Now code should be completed. Status messages are shown in the restAPI plugin page, instead of using the logging mechanism as I've planned to use in my first commit

Redor:
Great work. Thy !

marcorm:
great

PR #367, opened 2015-11-08
OF-967 add option to use Name as nickname for muc bookmark

akrherz:
Untested by me.


wrooot:
I have tested only the resulting jar provided by the patch author. Looks very simple.

PR #368, opened 2015-11-10
Tests should retrieve resources from the classpath rather than files.

sco0ter:
These tests otherwise may throw FileNotFoundException and are harder to make them run.

Also delete TestUtils.java because it's not needed now anymore.


akrherz:
Filed https://igniterealtime.org/issues/browse/OF-973

PR #371, opened 2015-11-13
Various improvements in Version class

sco0ter:
- Make it final and immutable.
- Use pattern-based parsing, which allows to also parse the ReleaseStatus and the statusVersion.
- Implement toString() method.
- Implement equals() and hashCode(), which is a strong recommendation when implementing Comparable: From the JavaDoc: "It is strongly recommended that (x.compareTo(y)==0) == (x.equals(y))", which was not the case.
- Allow CharSequence
- Simplify getVersionString() method (less duplicate code).
- Improve compareTo() method by properly comparing more than 2 digits and also comparing ReleaseStatus and statusVersion

### DIFF
--- a/src/java/org/jivesoftware/util/Version.java
+++ b/src/java/org/jivesoftware/util/Version.java
@@ -19,39 +19,43 @@
 
 package org.jivesoftware.util;
 
-import java.util.StringTokenizer;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Holds version information for Openfire.
  *
  * @author Iain Shigeoka
  */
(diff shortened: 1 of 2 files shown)


tevans:
Looks good to me. Nice to have some unit tests for a change!

@sco0ter FYI I seem to recall there is a corresponding JSP or class somewhere in the Ignite community website also, although I forget exactly where. I believe it is called by the admin console to determine whether there is a newer version of Openfire available for download. Maybe @akrherz can remember if this component still exists?

tevans:
See more information here: [OF-796|https://igniterealtime.org/issues/browse/OF-796] including a reference to the corresponding website JSP.

PR #374, opened 2015-11-13
OF-881: Simplify connection close semantics

tevans:
Avoid recursion and synchronization issues

### DIFF
--- a/src/java/org/jivesoftware/openfire/Connection.java
+++ b/src/java/org/jivesoftware/openfire/Connection.java
@@ -144,28 +144,8 @@ public interface Connection {
      *      <li>Call notifyEvent all listeners that the channel is shutting down.
      *      <li>Close the socket.
      * </ul>
-     *
-     * An invocation of this method is equal to invoking {@link #close(boolean)} with a parameter
-     * that is false.
      */
     public void close();
-
-    /**
-     * Close this session including associated socket connection. The order of
-     * events for closing the session is:
-     * <ul>
-     *      <li>Set closing flag to prevent redundant shutdowns.
-     *      <li>Call notifyEvent all listeners that the channel is shutting down.
-     *      <li>Close the socket.
-     * </ul>
-     *
-     * This method takes into account the connection state of the peer. Specifically,
-     * when the peer is known to be in a disconnected state, no data will be sent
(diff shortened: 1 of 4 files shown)


tevans:
Would like a review of this from @guusdk @sco0ter @dwd @akrherz before merge please.

dwd:
Well... It *looks* right, but we went down a long path to get to the code @guus wrote that you're now removing. I wonder if later changes have removed the need for that as well?

Also, does it work (ie, have yo run a non-trivial deployment of this)?

sco0ter:
Some random thoughts on this:
- I wonder if we could use a `AtomicReference<State> state` to get rid of the synchronized blocks, something like:
```java
                if (!state.compareAndSet(State.RUNNING, State.CLOSING)) {
                    return;
                }
```
- Having both `Session.STATUS_CLOSED` and `State.CLOSED` is confusing, isn't it?
- We should consider implementing `java.lang.AutoCloseable`... just because.
- The implemented behavior is called `idempotent` (closing a closed resource has no effect). Maybe it should be clear in the JavaDoc.
- It looks well otherwise. I think it works, `close()` should only run once, therfore no recursion, no deadlocks.

guusdk:
Reduced complexity is good. I cannot see anything wrong with the change either (although I did not run tests - do we know how to get in touch with Andi? He used to be able to shake things down thoroughly). As for my original fix: It was the result of a trial-and-error session more than a solution that was well designed and executed. A good part of the problems that my changes tried to fix related to the issues that we had after upgrading MINA - which has since been rolled back.

sco0ter:
@fowlerp I think you are right. `state = State.CLOSED;` is called outside the synchronization block as well. Read/Write access to class variables should always be synchronized on the *same* monitor, otherwise thread visibility issues may occur. Although it's surely hard to reproduce and proof here.
Alternatively the mentioned `AtomicReference` would help as well.

tevans:
Ok - thanks all. I think we are agreed on the basic concept here, but I will make another update to incorporate some of this helpful feedback.

tevans:
Review feedback applied. Note that we have done some initial testing to validate basic connection cases, but further testing (especially under load/concurrency scenarios) would be appreciated.

PR #375, opened 2015-11-14
ofmeet plugin version 0.2.3

deleolajide:
Simplified build, removed classes folder in WEB-INF
Changed default value for voip gateway to false
Removed some redundant and duplicate jar file in jitsivideobridge lib
Removed tinder.jar and whack.jar which caused ComponentManagerFactory.getComponentManager() to fail with OF 3.11


deleolajide:
This has been tested against OF 3.10.2 on windows and linux and works fine. However,recent PRs that changed behavior of Plugin Manager have caused it to fail with OF 3.11 Alpha. When those PRs are sorted out, I revisit it.

PR #381, opened 2015-11-17
StringUtils: Simplify verbose/duplicate and obsolete code constructs

sco0ter:
- The 4 replace methods were nearly identical.
- Deprecate String utility, which has a simple one-liner equivalent in the JDK
- Use DatatypeConverter for byte[] to hex. It's less code and even 2-3 times faster than the previous method.


tevans:
Looks good to me.

PR #383, opened 2015-11-17
Modernize XMLProperties with Java NIO.2 File API

sco0ter:


Redor:
Would it also increase the Performance?

sco0ter:
Generally the new API is more performant, but I doubt it would make significant difference in this case with just three property files.

PR #384, opened 2015-11-17
byte[].equals() should probably be Arrays.equals()

sco0ter:
This is highlighted as "probably a bug" in my IDE.

E.g. `new byte[]{1}.equals(new byte[]{1})` is always false.
But `Arrays.equals(new byte[]{1}, new byte[]{1});` is true.


tevans:
+1

PR #387, opened 2015-11-18
OF-974: Migrate cache content

tevans:
When switch to/from clustered deployment, ensure that critical cached
items such as components and remote servers are not inadvertently
dropped from the cache.


tevans:
Incorporated review feedback.

PR #389, opened 2015-11-18
Generify ClusterTask interface

sco0ter:
in the same fashion as e.g. `java.util.concurrent.Callable`

Please note: I haven't tested its runtime behavior, but this change felt like a natural fit, especially due to the `CallableTask` class.

@tevans: Can you please review and maybe even test it or otherwise comment on it before merge?


tevans:
I like this approach - the expected task result is now immediately visible. Let's go ahead and merge it in. I will also try to run it through a quick test cycle with the HZ plugin.

PR #390, opened 2015-11-19
JDBCAuthProvider: adding support for bcrypt and more

trumpetx:
JDBCAuthProvider: adding support for bcrypt, direct hashes, chained hashes, adding tests


trumpetx:
I realize you guys like 1 change per pull request; however, these changes are all related to each other and it would be weird to pull them apart, but I can do it if that's the only way you will accept them.

src/java/org/jivesoftware/openfire/auth/JDBCAuthProvider.java
```diff
@@ -53,8 +63,35 @@
...
+ *
+ * <p>External systems integrating with Openfire may only have access to a hashed password.  In this scenario, setting 
+ * jdbcAuthProvider.acceptPreHashedPassword = true will allow this AuthProvider to directly compare the input to 
+ * the hash stored in the database.  This configuration is the rough equivalent to allowing the scenario where EITHER 
+ * of the following authentication situations are allowed at the same time:<br>
+ * jdbcAuthProvider.passwordSQL = SELECT MD5(password) FROM user_account WHERE username=?<br>
+ * jdbcAuthProvider.passwordType = plain<br>
+ * -OR-<br>
+ * jdbcAuthProvider.passwordSQL = SELECT password FROM user_account WHERE username=?<br>
+ * jdbcAuthProvider.passwordType = md5<br></p>
```
trumpetx: Explanation of the new acceptPreHashedPassword configuration

src/java/org/jivesoftware/openfire/auth/JDBCAuthProvider.java
```diff
@@ -53,8 +63,35 @@
...
+ * <p>Example Setting: "md5,sha1"<br>  
+ * Usage: password -><br>
+ * (md5)&nbsp;286755fad04869ca523320acce0dc6a4&nbsp;-><br>
+ * (sha1)&nbsp;0524b1fc84d315b08db890413e65260040b08caa&nbsp;-></p>
+ * 
+ * <p>Bcrypt is supported as a passwordType; however, when chaining password types it MUST be the last type given. (bcrypt hashes are different 
+ * every time they are generated)</p>
+ * <p>Optional bcrypt configuration:</p>
+ * <ul>
+ * <li><b>jdbcAuthProvider.bcrypt.cost</b>: The BCrypt cost.  Default: BCrypt.GENSALT_DEFAULT_LOG2_ROUNDS  (currently: 10)</li>
```
trumpetx: Explanation of bcrypt configuration

akrherz:
How is the `.gitignore` change related?  Yeah, it is no big deal to do it.

src/java/org/jivesoftware/openfire/auth/JDBCAuthProvider.java
```diff
@@ -53,8 +63,35 @@
...
+ * jdbcAuthProvider.passwordSQL = SELECT password FROM user_account WHERE username=?<br>
+ * jdbcAuthProvider.passwordType = md5<br></p>
+ * 
+ * <p>jdbcAuthProvider.passwordType can accept a comma separated string of password types.  This can be useful in 
+ * situations where legacy (ex/md5) password hashes were stored and then "upgraded" to a stronger hash algorithm.
+ * Hashes are executed left to right.</p>
+ * <p>Example Setting: "md5,sha1"<br>  
+ * Usage: password -><br>
+ * (md5)&nbsp;286755fad04869ca523320acce0dc6a4&nbsp;-><br>
+ * (sha1)&nbsp;0524b1fc84d315b08db890413e65260040b08caa&nbsp;-></p>
```
trumpetx: Explanation of passwordType configuration (new, allowing comma separated - backwards compatible)

src/test/java/org/jivesoftware/openfire/auth/JDBCAuthProviderTest.java
```diff
@@ -0,0 +1,79 @@
+package org.jivesoftware.openfire.auth;
+
+import java.util.HashMap;
+import org.junit.Test;
+import static org.junit.Assert.*;
+import org.mindrot.jbcrypt.BCrypt;
+
+public class JDBCAuthProviderTest {
```
trumpetx: I was unable to write as many tests as I would like because I had a difficult time bootstrapping the JiveGlobals stuff.  I also didn't feel like making changes to include PowerMock.

trumpetx:
@akrherz - I'm on the road for a few minutes, but when I get into the office I'll make a post on the dev forum detailing this PR per the workflow.

trumpetx:
Post created:

https://community.igniterealtime.org/thread/57551

I've also tested this on my dev system against 3.10 codebase (minimal code changes from there to 3.11).

akrherz:
thanks, filed as https://igniterealtime.org/issues/browse/OF-975

src/java/org/mindrot/jbcrypt/BCrypt.java
```diff
@@ -0,0 +1,777 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+// WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+// ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+// WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+// ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+// OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+package org.mindrot.jbcrypt;
```
tevans: If we update to the latest bouncybastle (BC), we could use either of their implementations of this bcrypt generator. I believe we build and distribute with 1.51 currently, and the latest is 1.53. The bcrypt implementation was added in 1.52. I would prefer to reuse the existing library implementation rather than adding this class, if possible.

trumpetx: @tevans I agree completely.  I didn't know that BC provided a BCrypt implementation.  I'll take a peek and update this PR accordingly.

tevans:
Thanks for the helpful documentation and your community post. Please see my earlier comment about reusing the BC implementation, but overall I think this is a fine approach.

src/java/org/mindrot/jbcrypt/BCrypt.java
sco0ter: Alternatively it seems that this class is in:
http://mvnrepository.com/artifact/org.mindrot/jbcrypt/0.3m

dwd:
I've commented in the forums, but basically the "use the hash as a password" bit scares me a lot. The rest looks OK.

trumpetx:
@dwd I'd be concerned with the non-offcial release from "some random guy at github" (though, very minor concerns).  But then again, why trust me to copy it into the project!  I think it's probably best to trust BCrypt as a provider (since you're already trusting them with all the other crypto stuff).  What is submitted here is char for char directly from the jbcrypt distribution: http://www.mindrot.org/projects/jBCrypt/.

trumpetx:
@tevans - I bumped up the BC jars and don't see equivalent methods:gensalt / chekpw in BC's BCrypt implementation.  Are you aware of utility methods that mock the jbcrypt BCrypt API, but use BC instead?  I'd hesitate to write my own since I'm sure someone, somewhere has done this already (and probably better than I would whip up).

tevans:
Please also update build/lib/versions.txt for the new BC version.

tevans:
I think this class has the methods you will need to generate a bcrypt hash and to check a password against a hash: org.bouncycastle.crypto.generators.OpenBSDBCrypt

Ideally you would not need to include the class copied from the mindrot package.

trumpetx:
@tevans Done re: Bouncycastle > jBCrypt

dwd:
I think this PR still includes the hash-as-password solution; removing that and I think the rest can be committed. (Or tell me it's already gone).

trumpetx:
@dwd it does.  I was looking into the alternative that you were discussing on the chat thread before removing it.

In short, if there's a viable Openfire configurable alternative to my problem, I'll gladly use it, but if not I'd argue that the hash comparison solution is just a configuration to do what is _already_ configurable via the existing implementation.  It's just adding the ability to compare hashes to hashes OR passwords to hashes.

If I take more than a day or so I'll pull that specific change and resubmit it if I feel like arguing the point any further.

trumpetx:
@dwd @tevans - the offending feature has been removed.  This should be ready for a merge.

trumpetx:
/bump - this is ready.

I looked at changing the label to bring visibility, but I don't think I have permissions.

dwd:
Thanks, totally missed this.

tevans:
+1 - thanks for hanging in there with us on this one, @trumpetx

PR #391, opened 2015-11-19
Correct version status comparison

tevans:
bugfix for recent Version refactor


tevans:
Bad PR ... will close.

PR #395, opened 2015-11-20
Update web.xml schema namespaces to version 3.1 (Java EE 7)

sco0ter:
Old version 2.4 was J2EE 1.4

Also remove <taglib> from web.xml. This was required in JSP 1.2, but is no longer in JSP 2.0 (Servlet API 2.4) and gives warnings since it has been removed from the XSD.
See http://wiki.metawerx.net/wiki/RemovingTaglibFromWeb.xml


Redor:
+1

PR #397, opened 2015-11-21
Update admin.tld to use the latest taglib version.

sco0ter:
This is related to the JSP upgrade.

Some tags have changed after moving from the DTD to the XSD.
See also https://blogs.oracle.com/jluehe/entry/validate_your_tag_library_descriptor


dwd:
+1

tevans:
+1

PR #398, opened 2015-11-21
Delete unused classes

sco0ter:
These classes are unused, can we delete them to get a cleaner code base?


dwd:
Sounds reasonable to me.

tevans:
FYI - I believe @guusdk was recently doing some work around SSLJiveTrustManager.java. Otherwise +1 from me.

guusdk:
Although I'm deleting stuff myself, shouldn't we mark things as "deprecated" instead? That will give third party developers some leeway

dwd:
I'm thinking these are things that if plugin developers are using, they're probably doing something wrong if Openfire isn't using them at all. So I'm merging and seeing what breaks...

PR #399, opened 2015-11-21
Simplify/improve NIOConnection.deliver() by removing ByteBufferWriter.

sco0ter:
For every packet being delivered the `ByteBufferWriter` creates a new String for every few chars written and puts it to the buffer.

According to my tests it is more efficient to create only one string and put it into the buffer once.
(10-20% better performance)

This is the test I used:
```java
        Message message = new Message();
        message.setType(Message.Type.chat);
        message.setBody("Tests");
        Forwarded forwarded = new Forwarded(message);
        Sent packet = new Sent(forwarded);

        long now = System.currentTimeMillis();
        CharsetEncoder encoder = StandardCharsets.UTF_8.newEncoder();
        for (int i = 0; i < 100000; i++) {
            IoBuffer buffer = IoBuffer.allocate(1024);

            // Old method:
            //XMLWriter xmlSerializer = new XMLWriter(new ByteBufferWriter(buffer, encoder), new OutputFormat());
            //xmlSerializer.write(packet.getElement());
            //xmlSerializer.flush();

            // New method:
            buffer.putString(packet.getElement().asXML(), encoder);
        }
        System.out.println(System.currentTimeMillis() - now);
```


Redor:
+1

PR #400, opened 2015-11-21
OF-976 Language is not properly set in HttpSession

sco0ter:


dwd:
I *think* this is effectively a no-op; the XML namespace is bound to the prefix "xml" by standards action and cannot be rebound. I'm happy for this to go in, though:

+1

(Just not sure it makes any difference).

sco0ter:
´rootNode.attributeValue("xml:lang")´ returned null, even if this attribute was set.

dwd:
Ah, so I'd misunderstood. Thanks for the clarification.

tevans:
+1

PR #401, opened 2015-11-21
WebSocket plugin: "lang" attribute should be "xml:lang"

sco0ter:
Also be more correct when reading the "xml:lang" attribute and remove superfluous whitespaces.

Fixes OF-977


dwd:
Looks good to me.

tevans:
+1

dwd:
Merged manually and pushed as 43e3c5f5be7d26d8906613762bb9ebc923d742ae

PR #403, opened 2015-11-21
Language info should be in the Session, not in the Connection

sco0ter:
I feel the language (the stream's "xml:lang" attribute) should be in the Session interface rather than the Connection (similar as the stream id), because it's unrelated to the session's transport.

The HttpSession already did that.

Otherwise the used language is not easy to retrieve, e.g. when using SessionManager to get the session and OF-208 is hard to solve.

Alternatively we can expose the Connection in the Session interface, but this approach feels wrong to me.

Also use `java.util.Locale` rather than String, because it better reflects what can be in `xml:lang` (it can also have country codes, e.g. "de-CH")

### DIFF
--- a/src/java/org/jivesoftware/openfire/session/Session.java
+++ b/src/java/org/jivesoftware/openfire/session/Session.java
@@ -215,10 +215,9 @@ public interface Session extends RoutableChannelHandler {
     public String getCipherSuiteName();
 
     /**
-     * Returns the language code that should be used for this connection
-     * (e.g. "en").
+     * Returns the locale that is used for this session (e.g. {@link Locale#ENGLISH}).
      *
-     * @return the language code for the connection.
+     * @return The language for the session.
      */
     Locale getLanguage();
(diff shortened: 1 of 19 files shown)


dwd:
Looks good to me.

src/java/org/jivesoftware/openfire/session/Session.java
```diff
@@ -212,4 +213,12 @@
      * @return cipher suite name.
      */
     public String getCipherSuiteName();
+
+    /**
+     * Returns the language code that should be used for this connection
```
tevans: Might prefer "session" here rather than "connection" (also below), but not a blocker.

tevans:
+1 - thanks for including the plugins :)

sco0ter:
Thanks, I've fixed the JavaDoc and also increased the WebSocket plugin minVersion because it requires the new `createClientSession(Connection, Locale);` API.

PR #404, opened 2015-11-23
OF-115 Process S2S stanzas from a session serially

dwd:
This patch removes the thread pool from the ServerSocketReader, causing
stanzas from a particular S2S session to be processed serially on the same
thread used by the ServerSocketReader.

Each ServerSocketReader currently has a thread to itself, so this is unlikely
to have a serious impact on performance, but does have a very positive impact
on correctness.


dwd:
Backported fix.

tevans:
+1 ... are we planning to release a 3.10.4 at some point?

PR #405, opened 2015-11-23
OF-81: Change default BOSH port to 5280

sco0ter:
As per the IANA Port Numbers Registry.

Fixes issue OF-81

### DIFF
--- a/src/java/org/jivesoftware/openfire/http/HttpBindManager.java
+++ b/src/java/org/jivesoftware/openfire/http/HttpBindManager.java
@@ -84,7 +84,7 @@ public final class HttpBindManager {
 
     public static final String HTTP_BIND_PORT = "httpbind.port.plain";
 
-    public static final int HTTP_BIND_PORT_DEFAULT = 7070;
+    public static final int HTTP_BIND_PORT_DEFAULT = 5280;
 
     public static final String HTTP_BIND_SECURE_PORT = "httpbind.port.secure";
 


dwd:
What would this do for installations already deploying BOSH in the default port? I really don't want to break existing deployments in a weird way (although possibly 4.0 is a good time to do this).

tevans:
Agree we should give some consideration to the legacy deployments, perhaps setting the IANA default for new installations only (via setup)? Also, what are we doing about the SSL-enabled BOSH port (currently 7443)?

sco0ter:
I didn't test this with old installations, but I think it doesn't harm. We have set the port to 80 since at least version 3.8.2 and I can't remember somebody complained about the port being resetted to 7070 after installing a new Openfire version. But maybe this behaves different if the port hasn't been changed at all.

The issue was scheduled for 4.0.

guusdk:
I agree with Tom and dwd here. To me, it's more important to not break existing deployments than conforming to IANA. 

I like Toms reasoning of applying this as part of the setup, although I'd like to find a way to turn that around: upgraded installations should re-configure the new default back to 7070 (instead of having new installations use a different setting than the old 7070 default). 

Perhaps we could employ the incremental database upgrade scripts for this. Those are only executed when there is an update (fresh installations use the 'bulk' script instead). This arguably is not the nicest way of doing this, as it would create differences between the 'bulk' and the incremental database installation scripts. I'd welcome an alternative.

akrherz:
So as I look at my deployments, I don't see `httpbind.port.plain` nor `httpbind.port.secure` persisted to the database.  So for an upgrade situation, it would be difficult to account for this situation.  I'd think clearly denoting to people that they need to set the system properties prior to upgrading, if they wish to keep these ports would be sufficient?

guusdk:
When they're not in the database, you never changed the default. Actually, when they *are* in the database, we shouldn't modify a thing in the first place!

Not having them in the database shouldn't prevent this from working. We don't need to explicitly check for their value. Rather: when we're executing an _incremental_ database upgrade, we know that this installation is an older installation being upgraded. In that case, we could check for the value, and if it's *not* there, set a value that configures 7070 to be used. As long as we're not doing that in the database script that's executed by a fresh install, this should work. As I said, not the cleanest solution, so if we can find a better way, I'm all for that - but I think this works and it beats keeping 7070 as the default which is overridden by every fresh setup with a property.

tevans:
So I've been thinking more on this ... the IANA registry of just a single BOSH port implies the installation must choose between HTTP or HTTPS, rather than supporting both. This is also supported by [the guidance in the spec] (http://xmpp.org/extensions/xep-0124.html#security-client).

Perhaps we can modify the post-4.0 setup to set a new property (`httpbind.secure`) and use this as an indicator of a post-4.0 deployment. I think we could reasonably support legacy configuration (7070/7443) if this new property is not defined.

Neustradamus:
+1 for integration.

Neustradamus:
The 4.0.3 out but there is not this...

https://xmpp.org/extensions/xep-0124.html#iana
https://xmpp.org/extensions/xep-0156.html

Neustradamus:
For inform users: It is not in 4.0.4 and 4.1.0.
You must change one file.

PR #407, opened 2015-11-23
OF-473: Revert original patch

tevans:
Received reports of unexpected roster caching caused by this change (not
verified). Additional investigation is needed to determine whether/how
to fix original issue reported via OF-473.


dwd:
+1, I'm going to merge this straight away because it's just reverting back.

PR #410, opened 2015-11-25
Simplify and shorten pubsub message id

tevans:
The specification (XEP-0060) recommends generating unique message ids
within a pubsub instance to aid troubleshooting for notification
delivery issues when needed. This replaces a ridiculously long and
inefficient string concatenation with something a bit more tidy and
concise.


sco0ter:
+1

PR #411, opened 2015-11-25
Fix NPE when handling routing failures

tevans:
Avoids NPE by dropping (and optionally logging) error response stanzas
if the original packet has no sender JID.

### DIFF
--- a/src/java/org/jivesoftware/openfire/IQRouter.java
+++ b/src/java/org/jivesoftware/openfire/IQRouter.java
@@ -433,6 +433,12 @@ public class IQRouter extends BasicModule {
             Log.error("Cannot reply an IQ error to another IQ error: " + originalPacket.toXML());
             return;
         }
+        if (originalPacket.getFrom() == null) {
+        	if (Log.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


sco0ter:
How could this happen? AFAIK Openfire overwrites the 'from' attribute on every inbound stanza.

tevans:
@sco0ter Good question ... the specific case I've seen occurs during the resend of unacknowledged stanzas from the new stream manager. Likely this condition is caused by a system-generated packet. I'll dig a little deeper to see if I can figure out which packets are causing the issue.

tevans:
Following up on this - the specific packets we are seeing are server-generated Roster broadcasts, e.g. https://github.com/igniterealtime/Openfire/blob/master/src/java/org/jivesoftware/openfire/roster/Roster.java#L712
I suspect there are other type of packets sent from the server. Perhaps we should take a pass through to set the 'from' attribute to the domain, but this PR would act as an NPE guard in the meantime.

sco0ter:
Ok, just read the spec. These are indeed ok to have no from attribute.
http://xmpp.org/rfcs/rfc6120.html#stanzas-attributes-from-c2s

PR #412, opened 2015-11-27
Assume a client with MUC elements in presence needs to join

dwd:
XEP-0045 discusses how a request to join a room is the only case where a MUC
element should exist in presence. Therefore, a conformant client will always
send this whenever it believes it is not joined. This can happen in cases of
stanza loss (due to network outage) or remote server crash.

This patch assumes that a client sending a MUC element is trying to join, and
will go through much of the motions of the join, including sending history as
required, existing occupants' presence, and so on. To other existing occupants,
however, it'll look like an ordinary presence update.


dwd:
@tevans reviewed a copy of this ages ago, and I'd clean forgotten I'd never tidied it up.

tevans:
+1

PR #413, opened 2015-11-27
Fix NPE in Node.java when changing node configuration.

sco0ter:
This NPE caused two issues:
- subscribers are not correctly notified
- change is not persisted to the DB

https://community.igniterealtime.org/thread/57584


dwd:
+1

tevans:
+1

PR #417, opened 2015-12-02
OF-983 Do not bounce packets while lock is held

dwd:
OF-983 appears to be caused by a complex bouncing process run while the
authentication check lock is held. This patch moves the actual bounce to
outside the synchronized block, so the code no longer deadlocks.

Untested, but based on the stack traces in OF-983 reports.

### DIFF
--- a/src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
+++ b/src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
@@ -590,17 +590,20 @@ public class LocalOutgoingServerSession extends LocalServerSession implements Ou
     @Override
 	boolean canProcess(Packet packet) {
         String senderDomain = packet.getFrom().getDomain();
+        boolean processed = true;
         if (!getAuthenticatedDomains().contains(senderDomain)) {
             synchronized (senderDomain.intern()) {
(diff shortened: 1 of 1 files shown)


src/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
```diff
@@ -491,7 +491,7 @@ private boolean routeToRemoteDomain(JID jid, Packet packet,
 		    boolean retry = false;
 		    // If we're here, it means we have no functional route. Sort it out.
 		    final String domain = jid.getDomain();
-		    synchronized (domain.intern()) { // Only create one route at a time.
+		    synchronized (("AddRoute::" + domain).intern()) { // Only create one route at a time.
```
tevans: I know it's beyond the scope of this PR, but it's worth noting that S2S connections are shared by all nodes in a cluster configuration. This should really be constrained using a shared cache lock rather than a local synchronized block. There has been some recent chatter related to this in the community forum.

dwd:
Oh, this looks really ugly now. I'll rebase this first...

dwd:
OK, better now.

sco0ter:
+1

tevans:
+1

PR #418, opened 2015-12-02
OF-984 / OF-985 Dialback fixes for to attribute and TLS

dwd:
These two patches add the missing "to" attribute from the stream open for verify-only connections, and
also negotiate TLS if offered, using the S2S configuration.

Tested by running it up in production on d.c.n


tevans:
+1

PR #419, opened 2015-12-02
OF-988: Don't queue packets twice.

guusdk:


dwd:
No idea how this could possibly fail. Looks good to me.

guusdk:
Yeah, I can't figure that out either. The test runs fine locally. Maybe a fluke?

akrherz:
I asked travis to attempt to rebuild and it failed again in the same matter, strange.

guusdk:
Ther's no need for this any longer, as the code that is fixed by this PR was reverted by https://github.com/igniterealtime/Openfire/pull/423

PR #423, opened 2015-12-04
Revert "OF-115 Remove race during outgoing session creation"

dwd:
This reverts commit d0667e214def2576b9818de2b1dcdde5a2724bfc.

While I think the general code is right, it's clearly got serious shortcomings
so I think it's best to back this out for now.


guusdk:
I concur that rolling back to stop the existing issues is preferable.

PR #424, opened 2015-12-04
OF-631 Fix all the SCRAM faults

dwd:


dwd:
Not really ready. I think this catch must be being hit, but I don't understand why.

PR #426, opened 2015-12-04
Prevent WritePendingException during delivery of BOSH packets.

sco0ter:
This exception caused serious trouble, e.g.:
- Clients got an invalid/incomplete packet, which breaks them.
- Openfire logged related exceptions like org.jivesoftware.openfire.http.HttpConnectionClosedException: The http connection is no longer available to deliver content

It was caused by calling complete() although the previous write() has not completed (due to its async nature).

I also restructured that the Content-Length header is only set once.

Fixes OF-989.
See also https://community.igniterealtime.org/thread/57622

### DIFF
--- a/src/java/org/jivesoftware/openfire/http/HttpBindServlet.java
+++ b/src/java/org/jivesoftware/openfire/http/HttpBindServlet.java
@@ -294,12 +294,13 @@ public class HttpBindServlet extends HttpServlet {
         }
 
         final byte[] byteContent = content.getBytes(StandardCharsets.UTF_8);
+        // BOSH communication should not use Chunked encoding.
+        // This is prevented by explicitly setting the Content-Length header.
+        response.setContentLength(byteContent.length);
+
         if (async) {
             response.getOutputStream().setWriteListener(new WriteListenerImpl(context, byteContent));
         } else {
-            // BOSH communication should not use Chunked encoding.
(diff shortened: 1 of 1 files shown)


src/java/org/jivesoftware/openfire/http/HttpBindServlet.java
```diff
@@ -447,14 +449,22 @@ public WriteListenerImpl(AsyncContext context, byte[] data) {
...
-            Log.trace("Data can be written to [" + remoteAddress + "]");
-
-            // BOSH communication should not use Chunked encoding.
-            // This is prevented by explicitly setting the Content-Length header.
-            context.getResponse().setContentLength(data.length);
-
-            context.getResponse().getOutputStream().write(data);
-            context.complete();
+            // This method may be invoked multiple times and by different threads, e.g. when writing large byte arrays.
+            Log.info("Data can be written to [" + remoteAddress + "]");
```
dwd: Should this be info level logging?

tevans:
+1

src/java/org/jivesoftware/openfire/http/HttpBindServlet.java
tevans: Agree, let's set this back to trace (or debug).

sco0ter:
Change log level back to trace.

PR #427, opened 2015-12-04
OF-631 Reduce salt size to fit database

dwd:
The salt size was 32, and the database column was a VARCHAR(32).

However, the salt was base64 encoded prior to insertion, and this caused it to
be too large. Therefore this patch reduces the size to 24 bytes, which expands
to 8 units of three bytes encoded as four characters, hence 32 characters total.


guusdk:
Looks good to me.

PR #428, opened 2015-12-04
Fix XEP-0012 logic.

sco0ter:
If a query has no 'to' attribute it should be treated on behalf of the sender.
It should not be treated as if it were targeted to the server (and return the server uptime).

The former logic also threw an UserNotFoundException, if the user queried itself, because the user was not on his own roster.
 I've fixed that in the canProbePresence method.

 Fixes issue OF-982


dwd:
+1

tevans:
+1

PR #429, opened 2015-12-06
Subscribe with Response Unsubscribed Causes Roster Push to Responding Client

sco0ter:
This should fix OF-317.

During testing also discovered an UserNotFoundException thrown by canProbePresence, because it was passed the JID instead of the username.


dwd:
+1

PR #431, opened 2015-12-07
OF-992: Re-init on restart

guusdk:
When the BOSH manager is restarted, the thread pool (and other components that are used) should be recreated or re-initialized. Currently, restarting the manager won't actually work (as the executor pool isn't accepting any new tasks).

### DIFF
--- a/src/java/org/jivesoftware/openfire/http/HttpSessionManager.java
+++ b/src/java/org/jivesoftware/openfire/http/HttpSessionManager.java
@@ -81,39 +81,13 @@ public class HttpSessionManager {
     	
         JiveGlobals.migrateProperty("xmpp.httpbind.worker.threads");
         JiveGlobals.migrateProperty("xmpp.httpbind.worker.timeout");
-    	
-        this.sessionManager = SessionManager.getInstance();
-        init();
     }
 
-    public void init() {
-        Log.warn("HttpSessionManager.init() recreate sendPacketPool");
-        // Configure a pooled executor to handle async routing for incoming packets
-        // with a default size of 16 threads ("xmpp.httpbind.worker.threads"); also
-        // uses an unbounded task queue and configurable keep-alive (default: 60 secs)
-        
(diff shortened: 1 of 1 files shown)


src/java/org/jivesoftware/openfire/http/HttpSessionManager.java
```diff
@@ -81,39 +81,13 @@ public HttpSessionManager() {
...
-	                thread.setDaemon(true);
-	                return thread;
-	            }
-	    	});
-    }
+    /**
+     * @deprecated As of Openfire 4.0.0, the functionality of this method was added to the implementation of #start().
+     */
+    @Deprecated
+    public void init() {}
```
sco0ter: I don't think it does make sense to deprecate a method, but at the same remove all the logic from it. Otherwise users of the method *think* it still works, but it doesn't.
Either keep the logic or remove the method.

dwd: Does this matter? Previous behaviour was to call init() once, then start() multiple times. This call pattern still works, but the init() call isn't actually needed anymore. I'm not sure I care either way, mind.

guusdk: I did not want to remove init() because it is declared public. That's just me being cautious, perhaps overly so. Doesn't hurt to keep it around for a release or so.

sco0ter: If it was me, I'd remove it. It should have been private anyway.

guusdk: I agree with you that it was probably best to have it private from the start, but it wasn't. For reasons that I don't know, people might be using this now. 

As long as it doesn't hurt us in any significant way, I feel that we should prefer backwards compatibility over cleanliness - at least for one release. We should prefer keeping our existing users happy over ourselves being happy.

PR #433, opened 2015-12-07
OF-122: Fix inconsistent MUC subject handling

tevans:
Ensure that the MUC room subject is set correctly from initial request and upon subsequent reload.

Initial testing looks good, but would welcome additional validation and testing with various XMPP clients if possible.


guusdk:
I reviewed the code without running it - can't argue with anything.

PR #434, opened 2015-12-07
Avoid using classes from sealed packages

tevans:
This fix was applied a while back on an earlier incarnation of this test class.


dwd:
+1

sco0ter:
+1

PR #435, opened 2015-12-07
Fix NPE when accessing AuditManager module

tevans:
Not sure when this got changed, but the module loader uses the interface class name as a key, not the implementation class name.


dwd:
That looks embarrassingly familiar. I think I may have seen that in a merge I did recently. +1

Redor:
+1

PR #436, opened 2015-12-08
OF-995: Parent plugin identification should not be case-sensitive.

guusdk:


sco0ter:
Not sure, if the loss of ConcurrentHashMap does any harm (TreeMap is not thread-safe).
This is also somewhat related to OF-952, which suggests to have the plugin directories not lower-cased. But I think these two issues don't interfere with each other.

guusdk:
ConcurrentHashMap was there only because at one point, I used that as a drop-in replacement. The class itself was never thread safe, so we're not loosing anything there.

I had not noticed OF-952, but I agree with you, that's likely unrelated (although using the case insensitive mapping will help there as well, I think).

tevans:
+1

PR #439, opened 2015-12-08
OF-973: TLS negotiation failures should come in two flavors.

guusdk:
When STARTTLS fails, there are two distinct types of failure. Each type should
be handled in a distinct manner.

Note that this commit applies a fix to the S2S code only. C2S may also need work.


dwd:
I'm not convinced this is entirely right, but it's certainly a massive improvement - let's merge.

guusdk:
We've been running this on igniterealtime.org for a couple of days now. So far, things look ok.

tevans:
+1

PR #440, opened 2015-12-09
Admin Console: External Component Settings

guusdk:
The page on which a user can manage the settings for External Components should offer
all features that are available now that the new ConnectionManager has been implemented.

I am aware that eventually (before the next release), some more work is needed. Notably:
* hardcoded English texts should be moved to i18n property files;
* code duplication between this page and other pages that manage connections (s2s, c2s) should be reduced.

I also think we should hide some of the configuration complexity somehow - perhaps introduce a second page, or add a 'show advanced settings' button somewhere.

(The same goes for several other pages that are already committed, not part of this PR).

As especially managing i18n properties have proven to be very time consuming, I am postponing that effort, allowing all changes to have settled (I initially changed i18n property files, only to change them later again, when the layout of the pages was improved).


deleolajide:
+1

PR #441, opened 2015-12-10
Upgrading the Ant Contrib & XmlTask libraries

guusdk:


sco0ter:
+1
I only wonder if there's a special reason behind the upgrade.

dwd:
Is there any special reason? Can it be documented? If it's just an upgrade because we can, that's fine too.

sco0ter:
The only reason against the "because we can" is that I don't like binaries (jars) in the repository because it only crams it and cloning it will take longer and longer with every new binary (I hope I understood git correctly here). One of my motivation to move to Maven.

dwd:
Yeah. When we move to Maven I might consider cloning the repository for
history, and then stripping the binaries out of it completely.

Cloning takes long enough as it is, and it'll be really annoying if it's
mostly stuff that's been removed.

guusdk:
Ant Contrib was updated because locally, we had a fix for a build process that got rolled back right after I created this pull request. 

XmlTasks was updated because its changelog says it offers better support / logging for XmlCatalogs, which relates to OF-998

Both libraries are used by the build process, not by Openfire itself.

akrherz:
To note, when we first moved Openfire to github, I rewrote the history to strip out the JRE artifacts found in the repo, so the current situation could have been much worse :)  I dunno if stripping out old JARs is a good idea as it would mean older tags would not build properly.  Stripping JREs only impacted ant build artifacts and not building openfire itself.

dwd:
Yes, we'd need to maintain an old-style repository for building older
builds. So I'm thinking it'd be best to do this when we drop support for
3.10, and Maven is in place (so old-school builds would be a rarity).

PR #442, opened 2015-12-10
OF-997 Deny framing to avoid clickjacking

dwd:


guusdk:
Perhaps SAMEORIGIN instead of DENY is more suitable here? I bet that there are instances out there that are embedded in some kind of larger management console.

dwd:
Perhaps it's most sensible to make it an option.

deleolajide:
YES, there are few instances that we have of Openfire embedded in a customer portal. Please put this behaviour behind JiveGlobal property for backward compatibility.

dwd:
That better?

guusdk:
works for me. +1

deleolajide:
+1 Excellent. many thanks :-)

PR #446, opened 2015-12-11
Reflected XSS Issues in Admin Console

dwd:
Both found by @tjdurden using BurpSuite


dwd:
Running this locally and tested.

tevans:
+1

PR #448, opened 2015-12-11
OF-1003 OF-1002 Make XEP-0198 thread safer

dwd:


src/java/org/jivesoftware/openfire/SessionManager.java
```diff
@@ -1269,22 +1269,7 @@ public void onConnectionClose(Object handback) {
...
-	                    	        Message m = (Message)unacked.packet;
-        	                        Element delayInformation = m.addChildElement("delay", "urn:xmpp:delay");
-                                        delayInformation.addAttribute("stamp", XMPPDateTimeFormat.format(unacked.timestamp));
-                                        delayInformation.addAttribute("from", serverAddress.toBareJID());
-	                    	    }
-    	                            router.route(unacked.packet);
-	                    	}
-	                    }
-                    }
+                    session.getStreamManager().onClose(router, serverAddress);
```
tevans: Maybe we should consider using SessionEventDispatcher/Listener for stream management? That way we could clean up any unacknowledged stanzas regardless of what caused the session to end.

dwd: I was expecting this to get called during any session ending, anyway - and yes, there are quite possibly better ways to hook into the cleanup state. Ideally, I think we'll need to move more of the Session state either into StreamManager or something else, so resumption will have a change of working.

tevans:
+1

PR #453, opened 2015-12-14
openfire social plugin

deleolajide:
Openfire Social is a new plugin (for OF 4.0 and above) which combines WordPress and BuddyPress with Openfire chat/group chat to provide a fully collaborative online social community.


deleolajide:
With Clearspace officially removed from Openfire, I thought now was the time to bring an old favourite plugin into igniterealtime.org. See https://code.google.com/p/inspired-social/

dwd:
Why do this work in the main openfire repository? Would it be easier to do this in a new git repository?

deleolajide:
Good question :-)

I don't have permissions to create a new repository 

It does make sense to put all external integration plugins in a half-way house repository that allows them to remain synced up with OF changes and delivery and at the same time remove the bloat on the increasing number of plugins. Unfortunately we don't have that in place

This plugin does have a fair bit of Java code integrating Openfire with PHP/WordPress and I would prefer to fix issues as they happen being a part of the main repository rather than having to tackle a load of OF changes that break the plugins with no historical perspective much later.

guusdk:
Ugh. That is a _lot_ of files. I'd prefer finding a solution where we don't need to include all of WordPress' sources in our repository.

The repository is bloated as it is. I get that including the sources makes it easy to integrate stuff, but from a development point of view, I think it's not ideal: It introduces lots of duplication, and loading things in an IDE is bound to kick of all kind of 'automatic language support' triggers. My IDE is already having trouble doing basic indexing/searching as it is - we shouldn't add to that.

Perhaps Git Modules offer a suitable approach for us to reference sources in other repositories?

dwd:
Would you put this in a different repository if we set one up for you? I'd like to move ofmeet as well, if that's a good solution.

deleolajide:
I agree. There are pros and cons to same or different repositories.

@guusdk , git modules looks the way forward and moving openfire to maven might be a better first step.

@dwd , **yes**, but it would be nice to still have current binary versions of both plugins built as part of the Openfire delivery and appear in the plugin list at igniterealtime.org in order to build a community around them. I am not sure on how to do that. Maybe @akrherz would know :-)

tevans:
Agree I'd like to see some of these very large plugin code bases in their own Ignite repositories in github if possible. Perhaps some of our "plugin emeritus" components (kraken, fastpath, etc.) could be dealt with in this manner also?

deleolajide:
ok. I will work on a community plugins repository as the new home for ofmeet, ofsocial and other large integration plugins

PR #454, opened 2015-12-15
OF-1004: Reduce complexity of connection configuration

guusdk:
This pull request has two major changes:
1. The configuration of protocols and cipher suites now depends on one setting, instead of having both a 'enabled' as well as a 'disabled' list. As a side effect, this improves the default values, which up until this change were either to narrow (nothing) or to wide (every possible option was enabled). Now, the Java-configured default is used when no Openfire-specific setting is available.
2. Most of the TLS settings have been moved to an 'advanced settings' page on the Admin console. There, you can now also configure encryption protocols en cipher suites.

### DIFF
--- a/src/web/connection-settings-advanced.jsp
+++ b/src/web/connection-settings-advanced.jsp
@@ -1,0 +1,426 @@
+<%@ page import="org.jivesoftware.openfire.XMPPServer" %>
+<%@ page import="org.jivesoftware.util.ParamUtils" %>
+<%@ page import="org.jivesoftware.openfire.Connection" %>
+<%@ page import="org.jivesoftware.openfire.spi.*" %>
+<%@ page import="java.util.*" %>
+<%@ page errorPage="error.jsp" %>
+
+<%@ taglib uri="admin" prefix="admin" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/fmt" prefix="fmt" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn" %>
+<jsp:useBean id="webManager" class="org.jivesoftware.util.WebManager"  />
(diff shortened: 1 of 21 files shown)


src/web/connection-settings-socket-s2s.jsp
```diff
@@ -320,61 +263,23 @@
 
 <form action="connection-settings-socket-s2s.jsp" method="post">
 
-    <admin:contentBox title="Encrypted (legacy-mode) connections">
+    <admin:contentBox title="Plain-text (with STARTTLS) connections">
 
-        <p>Connections of this type are established using encryption immediately (as opposed to using STARTTLS). This type of connectivity is commonly referred to as the "legacy" method of establishing encrypted communications.</p>
+        <p>Openfire can accept plain-text connections, which, depending on the policy that is configured here, can be upgraded to encrypted connections (using the STARTTLS protocol).</p>
```
tevans: These strings should eventually find their way into the i18n properties.

src/web/connection-settings-advanced.jsp
```diff
@@ -0,0 +1,426 @@
...
+                    </td>
+                </tr>
+                <tr valign="middle">
+                    <td>
+                        <input type="radio" name="tlspolicy" value="optional" id="tlspolicy-optional" ${configuration.tlsPolicy.name() eq 'optional' ? 'checked' : ''} onclick="applyDisplayable()"/>
+                        <label for="tlspolicy-optional"><b>Optional</b> - Encryption may be used, but is not required.</label>
+                    </td>
+                </tr>
+                <tr valign="middle">
+                    <td>
```
tevans: More i18n needed (at some point) at various places in this page.

tevans:
+1 - lots happening here, but looks benign ... will need testing, including xss/sanitization for new http parameters.

guusdk:
Indeed, i18n is still needed. I am explicitly putting that off until we're done refactoring those pages. Applying i18n has proven to be very time consuming - I want to do that at the last possible moment to avoid me going back and forth.

PR #456, opened 2015-12-15
Do not depend on platform-provided junit integration for Ant.

guusdk:
When executing junit, Openfire used to depend on the Ant/JUnit integration
as provided in the home directory of ant. This leads to platform-specific
build results. By providing the ant/junit libraries, this effect is reduced.


tevans:
ok by me +1

sco0ter:
+1

PR #460, opened 2015-12-16
OF-1007: Supplement existing client-to-server whitelisting with blacklisting

guusdk:
This PR is created in response to current flooding of igniterealtime.org. For regular client-to-server connections, Openfire allows for a whitelist to be defined. However, there's no blacklisting.

The last commit in this PR adds blacklisting to those places where the whitelist is already in effect (note that this might be generally improved upon in the future)

The first and second commit in this PR do not intend to make any functional change.


tevans:
ok by me +1

PR #461, opened 2015-12-17
OF-793: Improve exception message

guusdk:
When a peer sends a non-encrypted data where we expected encrypted,
a confusing exception is logged. This commit wraps that exception,
giving a clear explanation.

This typically occurs during a failed TLS handshake, where the peers
should be closing the connection without sending data, but does not
so (typically, the exceptions that are logged indicate that
</failure or </stream is received in plain text).


dwd:
+1

sco0ter:
+1

PR #462, opened 2015-12-17
OF-1010: Cannot re-init default SSLContext

guusdk:
A default SSLContext object cannot be re-initialized. This causes LDAPS to fail.


dwd:
+1

sco0ter:
+1

PR #466, opened 2015-12-17
OF-1009: Ignore the exception indicating an error

guusdk:
Although not 'properly ended', the handshake failed in an expected manner.
This should be processed as a 'normal' handshake failure (not an exceptional
one).


dwd:
+1

guusdk:
csh gave his blessing (under conditions) in open_chat ;)

PR #467, opened 2015-12-17
Admin console: SSLv2Hello isn't a protocol, but an option.

guusdk:
This commit is a cosmetic change to the admin panel. If the JVM supports
SSLv2Hello, it is no longer presented as an encryption protocol, but rather
as an option.


dwd:
+1, though I think we need to check if this is sending an SSLv2 Hello and fix it if it is.

guusdk:
So you want to accept (for inbound connections, where we are in a server-role), but not transmit (when we're a client, ie outbound server-to-server connections)?

dwd:
Exactly that. Accepting an SSLv2 Hello is not ideal, since it restricts the
possible negotiation options (though it still allows TLSv1.2 with the usual
cipher suites). Sending one can cause us not to be allowed to connect.

tevans:
+1

PR #470, opened 2015-12-21
Fix twice sending of messages (invite/decline) on route.really-all-resources=true and 	sendingInvitationRejection event to MUCEventDelegate

ishmakov:
If route.really-all-resources enabled it's not requere process message again via priority algorithm.
We already processed in 560 line. 
If we not stop (return) than session.process() call twice (first in line 560 and next in next other)

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/MUCEventDelegate.java
+++ b/src/java/org/jivesoftware/openfire/muc/MUCEventDelegate.java
@@ -36,6 +36,12 @@ public abstract class MUCEventDelegate {
         HANDLED_BY_OPENFIRE,
         REJECTED;
     }
+    
+    public enum InvitationRejectionResult {
+        HANDLED_BY_DELEGATE,
+        HANDLED_BY_OPENFIRE,
+    }
+
 
     /**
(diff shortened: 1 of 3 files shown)


ishmakov:
1. Fix twice sending of messages (invite/decline) on route.really-all-resources=true
2. Extend MUCEventDelegate by sendingInvitationRejection method similarly as sendingInvitation

tevans:
This feels like two unrelated patches/PRs. They should probably be separated as such. 

Also, I did not see an implementation of the new abstract method declared in the MUCEventDelegate class.

ishmakov:
I agree. I don't understand how last changes has been added to this comit.
MUCEventDelegate in 4196966e6c16b0366330249ce37a8b991e224b1c
https://github.com/ishmakov/Openfire/commit/4196966e6c16b0366330249ce37a8b991e224b1c

tevans:
You should create a separate branch in your cloned repository for each patch you would like to submit via PR. Please do not use your "master" branch for upstream pull requests. 

I will close this PR for now, but invite you to resubmit each of your proposed changes again.

PR #471, opened 2015-12-23
Update BasicStreamIDFactory.java

microdust2010:
The random number to use streamId, can create a lot of conflict in the high concurrent connection.

### DIFF
--- a/src/java/org/jivesoftware/openfire/spi/BasicStreamIDFactory.java
+++ b/src/java/org/jivesoftware/openfire/spi/BasicStreamIDFactory.java
@@ -22,7 +22,7 @@ package org.jivesoftware.openfire.spi;
 
 import org.jivesoftware.openfire.StreamID;
 import org.jivesoftware.openfire.StreamIDFactory;
-import java.util.Random;
+import java.util.UUID;
 
 /**
  * A basic stream ID factory that produces id's using java.util.Random
@@ -32,14 +32,12 @@ import java.util.Random;
  */
 public class BasicStreamIDFactory implements StreamIDFactory {
 
-    /**
-     * The random number to use, someone with Java can predict stream IDs if they can guess the current seed *
-     */
-    Random random = new Random();
-
     @Override
     public StreamID createStreamID() {
-        return new BasicStreamID(Integer.toHexString(random.nextInt()));
+    /**
+     * The random number to use, can create a lot of conflic in the high concurrent connection.
+     */
+        return new BasicStreamID(UUID.randomUUID().toString());
     }
 
     public StreamID createStreamID(String name) {


dwd:
I'm not against this in principle, but:

* Please update comments along with code.
* Please detail what the bug you're fixing is - what the effects are, how one might identify that they're affected.

tevans:
One concern I have is the large size of the UUID creating a tax on overall network load and memory footprint, since the stream ID is used as the resource component of many JIDs.

I would prefer to see this change implemented as a new StreamIDFactory (UUIDStreamIDFactory) that could be swapped in at runtime based on a configuration parameter, and perhaps also for C2S streams only. Please see the SessionManager class for an example of this approach (AuditStreamIDFactory).

microdust2010:
# random #
	@Test
	public void random() {
		Random random = new Random();
		int i = 0;
		HashMap<String, String> sessions = new HashMap<>();
		while (i < 200000) {
			sessions.put(Integer.toHexString(random.nextInt()), "");
			i++;
		}
		System.out.println(sessions.size());
	}
 
- run times: 290ms	
- lost Not-Auth session:7
- There is a risk of loss of not-auth session when have high concurrent not-auth session.I think the bug is high level in the BasicStreamIDFactory class.
# UUID #
	@Test
	public void uuid() {
		int i = 0;
		HashMap<String, String> sessions = new HashMap<>();
		while (i < 200000) {
			sessions.put(UUID.randomUUID().toString(), "");
			i++;
		}

		System.out.println(sessions.size());
	}

- run times:1446ms 
- not exist lost Not-Auth session.
- The large size of the UUID creating a tax on overall network load and memory footprint is within the acceptable leve.

guusdk:
Perhaps we can find middle ground in switching to an implementation that generates IDs of equal length, but using the full alphabet (+digits) instead of just hexadecimals.

Perhaps we should switch to SecureRandom while we're at it.

guusdk:
Quick stab at a full-alphanumeric solution using SecureRandom:

        final SecureRandom random = new SecureRandom();
        final int maxStringSize = 8; // maximum amount of characters in the generated string.
        final String id = new BigInteger( maxStringSize * 5, random).toString( 36 );

This still used eight characters (easily changed on the second line), but is a lot less likely to generate duplicates (which, obviously, will eventually start to occur even with this implementation).

I have executed an iterative generation a couple of times. After about 14 to 15 million generations, my JVM begins to run out of memory. I've done that a couple of times - only once I ran into a collision (at around 13 million generations). That appears to be a vast improvement over the original test (which had multiple collisions for only 200,000 iterations.

tevans:
+1 for the shorter stream id proposed by @guusdk 

I am still open to a UUID-based stream id if implemented via a separate StreamIDFactory.

guusdk:
I've created https://github.com/igniterealtime/Openfire/pull/480 that includes the change that I suggested earlier (but not the pluggable StreamIDFactory, which can still be added as a complementary feature).

dwd:
Gone with #480, but thanks for raising this issue.

PR #472, opened 2015-12-24
Fix twice sending of messages on route.really-all-resources=true

ishmakov:
If route.really-all-resources enabled it's not requere process message again via priority algorithm.
We already processed in 560 line. 
If we not stop (return) than session.process() call twice (first in line 560 and next in next other)


dwd:
+1

PR #474, opened 2015-12-29
Fixed group population for colliding full names.

anuko:
When LDAP search returns a multiple result set for a user CN,
we now iterate through the entire set to find a matching user DN.

### DIFF
--- a/src/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
+++ b/src/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
@@ -227,7 +227,7 @@ public class LdapGroupProvider extends AbstractGroupProvider {
                 Pattern.compile("(?i)(^" + manager.getUsernameField() + "=)([^,]+)(.+)");
 
         SearchControls searchControls = new SearchControls();
-        searchControls.setReturningAttributes(new String[] { manager.getUsernameField() });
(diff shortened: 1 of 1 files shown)


guusdk:
I don't see how this PR breaks the unit test (as Travis reported).

Otherwise, +1. 

@speedy01 care to review?

guusdk:
Re-fired Travis, now all green.

sco0ter:
+1

tevans:
+1

wrooot:
Filed as https://igniterealtime.org/issues/browse/OF-1041

guusdk:
Hmm. In hindsight: would a better solution to the problem that this PR tries to solve be: make sure that the ```manager.getUsernameField()``` method returns a value that uniquely identifies a user? The value for that method is configurable through the ```ldap.usernameField``` property. Would the PR be unneeded if @anuko had set ```ldap.usernameField``` to "distinguishedName" for their setup?

guusdk:
I've created a PR to revert the change here: https://github.com/igniterealtime/Openfire/pull/509

guusdk:
Reading up on the subject a bit more, I don't believe that my suggested workaround would work. Nonetheless, what's in place now is not portable to other LDAP providers than Active Directory, and should be reverted.

PR #475, opened 2015-12-30
Add explanation to setup about default admin password

wrooot:
Add explanation to setup about default admin password
https://igniterealtime.org/issues/browse/OF-1016


guusdk:
I'm not to happy with the fact that we don't require a password for new installations, but that's out of scope for this PR. This PR improves the existing situation. +1

sco0ter:
+1

PR #477, opened 2016-01-04
xml:lang in stream header should conform to a language tag.

sco0ter:
toString() was called on the Locale here, resulting e.g. in "de_DE".
Instead it should be "de-DE".


dwd:
Well, obviously it *should* be en-GB, but I'll forgive you that one.

+1

guusdk:
+1

When doing a search for "xml:lang" in all java files, I'm finding a couple of others (StanzaHandler.java, SocketReadingMode.java) that are also populated with a Locale. Can you fix those as well?

sco0ter:
Good find, I've changed them.

PR #478, opened 2016-01-04
OF-122: MUC subject change policy

tevans:
Default to lenient policy, allowing override to strict (conforming)
policy. Tested using Pidgin, Psi, and Exodus. Refer to details in
OF-122.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/HistoryStrategy.java
+++ b/src/java/org/jivesoftware/openfire/muc/HistoryStrategy.java
@@ -336,8 +336,8 @@ public class HistoryStrategy {
 		// where the <message/> MUST contain a <subject/> element that specifies the new subject 
 		// but MUST NOT contain a <body/> element (or a <thread/> element).
 		// Unfortunately, many clients do not follow these strict guidelines from the specs, so we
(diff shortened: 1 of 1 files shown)


sco0ter:
To be honest, I don't like to violate the spec and let buggy clients think they are doing it right. This surely won't encourage them to fix their bugs.

I think XEP-0045 is clear enough on the subject change, not vague.

At least make the spec-conform behavior the default one.

tevans:
Agree in principle ... however in this case enforcing the strict behavior (in the server) effectively allows non-conforming clients to change the room subject without permission. This is because groupchat messages having both a subject and a body are (incorrectly) treated as subject change requests by these clients.

I think the proposed implementation is correct given the problems with the specification.

sco0ter:
But this is still a client issue. Why should we care about some non-conforming clients displaying a subject change, which isn't a real subject change as per the spec.
It sounds as if these clients treat "subject + body" in messages as subject change, but I fail to see, why this Openfire's concern.

tevans:
OK - I actually have made the same assertions in the comments of OF-922. I will adjust this patch to conform to the spec by default, allowing strict=false via a configuration override.

dwd:
+1

PR #476, opened 2016-01-04
Improving logging for S2S / dialback

guusdk:
The S2S (federation) code was not big on logging. This commit introduces new logging, and structures the logged messages. This commit does not introduce functional changes.

PR #479, opened 2016-01-04
S2S logging

guusdk:
This builds on https://github.com/igniterealtime/Openfire/pull/476 and further improves logging. Also, code complexity is reduced by removing unused code, and refactoring duplicated code into a utility class.

### DIFF
--- a/src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
+++ b/src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
@@ -248,14 +248,6 @@ public class LocalOutgoingServerSession extends LocalServerSession implements Ou
 
         log.debug( "Creating new session..." );
 
-        String localDomainName = XMPPServer.getInstance().getServerInfo().getXMPPDomain();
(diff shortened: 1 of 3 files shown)


src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
```diff
@@ -449,77 +433,15 @@ private static LocalOutgoingServerSession secureAndAuthenticate(String hostname,
             // Get new stream features
             features = reader.parseDocument().getRootElement();
             if (features != null) {
-                // Check if we can use stream compression
-                final Connection.CompressionPolicy compressionPolicy = connection.getConfiguration().getCompressionPolicy();
-                if (Connection.CompressionPolicy.optional == compressionPolicy) {
```
tevans: Are we dropping support for S2S compression?

guusdk: It was never used in the first place. A couple of lines down, it reads: if
(zlibSupported) { zlibSupported = false; }

tevans:
ok by me +1

PR #481, opened 2016-01-04
OF-1023: Fire contactChanged/Deleted events

tevans:
When a shared group is removed from a roster item, fire an update or
delete event to keep the corresponding Roster cache in sync.


guusdk:
Seems to be inline with what happens with an add (line 861 onwards). +1

PR #482, opened 2016-01-05
solve the memory leak problems which is caused by add LocalMUCRoom ob…

VicoWu:
LocalMUCRoom implements interface GroupEventListener and was add to listeners of GroupEventListener when constructed. But it is never removed from the listeners. According to the result of experiment, I do find that after a long time run ,openfire will consume a lot of memory.After analysed 
by MAT Tools, I find out that most of memory is consumed by the LocalMUCRoom object.Even user leave the room , it still cannot be reclaimed by GC.This will lead the result that each packet of workgroup will be broadcasted to all rooms in listeners. So, it should be removed from listeners when LocalMUCRoom is removed.
experiments has been done after I modified the code.Memory leak disappeared and GC can remove LocalMUCRoom objects which are destroyed.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -45,6 +45,7 @@ import org.jivesoftware.openfire.disco.DiscoItem;
 import org.jivesoftware.openfire.disco.DiscoItemsProvider;
 import org.jivesoftware.openfire.disco.DiscoServerItem;
 import org.jivesoftware.openfire.disco.ServerItemsProvider;
+import org.jivesoftware.openfire.event.GroupEventDispatcher;
 import org.jivesoftware.openfire.group.ConcurrentGroupList;
 import org.jivesoftware.openfire.group.GroupAwareList;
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks for taking the time to fix this!

Your change makes sense. Can it be improved by having the calls to GroupEventDispatcher's addListener() and removeListener() in the same class somehow? If we can, we should keep those together. That will cause less problems with maintaining the code later.

VicoWu:
Sure.I will modify my code according to your suggestion.Thank you for you reply and attention.



583424568@qq.com
 
Sender: Guus der Kinderen
Send Time: 2016-01-05 17:40
Receiver: igniterealtime/Openfire
cc: Vico.Wu
Subject: Re: [Openfire] solve the memory leak problems which is caused by add LocalMUCRoom ob… (#482)
Thanks for taking the time to fix this!
Your change makes sense. Can it be improved by having the calls to GroupEventDispatcher's addListener() and removeListener() in the same class somehow? If we can, we should keep those together. That will cause less problems with maintaining the code later.

guusdk:
I'd prefer if this PR makes it in the upcoming release, even if the suggested improvement has not been provided. Fixing the memory leak is a significant improvement.

PR #483, opened 2016-01-05
OF-999: Allow configuration of BOSH worker thread pool

guusdk:
The thread pool that performs processing of data received through the HTTP-based
connections of BOSH (as provided by Jetty) should not have a hardcoded upper bound.
Instead, the pool size should be configurable.

For other connection types, this is managed by properties named as such:
 - xmpp.client.processing.threads (client sockets)
 - xmpp.multiplex.processing.threads (connection managers)
 - xmpp.component.processing.threads (components)

BOSH does have a similar property: httpbind.client.processing.threads
It's value however is not used to configure the amount of worker threads. It is
used to configure the amount of (NIO) acceptors instead. In NIO, acceptor
threads handle network IO, and hand off processing to the worker threads.

There's little benefit in being able to configure the amount of acceptors, as
Jetty will default to a platform-specific default (based on the number of CPUs
that are available). There is greater benefit in being able to configure the
amount of worker threads (as that's more likely to affect throughput).

### DIFF
--- a/src/java/org/jivesoftware/openfire/http/HttpBindManager.java
+++ b/src/java/org/jivesoftware/openfire/http/HttpBindManager.java
@@ -96,7 +96,7 @@ public final class HttpBindManager {
 
     public static final String HTTP_BIND_AUTH_PER_CLIENTCERT_POLICY = "httpbind.client.cert.policy";
 
-    public static final int HTTP_BIND_THREADS_DEFAULT = 8;
+    public static final int HTTP_BIND_THREADS_DEFAULT = 200;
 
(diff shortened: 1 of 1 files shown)


guusdk:
The interesting bit is that we used to have the code in this state. It got changed by @tevans in commit 3032cc8484ff2d322a380cf9a05669c5a9527083 which mentions OF-857 (I believe that this is a typo - that issue seems unrelated). Tom, could you do a review please?

tevans:
Truth be told, I don't recall exactly what I was attempting to fix with my earlier commit. Likely it has been obviated by the recent move to NIO anyway. Let's go with this, +1.

PR #484, opened 2016-01-05
OF-1000 Prevent NPE and improve logging

guusdk:
When a log directory cannot be created, a warning should be logged,
and null pointers should be prevented when the directory is being
referenced.


dwd:
+1

tevans:
+1

PR #485, opened 2016-01-05
OF-998: Ant should resolve DTDs locally

guusdk:
Ant (or rather, the XMLTask in our build) attempts to lookup DTDs
online. The build (of plugins) will fail if no internet connectivity
is available. This commit references DTDs locally, allowing Ant to
skip the online lookup.


tevans:
+1

PR #486, opened 2016-01-05
Only use SSLv2Hello for client-mode

dwd:
Sending an SSLv2 Client Hello message never makes sense, since we don't support the SSLv2 protocol at all. Receiving one, however, is useful to support a handful of legacy servers, typically those using older Java and/or OpenSSL's SSLv23 method.

This PR therefore strips out SSLv2Hello support from client-mode TLS sessions if configured.


guusdk:
Some minor tweaks:
- It would be best to preserve the order of the set (not use HashSet)
- To make sure that we're not missing some default / edge case, check the content of the enabled protocols, not the configuration for SSLv2Hello

I think this alternative would do the trick:
```
final Set<String> protocols = new LinkedHashSet<>( Arrays.asList( sslEngine.getEnabledProtocols() ) );
protocols.remove( "SSLv2Hello" );
sslEngine.setEnabledProtocols( protocols.toArray( new String[ protocols.size() ] ) );
```

Lastly, we should also update the text in ```connection-settings-advanced.jsp``` (it should no longer mention the outgoing SSLv2Hello option).

guusdk:
+1

tevans:
+1

PR #487, opened 2016-01-05
OF-1025: Restore web-custom.xml functionality

guusdk:
For some reason, XPath-based selection of nodes no longer works. Traversing
the Dom4j object resolves that problem.


tevans:
+1

PR #488, opened 2016-01-05
Remove verbose logging

guusdk:
This class logged to debug on any property change (including properties
that do not relate to the class itself).


tevans:
+1

PR #490, opened 2016-01-06
WebSocket plugin: Fix wrong namespace adding to empty elements

sco0ter:
The old logic produced XML like:
&lt;presence xmlns="jabber:client" from="..." to="...">&lt;show xmlns="">xa&lt;/show>&lt;status xmlns=""/>&lt;/presence>

This is basically the same fix as in 6965c13.

Relates to OF-938.


guusdk:
+1

PR #492, opened 2016-01-06
Don't add multiple delayed delivery extension to resent messages.

sco0ter:
The issue is that messages, which get resent from offline storage already have an older timestamp.

There should not be another delayed delivery extension, which "overwrites" the old one.

Messages looked like that:
&lt;message>&lt;body>...&lt;/body>&lt;delay xmlns="urn:xmpp:delay" stamp="2016-01-06T13:48:45.312Z">&lt;/delay>&lt;delay xmlns="urn:xmpp:delay" stamp="2016-01-06T13:49:56.674Z">&lt;/delay>&lt;/message>

Instead the original timestamp should be preserved.


dwd:
Sounds good. +1

wrooot:
I think this change is causing troubles with offline messages in Spark (maybe other clients (not in Pidgin, Exodus)). Any ideas? https://community.igniterealtime.org/message/254361

PR #493, opened 2016-01-06
OF-1028 Stream Management refactoring

guusdk:
This started as a fix for OF-1028, but along the way, I refactored quite a bit of the code. I left the individual commits in place, as each of them are pretty isolated (makes for easier reviewing).


sco0ter:
Commit messages and code changes are reasonable, looks good to me.
I have only commented some questionable lines.

1a1efa27 seems to invalidate the previous work of 422e7ff, so these could be squashed theoretically.

+1

tevans:
+1 looks good to me

PR #495, opened 2016-01-06
OF-1033: Users should be notified of MUC service shutdown

guusdk:
This commit notifies all users that are in a MUC on the local server node of
a shut down of the service. The notification conforms to XEP-0045.

In order to prevent a prolonged shut down, this commit:
- ensures that a shutdown routine is executed just once (instead of for each
  type of interface being shut down: module,component and listener)
- the presence stanza that is sent is constructed using in-memory objects
  only (no database lookups)
- all notification occurs in a thread pool, which will concurrently process
  them, but will terminate forcibly after a certain amount of time.


src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -416,6 +429,75 @@ public void run() {
...
+                    }
+                }
+            } );
+        }
+
+        // Try to shutdown - wait - force shutdown.
+        service.shutdown();
+        try
+        {
+            service.awaitTermination( 500, TimeUnit.MILLISECONDS );
```
tevans: Should this value be configurable?

tevans:
+1 in spirit, although I'm not sure 500 ms will be sufficient in all cases

guusdk:
By special request...

tevans:
:+1:

PR #496, opened 2016-01-07
Fix implementation bug for XEP0049.

aztaru:
The comment in Line 105: "If the name and namespace of the element matches another" is correct, but the implementation is wrong, it checks "username and namespace" rather than "name and namespace".


sco0ter:
This looks like everybody can now retrieve all the private data of all users, if they use the same name/namespace (usually if they use the same client). Certainly not what we want.
I think it should ask for "username", "name" and "namespace.

aztaru:
You are right, The latest commit should meet the requirement.

tevans:
Thanks for your contribution.

I believe you will also need to update the `get()` method to provide the additional parameter for the `LOAD_PRIVATE` sql statement.

aztaru:
Done. :)

tevans:
OK - looks good to me. :+1:

PR #498, opened 2016-01-07
OF-821: Prevent duplicates for MUC item query

tevans:
Corrects issue with multiple JIDs sharing a nickname (masquerading)

### DIFF
--- a/src/java/org/jivesoftware/openfire/disco/DiscoItem.java
+++ b/src/java/org/jivesoftware/openfire/disco/DiscoItem.java
@@ -203,7 +203,7 @@ public class DiscoItem implements Result {
 
 	@Override
 	public String toString() {
-		return element == null ? null : element.asXML();
+		return element.asXML();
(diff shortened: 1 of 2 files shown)


sco0ter:
I don't know, if it's a good idea to return null from the toString() method. Maybe better return super.toString()?

Otherwise +1

tevans:
That's a fair point ... actually upon closer inspection it seems element can't actually be null (under normal circumstances), since there is no default constructor. I'll just drop the null check instead.

dwd:
This partially solves the bug by removing/suppressing duplicate disco#item entries, but it leaves the number of occupants reported by Disco as including any nick shares. I'm happy that this is useful, but I'm just checking how easy it'd be to fix the count reported, too.

dwd:
It's not clear whether getOccupantsCount() would be safe to update, so I'm going to leave this for another time.

PR #499, opened 2016-01-07
OF-1004: Apply i18n to admin console

guusdk:
This PR contains non-functional changes only. Text that was hard-coded in the JSP pages got replaced by i18n bundle usage.


tevans:
:+1:

PR #501, opened 2016-01-08
OF-821: Count occupants by nickname

tevans:
Returns the count of nicknames rather than the number of connections
(full JIDs), which will now also match the list of occupants.


dwd:
+1

PR #503, opened 2016-01-08
OF-454 Reflect presence to originator

dwd:


tevans:
Agree with @sco0ter here - I think you can instead remove the conditional statement that excludes the current session from the broadcast: `if (!originatingResource.equals(address)) { ... }`

Not as sure about the initial presence broadcast referred to at line 626 - seems like that takes a different path, although that code also mentions the same section of the RFC.

sco0ter:
There are two very similar sections in RFC 6121: 4.2.2 and 4.4.2.
Openfire follows 4.2.2 (I think line 624 should actually refer to 4.2.2), but not 4.4.2.

Initial presence logic first traverses broadcastPresenceToOtherResources() and later sessionAvailable(), making the initial presence being reflected twice.

dwd:
@tevans I can't remove the conditional because there's an early exit based around an option; so I need to do it in two steps - however, it looks like @sco0ter is again right, there's already reflection for initial presence.

guusdk:
+1

dwd:
Anyone else?

sco0ter:
+1, squash if you want.

PR #504, opened 2016-01-08
OF-872 Treat inbound IQ with no to as bare jid

dwd:


tevans:
This might be ok as-is, seeing that the `recipientJID` is used in local scope for subsequent routing requests. However, it also might not cover all the IQ cases ... not sure if that is intended here.

Tentative +1

dwd:
I think it should be this simple. Only iq:last has this odd behaviour in the specifications, and that's a historical wart that's crept in. That said, @sco0ter (correctly) has me worried that it's possible that existing IQHandlers might treat the bare jid and empty cases differently.

Given that this is currently not causing any interop issues, as far as I know, let's park this and review the existing cases before the *next* release.

PR #505, opened 2016-01-08
Denote Openfire 4.0.0 Release

akrherz:
This assumes pull requests #486 , #503, #504 are approved and merged first.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -26,7 +26,7 @@
     <property name="version.major" value="4"/>
     <property name="version.minor" value="0"/>
     <property name="version.revision" value="0"/>
-    <property name="version.extra" value="beta"/>    <!-- For 'beta' or 'alpha' -->
+    <property name="version.extra" value=""/>    <!-- For 'beta' or 'alpha' -->
 
 	<property name="javac.source" value="1.7"/>
 	<property name="javac.target" value="1.7"/>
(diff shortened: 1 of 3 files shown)


tevans:
:+1:

guusdk:
+1

On a side-note: Perhaps we should rethink our issue logging / code modification strategy. Currently, we keep issues in JIRA (from which the changelog is generated). Since we moved from SVN to Github, more and more pull requests got accepted that don't relate to any JIRA issue. That causes us to miss changes in the release notes. 

I'm not advocating for any particular change. I don't dislike JIRA, I don't dislike Github, and I certainly don't want to enforce "paperwork" - however as things are now, we're presenting a changelog as if it were a complete overview, while it isn't.

sco0ter:
On the other side, there are issues, which should not appear in the changelog, because they only reflect issues, which have occurred during work on 4.0.0, but not with any previous release. It will clutter the changelog and confuse users (e.g. if there are 3-4 bugs for XEP-0198, which is actually a new feature).

Some of theses issues are (among others): OF-1002, OF-1003, OF-1005, OF-1009, OF-1025, OF-1028, OF-1036.

IIRC, most PRs have an issue associated, unless they were either some internal code cleanup, refactoring or rework of other unreleased work.

akrherz:
Okay, updated PR:
1. Changed release date to date
2. Removed OF-872 from changelog

sco0ter:
I am still in favor of removing these other issues as well, which refer to "work-in-progress" of 4.0.0-only work.

PR #508, opened 2016-01-12
Fixes for banning problems in MUC rooms

bmarkwar:
Two commits that fix issues related to banning users from MUC rooms.

1. Openfire sends presence of "none" instead of "outcast" when a user is banned.
2. A user is not kicked from a non members-only room when they are banned.

Looks like both were introduced in commit 822abb014741b82d9248c3f747bea64779546999


guusdk:
+1

Thanks for taking the time to look into this. We're accepting pull requests using a policy that requires two regular contributors to sign off on it, before it gets merged. You've got my vote. One to go.

dwd:
Not on trunk anymore, we're post-release. Just one other contributor -
release branch remains two. I should note that somewhere.

dwd:
Although, having merged this, I realise we should probably make a Jira ticket before it goes on the release branch at least.

guusdk:
I was actually busy creating a JIRA ticket :) https://igniterealtime.org/issues/browse/OF-1040

PR #510, opened 2016-01-12
OF-1042: Don't NPE when a session is already closed when </stream>

guusdk:
I did not reproduce the NPE, but having a session already closed when </stream> is received sounds plausible to me. Some kind of race between handling the last data and following socket disconnect perhaps?


tevans:
+1

PR #512, opened 2016-01-14
add LocalMUCRoomManager to encapsule the simple management for LocalM…

VicoWu:
When a room is removed , we should remove it from GroupEventDispatcher's listeners,otherwise a memory leak problem will happen.So I create LocalMUCRoomManager to encapsule the room manage functions including remove/query/add which will avoid a direct operation to Map<String,LocalMUCRoom> .This class will duty on removing rooms from GroupEventDispatcher's listeners when a room is removed.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
@@ -1,0 +1,49 @@
+package org.jivesoftware.openfire.muc.spi;
+
+import java.util.Collection;
+import java.util.Date;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.jivesoftware.openfire.event.GroupEventDispatcher;
+import org.jivesoftware.openfire.muc.MUCRoom;
+
+/**
+ * this class supports the simple LocalMUCRoom management including remove,add and query.
+ * @author wuchang
+ * @email 583424568@qq.com
+ * 2016-1-14
+ */
+public class LocalMUCRoomManager {
+    private Map<String, LocalMUCRoom> rooms = new ConcurrentHashMap<>();
+     
+    public int getNumberChatRooms(){
+    	return rooms.size();
+    }
+    public void addRoom(String roomname,LocalMUCRoom room){
+    	rooms.put(roomname, room);
+    }
+    
+    public Collection<LocalMUCRoom> getRooms(){
+    	return rooms.values();
+    }
+    
+    public LocalMUCRoom getRoom(String roomname){
+    	return rooms.get(roomname);
+    }
+    
+    public MUCRoom removeRoom(String roomname){
+    	//memory leak will happen if we forget remove it from GroupEventDispatcher
+    	if(rooms.containsKey(roomname))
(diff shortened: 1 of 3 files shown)


guusdk:
Is this an improvement of #482 or a completely new change?

VicoWu:
From the aspect of 'solve memory leak problem ' , It is an improvement of #482. But it is also a completely new change , because I move the LocalMUCRoom management into another independent class whose duty is to manage the LocalMUCRoom objects and avoid the possibility of forgetting remove it from GroupEventDispatcher's listeners.


583424568@qq.com
 
Sender: Guus der Kinderen
Send Time: 2016-01-15 17:54
Receiver: igniterealtime/Openfire
cc: Vico.Wu
Subject: Re: [Openfire] add LocalMUCRoomManager to encapsule the simple management for LocalM… (#512)
Is this an improvement of #482 or a completely new change?

src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
```diff
@@ -0,0 +1,49 @@
...
+ * 2016-1-14
+ */
+public class LocalMUCRoomManager {
+    private Map<String, LocalMUCRoom> rooms = new ConcurrentHashMap<>();
+     
+    public int getNumberChatRooms(){
+    	return rooms.size();
+    }
+    public void addRoom(String roomname,LocalMUCRoom room){
+    	rooms.put(roomname, room);
```
aberenguel: In #518 I mentioned a case when the room is added in GroupEventDispatcher and is not removed.

That's why the GroupEventDispatcher is called inside the LocalMUCRoom constructor. 

So, I think the GroupEventDispatcher.addListener(room) should be called here, not in https://github.com/VicoWu/Openfire/blob/master/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java#L385

VicoWu: I approve that.

tevans:
Given the latest changes, is this PR now in sync with PR #518?
If so, I agree we should try to encapsulate the room lifecycle like this. 
+1

aberenguel:
@tevans Yes, it is synced with #518 now.

cmd6841:
@VicoWu - Could you tell how you verified this change please?

VicoWu:
Hi .What do you mean by verify?

cmd6841:
Meaning, did you run any kind of load tests, and performed memory analysis with and without this fix?

VicoWu:
The openfire server is used by my online system for which 500+ users online at the same time. before this fix, my openfire system's  memory usage become more and more and finally , about one month's running, oom happened. So I dumped the whole memory and find that so many LocalMUCRoom objects consumed most of the memroy. So I added this fix and the oom problem never happened again .Also , I dumped the whole memory once a week to monitor it ,everything is ok.

cmd6841:
Sweet, that's what I wanted to confirm. I am seeing similar behavior with ~1000 users. The difference being that rooms are created much more frequently in my case. I was going to cherry pick your change, and run these tests myself. But it helps to know that you have been doing that already.

Thanks!

PR #513, opened 2016-01-14
Should not store chat state messages, which contain a thread element.

sco0ter:
The problem was, that the logic stored empty chat messages, which contained a thread and chat state. The loop first checked the namespace of the thread, which was empty and therefore thought it should store it.


guusdk:
+1

Should his go in 4.0.1?

sco0ter:
sure, why not.

guusdk:
I've created https://igniterealtime.org/issues/browse/OF-1048 to track this change in the changelogs (which we generate from JIRA).

PR #516, opened 2016-01-19
OF-462 PEP should process IQ-gets that have no 'to' attribute.

sco0ter:
I've removed the "userFeatures", because I think a user cannot have features.

As per XEP-0163:

The server MUST return an identity of "pubsub/pep" on behalf of the account (as well as a list of the namespaces and other features it [the server?)] supports)
Anyway, the old logic only returned one feature (disco#info) for the user,
which was wrong, too (the pubsub features were missing at least)

disco#items addressed to the account (bare jid) or without 'to' attribute are treated the same now, too.

Also partially addresses OF-872.


tevans:
:+1:

PR #517, opened 2016-01-19
OF-144 History type change should be applied without a restart

sco0ter:
Basically the issue was, that each room had it's own HistoryStrategy copied on room creation (startup), while the UI only modified the MUC service's history (i.e. they worked on different objects).

It also fixed a minor oversight (> should be >=), which sent 5 messages from history to the client, although 4 were configured.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -373,7 +373,7 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
         this.canChangeNickname = MUCPersistenceManager.getBooleanProperty(mucService.getServiceName(), "room.canChangeNickname", true);
         this.registrationEnabled = MUCPersistenceManager.getBooleanProperty(mucService.getServiceName(), "room.registrationEnabled", true);
         // TODO Allow to set the history strategy from the configuration form?
-        roomHistory = new MUCRoomHistory(this, new HistoryStrategy(mucService.getHistoryStrategy()));
+        roomHistory = new MUCRoomHistory(this, mucService.getHistoryStrategy());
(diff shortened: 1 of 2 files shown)


src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -373,7 +373,7 @@ public LocalMUCRoom() {
         this.canChangeNickname = MUCPersistenceManager.getBooleanProperty(mucService.getServiceName(), "room.canChangeNickname", true);
         this.registrationEnabled = MUCPersistenceManager.getBooleanProperty(mucService.getServiceName(), "room.registrationEnabled", true);
         // TODO Allow to set the history strategy from the configuration form?
-        roomHistory = new MUCRoomHistory(this, new HistoryStrategy(mucService.getHistoryStrategy()));
+        roomHistory = new MUCRoomHistory(this, mucService.getHistoryStrategy());
```
tevans: I think each room needs its own instance, at least for the cached room subject property. Additional refactoring will likely be needed to make the history instance shareable across rooms within the MUC service.

sco0ter: Yes, I think you are absolutely right. I've overlooked that. `HistoryStrategy` stores the messages as well. I think to solve it, `roomSubject` and the messages should be attached to `LocalMUCRoom` maybe, and `HistoryStrategy` should only hold information about the actual strategy.

guusdk:
@sco0ter Am I right to conclude that this PR still needs work?

akrherz:
@sco0ter you think you can update this PR prior to us doing a 4.2 release?  We don't have a date in mind, but hopefully sooner than later!

sco0ter:
Sorry folks, I have no time currently for this. We could as well close this one.

akrherz:
closing as per last comment, can reopen if there is a desire to make it happen.

PR #518, opened 2016-01-19
Fix memory leak when getChatRoom() calls return null or throw exception

aberenguel:
Every instantiation of LocalMUCRoom registers itself in the
GroupEventDispatcher. However, when the instantiation is done in the
getChatRoom methods and so after it fails, the room is not returned and
it stays registered in the GroupEventDispatcher. This fix unregister the
room from GroupEventDispatcher in these cases.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -645,6 +645,7 @@ public class MultiUserChatServiceImpl implements Component, MultiUserChatService
                         else {
                             // Room does not exist and delegate does not recognize it and does
                             // not allow room creation
+                            GroupEventDispatcher.removeListener(room);
                             throw new NotAllowedException();
 
                         }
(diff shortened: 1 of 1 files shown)


aberenguel:
You can reproduce the leak creating some MUC rooms via admin console. 
For example, if you create 3 rooms, 6 rooms will be registered in GroupEventDispatcher. after destroying these 3 rooms, 3 rooms objects still stay in GroupEventDispatcher listeners.

guusdk:
Is this a duplicate of https://github.com/igniterealtime/Openfire/pull/512 ?

aberenguel:
Yes it is duplicated issue. 

However I think that the solution does not solve the problem when MUC rooms are added via admin console.

The LocalMUCRoom is added into GroupEventDispatcher listeners in its constructor.
The #512 solution will be OK if the LocalMUCRoom is added in GroupEventDispatcher in the LocalMUCRoomManager.addRoom()

tevans:
Seems safe enough to me, +1

aberenguel:
The commit ae0ffbfd9dd7f347464eb11b9111770a29d95c01 in the #512 fixed this issue.

dwd:
My reading of the last comment is that this can be closed.

PR #521, opened 2016-01-21
OF-1053: Don't use apostrophe in i18n

guusdk:
Replaced all apostrophes with HTML-encoded entities to prevent
formatting problems.


sco0ter:
+1

PR #522, opened 2016-01-21
OF-1046: Improve offline-storage usage

guusdk:
Existing code already tried to prevent message bounces when a stanza was
sent by the server itself. This commit builds on that solution.
Also, debug logging was added for the store-and-* strategies when storage
was no longer possible as a result of quota limitations (we had a hard time
debugging this)


akrherz:
+1

sco0ter:
+1

PR #523, opened 2016-01-21
OF-1045: Fix NPE in cluster info page

tevans:
Ensure cluster listener is properly initialized (and destroyed).


tevans:
Thanks for the feedback ... let me know if you agree this is acceptable as-is.

sco0ter:
Yes, of course, that's fine.

PR #528, opened 2016-01-25
OF-1056: Redirect commons logging over slf4j properly.

guusdk:
Instead of using our proprietary commons logging wrapper for slf4j, use
the one provided by slf4j. Not only does is reduce code, it also does a
better job logging things (using the correct log category).

The jcl-over-slf4j.jar is a drop-in replacement for commons-logging.jar
(for version 1.1.1 commons-logging - our copy was probably older).


sco0ter:
+1

build/lib/versions.txt
```diff
@@ -8,13 +8,13 @@ bcpkix-jdk15on.jar                  | 1.53
 bcprov-jdk15on.jar                  | 1.53                                                         | See http://www.bouncycastle.org/licence.html
 cglib.jar                           | 2.1.3 (JMock 2.1.0)                                          |
 commons-lang.jar                    | 2.3                                                          | Apache 2.0
-commons-logging.jar                 | Jetty 5.1.10                                                 | Apache 2.0
```
tevans: I think that the `CommonsLogFactory` class still depends on this library. If we remove this, will we break the build?

guusdk: No, it will not (Travis successfully builds the build, which is a good indication that the build isn't broken ;) ).  The API signature provided by jcl-over-slf4j.jar is similar to commons-logging 1.1.1.

PR #530, opened 2016-01-27
OF-1063: Prevent TimerThread from exiting

tevans:
Do not allow TimerThread to exit when the server becomes heavily loaded,
preferring instead to use the caller's thread to execute a scheduled
task when needed. This may result in a temporary slowdown, but is better
than a hard failure which requires a restart of the server.

### DIFF
--- a/src/java/org/jivesoftware/util/TaskEngine.java
+++ b/src/java/org/jivesoftware/util/TaskEngine.java
@@ -23,7 +23,15 @@ import java.util.Date;
 import java.util.Map;
 import java.util.Timer;
 import java.util.TimerTask;
-import java.util.concurrent.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.ThreadFactory;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Performs tasks using worker threads. It also allows tasks to be scheduled to be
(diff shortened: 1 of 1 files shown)


dwd:
This looks good - has it failed to build due to a Travis bug?

tevans:
Hmm ... yes I have seen intermittent failures of the trusted chain test. I will trigger the build again, see if it clears itself up.

tevans:
@guusdk - Appreciate the thoughtful feedback, agree that in some cases a failed task execution might be preferable to a potential delay or lack of stability. However, it is difficult to determine which tasks would be suitable for failure vs. delay. This approach effectively tells the server to make "best effort" by allowing the caller's thread to handle the requested work, which has the added benefit of slowing down the caller when resources become constrained.

My primary concern is that the Timer thread does not exit, as this creates an error condition that is unrecoverable. In our stress testing with this fix applied, we do see that the server will slow down under extreme loads, but it will now also recover after the load diminishes. At peak loads we observed that new connections were refused (usually via timeout), while the existing connections continued to operate, albeit with somewhat diminished responsiveness.

guusdk:
@tevans this change having proved itself as a fix for the problems that you had is a powerful argument in its favor. I can see how gracefully slowing down the submission of task is a nice auto-throttle, that can be very useful. Still, you haven't won me over on the way the change is implemented.

TaskEngine is a general-purpose implementation. Although the change works for the problem at hand, I am worried that it introduces problems elsewhere.

Concurrency is a complex beast. Changing the asynchronous implementation to an implementation that is asynchronous-expect-for-when-it-isn't is introducing another layer of complexity. Again, something that I fear will sooner or later introduce new issues.

Most of the thread pools in Openfire are size-limited. When it is such a thread that suddenly interacts in a blocking way with the TaskEngine, I fear that the risk of thread starvation or livelock is introduced.

PR #532, opened 2016-02-02
XEP-0016 error messages should set correct from attribute (the sender).

sco0ter:
See XEP-0016, example 50.


dwd:
Looks good to me. 4.0 branch as well?

sco0ter:
It's not critical, but as it's only a minor bug fix, we can put it in there.

tevans:
+1

PR #534, opened 2016-02-05
NullPointerException in RoutingTableImpl.

dchinmay:
In #routeToComponent, it is first checked if the component is being hosted in the local JVM. If it was not found, then all nodes in the ComponentsCache are checked to find if the component is hosted by one of those. In this process, the local routing table is again searched for the component, because it may have been added after the previous check. However, after this search, there is no null check done for the searched route. This results in an NPE when the component was not found locally and the #process(packet) method is called on the object. 

This also prevents the code from searching other nodes that may have been in the nodes set. Added a null check to prevent this scenario.

### DIFF
--- a/src/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/src/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -418,9 +418,12 @@ public class RoutingTableImpl extends BasicModule implements RoutingTable, Clust
 		                // This is a route to a local component hosted in this node (route
 		                // could have been added after our previous check)
 		                try {
-		                    localRoutingTable.getRoute(jid.getDomain()).process(packet);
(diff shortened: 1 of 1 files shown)


dchinmay:
Can anyone look at this request please?

akrherz:
Hi @dchinmay  , thanks so much for your contribution to Openfire!  We are always looking for new volunteers to help out with the code base and would love to hear how Openfire is working out for you. Please feel free to stop by our developers MUC (open_chat@conference.igniterealtime.org) or [here](http://www.igniterealtime.org/support/group_chat.jsp) to say hello.

I have filed [OF-1084](https://igniterealtime.org/issues/browse/OF-1084) to track this change.  Thanks again!

dchinmay:
Thanks @akrherz, I will!

PR #535, opened 2016-02-09
Fix unicode read on input

cliffchung:
In the current read scheme for Openfire, it is possible that due
to variable width encoding that we sometimes only create a
string with a partial character.

This breaks the message text and causes encoding issues when read.

This change modifies the way the characters are
read to use an bytearrayoutputstream and
only do the string conversion once.

Since we only do the string creation once after the full
buffer is read, we no longer have encoding
issues with broken unicode characters.


guusdk:
+1

akrherz:
Filed as https://igniterealtime.org/issues/browse/OF-1082

PR #536, opened 2016-02-09
Cannot join room in a cluster after an availability update.

derekmcl:
When a room occupant sends an availability update to a host in an
Openfire cluster, other users cannot join the room from other hosts in
the cluster. The availability update causes the other hosts to lose an
occupant's role and affiliation. New occupants encounter an NPE when
joining and are prevented from joining the room. Specifically, the NPE
occurs when Openfire attempts to send initial presences for current
occupants.

Remote hosts in the cluster lose the occupant's association because the
local room simply broadcasts the presence packet for the occupant's
availability update. Because it is from the client, this packet does not
have the occupant's role or affiliation. The remote hosts treat the
presence packet as the occupant's presence without modification.

This fix changes the order in which Openfire handles an availability
update. First, it updates its local view of an occupant's presence. This
populates the correct association. Then, it broadcasts the updated
presence to remote hosts in the cluster.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -1869,19 +1869,34 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
         return lockedTime > 0 && creationDate.getTime() != lockedTime;
     }
 
+    /**
+     * Handles occupants updating their presence in the chatroom. Assumes the user updates their presence whenever their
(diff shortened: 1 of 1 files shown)


akrherz:
Hi @derekmcl , thanks so much for your contribution to Openfire!  We are always looking for new volunteers to help out with the code base and would love to hear how Openfire is working out for you. Please feel free to stop by our developers MUC (open_chat@conference.igniterealtime.org) or [here](http://www.igniterealtime.org/support/group_chat.jsp) to say hello.

I have filed [OF-1083](https://igniterealtime.org/issues/browse/OF-1083) to track this change.  Thanks again!

PR #538, opened 2016-02-12
Set affiliation to 'none' after removing registration from room.

dchinmay:
Removing registration from room calls #addNone, which removes the requester
JID from all affiliation lists. But #applyAffiliationChange sets the affiliation
back to 'member' after it does not find the JID in any affiliation lists. It
should set the affiliation to 'none' before sending the updated presences.


dchinmay:
Ping!

akrherz:
Thanks, I filed [OF-1091](https://igniterealtime.org/issues/browse/OF-1091) to track this PR.

PR #539, opened 2016-02-15
OF-1087: Default namespace handling in BOSH.

guusdk:


sco0ter:
I really dislike(d) the old and new XML string manipulation. It's fragile and ugly (as can be now be seen with these fixes). Isn't there a clever way to do that in a clean way by working with dom4j elements?

The same logic applied to the BOSH implementation should also be fixed in the WebSocket implementation.

I also think there are more occurrences of `addNamespace`, which might cause trouble then.

PR #541, opened 2016-02-15
Outcasts should not be allowed to register with room.

dchinmay:
With the current implementation, any user can send the register IQ to register with
a room as a 'member'. This also allows outcasts, that were banned from a room, to be
able to register again as a member, and eventually re-join the room. The solution is
to prevent outcasts from registering with the room.


akrherz:
I filled [OF-1090](https://igniterealtime.org/issues/browse/OF-1090) on this, thanks for reporting

PR #543, opened 2016-02-16
Fixing NPE in RemoteSession

cliffchung:
The doSynchronousClusterTask method can return a null when the cluster
node info can't be found.  This will cause an NPE in some methods that
try to autobox the null to a primitive.

This change forces some default for those values that return primitive
types.


cliffchung:
*ping*

guusdk:
Is generating default values in such case wise? If something is off, it's not unreasonable that exceptions are raised. Perhaps we should improve exception handling instead of generating default values?

cliffchung:
I guess it seemed better to generate a default value rather than throw an NPE.  In the other methods that use strict objects we actually just return null.

Exception handling is also reasonable but its a much bigger change since we'd need to modify every method that calls these ones.

guusdk:
@tevans You know the hazelcast stuff inside out. What's your take?

tevans:
I'm OK with this proposed change. +1

PR #545, opened 2016-02-20
OF-210 - Support for Roster Versioning

aberenguel:
I implemented the Roster versioning ( http://xmpp.org/rfcs/rfc6121.html#roster-versioning ) so that if the roster version is not updated, it returns an empty IQ-result.

I tried to follow the spec in all aspects, and added comments copied from the spec.

The spec says:
"_In general, unless returning the complete roster would (1) use less bandwidth than sending individual roster pushes to the client (e.g., if the roster contains only a few items) or (2) the server cannot associate the version ID with any previous version it has on file, the server SHOULD send an empty IQ-result and then send the modifications (if any) via roster pushes._"

In this implementation, we have the case that "_the server cannot associate the version ID with any previous version it has on file_". So there is no way to send the modifications via roster pushes.

### DIFF
--- a/src/java/org/jivesoftware/openfire/roster/Roster.java
+++ b/src/java/org/jivesoftware/openfire/roster/Roster.java
@@ -34,7 +34,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.apache.commons.lang.StringUtils;
 import org.jivesoftware.database.JiveID;
 import org.jivesoftware.openfire.PresenceManager;
 import org.jivesoftware.openfire.RoutingTable;
@@ -700,13 +699,13 @@ public class Roster implements Cacheable, Externalizable {
         if (RosterManager.isRosterVersioningEnabled()) {
             try {
                 synchronized (this) {
-                    User user = UserManager.getInstance().getUser(username);
-                    String rosterVersionStr = user.getProperties().get("roster.version");
-                    long rosterVersion = StringUtils.isNumeric(rosterVersionStr) ? Long.parseLong(rosterVersionStr) : 0;
-                    rosterVersion++;
-                    user.getProperties().put("roster.version", Long.toString(rosterVersion));
+                    String latestRosterVersion = getLatestRosterVersion();
+                    String newRosterVersion = Long.toString(Long.parseLong(latestRosterVersion) + 1);
 
-                    roster.getChildElement().addAttribute("ver", Long.toString(rosterVersion));
+                    User user = UserManager.getInstance().getUser(username);
+                    user.getProperties().put("roster.version", newRosterVersion);
+
+                    roster.getChildElement().addAttribute("ver", newRosterVersion);
                 }
             } catch (UserNotFoundException e) {
                 // should not happen
@@ -1173,19 +1172,9 @@ public class Roster implements Cacheable, Externalizable {
         ExternalizableUtil.getInstance().readExternalizableMap(in, rosterItems, getClass().getClassLoader());
         ExternalizableUtil.getInstance().readStringsMap(in, implicitFrom);
     }
-    
-    public boolean hasRosterVersionBeenModified(String requestedVersion) {
-        if (StringUtils.isBlank(requestedVersion))
-            return true;
-
-        String currentVersion = User.getPropertyValue(username, "roster.version");
-        if (StringUtils.isBlank(currentVersion))
-            return true;
-
-        return !requestedVersion.equals(currentVersion);
-    }
 
     public String getLatestRosterVersion() {
-        return User.getPropertyValue(username, "roster.version");
+        String latestVersion = User.getPropertyValue(username, "roster.version");
+        return latestVersion != null ? latestVersion : "0";
     }
 }
--- a/src/java/org/jivesoftware/openfire/roster/Roster.java
+++ b/src/java/org/jivesoftware/openfire/roster/Roster.java
@@ -691,25 +691,15 @@ public class Roster implements Cacheable, Externalizable {
     private void broadcast(org.xmpp.packet.Roster roster) {
         JID recipient = server.createJID(username, null, true);
         roster.setTo(recipient);
-        if (sessionManager == null) {
-            sessionManager = SessionManager.getInstance();
-        }
+
         // When roster versioning is enabled, the server MUST include 
         // the updated roster version with each roster push.
         if (RosterManager.isRosterVersioningEnabled()) {
-            try {
-                synchronized (this) {
-                    String latestRosterVersion = getLatestRosterVersion();
-                    String newRosterVersion = Long.toString(Long.parseLong(latestRosterVersion) + 1);
-
-                    User user = UserManager.getInstance().getUser(username);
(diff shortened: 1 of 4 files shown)


aberenguel:
A fix was necessary in the RosterManager, so that the broadcast method is always called when there should be a roster push. (commit b058d77)

When the roster is updated via group renaming, group user adding or removing, the roster pushes only happen if there is a cache hit. If there is a cache miss (what can happen if the cache is full or if the admin cleaned up the cache) the user is not properly notified about the roster update. Thus only update rosters in memory can lead to this undesired behavior.

This commit avoids the use of the cache directly (where there can be a cache miss or a cache hit). It is using the method getRoster(username) that instantiante a new Roster in the case of a cache miss.

akrherz:
Hi @aberenguel , we are sorry that this PR got missed prior to 4.1 Release :(  Do you have any updates for it prior for consideration into 4.2 ?

akrherz:
@dwd, You have thoughts on this PR?

akrherz:
@dwd we really need a verdict / plan for this PR prior to getting close on 4.2 release, any ideas?

akrherz:
@guusdk do you have thoughts on what to do with this PR?

src/java/org/jivesoftware/openfire/roster/Roster.java
```diff
@@ -1153,4 +1162,21 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
         ExternalizableUtil.getInstance().readExternalizableMap(in, rosterItems, getClass().getClassLoader());
         ExternalizableUtil.getInstance().readStringsMap(in, implicitFrom);
     }
+
```
guusdk: I'm wondering if using a hash of the roster would be a suitable, less complex, less error prone alternative to maintaining manual bookkeeping of roster versions. By re-calculating a hash (implementing hashCode and equals for the Roster class, perhaps), the 'version' of the roster becomes a natural property of the roster. The chances of hash collisions are surely negligible, while the overhead of (re)calculation can probably be solved by a lazy-loaded field.

src/java/org/jivesoftware/openfire/roster/RosterManager.java
```diff
@@ -77,6 +77,15 @@ public static boolean isRosterServiceEnabled() {
         return JiveGlobals.getBooleanProperty("xmpp.client.roster.active", true);
     }
 
+    /**
+     * Returns true if the roster versioning is enabled.
+     *
+     * @return true if the roster versioning is enabled.
+     */
+    public static boolean isRosterVersioningEnabled() {
+        return JiveGlobals.getBooleanProperty("xmpp.client.roster.versioning.active", false);
```
guusdk: Why not have a default value of `true`? It'll only be used by clients that understand the protocol anyways, right?

src/java/org/jivesoftware/openfire/roster/RosterManager.java
```diff
@@ -354,11 +363,13 @@ else if ("sharedRoster.displayName".equals(keyChanged)) {
...
-                if (roster != null) {
-                    // Update the roster with the new group display name
-                    roster.shareGroupRenamed(users);
+                    try {
+                        roster = getRoster(updatedUser.getNode());
+
+                        // Update the roster with the new group display name
+                        roster.shareGroupRenamed(users);
+                    } catch (UserNotFoundException e) {
+                    }
```
guusdk: Please don't catch exceptions without doing anything with them. Even if these exceptions cannot happen, chances are that some future code change might (by accident) cause them to trigger. At the very least log an exception, but if the exception really is "impossible", apply the fail-fast strategy: rethrow it, or wrap it in a runtime exception that's thrown.

src/java/org/jivesoftware/openfire/roster/RosterManager.java
```diff
@@ -643,7 +650,10 @@ private void groupUserAdded(Group group, Collection<JID> users, JID addedUser) {
         // Get the roster of the added user.
         Roster addedUserRoster = null;
         if (server.isLocal(addedUser)) {
-            addedUserRoster = rosterCache.get(addedUser.getNode());
+            try {
+                addedUserRoster = getRoster(addedUser.getNode());
+            } catch (UserNotFoundException e) {
+            }
```
guusdk: Please don't catch exceptions without doing anything with them. Even if these exceptions cannot happen, chances are that some future code change might (by accident) cause them to trigger. At the very least log an exception, but if the exception really is "impossible", apply the fail-fast strategy: rethrow it, or wrap it in a runtime exception that's thrown.

src/java/org/jivesoftware/openfire/roster/RosterManager.java
```diff
@@ -652,23 +662,24 @@ private void groupUserAdded(Group group, Collection<JID> users, JID addedUser) {
...
+
                 // Check if the roster is still not in memory
                 if (addedUserRoster == null && server.isLocal(addedUser)) {
-                    addedUserRoster =
-                            rosterCache.get(addedUser.getNode());
+                    try {
+                        addedUserRoster =
+                                getRoster(addedUser.getNode());
+                    } catch (UserNotFoundException e) {
+                    }
```
guusdk: Please don't catch exceptions without doing anything with them. Even if these exceptions cannot happen, chances are that some future code change might (by accident) cause them to trigger. At the very least log an exception, but if the exception really is "impossible", apply the fail-fast strategy: rethrow it, or wrap it in a runtime exception that's thrown.

src/java/org/jivesoftware/openfire/roster/RosterManager.java
```diff
@@ -712,32 +723,28 @@ private void groupUserDeleted(Group group, Collection<JID> users, JID deletedUse
         // Get the roster of the deleted user.
         Roster deletedUserRoster = null;
         if (server.isLocal(deletedUser)) {
-            deletedUserRoster = rosterCache.get(deletedUser.getNode());
+            try {
+                deletedUserRoster = getRoster(deletedUser.getNode());
+            } catch (UserNotFoundException e) {
+            }
```
guusdk: Please don't catch exceptions without doing anything with them. Even if these exceptions cannot happen, chances are that some future code change might (by accident) cause them to trigger. At the very least log an exception, but if the exception really is "impossible", apply the fail-fast strategy: rethrow it, or wrap it in a runtime exception that's thrown.

src/java/org/jivesoftware/openfire/roster/Roster.java
guusdk: As Dave (an author of the original specification) mentions: "XEP-0237 details exactly that as a sane implementation strategy: https://xmpp.org/extensions/xep-0237.html#impl-exactmatch "

akrherz:
@aberenguel we are rapidly approaching a beta for 4.2 release.  Are you able to address Guus's feedback?  Sorry again about the delays with looking at this...

akrherz:
obsoleted by #923, thanks @aberenguel for getting this going, if you can test current master, that would be great

guusdk:
We are partially rolling back changes in this commit, as they appear to be causing severe performance issues. See https://issues.igniterealtime.org/browse/OF-1904 for details.

PR #546, opened 2016-02-25
Disallow addMember request with empty nicknames in rooms with reserved nicknames.

dchinmay:
If there is an addMember request, say from a room admin, with an empty nickname,
then an empty string is assigned as the nickname. For rooms with reserved nicknames,
this will create problems for the following scenario - A room admin grants membership
to a user JID, but with an empty nickname. The user then requests to join the room
with a non-empty nickname, but the room does not allow that since the reserved nickname
does not match with the provided one.


dchinmay:
Oops, that was sloppy. As my comment says, I intended to do nickname == null, not nickname != null. Thanks for pointing it out. I will fix it.

dwd:
This needed a manual merge, done so as #699 - sorry it took so long to get to this.

PR #547, opened 2016-02-26
Admin Console: prevent NPE on missing user creation date.

guusdk:
Some user providers might not provide a creation date for each user. There's no reason that the Admin Console should not be able to cope with that.


akrherz:
[OF-1093](https://igniterealtime.org/issues/browse/OF-1093)

PR #548, opened 2016-02-26
OF-1092: Pluggable SASL mechanism

guusdk:
A bit of rework of the SASL implementation. This commit changes the signature of SASLAuthentication,
so that it can be used to register/deregister new SASL mechanisms (represented by the new interface
SaslMechanism). Existing mechanisms have been pulled out of the original, bulky SASLAuthentication
class.

### DIFF
--- a/src/java/org/jivesoftware/openfire/net/SASLAuthentication.java
+++ b/src/java/org/jivesoftware/openfire/net/SASLAuthentication.java
@@ -20,26 +20,6 @@
 
 package org.jivesoftware.openfire.net;
 
-import java.io.UnsupportedEncodingException;
-import java.net.UnknownHostException;
-import java.nio.charset.StandardCharsets;
-import java.security.KeyStore;
-import java.security.Security;
-import java.security.cert.Certificate;
-import java.security.cert.X509Certificate;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Set;
-import java.util.StringTokenizer;
-import java.util.TreeMap;
-import java.util.regex.Pattern;
-
-import javax.security.sasl.Sasl;
-import javax.security.sasl.SaslException;
-import javax.security.sasl.SaslServer;
-
 import org.dom4j.DocumentHelper;
 import org.dom4j.Element;
 import org.dom4j.Namespace;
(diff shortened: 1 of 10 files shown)


sco0ter:
Why the new `SaslMechanism` interface? I am pretty sure we can use the Java standard `javax.security.sasl.SaslServer` for that.

We have implemented custom SASL mechanisms in the past using a custom implementation of `SaslServer` and then using `SASLAuthentication.addSupportedMechanism`.
It worked fine, but doesn't seem to work any longer.

I also suggest to deprecate the old method (`addSupportedMechanism(String)`) instead of removing it.

But my main concern is that I'd rather use `javax.security.sasl.SaslServer` instead. It's the standard interface for SASL and has served well.

There are already separate implementations, e.g. `SaslServerPlainImpl`, which implement `SaslServer`.
Why introduce a new abstraction/wrapper for them (`SaslProviderBasedMechanism`)? At the first glance it adds more complexity and forces developers to use a non-standard interface for SASL.

guusdk:
When you're right, you're right. It's a lot cleaner without the additional layer. I originally had that in place, as I am working on a custom ```SaslServer``` that needs configuration before being used - something I cannot do in the ```SASLAuthentication``` as it's very Provider-specific, but also wouldn't fit in the provider itself. The "middle layer" was born that way. Revisiting that solution, it seemed over-generic and over-configurable. I ripped it out.

I've added a new commit that transforms my earlier changes into standard ```SaslServer``` implementations (I can squash later - for the sake of this discussion, I thought it would be useful to keep the history). 

I also rebased, which introduced commit 5ec02ac.

sco0ter:
Thanks for taking care! I think it looks much better now (and even more so after squashing).

One last thing I've missed in my custom SASL implementation, is that any Exception thrown by `evaluateResponse` is treated as "not-authorized". There was no way to tell Openfire that it actually should be "account-disabled" (as an example), nor to pass a custom failure text to the client.

My idea back then, was to have an exception which derives from `SaslException` and which Openfire can understand.
You now introduced such an exception (`SaslFailureException`), which is a great idea, it only should derive from `SaslException` so that custom `SaslServer#evaluateResponse` can throw it and return a specific failure type + text to the client.

guusdk:
Like so?

guusdk:
@sco0ter I'm somewhat anxious to get this merged (I'm juggling with some branches that depend on each-other). Any last remarks?

sco0ter:
Yes, looks fine. Maybe consider squashing the commits.

guusdk:
Squashed.

sco0ter:
A bit awkward to review, but looks well I think. Maybe in a future commit we can also pass a failure string (SaslFailureException#getMessage()??) to authenticationFailed() and to the the client.

guusdk:
Yes. In a future commit. ;)

Thanks for your thorough reviewing!

PR #552, opened 2016-03-02
OF-1095: Improve plugin (un)loading related logging

guusdk:
This PR fixes a couple of gripes I had with the way that logging occurred related to plugins.


akrherz:
Not to be picky, but I would have preferred to see the log4j update handled in its own ticket, so to help the changelog.

guusdk:
I've created OF-1096 and reworded the relevant commit message.

akrherz:
@guusdk thank you!

PR #555, opened 2016-03-03
OF-1100: Allow for subjectAltNames of type 'DNS'

guusdk:
Please refer to OF-1100 for the rationale of this change.

This commit adds a couple of unit tests, of which two tests fail without the changes in this commit.

The way how I construct certificates in the unit tests (notably: the subjectAltName extensions) is copy/pasted from Google results. I'm not sure that this is the correct way of going about. I appreciate feedback on that, as well as generic feedback on the solution itself (although arguably, that should better be added as a comment on OF-1100).

### DIFF
--- a/src/java/org/jivesoftware/util/cert/SANCertificateIdentityMapping.java
+++ b/src/java/org/jivesoftware/util/cert/SANCertificateIdentityMapping.java
@@ -1,5 +1,9 @@
 package org.jivesoftware.util.cert;
 
+import org.bouncycastle.asn1.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.security.cert.CertificateParsingException;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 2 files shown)


guusdk:
@victorqhong-bah As the original author of the code I modified, I'd welcome your feedback on this.

victorqhong-bah:
@guusdk For the purpose of closing OF-1100, this looks good to me.

Though, I know there are other OtherNames used as identities (e.g. UserPrincipalName for client certificate authentication) and the code now doesn't look like it would easily support arbitrary OtherNames. That might not be an issue since users can register their own identity mappings but it might be nice to have.

guusdk:
Thanks, @victorqhong-bah . For client-based certificates, we should have the EXTERNAL SASL mechanism kick in. I would not be surprised if that needs work, but if needed, I prefer to address that in another issue / pull request.

PR #559, opened 2016-03-11
OF-1107 do not expose email, add readme and changelog

wrooot:


guusdk:
11 years. Must be some kind of record!

Perhaps we should have a property that allows you to override the default behavior. People that have had an email for the past 11 years, will probably still want one now :)

Also, @sco0ter will make you squash the three commits into one, or kill a kitten. ;) Use `git rebase -i` which will give you a nice help text. It's easy.

wrooot:
I don't think many are using this plugin. If it's internal company's server, there are probably other ways to find out the email. If it's a public one, then in my opinion it is a breach on privacy. Property would be nice, but i have no experience with coding that.

I'm sorry for 3 commits, but it is not like 3 completely different changes squashed together. Also, i'm not in cmd, so unless someone gives a step by step guide how to rebase in Eclipse and not break my setup completely so i wouldn't have to clone it again..

guusdk:
Oh, no worries. To use a property, simply replace that line in which the
message is composed with these lines:

if ( JiveGlobals.getBooleanProperty( "plugin.emailonaway.exposeemail",
false ) ) {
   message.setBody( "I'm currently away. Your message has been
forwarded to my service email address (" + emailTo + ")." );
} else {
   message.setBody( "I'm currently away. Your message has been
forwarded to my service email address." );
}

That will by default use the 'else' block, but allow you to override the
functionality with that property. Feel free to rename as you see fit.

As for the squashing: I don't use Eclipse myself, but from what I find in
Google, you can simply select all commits in your history, open the context
menu and select 'squash'.

PR #561, opened 2016-03-18
OF-929 prevent ghosts by removing ioSession check

akrherz:
Since the release of Openfire 3.9.3, users have reported problems with
"ghost" sessions left on the server. After reviewing the changes that
went into the 3.9.3 release, it seemed the fix for OF-464 may have had
some side effect causing this.

Since Igniterealtime's Openfire was reproducing OF-829, I did a test
whereby the functional changes of Igniterealtime/Openfire@ad08cae
were reverted.  After 36 hours, there were no ghosts!  Thinking about
this further, I think that only the change made to deliverRawText may
be to blame here.  Typically, this is used as a last gasp attempt
at delivering data to a connection that is closing/about to close.

The hope is that this change does not necessitate reopening OF-464.

### DIFF
--- a/src/java/org/jivesoftware/openfire/nio/NIOConnection.java
+++ b/src/java/org/jivesoftware/openfire/nio/NIOConnection.java
@@ -295,10 +295,6 @@ public class NIOConnection implements Connection {
             IoBuffer buffer = IoBuffer.allocate(4096);
             buffer.setAutoExpand(true);
             try {
-            	// OF-464: if the connection has been dropped, fail over to backupDeliverer (offline)
-            	if (!ioSession.isConnected()) {
-            		throw new IOException("Connection reset/closed by peer");
-            	}
                 buffer.putString(packet.getElement().asXML(), encoder.get());
                 if (flashClient) {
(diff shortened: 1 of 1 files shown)


akrherz:
This patch was not sufficient to stop the ghosts now on ignite, will try this again with `deliver()` check commented out too :(

akrherz:
So, no ghosts on igniterealtime's Openfire with this most recent commit applied.  So decision time!  Sacrifice OF-464 in favour of OF-829?

guusdk:
I'd rather have the fix for the ghost sessions, as that problem is a lot more visible. On top of that: are we sure that the original problem was fixed in the first place?

dwd:
To reiterate what I said in the chatroom:

OF-464 is obviated by XEP-0198, and exists as a result of the Two Generals problem - some kind of gap where messages can be lost is inherent and unavoidable, without something like XEP-0198 working to explicitly find out what traffic got across before the connection was lost.

The original patch would have reduced the possibility of loss, but it couldn't have eliminated it.

tevans:
This is OK by me also. +1

PR #562, opened 2016-03-18
OF-1111 Rethink RPM build and versioning

akrherz:
Ant RPM build configuration is modified such that:

 - rpmbuild arch of x86_64 is now default.  This will require a
   change to the Ignite's Bamboo to specify a 64bit JRE to bundle
 - the rpmbuild arch logic is simplified and one can now specify
   which arch to build for by setting `-Drpmbuild.arch=<arch>`
 - the resulting RPM filenames are changed, such to allow upgrades
   between alpha builds, beta/rc builds and GA builds without
   having to `rpm --force` these.  See note below on the new
   RPM filenames.
 - the changelog field in the RPM will contain the hash of the
   last commit in the local repository `git rev-parse --short HEAD`
 - A new `git.revision` Ant target is introduced so support the
   inclusion of this in the build metadata

New RPM filenames:
    For General Releases we get x.y.z-1
    For Alpha builds we get     x.y.z-0.1.{YYYYMMDD}alpha
    For Beta/RC builds we get   x.y.z-0.2.(beta|rc)

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -256,7 +256,27 @@
     <!-- TARGETs                                                                                 -->
     <!-- ======================================================================================= -->
 
-    <!-- init ================================================================================== -->
+	<!-- git.revision ========================================================= 
+	     http://stackoverflow.com/questions/2974106
+	-->
+	<available file=".git" type="dir" property="git.present"/>
+
+	<target name="git.revision" description="Store git revision in ${repository.version}" if="git.present">
+	    <exec executable="git" outputproperty="git.revision" failifexecutionfails="false" errorproperty="">
+	        <arg value="rev-parse"/>
+	        <arg value="--short"/>
+	        <arg value="HEAD"/>
+	    </exec>
+	    <condition property="repository.version" value="${git.revision}" else="unknown">
+	        <and>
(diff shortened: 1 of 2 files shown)


akrherz:
@dwd, could I get a verdict on this?  I'd like to follow up with a PR for better systemd support

guusdk:
I'm not knowledgeable enough to evaluate of the changes have the desired effect, but I'm in favor of the intended change.

dwd:
I did briefly wonder if it was worth bumping the Epoch number, but decided it's not (and if I'm wrong, it's easy to sort out later)

akrherz:
My experience with bumping the Epoch is that it should be avoided at all costs.  It is an absolute desperate measure when all else fails.  The reason being that every other downstream RPM builder would need to bump too, to prevent our RPMs from trumping theirs.  I may be incorrect about this.

akrherz:
For RPM users, one item to note...  Since the current openfire is distributed as i386 and this new one is either noarch or x86_64, you need to provide rpm a flag to get it to cleanly upgrade.  It appears `rpm -Uvh --ignorearch` will cleanly upgrade your pre 4.1 version to 4.1

PR #563, opened 2016-03-20
Adding indexes to PostgreSQL for message search query improvement

jcalderonzumba:
This is my first commit for this project so i do not know much about the rules when it comes to database changes.

The objective of this pull request is to improve the queries done on a PostgreSQL setup for the message archives.

I have checked that there are several WHERE clauses to the fields fromJID, toJID of the ofMessageArchive table.

The indexes i have added should improve performance on PostgreSQL setup and WILL avoid sequential scan over those fields. Sequential scan are bad when you have lots of messages, rooms and users, which is our company case.

The diff shows changes to the entire file but that is because the file was filled with "^M" chars at the end of every line.

The only changes are: 
CREATE INDEX ofMessageArchive_fromjid_idx ON ofMessageArchive (fromJID);
 CREATE INDEX ofMessageArchive_tojid_idx ON ofMessageArchive (toJID);

### DIFF
--- a/src/plugins/monitoring/changelog.html
+++ b/src/plugins/monitoring/changelog.html
@@ -43,6 +43,10 @@
 <h1>
 Monitoring Plugin Changelog
 </h1>
+<p><b>1.5.3</b> -- March 22, 2016</p>
+<ul>
+    <li>[<a href='https://igniterealtime.org/issues/browse/OF-1117'>OF-1117</a>] - Improve performance of monitoring plugin by adding database indexes.</li>
+</ul>
 
 <p><b>1.5.2</b> -- Feb 15, 2016</p>
 <ul>
--- a/src/plugins/monitoring/plugin.xml
+++ b/src/plugins/monitoring/plugin.xml
@@ -5,11 +5,11 @@
     <name>Monitoring Service</name>
     <description>Monitors conversations and statistics of the server.</description>
     <author>Jive Software</author>
-    <version>1.5.2</version>
+    <version>1.5.3</version>
     <date>2/15/2016</date>
     <minServerVersion>4.0.0</minServerVersion>
     <databaseKey>monitoring</databaseKey>
-    <databaseVersion>3</databaseVersion>
+    <databaseVersion>4</databaseVersion>
 
     <adminconsole>
         <tab id="tab-server">
--- a/src/plugins/monitoring/src/database/monitoring_db2.sql
+++ b/src/plugins/monitoring/src/database/monitoring_db2.sql
@@ -1,7 +1,7 @@
 -- $Revision$
 -- $Date$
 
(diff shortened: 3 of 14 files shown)


akrherz:
Greetings.  Thanks for your attempted contribution to Openfire.  Unfortunately, this commit is not sufficient.  Please address the following.

* database changes are made via the upgrade scripts, check out the numbered folders within `src/database/upgrade`
* Your text editor should understand files that use CRLF and not attempt to modify them.  It is a long standing issue of the Openfire code base that such files are mixed together.
* Database changes should be made in a consistent matter to all supported databases, so we have consistent behaviour with each.  You'll need to add these same indexes to the other databases with your commit.

Hopefully this all is not discouraging to your efforts.  Thanks again for your offer to help!

jcalderonzumba:
Not at all @akrherz I will do the needed changes, I just wanted some feedback because i had some doubts about the upgrade procedures.
Thanks for the reply.

jcalderonzumba:
@akrherz I think now is done. Let me know if something else is needed.

guusdk:
@jcalderonzumba Thank you for your modifications. This looks a lot better. I noticed that the index names are pretty long. Did you verify that they're within bounds of what is permissible (I ran into issues with that years ago, which makes me ask :P ). 

Also, could you perhaps squash your commits into one? That would simply look a little nicer. If you're unfamiliar with the procedure, have a look here: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/ It's pretty easy to do.

jcalderonzumba:
@guusdk Commit is squashed, (I'm used to squashing but this time i forgot) also I have done some research and for index names I think we are ok because the limits are as follow:

* DB2 index name limit 128 chars
* mysql 64 chars
* postgresql 64 chars
* oracle 30 chars
* sql server 128 chars

I did not find anything related to hsql but i would assume at least to b e as Oracle with 30 chars.

guusdk:
This looks great, thanks! I only now notice that you're updating a plugin, and not Openfire itself. It would be good to update the changelog file to write down what you've changed (and perhaps include a warning for uses that already have a huge table - adding the index will probably take some time).

Also, you should increase the version number of the plugin in the plugin.xml file. I'd go for a very minor update.

I have created the following JIRA issue for this change: https://igniterealtime.org/issues/browse/OF-1117 You can refer to it in the changelog.

jcalderonzumba:
perfect, I now understand all the changes process, at the begging i just had the fix. But not to worry in my company we follow similar procedures when fixing things that need to go to production systems. So i can related to all the needed changes.

jcalderonzumba:
@guusdk Friendly ping reminder about the issue being done (i think).

guusdk:
Ah, sorry, missed your last commit. yeah, this looks good!

PR #564, opened 2016-03-21
OF-1115 Use Java 8 for Debian packaging

akrherz:
untested by me, just a simple s/7/8/ with the debian build files


guusdk:
This probably helps to prevent OF-1116.

PR #566, opened 2016-03-23
OF-1093: Allow for empty/unknown user creation dates.

guusdk:
Some user providers might not be able to provide a creation date. There's no reason
that the admin page should fail over that.
Note that this commit completes commit 2746a7f409b075bbeb74fba37bda6765a0fe2b6d,
which had an correct but incomplete fix for the same problem.

### DIFF
--- a/src/plugins/search/src/web/advance-user-search.jsp
+++ b/src/plugins/search/src/web/advance-user-search.jsp
@@ -167,7 +167,7 @@
            <%= user.getName() %> &nbsp;
        </td>
        <td width="15%">
-           <%= JiveGlobals.formatDate(user.getCreationDate()) %> &nbsp;
+           <%= user.getCreationDate() != null ? JiveGlobals.formatDate(user.getCreationDate()) : "&nbsp;" %>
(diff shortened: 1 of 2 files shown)


jcalderonzumba:
@guusdk Why not have a default on database? the usual NOT NULL CURRENT_TIMESTAMP stuff? because when you do the "INSERT into ofuser" if the user does not provide creation date it is fair to assume that moment in time as the creation date of the user.

guusdk:
@jcalderonzumba I'm not worried by the users that are provided by our own database. However, Openfire can be configured to retrieve its users from an external source (such as LDAP, Crowd, etc). That framework is pluggable: you can write your own provider. When using that to extract existing users from some external repository, the creationdate might simply not be available. In such case, simply not displaying the date is better than making up a value.

jcalderonzumba:
Ok, i did not remember about that possibility, now i remember you can even write your own authentication provider, etc.

dwd:
Also, legacy data wouldn't have the default anyway.

PR #567, opened 2016-03-23
OF-1118: Don't try to use unsupported encryption settings.

guusdk:
Stored preferences of encryption protocols and/or cipher suites might include
non-supported items. To prevent issues, the list of preferences should be checked
against the currently supported items.


guusdk:
When merged, please cherry-pick for inclusion in 4.0.3.

PR #571, opened 2016-04-06
OF-1122: Restoring GSSAPI SASL support

guusdk:
The changes for OF-1092 accidentally removed a snippet of code that translates
a non-optional part of the GSSAPI SASL configuration. This commit restores that,
with a difference: the configuration is now checked continuously, intead of just
on server startup. This allows for configuration changes that need not be followed
by an Openfire restart.


guusdk:
Tests with speedy appear to be unsuccessful. I'll debug and recreate this PR.

PR #576, opened 2016-04-13
Add support for hMailServer's password salted hash structure

Zepmann:
[hMailServer](https://www.hmailserver.com/) by default does not use plain text passwords or hashed passwords. Instead, it uses salted SHA-256 hashes. With this patch, Openfire's JDBCAuthProvider is able to compare and create hMailServer passwords, to be read from and stored in hMailServer's database using MySQL. This allows a seamless integration with hMailServer for account management with the proper system properties.

[jdbcauth-hmailserver-systemproperties.txt](https://github.com/igniterealtime/Openfire/files/218580/jdbcauth-hmailserver-systemproperties.txt) has a list of system properties with sane values to use with this patch. Replace square brackets and the descriptions they contain with appropriate values. You must edit `jdbcProvider.connectionString`, `jdbcAuthProvider.passwordSQL`, and all `jdbcUserProvider.*` values.

The example properties use HybridProvider to allow admin to login locally, even when the hMailServer database is unavailable for some reason. Note that due to the absence of the keyword 'password', jdbcProvider.connectionString is not encrypted by default and will show a plain text password. Make sure to encrypt that property explicitly. 

Note that the MySQL account (which can be the same account as used by Openfire if it uses the same MySQL database server) will require SELECT access on five columns of hMailServer's hm_accounts table:
- accountaddress (to read the user's email address, XMPP username, and XMPP/mail domain name)
- accountpassword (to read and match the salted hash value upon authentication)
- accountactive (to only work with accounts that are active in hMailServer)
- accountpersonfirstname (to read the user's first name if set in hMailServer, used to compose the user's full name)
- accountpersonlastname (to read the user's last name if set in hMailServer, used to compose the user's full name)

An example to grant these permissions to MySQL user 'openfire' on a locally hosted MySQL instance, where hMailServer's database is named 'hmailserver':
```
GRANT SELECT(`accountaddress`)  ON TABLE `hmailserver`.`hm_accounts` TO 'openfire'@'localhost';
GRANT SELECT(`accountpassword`)  ON TABLE `hmailserver`.`hm_accounts` TO 'openfire'@'localhost';
GRANT SELECT(`accountactive`)  ON TABLE `hmailserver`.`hm_accounts` TO 'openfire'@'localhost';
GRANT SELECT(`accountpersonfirstname`)  ON TABLE `hmailserver`.`hm_accounts` TO 'openfire'@'localhost';
GRANT SELECT(`accountpersonlastname`)  ON TABLE `hmailserver`.`hm_accounts` TO 'openfire'@'localhost';
```

The reason why I do not use POP3AuthProvider is that (1) my hMailServer does not use POP3, (2) POP3AuthProvider uses plain [SASL](https://en.wikipedia.org/wiki/Simple_Authentication_and_Security_Layer) which is not supported by hMailServer, and (3) POP3AuthProvider does not integrate neatly when compared to JDBCAuthProvider and JDBCUserProvider.

### DIFF
--- a/src/java/org/jivesoftware/openfire/auth/JDBCAuthProvider.java
+++ b/src/java/org/jivesoftware/openfire/auth/JDBCAuthProvider.java
@@ -225,6 +225,11 @@ public class JDBCAuthProvider implements AuthProvider, PropertyEventListener {
             }
             return OpenBSDBCrypt.checkPassword(hashed, plainText.toCharArray());
         }
+        if (passwordTypes.get(lastIndex) == PasswordType.hmailserver) {
+            String salttext = hashed.substring(0, 6);
+            String plainTextSaltedHash = salttext.concat(StringUtils.hash(salttext.concat(plainText), "SHA-256")).toLowerCase();
+            return plainTextSaltedHash.equals(hashed);
+        }
 
         return hashPassword(plainText).equals(hashed);
     }
@@ -252,6 +257,11 @@ public class JDBCAuthProvider implements AuthProvider, PropertyEventListener {
                 new SecureRandom().nextBytes(salt);
                 int cost = (bcryptCost < 4 || bcryptCost > 31) ? DEFAULT_BCRYPT_COST : bcryptCost;
                 return OpenBSDBCrypt.generate(password.toCharArray(), salt, cost);
+            case hmailserver:
+                salt = new byte[3];
+                new SecureRandom().nextBytes(salt);
+                String salttext = javax.xml.bind.DatatypeConverter.printHexBinary(salt).toLowerCase();
+                return salttext.concat(StringUtils.hash(salttext.concat(password), "SHA-256")).toLowerCase();
             case plain:
             default:
                 return password;
@@ -416,7 +426,17 @@ public class JDBCAuthProvider implements AuthProvider, PropertyEventListener {
         /**
           * The password is stored as a bcrypt hash.
           */
-        bcrypt;
+        bcrypt,
+        /**
+         * The password is stored as a salted SHA-256 value. 
+         * To be more specific, the password is stored in the format: 
+         * concat("010a0f",sha256(concat("010a0f","plaintextPassword"))) , 
+         * where "010a0f" is a three byte salt in hex text format (lower case), 
+         * "plaintextPassword" is the plain text password, 
+         * and the output is a string of hex text (lower case). 
+         * This is the default format used by hMailServer to store passwords in its database.
+         */
+        hmailserver;
    }
 
     /**


akrherz:
Thanks for your contribution, I have filed [OF-1127](https://igniterealtime.org/issues/browse/OF-1127) to track this.

dwd:
Why should this be in the core? I can entirely understand a plugin - Openfire is specifically designed to allow plugin auth providers. But I don't think we could commit to maintaining this.

guusdk:
Hey, thanks for putting in the effort to contribute. Some questions:

1. Is that particular way of generating a password specific to hMailServer, or can we give it a more generic name?
2. Wouldn't it be more user-friendly to create a HMailServerAuthProvider?

Zepmann:
I do not think hMailServer should get its own AuthProvider plugin. Basically, the JDBCProvider is meant to interact with third-party databases to integrate authentication and user management. hMailServer provides a database and that was the chosen approach.

It is not the choice of JDBCAuthProvider which password storage strategy should be used. Rather, it is up to the stored values that JDBCAuthProvider needs to work with. JDBCAuthProvider misses some functionality regarding how passwords are stored. There is the option to store it as a hash (bad idea due to the existence of rainbow tables), plain text (an even worse idea) and more recently bcrypt. However, support for custom salted hashes is missing.

The method I implemented is specific to how hMailServer stores its passwords. I haven't seen it yet being used by any other product. If you want a more universal approach, there should be some kind of 'saltedhash' option. Extra parameters the user should specify would include:
- The used hash function.
- The structure and steps necessary to generate the salt. For hMailServer the structure is three bytes (24 bit), represented as lower case hex text. Other products might for example use a different salt length and upper case characters instead, maybe not even encoded as text but as raw (binary) data.
- The steps necessary to generate a salted hash. For hMailServer this is simply the salt and plain text password concatenated, but other products might put the salt at the end or process the plain text password first.
- The structure of how all values (salt and salted hash) are stored inside a single or separate database fields. hMailServer stores its salt and salted hash as lower-case hex text, concatenated in a single field. Other products might do it in many different ways (such as adding a separator character between the salt and the salted hash, or a different order in which the values are stores).

So to answer all your questions:
@dwd 
> Why should this be in the core?

I think it should be in core since there is no other place for it. Making a completely different plugin for it is overkill, since 99% of the code would be the same as the current JDCBAuthProvider. However, I do agree that it might be too specific to just add a salted hash function for hMailServer (although the password structure it uses hasn't changed in years).

@guusdk 
> 1. Is that particular way of generating a password specific to hMailServer, or can we give it a more generic name?

This method is specific to hMailServer. It is so specific that it can be referred to as 'a salted hash as used by hMailServer', since I am unaware of different products which do it in the exact same way. A generic 'saltedhash' option might be more useful, but would be quite complex.

> 2. Wouldn't it be more user-friendly to create a HMailServerAuthProvider?

I do not think an authentication provider should be created for every third-party product. Currently, the AuthProviders work from a general perspective to interface with different systems (native: authentication through different operating systems, POP3: authentication against different kinds of mail servers, JDBC: authentication against different kinds of databases). Rather, I think that JDBCAuthProvider misses an essential password storage option ('saltedhash') that is unfortunately quite complex to realize.

Of course, these are just my viewpoints on the matter. There are some problems with Openfire for why I took this approach: POP3 only works with plain SASL (which some mail systems do not support) and JDBCAuthProvider currently is not flexible enough to be configured with salted hashes. I needed something that would integrate with an hMailServer instance. It works with a little bit of added code and by handling everything else through the proper configuration of JDBCProvider. A small code addition might be preferable compared to creating another plugin with much more code to maintain.

godly:
From my point of view the development of the JDBCAuthProvider goes into the wrong direction. We have hash function in StringUtils & ScramUtils and now we add other salted hash functions (started with bcrypt) to this class. Currently this is the most easy option to extend the functionality of the JDBCAuthProvider.

I would prefer a separate class for hash functions and one for formatting and encoding of the hash and the salt. Hash functions should be added to Openfire core while the formatting and encoding should be plugable similar to the AuthProvider. Any thoughts on this?

PS: https://github.com/defuse/password-hashing has some secure methods we may want to add to Openfire.

dwd:
I'm going to close this unmerged.

* I don't think adding an application-specific password hashing method makes good sense in general, unless it's either extremely well-deployed, or a best of breed method - this is neither.
* In this particular case, I think hMailServer would benefit enormously from implementing a SCRAM compatible hash method, which we already support.

@godly  - The SCRAM method is actually based on PBKDF2, with some additions to allow it to be used within the SCRAM challenge-response SASL mechanism. It's basically as secure as @defuse's method at rest, plus it allows for some better on-the-wire security.

Fransebas:
@Zepmann  Hi I need help implementing an AuthProvider that uses salted passwords so I downloaded the source code and modify the jdbcAuthProvider the only problem is that I don't know how to pack the builded program once it's done to install it in my debian server. In other words I have the jar of the modified code but I don't know where to put it and if all the corresponding folders and files are in their right place.

Another option is to make a custom plug-in but I don't know anything about that and I believe it's a lot of work only to make my db work, please help understand what to do.

GregDThomas:
It should be sufficient to drop your .jar file in to openfire/lib - just make sure your classname is sufficiently unique. Then just ensure it's used by by setting `provider.auth.className` to your class.

PR #577, opened 2016-04-14
Update eclipse project file to reflect changed list of jars. 

danielhams:
Verified as building and launching successfully under Eclipse Luna.

Note that this still requires the eclipse "launcher" to be tuned to point to correct Openfire home and libdir plus references to resources like the sidebar.


guusdk:
Thanks for your contribution. Although I dislike having these files in our source in the first place, this change is likely to be an improvement.

PR #579, opened 2016-04-14
Eclipse project: Add extra plugins

danielhams:
Add their source directories and libraries where possible. 

Unable to add:

- clustering - seems dependant on coherence library not in repo
- rayo - duplicate classes cause clashes with jingleNodes and nodejs plugins so I've skipped adding it for now

e.g.

org.xmpp.jnodes.RelayChannel - in both jingleNodes and rayo
org.jitsi.util.OSUtils - in both nodejs and rayo

For the duplicate classes - I understand that plugins are isolated class-loader wise - but the clashes means you can't use a single eclipse project with them in. If these classes can't be moved to a common library the traditional approach with eclipse would be to have individual projects per plugin. Perhaps I should take this discussion to your forums - or you've indicated your reluctance to have such things in the repo, so maybe this is "Good Enough" _tm_.

### DIFF
--- a/build/eclipse/classpath
+++ b/build/eclipse/classpath
@@ -1,29 +1,38 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
 	<classpathentry kind="src" path="src/java"/>
+	<classpathentry kind="src" output="work/plugins-dev/bookmarks/target/classes" path="src/plugins/bookmarks/src/java"/>
 	<classpathentry kind="src" output="work/plugins-dev/broadcast/target/classes" path="src/plugins/broadcast/src/java"/>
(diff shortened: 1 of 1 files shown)


guusdk:
Good enough for me, indeed. :)

Although I don't think it relates to rayo/jingle, it helps to be aware of the 'parent-plugin' structure that plugins optionally have. When you define a parent for a plugin, the 'child' uses the classloader from the parent (which would allow you to circumvent duplicate libraries and such).

Also, generally, I don't like to have the plugin classpath enabled in my IDE by default. As the IDE typically does not limit the usage of libraries to the plugin, you run the risk of using a class provided by a plugin library in another plugin (or even the core). Also, having all plugins loaded can slow down the IDE considerably, as it needs to index a lot more things. But hey, to each his own, and, as I'm not using Eclipse anyway, I'm happy to accept this PR :)

PR #580, opened 2016-04-19
Occupants endpoint for restAPI

sailerinteractive:
I added an occupants endpoint for the restAPI plugin. The current participants endpoint only returns participants (no admins, owners,...)


Redor:
+1

PR #582, opened 2016-04-27
OF-1132: Ensure that retreived messages have a namespace.

guusdk:
If no 'xmlns' is set for the stanza then as per XML namespacing rules it would
inherit the default namespace, which is wrong (see XEP-0297).


thesabbir:
Please merge and release this fix asap :)
Thanks!

PR #583, opened 2016-04-29
Updated jmxweb plugin to be compatible with Openfire 4.x

deleolajide:
I had to use a branch on origin instead of branch of my clone because my clone is too complicated for this simple change. I need the jmxweb plugin for a project with OF 4.0.2


deleolajide:
Thank you :-)

PR #585, opened 2016-05-09
Allow JSP compilation with developmentMode=true

danielhams:
See https://community.igniterealtime.org/thread/58545

When launching openfire inside eclipse with developmentMode=true the JSP pages fail as the web app context is not being correctly initialised.

By switching to JasperInitializer (instead of JettyJasperInitializer) and using the container initializer from jetty-annotations - openfire will correctly compile and display JSP pages with or without developmentMode=true.

Note: This patch includes a change from "index.jsp" to "index.html" as the welcome page - when running in developmentMode=true accessing the root of the admin console will bypass the context security if the JSP page is listed as the welcome page.

I'd appreciate it if someone else would check this isn't breaking things for others (non eclipse users, for example).


danielhams:
Sorry for second commit. Want me to squash to one commit to avoid git log clutter?

akrherz:
@danielhams no worries, we can also squash with the merge action as well.

guusdk:
Hi, thanks for your contribution. Openfire typically ships with pre-compiled JSP pages (when *not* in development mode :P ) Will your change affect that?

Also: that `JettyJasperInitializer` class is used on other places as well. Should those now be modified too?

danielhams:
Hi Daryl + Guus,

> Will your change affect pre-compiled JSP pages

I tested launching openfire.sh from the target/bin directory after performing an "ant openfire" in the build directory and it still runs fine using the precompiled JSPs. I'm guessing this is where the shipped binaries come from.

> That JettyJasperInitializer is used other places...

Ouch, I should have spotted that :-) Good catch.

From a look at the code for those (JmxWebPlugin, HttpBindManager) they don't have any special casing for developmentMode, so they are assuming that everything is precompiled, I guess (and it doesn't look like either have JSPs to compile, just servlets)

You can indeed sensibly argue that using the same container initialiser would be good across the code base - but I'm not sure how to test that those pieces of functionality aren't broken by a change.

One thing I will mention is that when running in developmentMode=true I found that the context initialisation isn't performed _without_ adding the ServletContainerInitializersStartup managed wrapper to the context. I'm not certain if this is necessary or even useful given no special casing for precompiled/source files.

I could feasibly change those places too (to use JasperInitializer) in an additional commit if someone could either test with the change or point me in the direction of how to test and make sure things aren't broken (HttpBind requires a client like Candy via reverse apache I think - not sure what to use for Jmx).

danielhams:
Made the change as above and have validated that http-bind endpoints still work as expected - both when running precompiled (from an "ant openfire" created set of libraries) as well as launched inside eclipse with/without developmentMode=true.

I haven't verified if this breaks the JMXWeb plugin.

guusdk:
I have raised https://igniterealtime.org/issues/browse/OF-1133 to track this change.

guusdk:
I think we can take a calculated risk here. We'll know where to find you. ;-)

PR #593, opened 2016-05-20
OF-1141 Fix default configuration of PEP/PubSub

dwd:
Both PEP and Pubsub configure leaf nodes to store -1 items by default; however
this value is not corrected by the usual post-configuration checks to 1, and
this in turn means that when gathering the items to return, the single
item is removed.

This PR corrects the issue by setting the maxPublishedItems in the default
configuration to 1 instead in both cases.


jmankiewicz:
I suspect this to be the main cause why OMEMO-encryption (which needs PEP for public-key management) does not work with current releases of Openfire.

So: +1

PR #597, opened 2016-06-11
few more memory leak fixes caused by LocalMUCRoom

sara-github:
There is a mem leak fix already submitted:
https://github.com/igniterealtime/Openfire/pull/482

I found couple of other memory leaks caused by LocalMUCRoom and fixed it.

Please take a look.


aberenguel:
The PR #512 also fixed it, but it has not been merged yet.

dwd:
#512 is now merged, so I think this one can be closed.

PR #601, opened 2016-06-18
OF-1153 Optionally allow binding admin console to separate network interface

danielhams:
Uses the config variable "adminConsole.interface". If empty uses the standard interface configuration.

Useful for situations where the console must be bound to a different network interface than the regular openfire server ports (for example inside a DMZ).

### DIFF
--- a/src/java/org/jivesoftware/openfire/spi/ConnectionManagerImpl.java
+++ b/src/java/org/jivesoftware/openfire/spi/ConnectionManagerImpl.java
@@ -76,6 +76,8 @@ public class ConnectionManagerImpl extends BasicModule implements ConnectionMana
         super("Connection Manager");
 
         InetAddress bindAddress = null;
+        InetAddress adminConsoleBindAddress = null;
+
         try
         {
             bindAddress = getListenAddress();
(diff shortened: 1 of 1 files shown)


dwd:
This looks great - I've added http://issues.igniterealtime.org/browse/OF-1153 to track this, can you update to include the ticket ref at the front?

godly:
I'm sorry for commenting late but the name "adminConsole.interface" is misleading. It should be renamed to "adminConsole.address" as the code does not use "NetworkInterface.getByName(_adminConsole.interface_).getInetAddresses();" but "InetAddress.getByName(_adminConsole.interface_)".
Specifying the IP address to use is fine. But with this property name one will assume that 'adminConsole.interface=eth2' works which is not the case.

akrherz:
Hi @godly , I agree with you, but this does match what is found in openfire.xml
```
    <network>
        <interface></interface>
    </network>
```

godly:
I agree @ network.interface="An ip address to bind to. Generally only useful on multi-homed systems." (according to https://community.igniterealtime.org/docs/DOC-1061 )
It was/is already confusing and will not become better with this change. I prefer self-explaining property names. For most users/admins an interface is not an IP address and vice versa.
Here we have the chance to use better names before releasing the next version.
Migrating from 'network.interface' to 'network.address' may be a small challenge if this property is stored in the database.

danielhams:
Hi guys, no problem to do the work that changes the property names to be "address" as godly describes - but doesn't that have a potentially confusing impact for people upgrading?

There could of course be code that looks for the historical "network.interface" before looking for "network.address".

Let me know what your preference is - happy to write another patch if need be.

godly:
Here I'd make sure that "adminConsole.address" is used so we can finish this one.

Modifying "network.address" may have an impact. Personally I'd schedule this change for 5.0 (or 4.1) to keep the minor updates easy.

dwd:
Nobody has ever complained about this before, as far as I can tell.

I would want this left consistent with network.interface.

guusdk:
let's not change "network.interface" - at least not without a bulletproof, fully automatic upgrade. I fully agree that "interface" isn't the semantically correct word here, but I feel that we shouldn't introduce a problem for any existing installations, just to rectify this.

PR #602, opened 2016-06-19
	Optimization ClusterExternalizableUtil writeByteArray performance

wmz7year:
Optimization ClusterExternalizableUtil writeByteArray performance


dwd:
I wonder if @tevans could comment on this - looks fine to me, with the caveat that it also looks like a breaking change.

PR #603, opened 2016-06-20
OF-1152: Ensure that raw XML is logged;

GregDThomas:
o delay adding socket interceptors until the pluginManager is ready
  o Correctly interpret the sent/received messages

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -104,7 +104,7 @@
     <property name="copy.dbscripts" value="true"/>
     <property name="overwrite" value="true"/>
 
-    <property name="installer.install4j.home" value="/home/bamboo/install4j.6.0.4"/>
+    <property name="installer.install4j.home" value="/home/bamboo/install4j"/>
     <property name="installer.src" value="${basedir}/build/installer"/>
     <property name="installer.dest.dir" value="${release.dest.dir}"/>
     <property name="installer.install4j.srcfile" value="${installer.src}/openfire.install4j"/>


guusdk:
Thanks for this! It would be good to have that plugin fixed. Two questions:

1. Does 'addFirst' work when the stream is encrypted (both StartTLS on port 5222 as well as traditional TLS on port 5223)? Logging encrypted data isn't going to be very useful.
2. Why wait for the plugins to be loaded upon start? Plugins can be removed, added, or reloaded at any time after the initial boot anyway. Do we *need* to wait for something, or is that an optimization of sorts only?

GregDThomas:
> Does 'addFirst' work when the stream is encrypted (both StartTLS on port
5222 as well as traditional TLS on port 5223)? Logging encrypted data isn't
going to be very useful.

Sorry, I'm not too sure. However, previously it was "addBefore("xmpp" ..."
which was OK for received data, but sent data received an empty buffer. I
can only assume that a previous interceptor was doing that (not sure which
one), so the XML debugger had to get in there first.

> Why wait for the plugins to be loaded upon start?

connManager.getSocketAcceptor() returns null if the plugin manager hasn't
finished loading (in ConnectionManagerImpl the method startListeners()
won't run until the plugin manager completes).

Greg

GregDThomas:
NB. a previous version running against OF 3.6.4, probably 1.1.0 or 1.2.0, also logged the compressed data so I suspect functionality hasn't changed in this regard.

guusdk:
Fair enough. I'd be happy to merge this as-is. I'll leave the PR open for a bit to give others a chance to review. By-the-by: You're still at 3.6.4? Wowza.

GregDThomas:
> By-the-by: You're still at 3.6.4? Wowza

Well, if it ain't broke ... :)

guusdk:
> Well, if it ain't broke ... :)

[Well...](http://issues.igniterealtime.org/issues/?jql=resolution%20%3D%20Fixed%20AND%20fixVersion%20%3E%20%223.6.4%22%20AND%20project%20%3D%20Openfire )

guusdk:
I know we're slow to merge, but I'd still value the change. Can you reopen? :)

GregDThomas:
Apologies, that was my fault, I somehow managed to delete the branch when I created one for OF-1156. Unfortunately, the "Reopen" button is disabled for me :(

GregDThomas:
(Of course, I can't re-open the pull-request as the branch is deleted; I've created a new one at https://github.com/igniterealtime/Openfire/pull/610 - and linked back to this one)

PR #608, opened 2016-07-02
Going back to Java7 for Debian  OF-1115

akrherz:
Will build on bamboo + debian with java7 as well

### DIFF
--- a/build/debian/control
+++ b/build/debian/control
@@ -2,14 +2,14 @@ Source: openfire
 Section: net
 Priority: optional
 Maintainer: Ignite Realtime Community <admin@igniterealtime.org>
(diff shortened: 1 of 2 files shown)


akrherz:
I have adjusted the bamboo debian host to build JRE 1.7 (openjdk) now as well.  @Guusdk this is what we 'agreed' on, right?

guusdk:
Java 7, yes. I don't think we've agreed on the specific make/model of the JRE. So far, we've pushed the Oracle one. As a good chunk of our users will go with OpenJDK, it's not a bad idea to use that too, though.

Perhaps we should even consider testing with both OpenJDK as well as Oracle's - but that's another story.

PR #609, opened 2016-07-04
OF-1156: Be consistent with the contents of DefaultCache with the HZ cache

GregDThomas:
Hazelcast clustered cache implemented by MapProxyImpl by preventing null
keys and values being added. Also document this behaviour in the Cache
interface.

### DIFF
--- a/src/java/org/jivesoftware/util/cache/Cache.java
+++ b/src/java/org/jivesoftware/util/cache/Cache.java
@@ -30,7 +30,7 @@ package org.jivesoftware.util.cache;
  *
  * If the cache does grow too large, objects will be removed such that those
  * that are accessed least frequently are removed first. Because expiration
- * happens automatically, the cache makes <b>no</b> gaurantee as to how long
+ * happens automatically, the cache makes <b>no</b> guarantee as to how long
  * an object will remain in cache after it is put in.<p>
  *
  * Optionally, a maximum lifetime for all objects can be specified. In that
@@ -40,6 +40,9 @@ package org.jivesoftware.util.cache;
  * information from a database.<p>
  *
  * All cache operations are thread safe.<p>
+ *
+ * Note that neither keys or values can be null; A {@link NullPointerException}
+ * will be thrown attempting to place or retrieve null values in to the cache.
  *
  * @see Cacheable
  */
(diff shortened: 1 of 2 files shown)


guusdk:
Thanks! I've talked this over with Dave - he agreed that this seems to be a sensible improvement, even though it might break code that depends on null-keys for some reason (which would be third party code, and an edge-case at best, as far as I can tell now).

dwd:
Turns out that null LockOutFlags are stored against users (including the admin user during setup). This breaks setup... We can fix it case-by-case, of course, but we might need to revisit this.

GregDThomas:
Sounds to me like this re-enforces the case for consistency. If an attempt is made to store null LockOutFlags in a clustered setup the this will fail, and should be fixed. It's just being spotted quicker now, as the same result happens in a non-clustered environment.

dwd:
I don't know - but we need to be very careful before we release with this patch in, since breaking everything and finding out later isn't really an option. You get a null LockOutFlags object stored with pretty well every authentication, I think, so in theory I think the admin console would have broken and no users would be able to log into a cluster if what you're saying is literally true. Hence the suggestion that we might need to revisit this and see exactly what's going to break - maybe log for now and review later?

GregDThomas:
> You get a null LockOutFlags object stored with pretty well every authentication

Really? There may be an attempt to store them, but in a HZ cluster it will fail with a NPE. This is documented behaviour; http://docs.hazelcast.org/docs/latest/javadoc/com/hazelcast/core/IMap.html 

> This class does not allow null to be used as a key or value.

I'd say that the patch has had the desired effect by highlighting where a Cache is used inappropriately.

dwd:
It's causing the current trunk to be unusable.

That's not desirable.

GregDThomas:
HI wonder if this problem (in a non-clustered environment) is the same problem as that seen by https://community.igniterealtime.org/message/259162 - i.e. a null entry in a DefaultCache is preventing if from being updated to a ClusteredCache. How do you reproduce the problem you're seeing? I'll see if I can fix the one you're seeing.

guusdk:
Let's indeed take a step back, and retry a solution. For now, I've created a PR (#625) to revert this change, and re-opened OF-1156.

akrherz:
paging @tevans on this discussion :)

guusdk:
Note that OF-1167 is caused by the original solution. #625 prevents it from occurring.

guusdk:
Dave did #631 - does that prevent the issue that caused me to revert?

GregDThomas:
Just clearing up my backlog; now that Openfire is no longer attempting to store null in the Cache, can this be re-applied, or is the opinion now that it should be restricted to just a log message? My fear with that it won't ever be noticed until too late - at least with a NPE issues such as that in #631 are picked up pretty quickly.

PR #610, opened 2016-07-04
OF-1152: Ensure that raw XML is logged;

GregDThomas:
o delay adding socket interceptors until the pluginManager is ready
  o Correctly interpret the sent/received messages


GregDThomas:
NB. A second pull request as the original at https://github.com/igniterealtime/Openfire/pull/603 was deleted due to a PICNIC problem.

PR #612, opened 2016-07-05
max_items reading is fixed

amaembo:
This PR fixes reading of max_items parameter: currently it silently ignored due to typo in conditional statement.


akrherz:
I've filed https://issues.igniterealtime.org/browse/OF-1157

PR #613, opened 2016-07-08
Denote Openfire 4.0.3 Release

akrherz:
Lets end the pain that is the daily open_chat/forums complaints about debian 4.0.2 artifacts.  OF-477 remains open and we can turn around a 4.0.4 once somebody wants to own that change...


akrherz:
PR updated, awaiting Travis blessing...

PR #621, opened 2016-07-26
OF-1146 Update presence after setting a MUC role.

sco0ter:
This is related to f2386cc, where affiliation updates were not broadcast correctly to MUC occupants.

The error here is, when an owner/admin assigns the moderator role, the new role is not reflected in the presence broadcast.
So neither the actor, nor all other particpants get notifications about the new moderator.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRole.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRole.java
@@ -162,16 +162,7 @@ public class LocalMUCRole implements MUCRole {
         }
         this.presence = newPresence;
         this.presence.setFrom(getRoleAddress());
-        if (extendedInformation != null) {
-            // Remove any previous extendedInformation, then re-add it.
-            Element mucUser = presence.getElement().element(QName.get("x", "http://jabber.org/protocol/muc#user"));
-            if (mucUser != null) {
-                // Remove any previous extendedInformation, then re-add it.
-                presence.getElement().remove(mucUser);
-            }
-            Element exi = extendedInformation.createCopy();
-            presence.getElement().add(exi);
-        }
+        updatePresence();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -1439,6 +1439,8 @@ private Presence changeOccupantRole(JID jid, MUCRole.Role newRole) throws NotAll
             if (role.isLocal()) {
                 // Update the presence with the new role
                 role.setRole(newRole);
+                // Set the new presence, so that the updated role is reflected in the presence stanza.
+                role.setPresence(role.getPresence());
```
dwd: I see this works because setPresence() rebuilds LocalMUCRole.presence with the new extendedInformation - is it worth causing this to happen within setRole()/setAffiliation() instead?

sco0ter: Yes probably better. I can see if I can improve it.

sco0ter:
Improved logic by updating the presence from within `setRole` and `setAffiliation`

src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRole.java
```diff
@@ -162,16 +162,7 @@ public void setPresence(Presence newPresence) {
...
         this.presence = newPresence;
         this.presence.setFrom(getRoleAddress());
-        if (extendedInformation != null) {
-            // Remove any previous extendedInformation, then re-add it.
-            Element mucUser = presence.getElement().element(QName.get("x", "http://jabber.org/protocol/muc#user"));
-            if (mucUser != null) {
-                // Remove any previous extendedInformation, then re-add it.
-                presence.getElement().remove(mucUser);
-            }
-            Element exi = extendedInformation.createCopy();
```
aberenguel: I could fix it just changing this line by
``Element exi = extendedInformation;``.

aberenguel: In this case, any update in ``extendedInformation`` will be reflected in the mucRoom presence.

sco0ter: Yes, I even did this before and it worked, but I was unsure about possible hidden side effects, although I couldn't see any.
But I thought the original author surely had its reason to create a copy every time.

guusdk: @dwd appears to have added that a few months ago, to fix OF-928.

sco0ter: @guusdk Seems reasonable. I've seen this `IllegalAddException` before when not creating a copy.

sco0ter:
Related issue: https://issues.igniterealtime.org/projects/OF/issues/OF-1146

PR #623, opened 2016-08-08
OF-1168 fix Oracle DDL to use varchar2 and fix alter statements

hdeadman:
alter statements (at least in oracle 11g or earlier) do not allow the
COLUMN keyword after ADD in an ALTER statement

It looks like the update SQL for Oracle was copied from Postgresql? Or maybe it works on Oracle 12c? Anyway this fixes the DDL for Oracle 11g and I assume it would still work on Oracle 12c.


akrherz:
Thanks, I have filed [OF-1168](https://issues.igniterealtime.org/browse/OF-1168)

PR #626, opened 2016-08-12
OF-1116: Apply workaround for Java 7/8 compatibility issue

guusdk:
The JDK 7 signature of ConcurrentHashMap#keySet() returns a Set, while JDK 8 signature
returns a KeySetView, which is a new class in JDK 8. As a result, using Java 7 to run
an Openfire instance that is compiled with Java 8 (using the -source/-target 1.7
compilation flags) causes a java.lang.NoSuchMethodError.

To work around this problem, this commit applies a workaround as presented by
Martin Buchholz in https://bugs.openjdk.java.net/browse/JDK-8151366

This commit should be reverted after Openfire stops to support Java 7.

### DIFF
--- a/src/java/org/jivesoftware/openfire/roster/Roster.java
+++ b/src/java/org/jivesoftware/openfire/roster/Roster.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import org.jivesoftware.database.JiveID;
 import org.jivesoftware.openfire.PresenceManager;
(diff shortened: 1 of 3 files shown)


guusdk:
Note, this PR is based on the 4.0 branch. If accepted, this needs to go in master too.

src/java/org/jivesoftware/openfire/roster/Roster.java
```diff
@@ -625,7 +625,7 @@ public void broadcastPresence(Presence packet) {
             }
         }
         // Broadcast presence to shared contacts whose subscription status is FROM
-        for (String contact : implicitFrom.keySet()) {
+        for (String contact : ((Map<String,Set<String>>) implicitFrom).keySet()) { // Cast prevents Java compatibility issue as described in OF-1116. Remove the cast when Java 7 support is dropped from Openfire.
```
sco0ter: I'd prefer to change the type of the local variable `implicitFrom` to `Map`. It's cleaner imo (see http://stackoverflow.com/q/147468) and should fix the problem as well, according to https://gist.github.com/AlainODea/1375759b8720a3f9f094

We also don't need to revert anything.

guusdk: Agreed. Shouldn't we go with the ConcurrentMap interface, rather than the Map interface?

sco0ter: It wouldn't make a difference here. But if we don't need its methods (like `putIfAbsent`), I'd always prefer the most generic interface, i.e. Map.

guusdk: That'd lock you out of the concurrent goodness, for any future usage. Seems a waste to hide it.

sco0ter: In Java 8, I don't see a difference anyway, because all the `ConcurrentMap` methods are default methods on `Map` now.

guusdk: It boils down to a matter of preference, I believe. Dave introduced me to the term 'bikeshedding' yesterday. I'll go ahead and merge, since the fix itself is desirable now.

PR #631, opened 2016-08-16
OF-1156 Do not store null LockOutFlag in Cache

dwd:
The LockOutManager stores a null LockOutFlag (the default) in the cache,
which shouldn't work (and doesn't in a cluster).

### DIFF
--- a/src/java/org/jivesoftware/openfire/lockout/LockOutManager.java
+++ b/src/java/org/jivesoftware/openfire/lockout/LockOutManager.java
@@ -249,7 +249,7 @@ public class LockOutManager {
 				// If group wan't found in cache, load it up and put it there.
 				if (flag == null) {
 					flag = provider.getDisabledStatus(username);
-					lockOutCache.put(username, flag);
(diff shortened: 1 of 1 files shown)


GregDThomas:
I suspect that this may be the cause of the problem highlighted at https://community.igniterealtime.org/message/259162

I've not yet had time to prove it, but my theory is that the following sequence of events occur

1. The node restarts
2. User refreshes their browser; the cookie offered by the browser is no longer valid, so this is essentially a failed login attempt.
2. A null LockOutFlag is stored in the DefaultCache, user is redirected to the login screen
3. Clustering starts, and attempts to migrate the DefaultCache -> ClusteredCache
4. Because DefaultCache contains null, it can't be converted to a ClusteredCache, so is left as a DefaultCache
5. There's a class cast exception where the code retrieves a Cache that is expected to be a ClusteredCache, but is in fact a DefaultCache

dwd:
That sounds plausible. This is the only case I ran into where the exception-on-null patch you did caused a user-visible error, so I patched it. There's probably others.

PR #632, opened 2016-08-16
OF-1171 Update message routing to RFC 6120

dwd:
Various non-conformances in message routing caused some odd behaviour where
combinations of outgoing PEP events failing to be delivered could cause
errors to be delivered to all available clients.

This fixes message routing for three cases:
- Headline messages sent to the bare JID are now broadcast.
- Headline emssages sent to an offline session are discarded.
- Groupchat messages sent to a bare JID are handed to offline processing,
  however these should not occur.
- Errors sent to the bare JID are discarded.


guusdk:
Can we log data that's being discarded please? I've had to many endless searches for bugs that ended up in code that silently ignored stuff. Logging on debug level would be suffice.

PR #634, opened 2016-08-16
OF-1174: MUC should always respond to IQ requests.

guusdk:
IQ requests must always be responded to, if only with an error reply. For good
measure, IQ-Ping support was added too.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -80,11 +80,7 @@ import org.xmpp.component.ComponentManager;
 import org.xmpp.forms.DataForm;
 import org.xmpp.forms.DataForm.Type;
 import org.xmpp.forms.FormField;
-import org.xmpp.packet.IQ;
-import org.xmpp.packet.JID;
-import org.xmpp.packet.Message;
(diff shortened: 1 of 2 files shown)


dwd:
I don't think this is right - it's handling service-level IQ stanzas in the MucLocalUser - shouldn't this be handled in MultiUserChatServiceImpl instead?

guusdk:
Yes. I'm preparing a new fix.

src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -386,6 +398,9 @@ else if ("http://jabber.org/protocol/disco#items".equals(namespace)) {
             IQ reply = XMPPServer.getInstance().getIQDiscoItemsHandler().handleIQ(iq);
             router.route(reply);
         }
+        else if ("urn:xmpp:ping".equals(namespace)) {
+            router.route( IQ.createResultIQ(iq) );
```
dwd: I think we shouldn't be handling these when they refer to an occupant, only to a room or service. Otherwise a user can't reliably ping another occupant.

dwd: Ah, never mind, you've already told me I'm being daft in the chatroom.

PR #624, opened 2016-08-11
OF-1169 Allow java 7 or 8 as requirements for debian

akrherz:
completely untested by me, will attempt to find folks to comment on this.

PR #637, opened 2016-08-22
OF-1177 allow java7-runtime-headless to satisfy .deb requirement

akrherz:
See discussion on PR #624 , it is not clear to me what the equivalent of this is for build-deps, but this will at least help those who installed Oracle JRE and not OpenJDK

### DIFF
--- a/build/debian/control
+++ b/build/debian/control
@@ -9,7 +9,7 @@ Homepage: http://www.igniterealtime.org
 Package: openfire
 Section: net
 Priority: optional
-Pre-Depends: default-jre-headless (>= 1.7)
+Pre-Depends: default-jre-headless (>= 1.7) | java7-runtime-headless | java7-runtime | java8-runtime-headless | java8-runtime | java9-runtime-headless | java9-runtime
 Architecture: all
 Description: A high performance XMPP (Jabber) server.
  Openfire is an instant messaging server that implements the XMPP


guusdk:
Is there such a thing as java8-runtime-headless ? If so, it should probably be added.

akrherz:
@guusdk my understanding is that when you install Oracle JRE 8, this also provides the `java7-runtime-headless` as well. See https://github.com/igniterealtime/Openfire/pull/624#issuecomment-241257187

guusdk:
So, to my understanding, JavaX-runtime-headless are virtual packages, which is defined ([here](https://www.debian.org/doc/debian-policy/ch-binary.html#s-virtual_pkg)) as:

> Sometimes, there are several packages which offer more-or-less the same functionality. In this case, it's useful to define a virtual package whose name describes that common functionality. (The virtual packages only exist logically, not physically; that's why they are called virtual.) The packages with this particular function will then provide the virtual package. Thus, any other package requiring that function can simply depend on the virtual package without having to specify all possible packages individually.

In the example of Oracle JRE 8, which is a physical package, both java7-runtime-headless as well as java8-runtime-headless apparently are provided by it (if the comment you referred to is correct - I didn't verify this). That's not to say that _all_ physical packages that provide a Java 8 do this. To make sure that you allow for _any_ physical package that provides Java 8, I think you should explicitly define that dependency.

The [Authoritative List Of Virtual Package Names](https://www.debian.org/doc/packaging-manuals/virtual-package-names-list.txt) lists these Java packages:

-  java5-runtime           (a Java runtime environment, Java version 5)
-  java6-runtime           (a Java runtime environment, Java version 6)
-  java7-runtime           (a Java runtime environment, Java version 7)
-  java8-runtime           (a Java runtime environment, Java version 8)
-  java9-runtime           (a Java runtime environment, Java version 9)
-  java5-runtime-headless  (a non-graphical Java runtime environment, Java ver. 5)
-  java6-runtime-headless  (a non-graphical Java runtime environment, Java ver. 6)
-  java7-runtime-headless  (a non-graphical Java runtime environment, Java ver. 7)
-  java8-runtime-headless  (a non-graphical Java runtime environment, Java ver. 8)
-  java9-runtime-headless  (a non-graphical Java runtime environment, Java ver. 9)

Given this list, I'd say Openfire should depend on any of these (both headless and non-headless), that are Java 7 or higher.

guusdk:
Yeah, that's what I'd expect to work well.

PR #638, opened 2016-08-22
4.0.x branch - EXTERNAL auth with BOSH - avoid NPE when peer certificates array is null

hdeadman:
This only applies to 4.0.x branch b/c SASLAuthentication is very different in master, but if you put out a 4.0.4 this shouldn't hurt anything and it allows us to get back an error message rather than just timing out when doing external auth over https (with client certs).  

The second commit avoids a StackOverflowException that was happening when the code was retrieving keystores, see commit comment for details.

### DIFF
--- a/src/java/org/jivesoftware/openfire/net/SASLAuthentication.java
+++ b/src/java/org/jivesoftware/openfire/net/SASLAuthentication.java
@@ -549,16 +549,17 @@ public class SASLAuthentication {
             String principal = "";
             ArrayList<String> principals = new ArrayList<>();
             Connection connection = session.getConnection();
-            if (connection.getPeerCertificates().length < 1) {
+            Certificate[] peerCertificates = connection.getPeerCertificates();
+            if (peerCertificates == null || peerCertificates.length < 1) {
                 Log.debug("SASLAuthentication: EXTERNAL authentication requested, but no certificates found.");
                 authenticationFailed(session, Failure.NOT_AUTHORIZED);
                 return Status.failed; 
             }
 
-            final KeyStore keyStore   = connection.getConfiguration().getIdentityStore().getStore();
-            final KeyStore trustStore = connection.getConfiguration().getTrustStore().getStore();
-
-            final X509Certificate trusted = CertificateManager.getEndEntityCertificate( connection.getPeerCertificates(), keyStore, trustStore );
+            final CertificateStoreManager certificateStoreManager = XMPPServer.getInstance().getCertificateStoreManager();
+            final KeyStore keyStore   = certificateStoreManager.getIdentityStore( ConnectionType.SOCKET_C2S ).getStore();
+            final KeyStore trustStore = certificateStoreManager.getTrustStore( ConnectionType.SOCKET_C2S ).getStore();
+            final X509Certificate trusted = CertificateManager.getEndEntityCertificate(peerCertificates, keyStore, trustStore );
 
             if (trusted == null) {
                 Log.debug("SASLAuthentication: EXTERNAL authentication requested, but EE cert untrusted.");


guusdk:
I wonder if this relates to https://issues.igniterealtime.org/browse/OF-1191

guusdk:
Thanks for these fixes. I'm somewhat concerned by the latter part: this makes (this implementation of) SASL work only for (or with) the S2C keystores. Although obviously an improvement over "not working at all", it will break a feature that allows every type of connection to use it's own set of keystores.

Also, could you rebase the commit to get rid of that merge?

guusdk:
Apologies for the duplicate texts - I'm playing with the new reviewing functionality that was recently added to Github.

hdeadman:
I rebased the pull request to one commit. I think the code is looking up the C2S keystores in a block of code that is specific to client to server connections. Is openfire allowing different keystores for different C2S connections? The current code was getting a stack overflow exception. My changes allow client auth to work with BOSH (from strophe.js), at least in our setup.

guusdk:
No, Openfire does not allow for different keystores to be configured for different C2S connections (but does allow for different keystores for C2S and S2S, for example). The SASLAuthentication class is re-used in other parts (I'm seeing references to server dialback, for example), and ideally, should remain re-usable for any type of connection that need to do SASL. 

You mentioned that the 4.0.x branch is different from the master - did you happen to check if there are similar problems on master?

hdeadman:
I have not tried out openfire master. It looks like there are only deb files for the  [nightly builds](https://www.igniterealtime.org/downloads/nightly_openfire.jsp)? It would be easier to try it out if there were a tar.gz binary build although I suppose I could build it myself. I probably won't have time to test it this week. 

Regarding this pull request, the block of code being updated to use C2S keystores starts off with: 

```
        else if (session instanceof LocalClientSession) {
            // Client EXTERNAL login
            Log.debug("SASLAuthentication: EXTERNAL authentication via SSL certs for c2s connection");
```

It sounds like the C2S keystores are being used for a C2S connection if you can believe the debug.

hdeadman:
I see a 4.0.4 release is in the works, is it too late to include this pull request? Does my last comment not address your concern about using the C2S keystore (since the session from which the connection is retrieved has already been determined to be a LocalClientSession for a c2s connection for that block of code)? The patch diff doesn't show the else block that the code is in but here it is in context:

```
        else if (session instanceof LocalClientSession) {
            // Client EXTERNAL login
            Log.debug("SASLAuthentication: EXTERNAL authentication via SSL certs for c2s connection");
            
            // This may be null, we will deal with that later
            String username = new String(StringUtils.decodeBase64(doc.getTextTrim()), StandardCharsets.UTF_8);
            String principal = "";
            ArrayList<String> principals = new ArrayList<>();
            Connection connection = session.getConnection();
            Certificate[] peerCertificates = connection.getPeerCertificates();
            if (peerCertificates == null || peerCertificates.length < 1) {
                Log.debug("SASLAuthentication: EXTERNAL authentication requested, but no certificates found.");
                authenticationFailed(session, Failure.NOT_AUTHORIZED);
                return Status.failed; 
            }

            final CertificateStoreManager certificateStoreManager = XMPPServer.getInstance().getCertificateStoreManager();
            final KeyStore keyStore   = certificateStoreManager.getIdentityStore( ConnectionType.SOCKET_C2S ).getStore();
            final KeyStore trustStore = certificateStoreManager.getTrustStore( ConnectionType.SOCKET_C2S ).getStore();
            final X509Certificate trusted = CertificateManager.getEndEntityCertificate(peerCertificates, keyStore, trustStore );
```

hdeadman:
I tried out 4.1.1 (no patches) and the stack overflow error that I was getting on the 4.0 branch and that I resolved with this patch is still happening. This happens when I configure openfire with EXTERNAL sasl authentication (client certificate). 

Here is the stackoverflow in progress so you can see the stack trace (from debugger). 

image

```
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
HttpSession$HttpVirtualConnection.getConfiguration() line: 1146	
ExternalClientSaslServer.evaluateResponse(byte[]) line: 65	
SASLAuthentication.handle(LocalSession, Element) line: 324	
SessionPacketRouter.route(Element) line: 64	
HttpSession.sendPendingPackets() line: 639	
HttpSession$HttpPacketSender.run() line: 1271	
ThreadPoolExecutor.runWorker(ThreadPoolExecutor$Worker) line: 1142	
ThreadPoolExecutor$Worker.run() line: 617	
Thread.run() line: 745	
```

hdeadman:
Closing b/c it's for old branch.

PR #640, opened 2016-08-28
OF-1182 remove Release News, expand Blog

wrooot:


akrherz:
Thanks.  My thought on this was to replace it with the latest Openfire Forum discussion threads RSS.  What do you think of that?

wrooot:
Personally i have RSS disabled on my servers, as i have other means to keep up with the news and releases, but i think this Box is the main way for Openfire admins to find out about releases. Granted, Openfire will notify about a new version anyway, but here you have a quick link to the release blog post. You can also see info about Spark's releases and other news (like moving servers, attending conferences, etc.). I don't think general forum discussions feed will bring much value. One day it maybe flooded with Spark 2.8.0 login issue threads, another day it will be Ofmeet questions or in Portuguese :)

wrooot:
So, anything else or anyone else want to discuss this?

guusdk:
Can't we simply re-instate that RSS feed, instead of remove it? I think it's kind of nice to have that 'latest release' thingy in the admin panel. Restoring the feed surely isn't that hard?

wrooot:
What do you mean by restoring? To start cross-post in both blog and releases sub forum? Sounds like superfluous to me. I'm ok with just one source for the news. Sure sometimes we post something else in the blog, but often it will show same thing. Or do you suggest posting new versions only in that sub forum? Well, i kind of prefer blog and some folks may also watch it as the main source for the news. Anyway. Some solution should be applied to not show obsolete information in that box. It may be that sub forum if majority prefers it.

wrooot:
Ping ;) I wonder if forums post would get picked up by planetjabber (with a tag) the same way as blog post does.

src/web/index.jsp
```diff
@@ -69,9 +69,7 @@
 
 <%! long lastRSSFetch = 0;
     SyndFeed lastBlogFeed = null;
-    SyndFeed lastReleaseFeed = null;
     String blogFeedRSS = "https://community.igniterealtime.org/blogs/ignite/feeds/posts";
-    String releaseFeedRSS = "https://community.igniterealtime.org/community/feeds/messages?community=2017";
```
dwd: Should this be a configuration option? I can see it being useful to override for some closed-environment cases.

src/web/index.jsp
```diff
@@ -371,14 +369,13 @@
             <a href="<%= blogFeedRSS %>" class="jive-feed-icon"><img src="images/feed-icon-16x16.gif" alt="" style="border:0;" /></a>
             <h4><fmt:message key="index.cs_blog" /></h4>
             <% long nowTime = System.currentTimeMillis();
-                if (lastBlogFeed == null || lastReleaseFeed == null || nowTime - lastRSSFetch > 21600000) {
+                if (lastBlogFeed == null || nowTime - lastRSSFetch > 21600000) {
```
dwd: ... and likewise here, an option might be useful? Obviously this works as a default.

src/web/index.jsp
wrooot: I'm not a dev, so it was the fastest thing to change :) One can add an option here.

src/web/index.jsp
wrooot: Same as above.

dwd:
Opened OF-1237 to track making this configurable.

PR #642, opened 2016-09-09
Use equals-sign in zero-length SASL success elements

sco0ter:
Also don't use it in challenges.

RFC 6120 § 6.4.3. does not mention the equals-sign logic.

However the success case (§ 6.4.6) explicitly does.


dwd:
I'm reasonably confident that if the <challenge/> and <response/> part of the spec doesn't talk about the use of '=' for the zero-length string, it's an error in the spec and not our code. (success and auth both do, rightly, mention it)

dwd:
Closing in favour of #692, which includes this commit.

PR #644, opened 2016-09-26
OF-1194: New SMS service

guusdk:
This commit introduces a new SMS service that takes after the
existing email service. The new service utilizes SMPP to allow
for messages to be sent to a suitable service provider.


guusdk:
Travis currently fails, as this PR depends on https://github.com/igniterealtime/Openfire/pull/641

PR #646, opened 2016-09-27
OF-1196: Don't overflow system property name/value

guusdk:


guusdk:
I'm assuming it was added to create a tooltip.

akrherz:
@guusdk ah yes, you can mouse over and get the full text as a tooltip

guusdk:
I think we can do without, especially as this PR will make more of the text visible (assuming that most people use a near-fullscreen browser to look at the page.

PR #647, opened 2016-09-27
OF-1197: XEP-0227 support for User import Export plugin

annovanvliet:
Two commits: one to fix an type in the wildfire XSD, which prevented importing users
and a second to implement support for XEPP-0227


src/plugins/userImportExport/plugin.xml
```diff
@@ -6,8 +6,8 @@
    <name>User Import Export</name>
    <description>Enables import and export of user data</description>
    <author>Ryan Graham</author>
-   <version>2.5.0</version>
-   <date>10/12/2015</date>
+   <version>2.5.1</version>
+   <date>26/09/2016</date>
```
akrherz: Sadly, our date stamps are month/day/year.

guusdk:
Thanks for contributing this! Looks like quite a nice addition. I do think this warrants a bigger version-number update than the patch number though. Make it a 2.6 or 3.0 even. :)

Neustradamus:
@igniterealtime, @annovanvliet, @akrherz, @guusdk, @dwd, @Flowdalic, @deleolajide, @marclaporte, ...: It is possible to add the XEP-0227 1.1 support?
- https://xmpp.org/extensions/xep-0227.html

Thanks in advance.

PR #648, opened 2016-09-28
OF-1199: New JDBC based AdminProvider

robmv:
JDBCAdminProvider can read the list of admin users from a database

This was contributed years ago (before GitHub) at the same time when I added the _useConnectionProvider_ setting to all JDBC*Providers but it was never taken. This is another try. It is the only *Provider class that doesn't has a JDBC based implementation.

### DIFF
--- a/documentation/docs/db-integration-guide.html
+++ b/documentation/docs/db-integration-guide.html
@@ -150,6 +150,28 @@
     &lt;/jive&gt;
 </pre>
 
+<p>Another option is to use an AdminProvider. AdminProvider instances are responsible for listing
+    the administrators users dynamically. The default use the <tt>authorizedUsernames</tt> setting
+    previously explained. JDBCAdminProvider allows to list the administrators from a SQL query.
+    For example:</p>
+
+<pre>
+&lt;jive&gt;
+  ...
+  &lt;provider&gt;
+    ...
+    &lt;admin&gt;
+      &lt;className&gt;org.jivesoftware.openfire.admin.JDBCAdminProvider&lt;/className&gt;
+    &lt;/admin&gt;
+    ...
+  &lt;/provider&gt;
+  &lt;jdbcAdminProvider&gt;
+    &lt;getAdminsSQL&gt;SELECT userid FROM user_account WHERE administrator='Y'&lt;/getAdminsSQL&gt;
+  &lt;/jdbcAdminProvider&gt;
+  ...
+&lt;/jive&gt;
+</pre>
+
 <h3>User Integration</h3>
 
 <p>Optionally, Openfire can load user data from your custom database. If you enable user integration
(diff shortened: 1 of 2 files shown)


documentation/docs/db-integration-guide.html
```diff
@@ -150,6 +150,28 @@
     &lt;/jive&gt;
 </pre>
 
+<p>Another option is to use an AdminProvider. AdminProvider instances are responsible for listing
+    the administrators users dinamically. The default use the <tt>authorizedUsernames</tt> setting
```
guusdk: Typo: should by "dynamically"

src/java/org/jivesoftware/openfire/admin/JDBCAdminProvider.java
```diff
@@ -0,0 +1,147 @@
+/**
+ * $Revision$
+ * $Date$
+ *
+ * Copyright (C) 2005-2008 Jive Software. All rights reserved.
```
guusdk: You don't work for Jive, do you? :)

guusdk:
Thanks for this! I don't know why this wasn't added last time, but I see no reason why not to. Seems like a usable addition. I've created OF-1199 to track this new feature in our issue tracker.

dwd:
This looks good - update copyright and merge?

robmv:
@dwd Which file can be a good example of a header for the copyright? or do you mean only date? It is basically a copy of JDBCAuthProvider.java with changes to the interface

src/java/org/jivesoftware/openfire/admin/JDBCAdminProvider.java
robmv: It is based on JDBCAuthProvider, not written from scratch

guusdk:
You're now listing Jive Software as the copyright holder of your code. I'd put in your name instead.

src/java/org/jivesoftware/openfire/admin/JDBCAdminProvider.java
dwd: Yeah, so just add your own copyright statement. But don't put "All Rights Reserved" - just "Copyright 2012-2016 Robert Marcano" (in this example, I'm assuming you wrote it originally around 2012).

dwd:
... as well. If it's a derived work, at least.

src/java/org/jivesoftware/openfire/admin/JDBCAdminProvider.java
robmv: Done, hehe older than that, my local git repo has:

commit 5a7e21b3f995077746a499dc7d2ecb5b82b8954b
Author: Robert Marcano ....
Date:   Wed Nov 5 21:35:04 2008 +0000

PR #651, opened 2016-10-10
Minor improvements: more logging and removal of dead code.

guusdk:


src/java/org/jivesoftware/openfire/XMPPServer.java
```diff
@@ -587,14 +587,11 @@ private void startModules() {
         for (Module module : modules.values()) {
             boolean started = false;
```
dwd: Isn't this unused now?

guusdk: Yeah, and there's more redundant code. I'll squash an update.

PR #653, opened 2016-10-12
OF-1201: Complete context if an exception is thrown when setting read listener

jforsell:
We've yet to find the root cause of the state=EOF errors we've been seeing (which may be caused by an issue in our infrastructure or possibly client implementation), but this server-side change stopped file desciptors from being leaked when the error is thrown.

For more details see discussion in https://community.igniterealtime.org/thread/58399


jforsell:
Hi @guusdk - is it possible merge this change in and look at the refactoring of context#complete() separately (I'm afraid I haven't had any time to look at the bigger change).

PR #656, opened 2016-10-18
OF-1206: PrivacyLists on first run

Gugli:
From [here](https://community.igniterealtime.org/thread/59616)

There are 2 instances of PrivacyListProvider (one in PrivacyListManager and the other in IQPrivacyHandler)

The issue was : both instances have separate privacyListCount variables, which are filled during PrivacyListProvider constructor (== server startup).
Thus, if at server startup the privacyListCount is 0, and one of the instances adds a PrivacyList, the other will never be informed and discard all changes and reads.

Fix : make privacyListCount static.

### DIFF
--- a/src/java/org/jivesoftware/openfire/handler/IQPrivacyHandler.java
+++ b/src/java/org/jivesoftware/openfire/handler/IQPrivacyHandler.java
@@ -51,7 +51,7 @@ public class IQPrivacyHandler extends IQHandler
 
     private IQHandlerInfo info;
     private PrivacyListManager manager = PrivacyListManager.getInstance();
-    private PrivacyListProvider provider = new PrivacyListProvider();
+    private PrivacyListProvider provider = PrivacyListProvider.getInstance();
 
     public IQPrivacyHandler() {
         super("Blocking Communication Handler");
--- a/src/java/org/jivesoftware/openfire/privacy/PrivacyListManager.java
+++ b/src/java/org/jivesoftware/openfire/privacy/PrivacyListManager.java
@@ -34,7 +34,7 @@ public class PrivacyListManager {
     private static final PrivacyListManager instance = new PrivacyListManager();
     private static Cache<String, PrivacyList> listsCache;
 
-    private PrivacyListProvider provider = new PrivacyListProvider();
+    private PrivacyListProvider provider = PrivacyListProvider.getInstance();
 
     private List<PrivacyListEventListener> listeners = new CopyOnWriteArrayList<>();
 
--- a/src/java/org/jivesoftware/openfire/privacy/PrivacyListProvider.java
+++ b/src/java/org/jivesoftware/openfire/privacy/PrivacyListProvider.java
@@ -29,7 +29,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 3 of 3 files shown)


guusdk:
Additionally, we don't seem to need a 'count' at all. Perhaps you could modify that to be an AtomicBoolean, while you're at it?

Gugli:
Indeed. I also envisioned turning PrivacyListProvider into a singleton, but I was unsure about the correct way to do it so I went the "quick&easy" way (I'm not a Java developer, and I'm not used to this specific project).

Gugli:
I'll try to do both, and get back to you. I'll try to mimic the singleton logic from PrivacyListManager.

guusdk:
Oh, that's quite easy: To turn it into a singleton, add a static private field to hold the instance, make the constructor private, and have a static synchronized method return the instance, lazy initializing it when first called.

src/java/org/jivesoftware/openfire/handler/IQPrivacyHandler.java
```diff
@@ -51,7 +51,7 @@
 
     private IQHandlerInfo info;
     private PrivacyListManager manager = PrivacyListManager.getInstance();
-    private PrivacyListProvider provider = new PrivacyListProvider();
+    private PrivacyListProvider provider = PrivacyListProvider::getInstance();
```
guusdk: :: is not valid in Java. Static methods can be called like any other: `PrivacyListProvider.getInstance()` should do it.

src/java/org/jivesoftware/openfire/privacy/PrivacyListManager.java
```diff
@@ -34,7 +34,7 @@
     private static final PrivacyListManager instance = new PrivacyListManager();
     private static Cache<String, PrivacyList> listsCache;
 
-    private PrivacyListProvider provider = new PrivacyListProvider();
+    private PrivacyListProvider provider = PrivacyListProvider::getInstance();
```
guusdk: Again, replace `::` wit a single dot.

src/java/org/jivesoftware/openfire/privacy/PrivacyListProvider.java
```diff
@@ -86,11 +97,9 @@ public PrivacyListProvider() {
 
         // Load the total number of privacy lists in the database. We're looking
         // for the (very common) special case that there are no privacy lists stored.
-        // In that case, we can optimize away many database calls. In the future, a
-        // better general-case solution may be to cache all privacy lists defined
-        // if there are less than, say, 500.
-        privacyListCount = new AtomicInteger(0);
-        loadPrivacyListCount();
+        // In that case, we can optimize away many database calls. 
+        privacyListNotUsed = new AtomicBoolean(true);
```
guusdk: Holy double negation, Batman! It's probably easier to name this "databaseContainsPrivacyLists", which is what we check for. Also, the comment above this line refers to a 'number', which is no longer the case.

Gugli:
Holy guacamole ! I didn't expect the pull request to be changed when I pushed my amend. (I push to my fork to checkout on my build VM). Sorry for the mess....

guusdk:
No worries, we can squash this all later. :)

Gugli:
New version of the patch should meet the requirements and fixes the bug on my side (tested on Ubnutu 14. with an empty database).

PR #657, opened 2016-10-21
SPARK-1803 Can't use usernames with non-latin symbols

Alexander198961:
hope resolve https://issues.igniterealtime.org/projects/SPARK/issues/SPARK-1803


guusdk:
I suspect that at least some of the values that are affected are ASCII-encoding - but that certainly need not be the case for the password value in ScramUtils#createSaltedPassword.

@dwd, is there a risk in processing everything in UTF-8?

wrooot:
Alexander says this should fix login issue with unicode usernames in Spark 2.8.x (probably some related change in Smack 4). Can this be merged? I don't think it was logged already, so i have filed https://issues.igniterealtime.org/browse/OF-1220 but i'm not sure about the name and the type (bug or not).

PR #659, opened 2016-10-23
Add missing jsmpp jar to satisfy eclipse project dependencies for build.

danielhams:
Adds the built "libs/merge/jsmpp.jar" to the eclipse project dependencies.


guusdk:
Thanks!

PR #661, opened 2016-11-01
OF-1210 Correct the cache behaviour for Hazelcast backed caches

GregDThomas:
This pull request fixes the issues identified in the post at https://community.igniterealtime.org/message/261078 - i.e. a lack of time-to-live-seconds configuration, and an error when setting the MaxLifetime property

### DIFF
--- a/src/plugins/hazelcast/changelog.html
+++ b/src/plugins/hazelcast/changelog.html
@@ -44,10 +44,15 @@
 Hazelcast Clustering Plugin Changelog
 </h1>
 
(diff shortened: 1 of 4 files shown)


akrherz:
@GregDThomas Please consider bumping `plugin.xml` so that we can release this change

GregDThomas:
Have done - sorry for the delay. If I get time, I'll also look to see if it's possible to honour the programmatical setMaxCacheSize() / setMaxLifetime() which currently do nothing more then set a JiveProps.

PR #663, opened 2016-11-03
OF-1211 - Fix description of Broadcast disabling option in Client

BittyByte:
Currently Client Control plugin says "Allow for broadcasting of messages to specified groups or entire roster." by the option to disable Broadcasting in Spark. And it doesn't disable broadcasting option to selected users. This was fixed in Spark with SPARK-1827, so description should be updated also.

Note: This PR will only patch EN, CZ, ES, FR and PT_BR locales. Other locale files still need to be updated accordingly.


wrooot:
Well, lucky guess (and some google translate):

CZ - Povoluje vys\u00edl\u00e1n\u00ed zpr\u00e1v.
ES - Permitir difundir mensajes.
FR - Permet d'envoyer un message.
PT_BR - Permitir transmiss\u00e3o de mensagens.
zh_CN - no clue :D

I think it is safe to fix at least 4. I doubt anyone else will do this anyway.

BittyByte:
Ok, will do. Shall I delete the previous PR or just create a new one?

Sent with AquaMail for Android
http://www.aqua-mail.com

wrooot:
No, you can just add to this PR (same as you did when fixing 1827 and then adding 1822 fixes into same PR).

BittyByte:
OK, thanks.

wrooot:
@akrherz ? :)

BittyByte:
Updated PR so now it includes translations for CZ, ES, FR, PT_BR and English.

akrherz:
Do we need to bump plugin.xml, so that this change actually goes out to folks using it?

BittyByte:
Hi akrherz, do you mean to bump up the version and date? Yes, probably, I forgot about that. I'm new at this. Let me know which version and I'll be glad to update the PR. Thanks.

BittyByte:
I updated **plugin.xml** and **changelog.html** to reflect the changes as you requested. Let me know if there is anything else I may have missed. Thanks.

wrooot:
It's a trivial change, but i think we should still bump the version (2.0.1 is fine).

I wonder why the whole changelog file is replaced in the last commit though. It should be only new lines inserted. Also we usually list a JIRA ticket (as seen for 1.3.1 version in changelog).

BittyByte:
Yeah, I noticed that too. Don't know why it changed the whole file when I only added a couple of lines. I bumped the version to 2.0.1 and also updated to today's date (at least in my timezone).

BittyByte:
OK, will add the JIRA ticket...

BittyByte:
OK, all done. I think the whole changelog file got replaced because of the text editor that was used. Eclipse opens up notepad when editing HTML and it gets saved back as a plain text file (which it should).

wrooot:
Do you mean Windows Notepad? For me it opens in HTML Editor, though i think i have added it somehow recently, but before that i was opening it with its built-in Text Editor (with Open with).

wrooot:
Well, probably nothing critical for a changelog file which only updates from top to bottom.

BittyByte:
I'll have to change to an HTML editor in Eclipse preferences menu. You're right, it's just a text file anyway. No formatting codes should be present.

BittyByte:
Installed HTML Editor from Eclipse Marketplace. It looks great! Thanks again for your help and enjoy the rest of your day. Time for some sleep now!

wrooot:
It's 7 AM here, heading to work actually :) good rest ;)

guusdk:
I think that the line ending used in the changelog was changed, causing all lines of the file to be modified. Could you change that back to what it was?

BittyByte:
Hi Guus, I was able to revert the **changelog.html** file back to its original **Unix** text file formatting. The hazards of working on mixed platforms. Sorry about that. It should be ok now. Let me know if there is anything else I may have missed. Thanks!

PR #664, opened 2016-11-04
Adjust monitoring plugin for use with oracle database

Hammington:
since oracle does not permit two long columns in a single table and also
stated "long" column type as deprecated, long columns are changed to
clob

as distinct modifier is not allowed on clob columns by oracle, adjust
severall (not all) statements accordingly

add a switch for oracle 11g as it does not support keywords "LIMIT"
and "OFFSET"

add an option to the monitoring/archiving settings to store complete
stanzas for muc's

### DIFF
--- a/src/plugins/monitoring/src/database/monitoring_oracle.sql
+++ b/src/plugins/monitoring/src/database/monitoring_oracle.sql
@@ -28,15 +28,15 @@ CREATE INDEX ofConParticipant_conv_idx ON ofConParticipant (conversationID, bare
 CREATE INDEX ofConParticipant_jid_idx ON ofConParticipant (bareJID);
 
 CREATE TABLE ofMessageArchive (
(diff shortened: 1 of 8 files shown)


Hammington:
I splitted the change into four branches now so it should be easier to understand what is going on.

I also omitted all whitespace and code style changes.

I, as well, think that the database specific switch is ugly but from my point of view the change in the statements for oracle is so severe, that it should not affect other database systems.

The api change I added allows me to store the complete stanza into the archive system. The database already has a column for this purpose, but it was unused until this change. I also added a switch to the settings page so this change should only affect people that want the stanza to be stored for muc conversations.

I am not too sure about the changelog. Afterall I don't know which of my branches are going to be included into master. I also don't have an idea of the next version number. Maybe one of you could fill in these.

guusdk:
For future reference: this PR was replaced by #665, #666, #667 and #668.

PR #665, opened 2016-11-07
OF-1212 adjust ddl statements for oracle

Hammington:
since two long columns are not permitted in one table by oracle and the
datatype long is deprecated, the respective columns are changed to clob


akrherz:
If this script always failed from the beginning, then it would have never been applied to current installs?  Am I correct?

guusdk:
This PR replaces LONG with CLOB. LONG is  supported by Oracle, but its support has been deprecated.

akrherz:
@guusdk , but the point here was that Oracle never supported two LONGs in the same table, right?

guusdk:
Ah - that, I don't know. if that is the case, I'd consider to merge this PR as-is.

Hammington:
@guusdk I can confirm this, the ddl did never execute correctly with two "long" columns.

PR #668, opened 2016-11-07
add possibility to store complete stanza in archive

Hammington:
in the settings there is now a switch allowing to add complete stanza to
the archive for muc conversations


akrherz:
@Hammington Does this addres https://issues.igniterealtime.org/browse/OF-817 ?

Hammington:
@akrherz I think this is not the same as this issue refers to the internal muc-serrvice history and my change adresses the mam-service implementation included in the monitoring plugin.

Hammington:
@guusdk the stanza is stored in the table OFMESSAGEARCHIVE and can be retrieved through mam-queries, the code for storing the stanza was already available but the stanza was always set to empty string.

akrherz:
cherry-picked patch onto master

PR #669, opened 2016-11-08
OF-1213 Add extension points to MUC

dwd:
This PR adds extension points to MUC, allowing both Services and Rooms to be extended by plugins (and other code).

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -20,17 +20,7 @@
 
 package org.jivesoftware.openfire.muc.spi;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Queue;
-import java.util.Set;
-import java.util.TimerTask;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -169,6 +161,11 @@
     private IQMUCSearchHandler searchHandler = null;
 
     /**
+     * Plugin (etc) provided IQ Handlers for MUC:
+     */
+    private Map<String,IQHandler> iqHandlers = null;
```
guusdk: If you instantiate a new Map here, you can remove a couple of null checks.

src/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -402,6 +416,21 @@ else if ("urn:xmpp:ping".equals(namespace)) {
...
+            if (h != null) {
+                try {
+                    IQ reply = h.handleIQ(iq);
+                    if (reply != null) {
+                        router.route(reply);
+                    }
+                } catch (UnauthorizedException e) {
+                    IQ reply = IQ.createResultIQ(iq);
+                    reply.setType(IQ.Type.error);
+                    reply.setError(PacketError.Condition.service_unavailable);
```
guusdk: `service_unavailable`, not `not_authorized`'?

dwd: Ah, yes. It's the default error we should generally return, I think, since it's equivalent to an unknown namespace. A bunch of XEPs specify to return this in Unauthorized cases, too - not least because a private, unlisted room might otherwise unwittingly reveal its presence, for example.

PR #672, opened 2016-11-08
OF-817 Persist entire MUC stanza

dwd:
PR request made for easy comparison to #668, without prejudice.


dwd:
Seems I forgot the upgrade scripts, despite writing them.

akrherz:
So PR #668 got merged, so how does this code fit in with that code?  Can they coexist?

dwd:
Yes; this one hits the main Openfire tables instead of the Monitoring plugin's, and provides MUC history with full stanza replay. I'll also hit it in a future PR for MAM/MUC.

PR #678, opened 2016-11-18
OF-190 show java process owner on admin console

akrherz:
Introduces i18n string of index.process_owner "OS Process Owner:".
On Windows, I believe the value field will be 'SYSTEM'


wrooot:
If running as a service it most probably will be SYSTEM on Windows, but will show a user if running via launcher.

PR #679, opened 2016-11-20
OF-955 Update HSQL to 2.3.4

akrherz:
* adjust LOGSIZE and WRITE_DELAY settings
 * did simple smoke test of install, stop, start, OK
 * observed some errors on shutdown within PubSub, noted in OF-955


guusdk:
It's unlikely that we'll find any problems with this by looking at the patch. Lets merge, and see if we run into more issues.

PR #680, opened 2016-11-21
Fix typo on certificates store management screen

mgouline:
Fix minor typo on the certificates store management said "These stores are used for _erver-to-server_ XMPP communication".


guusdk:
Thanks!

PR #684, opened 2016-11-25
OF-1221 -- Add options from default.properties to Client Control plugin

BittyByte:
Added 23 more client features to Client Control plugin for admin to use on a network-wide basis.


wrooot:
I have already tested these changes with a new compiled clientcontrol.jar and updated Spark version. Everything works fine and this PR can be merged (i have already merged the Spark PR for this). @akrherz

PR #688, opened 2016-12-02
OF-1232 Bump fastpath minServerVersion to 4.1

akrherz:
Refactoring work done in 4.1 release of Openfire lead to an introduction of a requirement on `JiveSharedSecretSaslServer`


wrooot:
Probably will fix https://issues.igniterealtime.org/browse/OF-1230 as well.

PR #691, opened 2016-12-12
OF-1123 Respond to incoming streams with the right domain

dwd:
Openfire previously would respond to all incoming domains with the IM domain,
this patch examines and validates the "to" attribute on inbound streams and
responds accordingly.


dwd:
Removed extraneous import (thanks Daryl).

akrherz:
Uh oh, I don't think this is Jira issue OF-1123, or is it?

dwd:
Should have been OF-1233 - sorry!

PR #693, opened 2016-12-13
OF-1239 Prevent NPE in MUC#process(IQ) without NS

akrherz:


dwd:
I don't think this is right. this.iqHandlers can be null, but namespace can't be (at least, perhaps it can, but only theoretically - I think it'd be causing trouble way before this).

akrherz:
Yeah, #698 is right, my bad :(

PR #700, opened 2016-12-14
Add NT hashing for JDBC connections

FZachlod:
This allows using nt password hashes in JDBC connections. This is useful if you have users that are imported from or stored with compatibility to active directory in mind. E.g. this is done in the Alfresco database.


dwd:
Thanks!

PR #701, opened 2016-12-14
Now able to set a default privacy list at user creation.

Gugli:
The registration plugin allows to do several action on account registration.
This patch adds the possibility to add an XMPP PrivacyList to every registering user, and to set it as a Default.


Gugli:
Just so you know : there are 2 things in this patch that I'm not completely sure about.
- The first is the caching strategy for the parsed PrivacyList (to avoid parsing the String at every registration). Since there is only one object to store, I've not used openfire caching facilities.
- The second is the admin interface : I'm not sure a text entry is the best way to type in xml. But I don't know any better.

guusdk:
I like how you pragmatically continued the typo in the constant name. :) XML might not be the neatest way of providing a default list, but, as it's pretty flexible and optional, it's good enough for now. If at some point someone wants to improve on this, we can do that in a new PR.

Gugli:
Ooooops... I _may_ be the kind of guy who type code relying on copypaste/autocomplete :). And the ALLCAPS is not helping my eyes to see the typo.. Sorry !

PR #704, opened 2016-12-14
javadoc fixes and remove legacy SVN RCS tags

akrherz:
addresses warnings thrown by `ant javadocs`

### DIFF
--- a/src/java/org/jivesoftware/openfire/crowd/CrowdAuthProvider.java
+++ b/src/java/org/jivesoftware/openfire/crowd/CrowdAuthProvider.java
@@ -47,9 +47,6 @@ public class CrowdAuthProvider implements AuthProvider {
      * Returns if the username and password are valid; otherwise this
      * method throws an UnauthorizedException.<p>
      *
-     * If {@link #isPlainSupported()} returns false, this method should
-     * throw an UnsupportedOperationException.
-     *
      * @param username the username or full JID.
      * @param password the password
      * @throws UnauthorizedException if the username and password do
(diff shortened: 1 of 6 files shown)


src/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -633,8 +629,6 @@ public Presence kickOccupant(JID fullJID, JID actorJID, String actorNickname, St
     /**
      * Sets the minimal role of persons that are allowed to send private messages in the room. The provided value is
      * any one of: "anyone", "moderators", "participants", "none". If another value is set, "anyone" is used instead.
-     *
-     * @role The minimal role of persons that are allowed to send private messages in the room (never null).
```
guusdk: Instead of removing this, "@role" could be replaced by "@param role"

src/java/org/jivesoftware/openfire/crowd/CrowdAuthProvider.java
```diff
@@ -47,9 +47,6 @@ public CrowdAuthProvider() {
      * Returns if the username and password are valid; otherwise this
      * method throws an UnauthorizedException.<p>
      *
-     * If {@link #isPlainSupported()} returns false, this method should
-     * throw an UnsupportedOperationException.
-     *
```
guusdk: Was this removed because a technical javadoc issue (that I can't see), or because the statement itself is incorrect?

src/java/org/jivesoftware/openfire/muc/MUCRoom.java
akrherz: addressed, thanks

src/java/org/jivesoftware/openfire/crowd/CrowdAuthProvider.java
akrherz: Yes, incorrect, method doesn't throw `UnsupportedOperationException`, nor is isPlainSupported() a method.

PR #705, opened 2016-12-16
OF-477: Improve default XMPP domain name and FQDN

guusdk:
This commit intends to have better default values for two properties of Openfire:
- the XMPP domain name (eg: igniterealtime.org)
- the FQDN of the server (eg: xmpp.igniterealtime.org)

Older code did not always make a clear distinction (ambiguously referring to
'server name'). Having a proper FQDN has become more important, as newer SASL
implementations depend on the client using the same FQDN as what is configured
in the server).

This commit removes support for changing the XMPP domain name at run time (although
you can obviously still change the corresponding property). Instead, the FQDN can
now be changes (although will require a restart).

Additionally, the XMPPServerInfo implementation got a bit of a refactoring: values
that are not expected to be modified (or considered unmodifiable from a functional
perspective), are no longer passed around, but are rather hardcoded in the
implementation. Support for deprecated ServerPort was removed.

Finally, most access to the properties that hold the XMPP domain name and FQDN
are now encapsulated by XMPPServerInfoImpl.

### DIFF
--- a/src/i18n/openfire_i18n_en.properties
+++ b/src/i18n/openfire_i18n_en.properties
@@ -1037,8 +1037,8 @@ index.version=Version:
 index.home=Server Directory:
 index.certificate-warning=Found RSA certificate that is not valid for the server domain.
 index.certificate-error=Unable to access certificate store.  The keystore may be corrupt.
-index.server_name=Server Name:
-index.host_name=Host Name:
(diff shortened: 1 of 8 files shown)


guusdk:
Also: the 'host name' is now retrieved from InetAddress's "CanonicalHostname", which I expect gives better results.

PR #706, opened 2016-12-17
OF-942 et passim

dwd:
This closes off several issues found by Simon Waters (Surevine Ltd).


guusdk:
My only concern is that in places, the CSRF failure is hidden: an update is silently ignored, without any kind of error message. I've already encountered this in the wild before, which is terribly confusing. Some kind of error message would improve things a lot.

dwd:
It's even confusing to Simon. Let me see what I can do there.

PR #707, opened 2016-12-18
OF-1244: Admin console DNS SRV evaluation

guusdk:
This commit adds an admin console page that allows an administrator to evaluate
the DNS SRV configuration in context of the XMPP domain name and fully qualified
domain name of the sever that is running Openfire.

### DIFF
--- a/src/web/dns-check.jsp
+++ b/src/web/dns-check.jsp
@@ -1,0 +1,136 @@
+<%@ page import="java.util.*"
+         errorPage="error.jsp"
+%>
+<%@ page import="org.jivesoftware.openfire.net.DNSUtil" %>
+<%@ page import="org.jivesoftware.openfire.XMPPServer" %>
+
+<%@ taglib uri="admin" prefix="admin" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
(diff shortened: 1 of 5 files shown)


guusdk:
Note that I did not test this with a domain that had DNS SRV records set up. Let's do that on ignite. :)

src/web/dns-check.jsp
```diff
@@ -0,0 +1,136 @@
...
+            <fmt:message key="system.dns.srv.check.no-match.one-liner" />
+        </admin:infobox>
+        <fmt:message key="system.dns.srv.check.no-match.description" var="plaintextboxcontent"/>
+    </c:otherwise>
+</c:choose>
+
+<p>
+    <fmt:message key="system.dns.srv.check.info">
+        <fmt:param value="${xmppDomain}" />
+        <fmt:param value="${hostname}" />
```
dwd: I think these will need escaping to avoid an XSS.

dwd: You could probably escape the parameters set above, actually.

guusdk: Shouldn't be necessary, as this page takes no user input whatsoever?

dwd: If we validate the xmppDomain and hostname elsewhere then we're OK.

PR #708, opened 2016-12-18
OF-1245: Prevent error while parsing self-generated certificate

guusdk:
Openfire, during setup, generates a certificate, that it cannot parse later.

This commit changes the generation routine, changing the subjectAltName value
from a tagged object to non-tagged string.


guusdk:
@dwd I am uncertain if I evaluated the problem correctly. As you are an expert of many things, including this, could you have a look? Is the problem indeed in the generation of the certificate, or should I have modified the parsing  of certificates instead?

dwd:
Flattery will get you everywhere.

PR #709, opened 2016-12-18
OF-1188 Add remote domain access check in main thread

dwd:
By adding an additional access check fo the remote domain in the main loop,
we avoid the need to tie up an OutgoignSessionPromise thread during S2S
initiation.


src/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
```diff
@@ -477,7 +478,11 @@ private boolean routeToRemoteDomain(JID jid, Packet packet,
 		        }
 		    }
 		}
-		else {
+		else if (!RemoteServerManager.canAccess(jid.getDomain())) { // Check if the remote domain is in the blacklist
+            Log.info( "Unable to authenticate: Remote domain is not accessible according to our configuration (typical causes: server federation is disabled, or domain is blacklisted)." );
```
akrherz: Would logging the actual domain here be useful? :)

dwd: Yes... I just copied the existing error message, but I think that must have the domain in the logger. I'll fix this.

PR #713, opened 2016-12-19
OF-867 Wrong key in groupMetaCache evict

akrherz:
untested by me, patch by @aberenguel

### DIFF
--- a/src/java/org/jivesoftware/openfire/group/GroupManager.java
+++ b/src/java/org/jivesoftware/openfire/group/GroupManager.java
@@ -215,7 +215,7 @@ public class GroupManager {
                     JID userJid = new JID(user);
                     if (XMPPServer.getInstance().isLocal(userJid)) {
                         String username = userJid.getNode();
-                        synchronized (username.intern()) {
+                        synchronized ((getClass().getSimpleName() + username).intern()) {
                             groupMetaCache.remove(username);
                         }
                     }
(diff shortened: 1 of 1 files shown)


src/java/org/jivesoftware/openfire/group/GroupManager.java
```diff
@@ -213,11 +207,20 @@ public void adminRemoved(Group group, Map params) {
...
-                }
+                evictCachedUserForGroup(member);
             }
 
+            private void evictCachedUserForGroup(String user) {
+                if(user != null) {
+                    JID userJid = new JID(user);
+                    if (XMPPServer.getInstance().isLocal(userJid)) {
+                        String username = userJid.getNode();
+                        synchronized (username.intern()) {
```
dwd: I'm not crazy about synchronizing on an interned string. Was the lack of a lock a problem before?

dwd:
I'm not so much "requesting changes" as wanting to know some rationale, BTW.

src/java/org/jivesoftware/openfire/group/GroupManager.java
guusdk: The code appears to lock, always, based on the cache-key that is accessed. The same lock is used when data is added for a particular username in ```getSharedGroups()```, so in that sense it is "correct" to do that here too.

I do share @dwd's concern of using an interned string - more so when the interned string is a very generic one, such as a username. If any other code (unrelated to this functionality), applies the same technique of locking on a username, we're in trouble. A quick and dirty improvement would be to apply some kind of namespace. Instead of locking on "username", we could lock on "classname+username" - or something similar.

guusdk: let's replace ```synchronized (username.intern())``` with ```synchronized ( (getClass().getSimpleName() + username).intern())``` throughout this class.

guusdk:
With Daryl's change, this looks good to me.

PR #715, opened 2016-12-19
OF-1246 MAM for MUC

dwd:
Tidied:
* Lots of bugfixes to 1:1 MAM here as well.
* Some of this patch seems to have slipped into the existing master (basically the database upgrade scripts contain the new fields already).
* As per discussions, the MAM-MUC code is actually hitting the core's database, not the Monitoring Plugin's MUC tables, so we can move it when the 1:1 MAM code moves.

### DIFF
--- a/src/database/openfire_db2.sql
+++ b/src/database/openfire_db2.sql
@@ -261,6 +261,7 @@ CREATE TABLE ofMucMember (
 
 CREATE TABLE ofMucConversationLog (
   roomID              INTEGER       NOT NULL,
+  messageID         INTEGER NOT NULL,
   sender              VARCHAR(2000) NOT NULL,
   nickname            VARCHAR(255),
   logTime             CHAR(15)      NOT NULL,
@@ -269,6 +270,7 @@ CREATE TABLE ofMucConversationLog (
   stanza              CLOB
 );
 CREATE INDEX ofMucConvLog_time_idx ON ofMucConversationLog (logTime);
+CREATE INDEX ofMucConvLog_msg_id ON ofMucConversationLog (messageID);
 
 
 -- PubSub Tables
@@ -392,7 +394,7 @@ INSERT INTO ofID (idType, id) VALUES (19, 1);
 INSERT INTO ofID (idType, id) VALUES (23, 1);
 INSERT INTO ofID (idType, id) VALUES (26, 2);
 
-INSERT INTO ofVersion (name, version) VALUES ('openfire', 24);
+INSERT INTO ofVersion (name, version) VALUES ('openfire', 25);
 
 -- Entry for admin user
 INSERT INTO ofUser (username, plainPassword, name, email, creationDate, modificationDate)
--- a/src/database/upgrade/25/openfire_db2.sql
+++ b/src/database/upgrade/25/openfire_db2.sql
@@ -1,0 +1,1 @@
+UPDATE ofVersion SET version = 25 WHERE name = 'openfire';
--- a/src/database/upgrade/25/openfire_db2.sql
+++ b/src/database/upgrade/25/openfire_db2.sql
@@ -1,1 +1,3 @@
+CREATE INDEX ofMucConvLog_msg_id ON ofMucConversationLog (messageID);
+
 UPDATE ofVersion SET version = 25 WHERE name = 'openfire';
--- a/src/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
@@ -129,8 +129,8 @@ public class MUCPersistenceManager {
     private static final String DELETE_USER_MUCAFFILIATION =
         "DELETE FROM ofMucAffiliation WHERE jid=?";
     private static final String ADD_CONVERSATION_LOG =
-        "INSERT INTO ofMucConversationLog (roomID,sender,nickname,logTime,subject,body,stanza) " +
-        "VALUES (?,?,?,?,?,?,?)";
+        "INSERT INTO ofMucConversationLog (roomID,messageID,sender,nickname,logTime,subject,body,stanza) " +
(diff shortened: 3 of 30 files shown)


dwd:
Fixed the patches, rebased to make them (considerably) cleaner.

src/database/openfire_db2.sql
```diff
@@ -261,6 +261,7 @@ CREATE TABLE ofMucMember (
 
 CREATE TABLE ofMucConversationLog (
   roomID              INTEGER       NOT NULL,
+  messageID         INTEGER NOT NULL,
```
guusdk: *twitch*

src/database/upgrade/25/openfire_db2.sql
```diff
@@ -0,0 +1 @@
+UPDATE ofVersion SET version = 25 WHERE name = 'openfire';
```
guusdk: You're not creating the new index in any of the upgrade scripts, but you did add them to the initial installation scripts - is that intended?

src/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
```diff
@@ -129,8 +129,8 @@
     private static final String DELETE_USER_MUCAFFILIATION =
         "DELETE FROM ofMucAffiliation WHERE jid=?";
     private static final String ADD_CONVERSATION_LOG =
-        "INSERT INTO ofMucConversationLog (roomID,sender,nickname,logTime,subject,body,stanza) " +
-        "VALUES (?,?,?,?,?,?,?)";
+        "INSERT INTO ofMucConversationLog (roomID,messageID,sender,nickname,logTime,subject,body,stanza) " +
+        "SELECT ?,COUNT(*),?,?,?,?,?,? FROM ofMucConversationLog";
```
guusdk: Is that thread safe, in any SQL server?

src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/MucMamPersistenceManager.java
```diff
@@ -0,0 +1,203 @@
...
+                    }
+                    stanza = message.toString();
+                }
+                ArchivedMessage archivedMessage = new ArchivedMessage(sentDate, ArchivedMessage.Direction.from, null, null);
+                archivedMessage.setStanza(stanza);
+                archivedMessage.setId(id);
+                msgs.add(archivedMessage);
+            }
+        } catch (SQLException e) {
+            // TODO ???
```
guusdk: If anything, this exception should be logged.

src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/MucMamPersistenceManager.java
```diff
@@ -0,0 +1,203 @@
...
+            if (msgs.size() > 1) {
+                xmppResultSet.setLast(msgs.get(msgs.size() - 1).getId());
+            }
+        }
+        return msgs;
+    }
+
+    @Override
+    public Collection<Conversation> getActiveConversations(int conversationTimeout) {
+        return null;
```
guusdk: Perhaps return an empty collection instead?

src/database/upgrade/25/openfire_db2.sql
dwd: No, not really. Thanks.

src/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
dwd: Yes, entirely - it's within a single transaction.

src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/MucMamPersistenceManager.java
dwd: It can't be called - it's part of the '136 support I'm not attempting to handle for MUC (and not even sure it's possible to handle).

src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/MucMamPersistenceManager.java
dwd: OK

src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/MucMamPersistenceManager.java
guusdk: In that case, perhaps throw an UnsupportedOperationException?

dwd:
Anything else to do here?

src/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
akrherz: This SQL doesn't look right to me, the `COUNT(*)` needs to be in an aggregate function, right?

dwd: Those are all really database-specific, and often really hard to add to an existing table. I'd have had to do the upgrade by creating a new table and populating it from the old one (which I can do, but getting it right in every database we support....?)

akrherz: Yeah, but I worry the count(*) approach will be very resource expensive for the DB to run.  Not all DBs are optimized for this query to be fast and requires a full table sequence scan.  I'll investigate this some more now.

dwd: You think? A COUNT(*) should be fast everywhere; it's a COUNT(something) that might introduce a table scan. FWIW, I believe we're already doing similar for 1:1 messages in support of MAM.

PR #721, opened 2016-12-21
OF-1195 Correct SCRAM supported logic

dwd:
Agreed with the suggestion in the bug report - this should be OR instead of AND. The logging message even says as much.


dwd:
Actually, scrap this. I'm not convinced it's this simple when I look deeper, and without properly testing it I'd rather put this aside for now.

akrherz:
I think this PR is right, both capabilities need to be supported for SCRAM to be supported by the Provider.

PR #725, opened 2016-12-23
Add the ability to disabled delayed delivery (XEP-203)

GregDThomas:
This pretty small change allows an admin to disable delayed delivery by setting the sysprop `xmpp.pubsub.disable-delayed-delivery` to `true` (naturally, this defaults to false so will have no affect unless set).

Not all use-cases require delayed delivery, and this provides an easy way to disable it rather than require each client to ignore messages tagged with `<delay xmlns='urn:xmpp:delay' ... />`.


GregDThomas:
Is there any chance of getting this PR in the next release? Is there anything else that needs to be done to get it merged? TIA

akrherz:
This is targeted for Openfire 4.2.0 release, which does not have an ETA at the moment.

PR #732, opened 2016-12-30
OF-1253: Be more specific when detecting default username/password

guusdk:
Exception handling is used to determine if the default admin console username/password
applies. This exception handling should be more specific than depend on any Exception.
Instead, authentication-related exceptions should be used.


akrherz:
I was worried that this change would fail if the `admin` account does not exist, but it appears that the `AuthProvider#authenticate` will trap the `UserNotFoundException` and instead return the `UnauthorizedException`, so should be OK

PR #734, opened 2017-01-03
remove legacy CVS keywords

akrherz:
I'll certainly take the blame for false positives with this change.  I did a couple of diagnostic greps over the diff to look for erroneous modifications.  The procedure I followed was to grep for the following tags: `RCSfile` `Date` `Revision` `Id` and then run a python script over the results.

1. git grep '$CVStag' > todo.log
2. run a python script over that todo.log looking to squash lines that matched the regex

```python
import os
import re

TAGS = re.compile("\$(RCSfile|Date|Revision|Id)")

for line in open('todo.log'):
  if line.startswith("Binary"):
    continue
  fn = line.split(":")[0]
  lines = open(fn).readlines()
  o = open(fn, 'w')
  for line in lines:
    if not TAGS.search(line):
      o.write(line)
  o.close()
```


guusdk:
Doesn't look to bad to me, but time will tell, i guess ;)

Perhaps we should do a couple of follow up changes:

One that removes these two lines (followed by any whitespace-only or blank lines):

    

Also it would be good to replace:

    /**
     *

with:

    /**

(basically, remove any line following ```/**``` that only has a star and whitespace in it)

akrherz:
@guusdk hopefully I have now done just that.

guusdk:
Excellent. Can you do this for other repositories too?

akrherz:
@guusdk sure, I did igniterealtime/Smack#99 , igniterealtime/Spark#283, and igniterealtime/IgniteRealtime-Website#58

guusdk:
Much obliged. I've reviewed the website's and Smack's, which looked fine (@wrooot beat me to Spark's, which is already merged). I merged the website one, but will leave Smack's to @Flowdalic

wrooot:
Spark builds fine, so everything is fine (i suppose) :)

PR #735, opened 2017-01-04
OF-1263: Improve cache invalidation for group metadata

guusdk:
When evicting cache entries for group members, the members of other
groups that are affected should also be evicted.


guusdk:
@tevans My change applies to your changes for group management. As such, I'd love your feedback!

wrooot:
Maybe the issue originated from the older change in 3.8.0 but it was intensified by some other change in 4.1.0. I think applying this fix and letting affecting users try it out is better than nothing. Maybe it will also help with OF-1269

PR #743, opened 2017-01-18
OF-1275

speedy01:
This should allow Openfire to connect to LDAPs using TLS 1.0, TLS 1.1, or TLS 1.2, instead of capping the protocol to TLS 1.0

### DIFF
--- a/src/java/org/jivesoftware/util/SimpleSSLSocketFactory.java
+++ b/src/java/org/jivesoftware/util/SimpleSSLSocketFactory.java
@@ -52,7 +52,7 @@ public class SimpleSSLSocketFactory extends SSLSocketFactory implements Comparat
     public SimpleSSLSocketFactory() {
 
         try {
-            final SSLContext sslContext = SSLContext.getInstance("TLSv1");
+            final SSLContext sslContext = SSLContext.getInstance("TLSv1.2");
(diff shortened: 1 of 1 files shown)


guusdk:
Would it make sense to make this configurable?

speedy01:
I thought about that as well. I could see the benefits of it being configurable, and allowing an admin to guaranty the protocols used.  Would something like that be difficult to implement?

guusdk:
Not if you don't bother with a nice UI. You can simply use something like: ```JiveGlobals.getProperty( "property.name", "default value" );``` (from memory, might be slightly different. It is used all over the place, should be easy to find an example.

speedy01:
I don't think it would need a UI. It looks like SimpleSSLSocketFactory is also used with the email service. I might have to defer to you since I'm still learning.  Adding a specific ldaps property here probably wouldn't be a good idea!

akrherz:
I did some looking around and it appears to be that IBM, Oracle, and OpenJDK all support `TLSv1.2` in Java 7+.  Setting this to `TLSv1.2` should gracefully degrade back to `TLSv1` if necessary.  I think this change is fine without adding a property, but am not against it.  Just want this in for the 4.1.2 release :)

PR #748, opened 2017-02-11
OF-1287 Discard spoofed Carbons (CVE-2017-5589+)

dwd:
This patch detects Carbons which appear to be using the attack described in
the CVE-2017-5589 report, and discards them.

### DIFF
--- a/src/java/org/jivesoftware/openfire/session/LocalClientSession.java
+++ b/src/java/org/jivesoftware/openfire/session/LocalClientSession.java
@@ -19,6 +19,7 @@ package org.jivesoftware.openfire.session;
 import java.net.UnknownHostException;
 import java.util.*;
 
+import org.dom4j.Element;
 import org.jivesoftware.openfire.Connection;
 import org.jivesoftware.openfire.SessionManager;
 import org.jivesoftware.openfire.StreamID;
@@ -40,10 +41,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
-import org.xmpp.packet.JID;
-import org.xmpp.packet.Packet;
-import org.xmpp.packet.Presence;
-import org.xmpp.packet.StreamError;
+import org.xmpp.packet.*;
 
 /**
  * Represents a session between the server and a client.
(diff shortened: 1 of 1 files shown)


src/java/org/jivesoftware/openfire/session/LocalClientSession.java
```diff
@@ -932,6 +932,26 @@ public void setMessageCarbonsEnabled(boolean enabled) {
...
+        if (JiveGlobals.getBooleanProperty(DISCARD_CARBON_SPOOF, true)) {
+            Message msg = (Message)packet;
+            if (msg != null) {
+                PacketExtension carbon = msg.getExtension("received", "urn:xmpp:carbons:2");
+                if (carbon != null) {
+                    try {
+                        Element copy = carbon.getElement().element("forwarded").element("message");
+                        JID carbonFrom = new JID(copy.attribute("from").getValue());
+                        if (carbonFrom.asBareJID() != packet.getFrom()) {
+                            Log.info("Carbons message appears spoofed, dropping");
```
akrherz: I'd prefer to see a better INFO log here, this message does not provide the Openfire admin with any helpful info, like who the stanza is from or how it is making the spoof test.

src/java/org/jivesoftware/openfire/session/LocalClientSession.java
```diff
@@ -932,6 +932,26 @@ public void setMessageCarbonsEnabled(boolean enabled) {
...
 	public boolean canProcess(Packet packet) {
+        if (JiveGlobals.getBooleanProperty(DISCARD_CARBON_SPOOF, true)) {
+            Message msg = (Message)packet;
+            if (msg != null) {
+                PacketExtension carbon = msg.getExtension("received", "urn:xmpp:carbons:2");
+                if (carbon != null) {
+                    try {
+                        Element copy = carbon.getElement().element("forwarded").element("message");
+                        JID carbonFrom = new JID(copy.attribute("from").getValue());
+                        if (carbonFrom.asBareJID() != packet.getFrom()) {
```
guusdk: I'm not sure if this is correct. Message stanzas that have a carbon should always be from the intended recipients bare JID, and to a full JID of the intended recipient, correct? (Maybe checking that is enough for spoof prevention?) There are two types of carbons: 'received' carbons, and 'sent' carbons. The 'from' of the forwarded message in a 'received' carbon can be another entity address ("one of your full jids received a message from another user, all of your other full jids receive a carbon).

dwd: Yeah, you're right. No idea what I was(n't) thinking.

PR #749, opened 2017-02-16
OF-1288 Add missing attributes

dwd:
This adds attributes mandated by MAM, by missing from the rendering in the
PacketExtension.


dwd:
Found another issue; I'll fix this in the same PR.

PR #750, opened 2017-02-16
Support UPN OID extraction directly in SANCertificateIdentityMapping

hdeadman:
This change allows a sub-class to support additional OID values in the certificate by overriding a single method.

I am extracting the "otherName" from a client cert (smartcard) and the OID is 
[1.3.6.1.4.1.311.20.2.3](http://www.oid-info.com/get/1.3.6.1.4.1.311.20.2.3) which this class doesn't handle. 

With the change in this pull request my subclass is able to extract the string value of otherName with the following code:

```java
	protected String parseOtherName(ASN1ObjectIdentifier typeId, ASN1Primitive value) {
		String otherName = null;
		if (value instanceof ASN1TaggedObject) {
			ASN1TaggedObject taggedObject = (ASN1TaggedObject) value;
			ASN1Primitive primitive = taggedObject.getObject();
			if (primitive instanceof DERUTF8String) {
				otherName = ((DERUTF8String)primitive).getString();
			} else {
		    	logger.info("Object in tagged object is of type: " + primitive.getClass().getName() );
				otherName = primitive.toString();
			}
	    } else {
	    	logger.info("Other name in certificate is of type: " + value.getClass().getName() );
	    	otherName = value.toString();
	    }
    	logger.info("otherName from certificate has value of: " + otherName );
    	return otherName;
    }
```
If you would prefer I could add support for that OID directly to the case statement in this class, but in that case this change still might be helpful if someone needed handle a different OID in the future or if they needed to override one of the parse* methods for the OIDs that are handled (I would need to change two existing parse* methods in the class to not be static, they don't really need to be static).

### DIFF
--- a/src/java/org/jivesoftware/util/cert/SANCertificateIdentityMapping.java
+++ b/src/java/org/jivesoftware/util/cert/SANCertificateIdentityMapping.java
@@ -42,7 +42,15 @@ public class SANCertificateIdentityMapping implements CertificateIdentityMapping
      * @see <a href="https://tools.ietf.org/html/rfc4985">RFC 4985</a>
      */
     public static final String OTHERNAME_SRV_OID = "1.3.6.1.5.5.7.8.7";
+    
+    /**
+     * User Principal Name (UPN) Object Identifier.
+     *
+     * @see <a href="http://www.oid-info.com/get/1.3.6.1.4.1.311.20.2.3">User Principal Name (UPN)</a>
+     */
+    public static final String OTHERNAME_UPN_OID = "1.3.6.1.4.1.311.20.2.3";
 
+    
     /**
      * Returns the JID representation of an XMPP entity contained as a SubjectAltName extension
      * in the certificate. If none was found then return an empty list.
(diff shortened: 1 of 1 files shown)


dwd:
Happy with this change, although:
* I'd be quite happy with this particular case (UPN) being supported in the core.
* I've a vague preference for downloadable plugins, rather than subclassing which would - I think - require more from administrators.

hdeadman:
OK, hold off on merging this, I have made update to support UPN directly in the class but I need to test it and I am not in a position to do that at the moment.

hdeadman:
OK I updated this class to support UPN, Let me know if you want me to make any changes.

hdeadman:
@dwd I was hoping to get this merged for the next release if possible, let me know if any of the changes are holding it up. I made some methods non-static but they were things that probably shouldn't be static and it is unlikely that anyone was using them from anywhere else. If that is a problem I can keep them static. The class will no longer be extensible but I don't need to extend it now that it supports UPN.

PR #753, opened 2017-02-21
Allow code reuse from plugins

Gugli:
For example, PacketInterceptors may need to reuse these functions.


Gugli:
Here is an example of a plugin using these function. Writing such a plugin without access to these functions seemed tedious and risky to me.

https://github.com/Gugli/Openfire/blob/roster-restrictions-interceptor-plugin/src/plugins/rosterRestrictions/src/java/org/jivesoftware/openfire/plugin/rosterrestrictions/RosterRestrictionsPlugin.java

Gugli:
Still no review for this PR ? It has been in prod on my side for some weeks with between 200-3000 simultaneous users. No issues to be reported yet.

akrherz:
@Gugli sorry there has been no review, will set this against 4.2 release in hopes of getting one.

PR #754, opened 2017-02-22
Fix OF-1263 and OF-1269 by reverting OF-867

akrherz:
Reverting changes introduced in Openfire 4.1 by OF-867 fixed issues seen by at least two users that reported success in the forums.

### DIFF
--- a/src/java/org/jivesoftware/openfire/group/GroupManager.java
+++ b/src/java/org/jivesoftware/openfire/group/GroupManager.java
@@ -171,7 +171,9 @@ public class GroupManager {
                 
                 // Remove only the collection of groups the member belongs to.
                 String member = (String) params.get("member");
-                evictCachedUserForGroup(member);
+                if(member != null) {
+	                groupMetaCache.remove(member);
+                }
             }
 
             @Override
@@ -182,7 +184,9 @@ public class GroupManager {
                 
                 // Remove only the collection of groups the member belongs to.
                 String member = (String) params.get("member");
-                evictCachedUserForGroup(member);
+                if(member != null) {
+	                groupMetaCache.remove(member);
+                }
             }
 
             @Override
@@ -193,7 +197,9 @@ public class GroupManager {
                 
                 // Remove only the collection of groups the member belongs to.
                 String member = (String) params.get("admin");
-                evictCachedUserForGroup(member);
+                if(member != null) {
+	                groupMetaCache.remove(member);
+                }
             }
 
             @Override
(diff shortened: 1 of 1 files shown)


aberenguel:
After some analysis, I saw that the commit ffe14c90b18 exchanged all groupMetaCache keys by the user node (username).

The key = userJID was used by getGroups(JID user) and the key = username was used by getSharedGroups(String username).

Making the same key (username) be used in both places **was a mistake of mine**. The getSharedGroups is wider than getGroups. Take a look in https://github.com/igniterealtime/Openfire/blob/master/src/java/org/jivesoftware/openfire/group/AbstractGroupProvider.java#L181

The right thing to do when the user is removed from the group is to remove both keys: username and userJID.

I'll make a fix for it.

akrherz:
@aberenguel thank you so much for looking into this, shall I merge this revert for your new PR to then come in new overtop this or would you rather simply correct what's currently in master?  Also, what do you think of the additional change @guusdk made with your original change?

aberenguel:
@akrherz I've just fixed it in my of1263 branch https://github.com/aberenguel/Openfire/commit/7a31c5c115a4798044980202d251994824a6b0aa

aberenguel:
About the change @guusdk made is valid. BTW, I forgot to revert it in my commit.

aberenguel:
https://github.com/aberenguel/Openfire/of1263

akrherz:
I'll pulled your changes into this PR, thanks!

aberenguel:
Just one thing. The getClass() in evictCachedUserForGroup was returning the GroupEventListener class. 
I changed a little to make it less error prone.

akrherz:
pulled that change too

aberenguel:
Thanks @akrherz !

What do you thing about this commit? https://github.com/aberenguel/Openfire/commit/3017e39b7fa85806e5e6eeb5af380d0c0201b48e

This groupMetaCache is a lot confusing. I thought it is better make all keys prefixed with a constant. So we can understand for what a key is used.

akrherz:
@aberenguel , the development team discussed this today and here's what we wish to do.

1. Revert the previous changes and make an Openfire 4.1.3 release
2. Kindly ask that you resubmit your subsequent changes as a new PR and we'll consider for 4.2.0, I can certainly do this for you, if you would like.

Thanks again for your efforts on this.  I just want to get a 4.1.2 fix out the door and our main coding devs need more time to consider your additional changes.

aberenguel:
@akrherz  I totally agree, since the need for fixed 4.1.3 is urgent.

I found another bugs related to group chache that were not fixed in these commits.

So, what is the better approach? 

1. Describe how to reproduce each bug in the https://community.igniterealtime.org/community/developers/openfire_dev and wait for new Jira issue to be created; or
2. Directly create a PR for each bug, describing how to reproduce in each PR; or
3. Create a big PR fixing all bugs

akrherz:
@aberenguel I've elevated your Jira account privs to allow creation of issues.  So please create issues as you see fit and associate your PRs with them, thanks!

PR #757, opened 2017-03-02
OF-1295 add a note about service to upgrade guide

wrooot:


wrooot:
Not sure if this PR is right. Using GitHub via web and that's how it works (outs everything into one PR).

PR #761, opened 2017-03-12
New Candy plugin

guusdk:
A new Openfire plugin, that bundles the third-party Candy web client.


guusdk:
As inspired by @deleolajide 😉

dwd:
Should this be in the core repo? And do we intend (eventually) shipping this by default?

guusdk:
I would say: yes, respectively, no. But I shall be easily persuaded otherwise on both items.

akrherz:
This PR need an update for maven? :)

deleolajide:
I think it should be in the core repo simply because we use it at igniterealtime.org for open_chat. I am assuming that we can make PRs to add Candy plugins like the one we use for Openfire Meetings at work.

guusdk:
Adding common plugins would obviously be possible, but perhaps we should find a way in which administrators can add (proprietary) Candy plugins in their version of Openfire.

It'd be interesting to see how that would be managed (would it survive a re-install of the Openfire plugin?) I don't know if there is any part of the Admin Console where we have custom content like this. If we do, we could perhaps borrow from the mechanism that's used there.

PR #766, opened 2017-03-22
OF-1304 Correct query builder syntax

akrherz:
Suggested patch by David Liffmann

### DIFF
--- a/src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/JdbcPersistenceManager.java
+++ b/src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/JdbcPersistenceManager.java
@@ -178,7 +178,7 @@ public class JdbcPersistenceManager implements PersistenceManager {
 			+ "ofMessageArchive.sentDate, "
 			+ "ofMessageArchive.stanza, "
 			+ "ofMessageArchive.messageID "
-			+ "FROM ofMessageArchive";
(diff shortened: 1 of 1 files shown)


src/plugins/monitoring/src/java/com/reucon/openfire/plugin/archive/impl/JdbcPersistenceManager.java
```diff
@@ -456,7 +456,7 @@ private int bindConversationParameters(Date startDate, Date endDate, String owne
 			appendWhere(whereSB, "( ", MESSAGE_TO_JID, " = ? OR ", MESSAGE_FROM_JID, " = ? )");
 		}
 		if (whereSB.length() != 0) {
-			querySB.append(" AND ").append(whereSB);
+			querySB.append(" WHERE ").append(whereSB);
 		}
```
dwd: Not quite right - SELECT_MESSAGES_ORACLE has no WHERE clause, but SELECT_MESSAGES always does. I think the right fix is therefore to add a dummy WHERE clause to the Oracle statement (WHERE 1=1 or something).

akrherz: Thanks, I agree!

akrherz:
@dwd I think I addressed your feedback, PR ready to go now?

PR #767, opened 2017-03-22
OF-1305: Make escaping of multibyte characters in LDAP search optional.

guusdk:
A new property (ldap.encodeMultibyteCharacters) is introduced that controls if multibyte characters in LDAP search queries are escaped.

Escaping of these characters started with the fix for OF-830, which appears to have caused OF-1305. Although I can't say that escaping of characters is wrong (per RFC 4515), it does cause real-world problems.

This fix defaults to not encoding again (reverting back to the behavior pre OF-830, without affecting other changes made in OF-830).


akrherz:
[Tested and verified on forums](https://community.igniterealtime.org/thread/60433)

PR #769, opened 2017-03-22
OF-1306: Add a cache for UserDN search results.

guusdk:
Openfire repeatedly searches for the same UserDN values. This cache should alleviate the backend storage and improve performance.


guusdk:
I am not to sure about this one. The approach that I took was a naive one: "this takes a lot of time and repeats itself, let's cache it." I have not looked if we can reorganize the code in a more efficient pattern, nor did I look for similar improvements. Finally, I've added a default cache. I'm not sure what it's capacity is, nor what its eviction policy is (or should be).

guusdk:
@speedy01 reported that the change was effective: things speed up.

PR #770, opened 2017-04-06
Gracefully close the stream with </stream:stream>

sco0ter:
Fixes OF-1308


akrherz:
@guusdk and I discussed this patch today on our weekly hangout and are going to run with it and see what fun we have this time :)  The patch should be good, but it seems we were hitting side-effects within MINA that was causing this type of approach to fail.  Will gen up a 4.2 build for igniterealtime and run with it!

PR #772, opened 2017-04-12
Fixes for OF-1311 and OF-1312

guusdk:
The latter depends on (and is a showcase for) the former, which is why I combined them in one PR.


guusdk:
OF-1312 adds a section to the "Registration & Login" admin console page, which will look like this:
sasl-mechs-3

It will auto-detect available SASL mechanisms.

PR #775, opened 2017-04-18
OF-546: Plugins should not inherit from parent.

guusdk:
The build plans for the plugins should not inherit from the parent project, as this
ties their existence in with our parent project. Third party developers cannot do
this (without modifying the parent project structure). We should facilitate third
party developers, and use the same tools as we provide them.

This commit removes the parent/child dependencies between the top-level parent POM
and the plugins POM. There still as a parent/child dependency between the plugins
POM and each individual plugin.

Although I'm not sure this change is a complete solution, it is aimed to more closely
represent a development setup in which a third party developer might find itself. To
create a plugin of their own, they can now:

 - Use our 'plugins' module directly (define it as a parent)
 - Replace our 'plugins' module by a proprietary one (when developing multiple
   proprietary plugins).
 - Take most POM definition from our 'plugins' module and merge that with a new
   stand-alone plugin project.

### DIFF
--- a/plugins/pom.xml
+++ b/plugins/pom.xml
@@ -1,11 +1,8 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <parent>
-        <artifactId>parent</artifactId>
-        <groupId>org.igniterealtime.openfire</groupId>
-        <version>4.2.0-SNAPSHOT</version>
-    </parent>
+    <groupId>org.igniterealtime.openfire</groupId>
+    <version>4.2.0-SNAPSHOT</version>
     <artifactId>plugins</artifactId>
     <packaging>pom</packaging>
     <name>Openfire Plugins</name>
@@ -52,7 +49,25 @@
         <module>xmldebugger</module>
     </modules>
 
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <!-- Versions -->
+        <openfire.version>4.2.0-SNAPSHOT</openfire.version>
+        <jetty.version>9.2.14.v20151106</jetty.version>
+        <slf4j.version>1.7.7</slf4j.version>
+    </properties>
+
     <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.7</source>
+                    <target>1.7</target>
+                </configuration>
+            </plugin>
+        </plugins>
         <pluginManagement>
             <plugins>
                 <plugin>
@@ -62,7 +77,7 @@
                         <dependency>
                             <groupId>org.igniterealtime.openfire.plugins</groupId>
                             <artifactId>openfire-plugin-assembly-descriptor</artifactId>
-                            <version>${project.version}</version>
+                            <version>${openfire.version}</version>
                         </dependency>
                     </dependencies>
                     <executions>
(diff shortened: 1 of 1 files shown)


guusdk:
This PR is intended to improve on the issues as described in the bottom part of the discussion in #347.

guusdk:
There are likely further improvements possible, but with the changes in this PR, I intend to make it possible to create an Openfire plugin project with a POM like this:

```
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <artifactId>plugins</artifactId>
        <version>4.2.0-SNAPSHOT</version>
        <relativePath></relativePath>
    </parent>

    <groupId>org.example</groupId>
    <artifactId>my-openfire-plugin</artifactId>
    <version>1.0-SNAPSHOT</version>

    <build>
        <sourceDirectory>src/java</sourceDirectory>
        <plugins>
            <plugin>
                <artifactId>maven-assembly-plugin</artifactId>
            </plugin>
        </plugins>
    </build>
</project>```

GregDThomas:
You define both jetty.version and slf4j.version in the plugins pom.xml; given you've got xmppserver as a provided dependency, won't the plugins inherit the appropriate versions of these from the xmppserver ?

i.e. what I'm saying is - can/should these definitions be removed?

guusdk:
``slf4j.version`` isn't even used, so that can definitely be removed. I don't think we can remove the jetty version though, as properties defined in dependencies are not accessible in the project as far as I know (unless it's defined in a parent project, which it now is not).

GregDThomas:
I'm worried that each plugin must define both the version of Openfire (4.2.0-SNAPSHOT) and the version of Jetty (9.2.14.v20151106) - and these must match (or there is a risk of undefined and obscure bugs). 

It it worth having a incredibly simply parent pom that /only/ defines the versions of all these dependencies. Anything Openfire related could use this as a parent POM, and for any given version of Openfire you can be sure of getting the correct versions. i.e. reference

    ...
    <parent>
        <artifactId>properties</artifactId>
        <version>4.2.0-SNAPSHOT</version>
    </parent>
   ...

whilst the properties artifact contains just ...

    <properties>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <openfire.version>4.2.0-SNAPSHOT</openfire.version>
        <jetty.version>9.2.14.v20151106</jetty.version>
        <slf4j.version>....</slf4j.version>
        .... etc ..,
    </properties>

That way you can simply reference the Openfire properties for version x, and be confident all the correct versions are used.

GregDThomas:
(actually, having written that I realise the canonical way to do this is of course with a Maven BOM - e.g. http://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#overview-maven-bom

guusdk:
I'm not sure if it's worth the trouble. They only relevant property remaining is the jetty one, which is used for compilation of the JSPs. There's no real harm in doing that with a compiler provided by the Jetty project from a version that's different from the Jetty instance that's running the admin console. Even if the compiler version number changes, the compiled bytecode (from the JSPs) should be interchangeable. Write once, run anywhere, right? ;)

guusdk:
I like the BOM idea in principle, but I don't think we need it just yet. I've removed the jetty property, by in-lining it with the JSPC plugin. How does this look to you?

Another thing that I noticed is that the JSPC plugin very specifically targets a particular java package. I wonder why @sco0ter specified it like this. I'd guess that simply compiling any JSP would be fine?

GregDThomas:
The package thing makes sense to me; at the simplest level as each JSP page (e.g. `settings.jsp`) is compiled to a unique class (`org.jivesoftware.openfire.plugin.my-plugin-name.settings_jsp.java`) it means it's easy to identify the file at fault in a stack trace - better than leaving it to the default.

guusdk:
Yup yup - I mistakenly thought it was put in to select/limit the source JSPs, which is silly, as they don't have a package definition. My bad.

PR #776, opened 2017-04-19
OF-546: Exclude Openfire-provided libs from plugins

guusdk:
When building a plugin, libraries that are part of Openfire (and are thus provided)
should not be included in the plugin.

### DIFF
--- a/plugins/openfire-plugin-assembly-descriptor/src/main/resources/assemblies/openfire-plugin-assembly.xml
+++ b/plugins/openfire-plugin-assembly-descriptor/src/main/resources/assemblies/openfire-plugin-assembly.xml
@@ -63,6 +63,52 @@
         <dependencySet>
             <outputDirectory>/lib</outputDirectory>
             <useProjectArtifact>false</useProjectArtifact>
+            <useTransitiveFiltering>true</useTransitiveFiltering>
+            <excludes>
+                <!-- Exclude all dependencies that are provided by Openfire! -->
+                <exclude>org.igniterealtime:tinder</exclude>
+                <exclude>dom4j:dom4j</exclude>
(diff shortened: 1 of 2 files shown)


guusdk:
@GregDThomas @sco0ter is this right? Should we always exclude them (and if so, shouldn't we also exclude them from the compilation scope? Is there a scenario in which we'd *not* want to exclude them? Forcing a different version, something like that?

GregDThomas:
Is it necessary to explicitly exclude them now? As xmppserver is a provided dependency I would expect it and it's transitive dependencies to be automaticaly excluded from the bundle. e.g. when developing a traditional web app, tomcat (or jetty) is typically a provided dependency and hence not included in the built WAR file.

guusdk:
Hmm, it was not working for me, but only now I see this warning in my build output:

> The POM for org.igniterealtime.openfire:xmppserver:jar:4.2.0-SNAPSHOT is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details

Back to the drawing board...

by-the-by: feel free to pop in the open_chat group chat - that makes for easier discussions. :)

GregDThomas:
Not from behind the corporate firewall where I am :(

GregDThomas:
Ah yes, Fixing the POM for org.igniterealtime.openfire:xmppserver:jar:4.2.0-SNAPSHOT is probably the first thing; once that's done the transitives should automatically be dealt with, so no need to specify them in the plugins. I /may/ get a chance to look today ...

guusdk:
I yesterday fixed that POM (here: ef80834641a0dadef80c97533da225d3d798747f )

guusdk:
I'm closing this, as I'm doubtful if the explicit solution is needed, or even desired. Most of the problems that I set out to fix were caused by something else.

PR #778, opened 2017-04-19
OF-546: Plugin artifact rework

guusdk:
Previously, the openfire-plugin-assembly artifact replaced the default artifact. When using plugin project as a Maven dependency, this causes oddities: the dependency still is a JAR file, but does not conform to a standard archive. This results in missing classes. Instead, this commit retains the original project artifact. This is expected to be more useful.

Sadly, we appear to can pick only one of these options:

- Attach the openfire-plugin-assembly artifact (making it referable)
- Define a finalName for this artifact (\<artifactname\>.plugin)

Both options appear to be mutually exclusive.

More annoyance is prevented by giving the plugin jar file a 'proper' name, I suspect, which is why I opted for doing that. For the other option, set 'appendAssemblyId' to true, and remove the 'attach' element of the 'make-assembly' block.

The removed openfire-plugin-jar assembly does not appear to have a practical purpose: it builds a jar file that is pretty similar to the standard artifact (it misses some metadata, and its final name is different). Removing this assembly reduces complexity, at the cost of a benefit that I cannot see.


guusdk:
@sco0ter , @GregDThomas - thoughts?

PR #779, opened 2017-04-20
OF-1317: Upgrade dom4j from 1.6.1 to 2.0.0.

guusdk:
dom4j is what's powering most of our XML representation. The library that we're using now (1.6.1) is quite old. I'd love to see it updated to the latest 2.0.0 release, for two reasons:

- 1.6.1 pulls in an old, incompatible Xerces parser, which I want to get rid of.
- 2.0.0 makes use of Java generics.

This commit only replaces the binaries - it does not apply the generics throughout the source code.


sco0ter:
I am skeptical, quote from the official Dom4j site:

> Warning: dom4j 2.0 is in pre-alpha stage. It is likely it can't be compiled. In case it can be compiled > at random it is likely it can't run. In case it runs occasionally it can explode suddenly. If you want to > use dom4j, you want version 1.6.1. Really.

guusdk:
Although caution is wise, that notice seems to be outdated. A proper, non-alpha, 2.0.0 release was made last year: https://github.com/dom4j/dom4j/commit/a4d39926ff08656e4cf86c37f3246029c4e9122b

sco0ter:
Never mind, it seems [the site](http://dom4j.sourceforge.net/download.html) is outdated.

PR #786, opened 2017-04-26
OF-1322 Do not do EXTERNAL without trusted cert

dwd:


src/java/org/jivesoftware/openfire/net/SASLAuthentication.java
```diff
@@ -192,6 +192,24 @@ public static Element getSASLMechanismsElement( ClientSession session )
...
+                boolean trustedCert = false;
+                if (session.isSecure()) {
+                    final LocalClientSession localClientSession = (LocalClientSession)session;
+                    if (localClientSession != null) {
+                        final Connection connection = localClientSession.getConnection();
+                        final KeyStore keyStore = connection.getConfiguration().getIdentityStore().getStore();
+                        final KeyStore trustStore = connection.getConfiguration().getTrustStore().getStore();
+                        final X509Certificate trusted = CertificateManager.getEndEntityCertificate(connection.getPeerCertificates(), keyStore, trustStore);
+                        if (trusted != null) {
+                            trustedCert = true;
```
guusdk: Why not 'continue' here? Perhaps some debug logging - but otherwise, looks good to me.

dwd: Because continue just there isn't enough, and I decided I'd rather have a single "continue". Nothing more than a stylistic choice really.

guusdk: I think it is enough, there? regardless, style, indeed.

guusdk:
Note, this introduces a null reference when there are no client certs. Fixed here: #789

PR #790, opened 2017-05-03
Allow Components to subscribe

GregDThomas:
This PR allows Openfire components to subscribe to XMPP nodes. Given a component has to be installed on the server, it must already be trusted so there should be no security issues with this.


GregDThomas:
Wrong files added

PR #791, opened 2017-05-03
Allow components to subscribe to nodes

GregDThomas:
This PR allows a component to subscribe to XMPP pubsub nodes. Given a component is part of the Openfire deployment, it must already be trusted so there are no security issues.

### DIFF
--- a/src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
+++ b/src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
@@ -543,7 +543,7 @@ public class PubSubEngine {
             return;
         }
         // Check if the subscriber is an anonymous user
-        if (!UserManager.getInstance().isRegisteredUser(subscriberJID)) {
+        if (!UserManager.getInstance().isRegisteredUser(subscriberJID) && !isComponent(subscriberJID)) {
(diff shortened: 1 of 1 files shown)


src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
```diff
@@ -543,7 +543,7 @@ private void subscribeNode(PubSubService service, IQ iq, Element childElement, E
             return;
         }
         // Check if the subscriber is an anonymous user
-        if (!UserManager.getInstance().isRegisteredUser(subscriberJID)) {
+        if (!UserManager.getInstance().isRegisteredUser(subscriberJID) && !isComponent(subscriberJID)) {
```
dwd: Assuming Java does shortcuts, you'll want this as the first condition:

if (!isComponent(subscriberJID) && ...)

isRegisteredUser() will do a disco#info query (rather surprisingly) in order to find out whether a remote user is anonymous or not. Not sure this is right, but it's what's done.

GregDThomas: Thanks; Java does to shortcuts with `&&` (not with `&`) so I've re-ordered the check

PR #792, opened 2017-05-03
WIP: Core labelling

dwd:
This provides some basic concepts of labelling within the core of Openfire, such that a formal access control facility based on labelling (also known as a Policy Enforcement Point) can be written as a plugin. These are known as ACDFs, or Access Control Decision Functions.

A plugin built to this written around Spiffing is at surevine/openfire-spiffing.

Of note:

* Clearances are opaque strings. These are not interpreted in any way by Openfire, and simply passed onto the ACDF.
* Labels are XEP-0258 style blobs, where there is an outer format marker and human readable display marking, and an inner machine-readable label to be interpreted by the ACDF engine.

This WIP branch demonstrates basic message filtering and XEP-0314-like pubsub item filtering.

Missing from this is MAM filtering, MUC room (and similar) filtering, and so on.

### DIFF
--- a/src/java/org/jivesoftware/openfire/XMPPServer.java
+++ b/src/java/org/jivesoftware/openfire/XMPPServer.java
@@ -35,8 +35,12 @@ import org.jivesoftware.openfire.filetransfer.DefaultFileTransferManager;
 import org.jivesoftware.openfire.filetransfer.FileTransferManager;
 import org.jivesoftware.openfire.filetransfer.proxy.FileTransferProxy;
 import org.jivesoftware.openfire.handler.*;
+import org.jivesoftware.openfire.interceptor.InterceptorManager;
+import org.jivesoftware.openfire.interceptor.PacketInterceptor;
 import org.jivesoftware.openfire.keystore.CertificateStoreManager;
 import org.jivesoftware.openfire.keystore.IdentityStore;
+import org.jivesoftware.openfire.labelling.AccessControlDecisionFunction;
+import org.jivesoftware.openfire.labelling.MessageInterceptor;
 import org.jivesoftware.openfire.lockout.LockOutManager;
 import org.jivesoftware.openfire.mediaproxy.MediaProxyService;
 import org.jivesoftware.openfire.muc.MultiUserChatManager;
@@ -141,6 +145,13 @@ public class XMPPServer {
     private boolean shuttingDown;
     private XMPPServerInfoImpl xmppServerInfo;
 
+    /*
+     * Provider for fine-grained access control, based on:
+     * "clearances" (user authorization data) and
+     * "labels" (per-item, user-supplied, access control data).
+     */
+    private AccessControlDecisionFunction accessControlDecisionFunction;
+
     /**
      * Returns a singleton instance of XMPPServer.
      *
@@ -172,6 +183,35 @@ public class XMPPServer {
             throw new IllegalStateException("Not initialized yet");
         }
         return xmppServerInfo;
+    }
+
+    /**
+     * Returns an ACDF object, which can perform decisions.
+     *
+     * @return AccessControlDecisionFunction implementation
+     */
+    public AccessControlDecisionFunction getAccessControlDecisionFunction() {
+        return accessControlDecisionFunction;
+    }
+
+    /**
+     * Sets the AccessControlDecisionFunction implementation.
+     *
+     * @param acdf AccessControlDecisionFunction object.
+     */
+    public void setAccessControlDecisionFunction(AccessControlDecisionFunction acdf) {
+        this.accessControlDecisionFunction = acdf;
+        if (acdf == null) {
+            getIQDiscoInfoHandler().removeServerFeature("urn:xmpp:sec-label:0");
+            for (PacketInterceptor i : InterceptorManager.getInstance().getInterceptors()) {
+                if (i instanceof MessageInterceptor) {
+                    InterceptorManager.getInstance().removeInterceptor(i);
+                }
+            }
+        } else {
+            InterceptorManager.getInstance().addInterceptor(new MessageInterceptor());
+            getIQDiscoInfoHandler().addServerFeature("urn:xmpp:sec-label:0");
+        }
     }
 
     /**
--- a/src/java/org/jivesoftware/openfire/labelling/AbstractACDF.java
+++ b/src/java/org/jivesoftware/openfire/labelling/AbstractACDF.java
@@ -1,0 +1,63 @@
+package org.jivesoftware.openfire.labelling;
+
+import org.jivesoftware.openfire.XMPPServer;
+import org.jivesoftware.openfire.user.User;
+import org.jivesoftware.openfire.user.UserManager;
+import org.jivesoftware.openfire.user.UserNotFoundException;
+import org.jivesoftware.util.JiveGlobals;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xmpp.packet.JID;
+
+/**
+ * Created by dwd on 20/03/17.
+ */
+abstract public class AbstractACDF implements AccessControlDecisionFunction {
+    private static final Logger Log = LoggerFactory.getLogger(MessageInterceptor.class);
+    private String defaultUserClearance;
+    private String defaultPeerClearance;
+    private String serverClearance;
+
+    private final String PROP_PEER_PREFIX = "clearance.domain."; // + domain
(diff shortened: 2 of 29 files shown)


src/java/org/jivesoftware/openfire/XMPPServer.java
```diff
@@ -141,6 +145,13 @@
     private boolean shuttingDown;
     private XMPPServerInfoImpl xmppServerInfo;
 
+    /*
+     * Provider for fine-grained access control, based on:
+     * "clearances" (user authorization data) and
+     * "labels" (per-item, user-supplied, access control data).
+     */
+    private AccessControlDecisionFunction accessControlDecisionFunction;
+
```
guusdk: Could this go in a Module, instead of straight into XmppServer? It'd be good to separate concerns, where possible.

src/java/org/jivesoftware/openfire/XMPPServer.java
```diff
@@ -175,6 +186,35 @@ public XMPPServerInfo getServerInfo() {
...
+     * Sets the AccessControlDecisionFunction implementation.
+     *
+     * @param acdf AccessControlDecisionFunction object.
+     */
+    public void setAccessControlDecisionFunction(AccessControlDecisionFunction acdf) {
+        this.accessControlDecisionFunction = acdf;
+        if (acdf == null) {
+            getIQDiscoInfoHandler().removeServerFeature("urn:xmpp:sec-label:0");
+            for (PacketInterceptor i : InterceptorManager.getInstance().getInterceptors()) {
+                if (i instanceof MessageInterceptor) {
```
guusdk: This would also unload third party / plugin implementations, I think. Is that intended?

src/java/org/jivesoftware/openfire/labelling/AbstractACDF.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private final String PROP_PEER_PREFIX = "clearance.domain."; // + domain
+    private final String PROP_CLEARANCE_USER_DEFAULT = "clearance.user.default";
+    private final String PROP_CLEARANCE_PEER_DEFAULT = "clearance.peer.default";
+    private final String PROP_CLEARANCE_SERVER = "clearance.server";
+
+    protected AbstractACDF() {
+        this.defaultUserClearance = JiveGlobals.getProperty(PROP_CLEARANCE_USER_DEFAULT);
+        this.defaultPeerClearance = JiveGlobals.getProperty(PROP_CLEARANCE_PEER_DEFAULT);
+        this.serverClearance = JiveGlobals.getProperty(PROP_CLEARANCE_SERVER);
```
guusdk: No hard-coded defaults? Without properties, this results in null values. Do null imply some kind of default?

src/java/org/jivesoftware/openfire/labelling/AbstractACDF.java
```diff
@@ -0,0 +1,63 @@
...
+        if (XMPPServer.getInstance().isLocal(entity)) {
+            User user;
+            try {
+                user = UserManager.getInstance().getUser(entity.getNode());
+            } catch(UserNotFoundException e) {
+                return defaultUserClearance;
+            }
+            String clr = user.getClearance();
+            if (clr == null) {
+                Log.debug("Using default user clearance");
```
guusdk: When debug logging, it is often useful to provide the context of the log: in this instance, the argument passed in the argument.

src/java/org/jivesoftware/openfire/labelling/Clearance.java
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Created by dwd on 16/03/17.
+ */
+public class Clearance {
+    Element element;
+
+    Clearance(Element element) {
+        this.element = element;
+    }
+}
```
guusdk: This is an odd wrapper class. Shouldn't the class state be more standard Java types (String, int, boolean)? Perhaps also add `hashCode` and `equals` implementations?

src/java/org/jivesoftware/openfire/labelling/MessageInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+        AccessControlDecisionFunction acdf = XMPPServer.getInstance().getAccessControlDecisionFunction();
+        if (acdf == null) {
+            return;
+        }
+        Log.debug("Packet intercept: " + packet.toString());
+        Message msg = (Message)packet;
+        try {
+            boolean rewritten = false;
+            boolean need_rewrite = false;
```
guusdk: Underscore characters in variable names? How very not Java of you!

src/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
```diff
@@ -613,6 +617,16 @@ public boolean canSendPublicationEvent(LeafNode leafNode, PublishedItem publishe
         if (!canSendEvents()) {
             return false;
         }
+        AccessControlDecisionFunction acdf = XMPPServer.getInstance().getAccessControlDecisionFunction();
+        if (acdf != null) {
+            try {
+                acdf.check(acdf.getClearance(this.jid), publishedItem.getSecurityLabel(), null);
+            } catch (SecurityLabelException e) {
+                Log.debug("Dropped item publish due to SecurityLabel: ", e);
```
guusdk: Also log a reference to the item that's dropped.

src/java/org/jivesoftware/openfire/pubsub/PublishedItem.java
```diff
@@ -183,24 +192,24 @@ public Date getCreationDate() {
...
-		    			 log.error("Failed to parse payload XML", ex);
-		    		} finally {
-		    			if (xmlReader != null) {
-		    				xmlReaders.add(xmlReader);
-		    			}
-		    		}
-				}
-			}
-    	}
+        if (payload == null && payloadXML != null) {
```
guusdk: Is this just a whitespace change?

guusdk:
I missed any Admin Console integration, I think. A simple switch on/switch off perhaps?

src/java/org/jivesoftware/openfire/XMPPServer.java
dwd: That's entirely possible - I'll look into how to construct that.

src/java/org/jivesoftware/openfire/XMPPServer.java
dwd: I don't think so - it's explicitly only removing the org.jivesoftware.openfire.labelling.MessageInterceptor class objects, isn't it?

src/java/org/jivesoftware/openfire/labelling/AbstractACDF.java
dwd: null would imply a default. An empty string would (or might) mean something to the ACDF.

src/java/org/jivesoftware/openfire/labelling/AbstractACDF.java
dwd: Good call.

src/java/org/jivesoftware/openfire/labelling/Clearance.java
dwd: I think this isn't even used, actually. I originally intended making a wrapper construct much like the SecurityLabel, to contain clearances within a syntax-marked XML wrapper. I'll ditch this.

src/java/org/jivesoftware/openfire/labelling/MessageInterceptor.java
dwd: It's possible I was thinking in C++ at this point. :-)

src/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
dwd: Again, good catch.

src/java/org/jivesoftware/openfire/pubsub/PublishedItem.java
dwd: Seems like it, sorry.

dwd:
So the entire framework has no effect unless an AccessControlDecisionFunction is loaded - so that's the on/off switch. It could really use some clearance loading stuff, but the problem there is that it's just opaque strings at this level.

akrherz:
@dwd Is this still targeted for Openfire 4.2 release?

akrherz:
Closing this as I think it is diverged greatly from master and can always reopen if I am wrong :)

PR #793, opened 2017-05-03
Provide support for using CA's SiteMinder for authenticating users

GregDThomas:
CA's SiteMinder product  (aka CA Single Sign-On) provides a secure single sign-on mechanism to web apps and more. This PR provides support for Site Minder for the Admin console, removing the need to manually login once a SiteMinder token has been obtained.

Note 1; To enabled SiteMinder support, set the property `siteminder.auth.enabled` to `true`
Note 2; Do not enable SiteMinder support unless SiteMinder is installed. SiteMinder works as a reverse proxy, and injects the "SM_USER" http header to requests so the application can authorise the user. Unless SiteMinder is installed, an arbitrary client can set this header to any value.

### DIFF
--- a/src/java/org/jivesoftware/admin/AuthCheckFilter.java
+++ b/src/java/org/jivesoftware/admin/AuthCheckFilter.java
@@ -33,6 +33,8 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.jivesoftware.openfire.admin.AdminManager;
+import org.jivesoftware.openfire.auth.AuthToken;
 import org.jivesoftware.util.JiveGlobals;
 import org.jivesoftware.util.WebManager;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


dwd:
Sorry it's taken so long to get to this.

My concern with this would be that there is no obvious way we can test the work, nor easily evaluate it for security. Overall I'm leaning toward rejecting it - not because of anything to do with the code, but due to logistics.

GregDThomas:
> Sorry it's taken so long to get to this.

No problem; apologies for firing off a whole bunch of PRs over the last few months with little discussion about it - the forums don't always seem like the right place.

> My concern with this would be that there is no obvious way we can test the work, nor easily evaluate it for security. 

Understood - it does leave a bit of a bad taste in my mouth, too. FWIW, the way we tested it was using a Chrome plugin that allowed you to set an arbitrary header (https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj) and use it to set the header that the SM reverse proxy would otherwise do. 

For other similar uses, we've managed to implement interfaces (e.g. org.jivesoftware.openfire.auth.AuthProvider) but that doesn't work for the admin UI. As an alternative to this PR, do you think a PR which modifies AuthCheckFilter to call an optional interface, ServletRequestAuthenticator that can authenticate based on an an HTTP request would be more suitable? We can then implement our own ServletRequestAuthenticator that does the SiteMinder authentication without "polluting" the core Openfire code.

dwd:
That sounds like a fine idea.

FWIW, if you do want more discussion, the chatroom is the best place to engage with other developers.

GregDThomas:
Is that open_chat@conference.igniterealtime.org ?

akrherz:
@GregDThomas yes, or a [web interface](https://www.igniterealtime.org/support/group_chat.jsp)

GregDThomas:
Closing this, in preference to https://github.com/igniterealtime/Openfire/pull/879 as it's sufficiently different.

PR #796, opened 2017-05-05
OF-1325 -- Implement separate History settings in Client Control

BittyByte:
Currently Client Control plugin's setting "History Settings" not just hides the history settings, but also disables the history in Spark. Which might not be desired. One might need to have history, but hide settings, so users won't be able to disable it from a GUI. This change will implement two separate history features in Client Control - "History Settings" and "History Transcripts". Now an admin can select to enable/disable the history controls as well as enable/disable the logging of the history into transcript files. **NOTE: This is updated version 2.1.2 of the Client Control plug-in and requires a minimum of Spark version 2.9.0 for proper functionality.**


wrooot:
@akrherz @guusdk can somebody of you review and merge this? Spark's part has been already merged to work with this changes.

guusdk:
This change replaces one property (`history.enabled`) with two others (`historysettings.enabled` and `historytranscripts.enabled`). Perhaps, both new properties should default to the value of the old property, if it was set?

wrooot:
Yes, this change might affect current users once they update the plugin. If it is possible and easy, we can probably set both these settings to Disabled, if the old one was Disabled. Although, based on a number of threads in the forums complaining about missing menus once they update Spark i see that rarely somebody updates the plugins in Openfire (it is easy not to every go into Plugins menu and not see any updates). So the effect could be minimal and easy to fix, just disable the setting.

Also, the setting was actually called History Settings (history.enabled), but it was disabling the history itself and it was a confusing and not intended behavior. We are actually fixing the behavior here, so maybe it should reset actually? And then admins should choose whether they want to disable just settings or the history completely. Everything is explained in the changelog.

BittyByte:
Actually, **client.features.history** was renamed to **client.features.historysettings** and **client.features.history.description** was renamed to **client.features.historysettings.description**. Their original default values were kept. Then two new properties were added - **client.features.historytranscripts** and **client.features.historytranscripts.description**. I hope this is clear now.

wrooot:
I think that Guus was talking about the scenario where admin has old setting in Disabled mode. Once he updates the plugin, the old setting is gone and the clients will show the history and the history settings again. Unless admin goes to the settings menu and disables one of the two new settings. There can be a few annoyed admins who don't read changelogs and update on the whim. But i'm not sure if we can somehow read what the value was for the old setting and then if it was Disabled put both new settings to Disabled mode.

BittyByte:
I understand now. Well, as best practice, I would hope an admin would read the changelog when installing the new plugin and then set the features like he usually has them set in prior versions. If you want, I can add a note in the changelog, but I can't force any admin to read it.

wrooot:
So, Michael tried to find a way to set new settings with the value of an old setting and he couldn't find one. I say we go with the current version of this patch (maybe adding a warning into changelog, just not in red color, this will look bad :) ). It will affect a small number of installations, i believe. And some admins might be even glad about it as maybe they only wanted to disable the settings and not the history, but didn't had choice. Now, once it stops working, they will find two separate settings, which may be more convenient.

BittyByte:
Hey guys, I figured it out and I have added code to resolve this issue. It will now check for the existence of the previous "History Settings" and carry over those values to the new "History Settings" and "History Transcripts" features. I will resubmit this PR so Guus can review it and give a final OK. Thanks!

guusdk:
Looks good to me! Thanks for taking the time to look into these details!

BittyByte:
No problem. Thanks for bringing it to my attention and I'm glad to have found a solution.

PR #797, opened 2017-05-08
Fix OF-1159 - add database column to identify encryption status

GregDThomas:
OF-1159 identifies the problem that changes to the file `conf/security.xml` are not reflected across all nodes in the cluster when a property is encrypted. This patch adds a `encrypted` column to the database, and uses that instead of a local file to identify is the property is encrypted.

Note; if a property is already encrypted by the presence of the indicator in `security.xml` then the database column will not be updated until that property is updated/deleted. It is at this time that the encryption indicator in `security.xml` is removed, but that will only happen on nodes that are part of the cluster at the time of the change.

### DIFF
--- a/src/database/openfire_db2.sql
+++ b/src/database/openfire_db2.sql
@@ -123,6 +123,7 @@ CREATE TABLE ofID (
 CREATE TABLE ofProperty (
   name        VARCHAR(100) NOT NULL,
   propValue   VARCHAR(3000) NOT NULL,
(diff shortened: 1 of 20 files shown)


dwd:
Sorry this took so long. Been trying to find the time to really understand the code.

GregDThomas:
No problem; it wasn't the cleanest of patches - but I'll blame the original code, it can't be my fault ;)

PR #799, opened 2017-05-08
OF-1277 -- Change setting name to Invisible Login and Status

BittyByte:
After adding additionally blocking Invisible presence status, Client Control plugins needs to be updated. Changing the name of the setting to "Invisible Login and Status".


wrooot:
This PR only renames one setting and updates the changelog (actual changes were done on the Spark's side). So i'm going to merge it without waiting for another review.

PR #804, opened 2017-05-15
OF-1331 Warn user during setup when a session failure occurs

akrherz:
This is a complex problem.  During setup, fetching xmppSettings from the session may fail due to an invalid session state maintained between the client browser and the server.  This is likely caused by a previous HTTPS session to the same server hostname setting a `JSESSIONID` that subsequently can not be overwritten by the current HTTP openfire-setup session.

This introduces i18n string `setup.invalid_session`

### DIFF
--- a/src/i18n/openfire_i18n_en.properties
+++ b/src/i18n/openfire_i18n_en.properties
@@ -1844,6 +1844,7 @@ component.session.details.info=Below are details about the session with the exte
 # General Setup
 
 setup.title=Setup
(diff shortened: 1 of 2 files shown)


akrherz:
An alternative here may be to set a different session cookie variable name than `JSESSIONID` and perhaps have it be custom for both HTTP and HTTPS connections.  I could only figure this out globally by modifying `src/web/WEB-INF/web.xml`
```
<session-config>
    <cookie-config>
        <name>of_jsessionid</name>
    </cookie-config>
</session-config>
```
So I'll have to punt on any more elegant solution

guusdk:
I've had a look at setting different cookie names, but that's not a trivial change either, from what I can tell. For now, let's at the very least warn the users, as you implemented. Thanks!

PR #807, opened 2017-05-19
OF-1334 Don't do CSRF check unless actions are requested

akrherz:
The CSRF check was armed by default, which does not make sense when first visiting the page and no actions are requested.  This would result in an error shown that the 'Archive Index rebuild failed".  I also removed an unused css import


akrherz:
@guusdk I agree about your style suggestion and have modified the PR

PR #808, opened 2017-05-24
Maven update

guusdk:


sco0ter:
My initial intent was to actually move the plugins' source code to plugins/xyz/src/main/java evntually, in order to have a good Maven module structure and to follow the default Maven layout.

You know, I'd like to have it `module/src/main/java`, but as is, it's rather src/java (core) and src/plugins/pluginname/src, which is pretty messy.

I'd rather see the source code move to the default Maven directory layout instead of moving the poms back a poor structure.

guusdk:
I (nowadays) agree that we should move to a standard Maven directory structure - but we should not do that _now_. We will, for at least one release, support both the Maven as well as Ant-based build. This makes it impractical to move the source folders.

Additionally, some have voiced that the plugins should move to their own repository - taken out of the Openfire repository completely. I don't necessarily have an opinion on that, but it illustrates that the structure that we will end up with is still subject to debate.

Having an in-between solution, separating the Maven metadata from the source, has proven to be confusing (at least to me), while it does not add much apart from a glimpse of what the future structure could look like. That is why this PR removes that structure. This in no means is intended to be the end result though.

PR #812, opened 2017-05-31
OF-1336: Introduce UserPropertyProvider

guusdk:
This commit adds a new feature to Openfire, that allows one to store user properties in an external system. This is quite similar to concepts like UserProvider, AuthProvider, GroupProvider, and so on.

The default behavior, moved from the User class to DefaultUserPropertyProvider, continues to use the ofUserProp database table. Provider implementations matching those of other types of providers have been added.

### DIFF
--- a/src/java/org/jivesoftware/openfire/user/User.java
+++ b/src/java/org/jivesoftware/openfire/user/User.java
@@ -437,7 +437,7 @@ public class User implements Cacheable, Externalizable, Result {
             String keyString = key;
 
             try {
-                synchronized (getName() + keyString.intern()) {
+                synchronized ((getName() + keyString).intern()) {
                     if (properties.containsKey(keyString)) {
                         String originalValue = properties.get(keyString);
                         answer = properties.put(keyString, value);
--- a/src/java/org/jivesoftware/openfire/user/User.java
+++ b/src/java/org/jivesoftware/openfire/user/User.java
@@ -16,24 +16,6 @@
 
 package org.jivesoftware.openfire.user;
 
-import java.io.Externalizable;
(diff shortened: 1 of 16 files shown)


dwd:
"DefaultUserPRoeprtyProvider" looks a bit Dutch. ;-)

guusdk:
My favorite is `UserTypeBasedUserPropertyProviderMapper`

src/java/org/jivesoftware/openfire/user/User.java
```diff
@@ -472,28 +436,34 @@ public String put(String key, String value) {
...
-                    eventParams.put("originalValue", originalValue);
-                }
-                else {
-                    answer = properties.put(keyString, value);
-                    insertProperty(keyString, value);
-                    // Configure event.
-                    eventParams.put("type", "propertyAdded");
-                    eventParams.put("propertyKey", key);
+            try {
+                synchronized (getName() + keyString.intern()) {
```
dwd: Is this right? I think this would result in a non-interned String, which would be a temporary object here, and so not syncrhonize properly. Appreciate this is existing code, but might be nice to fix if needed.

guusdk: Ah, good catch.

guusdk:
I agree on the LDAP side of things, but that's to complicated for me to want to spend time on that now. Note that many LDAP-based properties can already be exposed via the VCard. Not ideal, but at least, somewhat accessible.

dwd:
Hence "for the future". I think it's a logical, later, addition - absolutely not something essential for this PR.

PR #816, opened 2017-06-09
OF-1339: Merge websocket plugin with Openfire core.

guusdk:


guusdk:
I've looked at porting some of the documentation, but a lack of a proper place to put it, as well as most of the text describing what I think are settings that relate to SM, not websockets, I thought it was OK to give up my attempt.

PR #820, opened 2017-06-12
OF-1343 update docs about new windows service behavior

wrooot:
Should be merged when 4.1.5 is released.


guusdk:
... why not merge now?

wrooot:
I guess there is not much difference. But could look weird for someone reading it online before 4.1.5 is released :) although we are probably very close to release.

PR #821, opened 2017-06-12
OF-1345 OF-1139: 'Mapped' providers should be documented.

guusdk:
Usage of Mapped providers requires a lot of configuration, that should be documented. The documentation provided in this commit describes a common use case, in which admins are to be retrieved from the default providers, while regular users are obtained from an external system.

### DIFF
--- a/documentation/docs/index.html
+++ b/documentation/docs/index.html
@@ -51,6 +51,10 @@ messaging (IM) services using the XMPP protocol.
         A guide to integrating Openfire authentication, user, and group data with a custom database.
     </li>
     <li>
+        <a href="separating-admin-users-guide.html">Separating Administative Users Guide</a> -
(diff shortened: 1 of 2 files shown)


guusdk:
Note that there's a bug in the 'mapper' implementations, that's fixed by OF-1327 / https://github.com/igniterealtime/Openfire/pull/800

guusdk:
http://htmlpreview.github.io will let you look at rendered HTML from commits. For example: http://htmlpreview.github.io/?https://raw.githubusercontent.com/guusdk/Openfire/86d92a64325c0ecece673c419fbc4e0dcd44ce7c/documentation/docs/separating-admin-users-guide.html

guusdk:
I'm merging this, as it's a documentation-only change, which Sean tried out, and was able to make work.

PR #822, opened 2017-06-12
OF-1347: Improve Admin Console Group Edit page

guusdk:
This commit adds some UX improvements to the 'edit group' page on the admin console.

### DIFF
--- a/src/i18n/openfire_i18n_en.properties
+++ b/src/i18n/openfire_i18n_en.properties
@@ -980,16 +980,20 @@ group.delete.delete=Delete Group
 group.edit.title=Edit Group
 group.edit.form_info=Edit group settings and add or remove group members and administrators using the \
         forms below.
+group.edit.error=Error updating group settings. Please check your error logs.
 group.edit.details_info=Edit group settings below.
 group.edit.update=Group information updated successfully.
 group.edit.update_add_user=User(s) added successfully.
 group.edit.update_del_user=User(s) deleted successfully.
 group.edit.update_user=User(s) updated successfully.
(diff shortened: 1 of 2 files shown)


akrherz:
Thanks @guusdk !  I pulled this in, built it successfully and played with the interface

 - The concept of Group admins remains, is this even used anywhere?
 - I like the disabled UI for shared group

I tried various things and didn't find any bugs.

guusdk:
Thanks for checking! I honestly don't know if the purpose of admins serves any purpose. I assumed so, and retained the functionality (there's quite a bit of code involved, actually). If they indeed serve no purpose, we might want to remove them - but lets do that in a dedicated issue/PR.

deleolajide:
Please don't pull out the group admins feature. We use it in telecoms. Groups are used for user roles and entitlement provisioning. Very effective  when Openfire is connected to Active Directory.

guusdk:
Good to see someone is using it after all. :) The current code in Openfire does not make use of this feature at all, I think. Is there any generic bit from your application that you can/want to push upstream, @deleolajide ?

deleolajide:
We are not doing anything generic that can be pushed upstream, we just get the group admin users of a group like this

````
			JSONArray adminsJSON = new JSONArray();
			int i = 0;

			for (JID memberJID : group.getAdmins())
			{
				JSONObject adminJSON = new JSONObject();
				adminJSON.put("jid", memberJID.toString());
				adminJSON.put("name", memberJID.getNode());
				adminsJSON.put(i++, adminJSON);
			}
			groupJSON.put("admins", adminsJSON);
````

PR #824, opened 2017-06-19
OF-1309 Authenticate based on domain pairs

dwd:
Openfire considers that if a session has authenticated as A, and has
authenticated to B, then the session is already authorized to send stanzas
from A to B. This is not always the case.

This patch stores domain pairs (encapsulated as a DomainPair object) for a
session, and always authenticates if it has previously not requested
authorization for this domain pair.

This should be sufficient to fix OF-1309, but testing is tricky.


dwd:
Haven't done any testing beyond "Does it still build?", but I could use the reviews...

src/java/org/jivesoftware/openfire/session/DomainPair.java
```diff
@@ -0,0 +1,38 @@
...
+public class DomainPair {
+    private final String local;
+    private final String remote;
+
+    public DomainPair(String local, String remote) {
+        this.local = local;
+        this.remote = remote;
+    }
+
+    public int hashCode() {
```
guusdk: Even though it is *probably* alright, the implementation of `hashCode` and `equals` aught to be consistent. By far the easiest way to guarantee this is to let your IDE generate these methods for you. In Intellij: [ALT-INSERT (select "generate equals() and hashCode()")](https://www.jetbrains.com/help/idea/generate-equals-and-hashcode-wizard.html). Generated code also lets you stop worry about null references and the like.

This is of particular importance here, as you're storing instances of `DomainPair` in a HashSet (in `LocalOutgoingServerSession`)

src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
```diff
@@ -698,4 +700,13 @@ public String getAvailableStreamFeatures() {
...
     }
+
+    public void addOutgoingDomainPair(String localDomain, String remoteDomain) {
+        outgoingDomainPairs.add(new DomainPair(localDomain, remoteDomain));
+    }
+
+    @Override
+    public boolean checkOutgoingDomainPair(String localDomain, String remoteDomain) {
+         return outgoingDomainPairs.contains(new DomainPair(localDomain, remoteDomain));
+    }
```
guusdk: Is there any point - or is it even safe - to let `#getAuthenticatedDomains()` exist? Isn't usage of that method inherently unsafe, in context of the issue that we're trying to fix. If so, `#addAuthenticatedDomain` can probably also go.

src/java/org/jivesoftware/openfire/session/OutgoingServerSession.java
```diff
@@ -79,4 +79,6 @@
      * @return True if the subdomain was authenticated by the remote server.
      */
     boolean authenticateSubdomain(String domain, String hostname);
+
```
guusdk: Tsk. All other definitions in this interface definition got nice documentation. :)

src/java/org/jivesoftware/openfire/session/DomainPair.java
dwd: Done.

src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
dwd: Yes, you're right - I've removed both getAuthenticatedDomains and getHostnames and so on, and replaced them purely with DomainPairs throughout.

src/java/org/jivesoftware/openfire/session/OutgoingServerSession.java
dwd: I have written nice documentation now.

dwd:
Tested this, now can't find a behaviour difference for simple cases. I haven't figured out how to consistently trigger OF-1309 yet, but I'm confident it won't affect this design.

PR #832, opened 2017-06-25
Restoring duplicate TLDs (used by Ant build)

guusdk:


guusdk:
This should fix the problem with the Ant build that I introduced in 93f818483a2bd7fe4dd8ab263ecdb8189f99bff0

PR #835, opened 2017-06-30
OF-1362: Websocket plugin has been broken since Openfire 4.1.4. This PR implem…

deleolajide:
…ents the missing code and fixes the issue

    @Override
   public ConnectionConfiguration getConfiguration()
   {
       // TODO Here we run into an issue with the ConnectionConfiguration introduced in Openfire 4:
       //      it is not extensible in the sense that unforeseen connection types can be added.
       //      For now, null is returned, as this object is likely to be unused (its lifecycle is
       //      not managed by a ConnectionListener instance).
       return null;
   }

### DIFF
--- a/src/plugins/websocket/plugin.xml
+++ b/src/plugins/websocket/plugin.xml
@@ -8,8 +8,8 @@
     <name>Openfire WebSocket</name>
     <description>Provides WebSocket support for Openfire.</description>
     <author>Tom Evans</author>
-    <version>1.2.0</version>
-    <date>05/08/2017</date>
+    <version>1.2.1</version>
+    <date>06/30/2017</date>
     <url>https://tools.ietf.org/html/rfc7395</url>
-    <minServerVersion>4.2.0 Alpha</minServerVersion>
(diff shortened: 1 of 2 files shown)


deleolajide:
> The question: this appears to relate to changes introduced in Openfire 4.0.0 (OF-946), but you mentioned that this only started to be a problem with Openfire 4.1.4 - even explicitly mentioned that it was not a problem with Openfire 4.0.2. That does not compute for me

I can confirm that the highest version of OF that runs the websocket plugin for me is 4.1.3.  That is because I always use Openfire meetings plugin or a derived version which adds the OFMEET external SASL mechansim which causes websockets plugin to generate a NPE.

deleolajide:
> Instead of lazy-loading the connection configuration, it might be better to not store the value in a field, but simply re-generate it every time.

Agreed, but I simply replicated the behaviour of the http-bind socket connection. You might want to revise that as well.

https://github.com/igniterealtime/Openfire/blob/a469be3e7a04cedc35551d776cd2a58e5d7a8550/src/java/org/jivesoftware/openfire/http/HttpSession.java

PR #836, opened 2017-07-03
OF-1359: PR to elevate webclients to top level menu in openfire admin UI

deleolajide:
We should aim to put the webclients menu item at the top level to enable each individual web client have a menu level to use for other app features beyond settings


guusdk:
Can't think of a reason why we should not do this. Perhaps ofmeet should be moved there too?

deleolajide:
> Perhaps ofmeet should be moved there too?

I was thinking the same thing.  The UI settings menu

guusdk:
I've merged this PR (thanks, by the way :P) and am following up with another commit that mentions this in the changelog.

PR #840, opened 2017-07-17
Jmxweb 0.0.7

deleolajide:
The jmxweb plugin web app does not work with OF 4.1.5.  This PR fixes the issue.


akrherz:
Hi @deleolajide Your PR mentions "fixed jmxweb for openfire 4.1.5", but the maven changes were never made to 4.1 branch, but 4.2.  Is this a typo or does this fix need to be made to make jmxweb work for 4.1.5 openfire?

deleolajide:
I have used the wrong words to describe my peculiar situation. Let me see if I can break it down:

1. The plugin does not work in OF 4.1.5 because of a 4.2 change that was back ported to 4.1.  This change breaks this and other plugins that have an embedded jetty web service. 
2. I am working with 4.2 source code and had to fix the issue and rebuild the plugin. In the process, I had to fix other issues that were caused by the maven restructuring in 4.2 

In other words, this fix makes the plugin work in 4.1.5+ and the source has to be applied to OF master which is based on 4.2.

deleolajide:
I am planning to add push notification on the plugin and may refactor the code.

PR #843, opened 2017-08-04
OF-1366: Fixing NullPointerException caused by NULLs in map

jgitlin-bt:
Fixing NullPointerException caused by having a mix of valid Groups and NULLs inside the result of getGroupsFromKeys. See: https://community.igniterealtime.org/message/265804


guusdk:
Thanks!

PR #847, opened 2017-08-10
Ensure that Hazelcast backed Cache objects have the correct settings

GregDThomas:
[NB. This is a re-submit of a previous pull request without superfluous commits]

I recently came across an issue whereby if you set a cache size/lifetime using Cache.setMaxCacheSize() / Cache.setMaxLifetime() it has no effect on Hazelcast backed Caches - configuration is taken solely from the hazelcast-cache-config.xml file.

This PR configures the underlying Hazelcast IMap based on the values that were previously set by these methods, defaulting to the values in hazelcast-cache-config.xml if not present. Some discussion on this can be found at https://groups.google.com/forum/#!topic/hazelcast/DYmHGo-sCnI

Note; currently it's not possible to modify these values after an IMap has been created -
the Javadoc for Cache has been updated to reflect this. hazelcast/hazelcast#592 suggests that this ability is coming in Hazelcast 3.9.

### DIFF
--- a/src/java/org/jivesoftware/util/cache/Cache.java
+++ b/src/java/org/jivesoftware/util/cache/Cache.java
@@ -72,6 +72,9 @@ public interface Cache<K,V> extends java.util.Map<K,V> {
      * than the max size, the least frequently used items will be removed. If
      * the max cache size is set to -1, there is no size limit.
      *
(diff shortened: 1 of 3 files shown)


tevans:
This looks good to me also. Thanks for the contribution!

GregDThomas:
I don't know if it's worth raising & closing a JIRA for this, so the changelog can be updated?

PR #850, opened 2017-08-14
OF-1368 Display more-useful admin console msg on user@ login failures

akrherz:
Here's the message displayed with the username contains '@'

> Login failed: You provided what appears to be an email address, consider using just the user part (everything before the @). Typical installations use just 'admin' as the username.

### DIFF
--- a/src/web/login.jsp
+++ b/src/web/login.jsp
@@ -1,6 +1,3 @@
-<%--
---%>
-
 <%@ page import="org.jivesoftware.admin.AdminConsole,
                  org.jivesoftware.openfire.admin.AdminManager"
     errorPage="error.jsp"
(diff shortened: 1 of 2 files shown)


src/web/login.jsp
```diff
@@ -144,8 +141,13 @@
         catch (UnauthorizedException ue) {
             Log.debug(ue);
             LoginLimitManager.getInstance().recordFailedAttempt(username, request.getRemoteAddr());
-            errors.put("unauthorized", LocaleUtils.getLocalizedString("login.failed.unauthorized"));
-        }
+            // Provide a special message if the user provided something containing @
+            if (username.indexOf("@") > 0){
```
guusdk: larger than -1, technically - but people starting a username with '@' have other issues :)

GregDThomas: Perhaps a simple `if (username.contains("@")) {` makes it explicit?

akrherz: updated PR to use `contains`

PR #852, opened 2017-08-16
OF-1208: Disallow (by default) S2S for anonymous users.

guusdk:
A new properyt, 'xmpp.server.allow-anonymous-outbound-data' controls if client sessions that are anonymous can send data to remote domains. By default, they cannot.

### DIFF
--- a/src/java/org/jivesoftware/openfire/session/ConnectionSettings.java
+++ b/src/java/org/jivesoftware/openfire/session/ConnectionSettings.java
@@ -59,6 +59,7 @@ public final class ConnectionSettings {
 
         public static final String PERMISSION_SETTINGS = "xmpp.server.permission";
         public static final String AUTH_PER_CLIENTCERT_POLICY = "xmpp.server.cert.policy";
+        public static final String ALLOW_ANONYMOUS_OUTBOUND_DATA = "xmpp.server.allow-anonymous-outbound-data";
     }
 
     public static final class Multiplex {
--- a/src/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/src/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -463,38 +463,37 @@ public class RoutingTableImpl extends BasicModule implements RoutingTable, Clust
             {
                 Log.info( "The anonymous user '{}' attempted to send data to '{}', which is on a remote domain. Openfire is configured to not allow anonymous users to send data to remote domains.", packet.getFrom(), jid );
                 routed = false;
+                return routed;
             }
         }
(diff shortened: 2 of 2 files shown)


guusdk:
First, untested stab at solving this. I'm not sure if the default error is appropriate. The code will ignore data sent from entities that have cannot be resolved to a client session - so there still might be an open vector there.

src/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
```diff
@@ -454,36 +454,49 @@ private boolean routeToComponent(JID jid, Packet packet,
...
-		            routed = remotePacketRouter.routePacket(nodeID, jid, packet);
-		        }
-		    }
-		}
-		else if (!RemoteServerManager.canAccess(jid.getDomain())) { // Check if the remote domain is in the blacklist
-            Log.info( "Will not route: Remote domain {} is not accessible according to our configuration (typical causes: server federation is disabled, or domain is blacklisted).", jid.getDomain() );
-            routed = false;
+	private boolean routeToRemoteDomain(JID jid, Packet packet, boolean routed)
+    {
+		if ( JiveGlobals.getBooleanProperty( ConnectionSettings.Server.ALLOW_ANONYMOUS_OUTBOUND_DATA, false ) )
```
akrherz: Wouldn't this cause outbound packets from non-anonymous sessions to not be routed in the case of this setting being enabled?

akrherz: Also, the default here is false, I thought from your PR message that it was defaulted to be true?  Sorry, maybe I am not awake yet this fine morning!

guusdk: I think I was the one not being awake. My intend was to default to a scenario where anonymous users cannot initiate s2s. I'll negate the condition in that if-statement...

guusdk:
Please squash when merging. :)

src/java/org/jivesoftware/openfire/session/ConnectionSettings.java
```diff
@@ -59,6 +59,7 @@ private ConnectionSettings() {
 
         public static final String PERMISSION_SETTINGS = "xmpp.server.permission";
         public static final String AUTH_PER_CLIENTCERT_POLICY = "xmpp.server.cert.policy";
+        public static final String ALLOW_ANONYMOUS_OUTBOUND_DATA = "xmpp.server.allow-anonymous-outbound-data";
```
dwd: Not a very obvious property name. But I can't think of a better one.

PR #854, opened 2017-08-17
MucRole code revised

yusufcakal:
Code Revised on Enum

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/MUCRole.java
+++ b/src/java/org/jivesoftware/openfire/muc/MUCRole.java
@@ -211,12 +211,10 @@ public interface MUCRole {
          * @return the associated affiliation.
          */
         public static Role valueOf(int value) {
-            switch (value) {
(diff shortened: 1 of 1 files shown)


guusdk:
Hi! Thanks for your contribution! Why is this change desirable?

yusufcakal:
Hi, It becomes a static code when the field number increases.
I chose such a way not to increase the number of switch cases at a later time.
The changes I have made are independent of the number of fields found in the enum and are more dynamic.

Thanks.

guusdk:
Where you say "field number", you are referencing to the enum values, right?

Those are defined in a specification, and will never change.

Although I'm happy to improve code to resolve issues, I am reluctant to apply changes for the sake of change. I'd like to avoid over-optimization.

yusufcakal:
Yes, I am at the threshold of enum values.
I see that I thought I could make more optimizations.
it does not matter
Have a good day.

PR #860, opened 2017-09-05
OF-1376 Use an event rather than an arbitrary delay before starting clustering

GregDThomas:
This PR switches to using a PluginManagerListener to wait for the plugins to load before attempting to start Hazelcast clustering, instead of an arbitrary delay of 5 seconds. 

We came across this when we were getting inconsistent behaviour on plugin startup; plugin A*** was working just fine, plugin W*** would, sometimes, get confused as clustering was started in the middle of it's plugin initialisation. Possibly less of an issue in OF version 4.2 as plugins will be loaded in parallel, but even so ...

I also added a missing changelog entry for version 2.2.2.

### DIFF
--- a/src/plugins/hazelcast/changelog.html
+++ b/src/plugins/hazelcast/changelog.html
@@ -44,6 +44,16 @@
 Hazelcast Clustering Plugin Changelog
 </h1>
 
+<p><b>2.2.3</b> -- September 5, 2017</p>
+<ul>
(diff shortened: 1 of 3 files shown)


GregDThomas:
Closeing/re-opening to ensure the CI job works OK

guusdk:
Should we have JIRA issues for these changes?

PR #861, opened 2017-09-06
OF-1250: Old DWR causes CSRF, XSS in Admin Console

richmidwinter:


dwd:
I think this is failing to build due to the Ant build not having the jars. Do we care?

guusdk:
Yes, we should care if that's the case. 4.2 should be buildable using both Ant as well as Maven (after 4.2, we can drop the Ant build).

src/plugins/monitoring/src/java/org/jivesoftware/openfire/exceptions/ServiceException.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * The Class MUCServiceException.
+ */
+public class ServiceException extends Exception {
+
+	/** The Constant serialVersionUID. */
+	private static final long serialVersionUID = 4351720088030656859L;
+
+	/** The ressource. */
+	private String ressource;
```
guusdk: ressource <--> resource?

PR #862, opened 2017-09-07

richmidwinter:


src/web/plugin-admin.jsp
```diff
@@ -92,7 +92,7 @@
 
     if (csrf_check && uploadEnabled && uploadPlugin) {
         Boolean installed = false;
-
+
```
guusdk: ooh! aah!

src/web/plugin-admin.jsp
```diff
@@ -106,7 +106,14 @@
             for (Object objItem : items) {
                 FileItem item = (FileItem)objItem;
                 String fileName = item.getName();
+                
                 if (fileName != null) {
+                    // Ensure we're dealing with a jar file.
+                    if (!"application/java-archive".equals(item.getContentType())) {
```
guusdk: Openfire accepts not only `.jar`, but also `.war` files.

The `FileItem` instance is a representation of data in a HTTP request, right? I'm not sure if checking the content type will bring any benefit. A malicious user can easily forge that.

guusdk: Ah, perhaps not so much with the CSRF checks that we now have. Does this make it fully safe? If not, I'm worried that this might introduce more issues than solve any.

src/web/plugin-admin.jsp
```diff
@@ -533,7 +540,7 @@ tr.lowerhalf > td:last-child {
         <h3><fmt:message key="plugin.admin.upload_plugin" /></h3>
         <p><fmt:message key="plugin.admin.upload_plugin.info" /></p>
         <form action="plugin-admin.jsp?uploadplugin&amp;csrf=${csrf}" enctype="multipart/form-data" method="post">
-            <input type="file" name="uploadfile" />
+            <input type="file" name="uploadfile" accept="application/java-archive"/>
```
guusdk: Does this pick up `.jar` as well as `.war` in the relevant browsers? I'm worried that we're excluding the latter.

src/web/plugin-admin.jsp
GregDThomas: Isn't the content-type supplied by the browser? Can you be sure that all (reasonable) browsers will set the content-type to exactly application/java-archive and not application/jar or something else?

PR #864, opened 2017-09-07
OF-1020: Admin Console Remote File Inclusion

richmidwinter:


src/java/org/jivesoftware/openfire/update/UpdateManager.java
```diff
@@ -73,7 +73,7 @@
     /**
      * URL of the servlet (JSP) that provides the "check for update" service.
      */
-    private static String updateServiceURL = "http://www.igniterealtime.org/projects/openfire/versions.jsp";
+    private static String updateServiceURL = "https://www.igniterealtime.org/projects/openfire/versions.jsp";
```
akrherz: The only concern here would be that the JRE supports LetsEncrypt signed certs that igniterealtime uses.

dwd: True. We could (possibly) use a specific truststore for this that explicitly (and maybe only) included the LetsEncrypt CA.

guusdk: Oh, let's not go down the road of maintaining more certs, keeping it in sync with whatever we use on the website.

What exactly are we trying to achieve here? Is doing HTTPS with certificate-checks disabled acceptable?

guusdk:
This looks like a sensible fix. Thanks!

PR #865, opened 2017-09-08

richmidwinter:


guusdk:
Could you rebase to the current master? I've just added a fix there that hopefully prevents the continuous Travis failures.

PR #866, opened 2017-09-08

richmidwinter:


dwd:
Should this delete the .part file if an exception is generated?

richmidwinter:
Yes. I'd tested for that and thought it worked so assumed it was done elsewhere... but it's not. I'll update.

PR #867, opened 2017-09-11
OF-35 - PubSub Admin console

ian-green:
Initial work on OF-35 - Pub sub admin console

Added 'PubSub' tab to admin console (modelled after the Group Chat tab)
Current functionality:
* View and delete Leaf Nodes on the pubsub service (Note Collection nodes and node hierarchy are currently ignored)
* View and delete subscriptions on Leaf nodes
* View and delete items on Leaf nodes

Further functionality to follow.

### DIFF
--- a/build/eclipse/classpath
+++ b/build/eclipse/classpath
@@ -32,7 +32,6 @@
 	<classpathentry kind="src" output="work/plugins-dev/userCreation/target/classes" path="src/plugins/userCreation/src/java"/>
 	<classpathentry kind="src" output="work/plugins-dev/userImportExport/target/classes" path="src/plugins/userImportExport/src/java"/>
 	<classpathentry kind="src" output="work/plugins-dev/userservice/target/classes" path="src/plugins/userservice/src/java"/>
-	<classpathentry kind="src" output="work/plugins-dev/websocket/target/classes" path="src/plugins/websocket/src/java"/>
 	<classpathentry kind="src" output="work/plugins-dev/xmldebugger/target/classes" path="src/plugins/xmldebugger/src/java"/>
 	<classpathentry kind="src" path="src/test/java"/>
 	<classpathentry kind="src" path="src/test/throttletest/src"/>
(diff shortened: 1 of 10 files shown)


guusdk:
I'm looking forward to having this in the admin console!

Could I bother you to do the JSP bit in JSTL? We're (well, me), are trying to move away from scriptlets. There are a couple of custom taglibs that reduce some of the boilerplate HTML as well. You'll see plenty examples by looking at those JSPs that have been edited most recently: 
 https://github.com/igniterealtime/Openfire/blob/master/src/web/group-edit.jsp for example

ian-green:
Seems like a sensible move, wIll do.

guusdk:
Thanks for doing that. This pleases my compulsive scriptlet removal urge. :)

This PR adds a new "Pubsub" tab, on the highest level (next to Server, Users/Groups, Sessions, Group Chat, and Plugins). My concern is that this confuses users that aren't well versed in XMPP (but just want to admin a chat server). Pubsub as a term is not something that many people will recognize.

I've noticed that the Pubsub tab has but one sub-tab: node administration. That in itself has, now, one menu item: node summary. All in, it's pretty empty, for a complete menu structure.

Perhaps we could move the "node administration" sub-tab to the Server tab, renaming it "Pubsub"? That'd both condense things a little, and keep the top menu clean-ish.

This again is not a hill I'm willing to die on (or even do battle on), but I think it'd make for a nice improvement, that takes little effort. Thoughts?

deleolajide:
First of all,  thank you @ian-green for doing this. It is most appreciated.

We are heavy users of xmpp pub sub and will prefer it to be a top level menu item. This gives us a lot of scope to add more features. These features are currently buried inside a few commercial plugins, but with this pubsub place holder available, you can expect a few more PRs over the coming months to fill up the menu structure :-)

guusdk:
I'm very interested in what kind of things you might be providing, @deleolajide  :)

Still, I'm concerned with the increased complexity of the Admin Console. One of the bigger plusses of Openfire used to be that it was super easy to setup and maintain. We've slowly started to drift away from that in the past few years. I'd like to prevent us from making things more complex.

I'd still prefer to move the new Pubsub menu down a notch, and have it be part of the Server tab. We can always re-arrange the menu, when new features pop up that warrant more Admin Console space.

deleolajide:
>We can always re-arrange the menu, when new features pop up that warrant more Admin Console space.

I disagree :-) Nothing more confusing to a sysadmin is a menu that disappears and re-appears elsewhere. 

The addition of pubsub admin features in Openfire is not additional complexity, but a glaring omission in the first place. Pubsub is an approved and widely used XEP with full support from Smack.

I have not looked lintonthe detail of what Ian has done, but we would add the management of subscriptions including presence tracking of subscribers among other things.

guusdk:
Merged with #875

PR #868, opened 2017-09-15
OF-159: Add an s2s testing feature

richmidwinter:
Sends an s2s ping and presents some debug info.


src/java/org/jivesoftware/util/S2STestService.java
```diff
@@ -0,0 +1,213 @@
...
+		// Intercept logging.
+		final StringBuilder logs = new StringBuilder();
+		Appender appender = interceptLogging(logs);
+
+		// Intercept packets.
+		PacketInterceptor interceptor = new S2SInterceptor();
+		InterceptorManager.getInstance().addInterceptor(interceptor);
+
+		// Send ping.
+		Log.info("Sending server to server ping request.");
```
akrherz: could the server being sent to please be denoted in the log message, if not already via some magic I am unaware of :)

PR #869, opened 2017-09-15
Disable External Entities in plugin.xml

dwd:


akrherz:
Should [OF-1302](https://issues.igniterealtime.org/browse/OF-1302) appear in this commit?

guusdk:
I don't think that @dwd's change fixes OF-1302 - but it's probably fixed with a very similar change. Dave, want to have a look?

dwd:
Similar commit, different code.

PR #870, opened 2017-09-15
Revert "OF-1376 Use an event rather than an arbitrary delay before starting clustering"

guusdk:
Reverts igniterealtime/Openfire#860

Reverting this change, as it appears to break the plugin. This is likely related to #824.


guusdk:
This doesn't fix anything. :/

PR #871, opened 2017-09-15
OF-1309: Make Hazelcast plugin use new API

guusdk:
This is a fix for API changes introduced with the fix for OF-1309.


guusdk:
@dwd The last commit in #824 made changes that were not applied to the Hazelcast plugin. As we had plugin build failures in CI because of other reasons, I've missed that, and merged. That leaves Master in a state of utter brokenness. 

This PR is an attempt to rectification. Although it now compiles, I've not ran this code.

dwd:
I think this should work, but it looks like lgtm had a misbuild - it's only picked up the changes to the plugin.xml

PR #872, opened 2017-09-18
WIP : 198 resumption

dwd:
Do not push this.

It should build OK, but will probably (read: certainly) break in anything approaching real use.

Comments very welcome.


YunlongYang:
Hope Openfire can build on gradle some day.

guusdk:
Please do not misuse PR's for random comments.

dwd:
Argh. CRLF.

dwd:
Thanks for the review, @ganeshbms 

The reasoning behind the deliberate NPE is that I'm trying to ensure that use of getConnection() on a detached session (that is, one that contains this.conn == null) is logged very explicitly. It shouldn't be occurring at all, since the caller can check isDetached() first.

Left to its own devices, it'd generate an NPE anyway, and there's an outside chance it might do the check for null on the return (though I don't think so, I checked each caller), so I don't want to generate an exception without catching it.

So basically it's a mild hack to generate logging and a stack trace.

PR #874, opened 2017-09-23
OF-159: Various tweaks

guusdk:
I'm sorry for the whitespace commit, but let's not add more new files with different line-endings or indentation. Individual commits will expose the exact changes.

@richmidwinter Could you consider installing http://editorconfig.org/ for your IDE / editor of choice? It'll automatically set-up the little bit of code convention that we have (tabs and line-endings)


guusdk:
@dwd, could you have a look at the OF-1379 fix? Do we need to worry about side-effects?

dwd:
We do need to worry about side-effects, of course, but we'll see what we can do in testing to locate these.

PR #875, opened 2017-09-24
OF-35 - PubSub Admin console + PEP

guusdk:
This PR builds on @ian-green's work from #867 and adds support for PEP. 

Most of the refactoring that I did wasn't because there was a problem with the original code. Rather this refactoring made it easier to add PEP support.


deleolajide:
Thank you all for making this happen 😀

PR #877, opened 2017-09-25
Determine the class name from the class, rather than hard-coding it

GregDThomas:
This PR actually does very little; instead of hardcoding the class name - `"com.sun.jndi.ldap.LdapCtxFactory"` it is determined from the class itself - `LdapCtxFactory.class.getName()`

The only reason for this is to deliberately introduce an explicit warning at compile time, as the compiler now knows we're using an internal API. 

```
[of.javac] Openfire\src\java\org\jivesoftware\openfire\ldap\LdapManager.java:87: warning: LdapCtxFactory is internal proprietary API and may be removed in a future release
[of.javac]     private static final String DEFAULT_LDAP_CONTEXT_FACTORY = LdapCtxFactory.class.getName();
```

This may become more of an issue if/when Openfire moves to Java 9.

### DIFF
--- a/src/java/org/jivesoftware/openfire/ldap/LdapManager.java
+++ b/src/java/org/jivesoftware/openfire/ldap/LdapManager.java
@@ -84,6 +84,10 @@ import java.util.regex.Pattern;
 public class LdapManager {
 
     private static final Logger Log = LoggerFactory.getLogger(LdapManager.class);
+    // Determine the name of the default LDAP context factory.
+    // NOTE: Extracting the name from the class rather than hard coding it allows the compiler to detect the use of this
+    // internal class and emit an appropriate warning ("warning: LdapCtxFactory is internal proprietary API and may be removed in a future release")
+    // This is deliberate, to highlight use of classes that may be removed in the future.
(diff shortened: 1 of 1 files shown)


guusdk:
Good idea. I think you should add a javadoc comment in the code, mentioning this, to prevent people from reverting this change by "cleaning up" the code.

GregDThomas:
A good point; I've added the comment as you suggest.

guusdk:
Thanks!

GregDThomas:
Hmm, the merge seems to have broken the bamboo build (https://bamboo.igniterealtime.org/browse/OPENFIRE-NIGHTLYDEB-JOB1-2406/log) - I don't think it's as a result of these changes  is it just one of those things, or should I be worried?

GregDThomas:
Actually, I just noticed that all the bamboo builds are failing - so I'll just ignore it!

akrherz:
@GregDThomas Sorry about the confusion, our debian bamboo build agent is fouled up at the moment and I can't seem to figure out how to fix it :(

PR #880, opened 2017-09-26
OF-1380 Add all.log to Admin Console log viewer

akrherz:
* Did not implement "set mark" for all.log, unsure how to do that or if it can be done


guusdk:
Thanks!
I would not mind this replacing all other tabs (which add nothing but confusion, in my opinion).

akrherz:
@guusdk I think the other tabs are good as they are effectively filters on the `all.log` information.  Especially when the server is in debug mode

PR #881, opened 2017-09-26
OF 35 - Affiliates pages for pub sub admin console.

ian-green:
* Page for viewing existing affiliates
* Ability to delete affiliates (will remove all subscriptions)
* Ability to edit the affiliation type (subscriptions are updated
  as appropriate)

Rebased on top of the PEP stuff so works there as well.

Note: I am not proud of the nested switch statement in pubsub-node-affiliates-edit.jsp, but it is just a reflection of how the current affiliate implementation works and I didn't want to mess with that.

### DIFF
--- a/src/i18n/openfire_i18n_en.properties
+++ b/src/i18n/openfire_i18n_en.properties
@@ -555,13 +555,15 @@ tab.server.descr=Click to manage server settings
     sidebar.sidebar-pubsub=PubSub
         sidebar.pubsub-node-summary=Node Summary
         sidebar.pubsub-node-summary.descr=Click to manage PubSub nodes
-        sidebar.sidebar-node-options=Node Options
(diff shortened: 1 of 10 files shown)


guusdk:
The pubsub admin pages are starting to look really good! Thanks!

A couple of observations:

- `pubsub-node-items.jsp` has a summary table, that displays counts for items, affiliates and subscribers, just like other pages. On `pubsub-node-items.jsp`, the user cannot click on those numbers.
- Some pubsub items contain base64 encoded binary data. That currently breaks the display a bit.
- Listing on the node summary can be lengthy. The pagination is great, but it'd be nice to be able to toggle the sort by clicking on the column headers.

guusdk:
Where subscribers and affiliates are users of the local domain, it'd be neat to link to their userpage ( `/user-properties.jsp?username=<username>` )

PR #883, opened 2017-09-28
OF-1335 Do not rewrite namespaces on forwarded messages

dwd:
This patch works by altering the content namespace
check to only rewrite the element if its ancestors all
share the same namespace. Thus a forwarded
message will no longer be changed.


dwd:
#902 is way better than this. And has Tests. TESTS!!!

PR #886, opened 2017-10-02
UserImportExport plugin: users where not created

annovanvliet:
With the default import, users where not created when it was possible.

### DIFF
--- a/src/plugins/userImportExport/src/java/org/jivesoftware/openfire/plugin/OpenfireExporter.java
+++ b/src/plugins/userImportExport/src/java/org/jivesoftware/openfire/plugin/OpenfireExporter.java
@@ -232,11 +232,11 @@ public class OpenfireExporter implements InExporter {
         }
       }
 
-      if ((userName != null) && (password != null)) {
+      if (userName != null) {
         try {
           userName = Stringprep.nodeprep(userName);
 
-          if (isUserProviderReadOnly) {
+          if (!isUserProviderReadOnly && (password != null)) {
             userManager.createUser(userName, password, name, email);
           }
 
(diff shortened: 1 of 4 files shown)


src/plugins/userImportExport/src/java/org/jivesoftware/openfire/plugin/OpenfireExporter.java
```diff
@@ -236,7 +236,7 @@ public boolean validate(InputStream file) {
         try {
           userName = Stringprep.nodeprep(userName);
 
-          if (isUserProviderReadOnly) {
+          if (!isUserProviderReadOnly && (password != null)) {
```
akrherz: Isn't the `password != null` check already done on line 235?

annovanvliet: You are right, but then is the question, which can be removed? Password is needed to create a user. but not for adding Roster entries. If you compare this to the Xep227Exporter, it would be logical to remove the first check.

akrherz: if the user does not get created, it would seem to not made sense to add roster items to a user that does not exist?  I am unsure though

annovanvliet: This is handled. When the user is not known, an exception is thrown by the usermanager.getUser(), which is handled later on. With a readOnly UserProvider it would be used to update the roster information, where the Users are already are migrated in the LDAP for example.

PR #889, opened 2017-10-03
gojara Fixed ofGojaraSessions missing after install

Fank:
This table does only exist in upgrade folder, and was did not exist.

It also contain a deprecation fix for SQL-Server 2016+ https://docs.microsoft.com/en-us/sql/database-engine/deprecated-database-engine-features-in-sql-server-2016#features-not-supported-in-a-future-version-of-sql-server

### DIFF
--- a/src/plugins/gojara/src/database/gojara_sqlserver.sql
+++ b/src/plugins/gojara/src/database/gojara_sqlserver.sql
@@ -1,12 +1,19 @@
-
+-- Create SessionEntry
+CREATE TABLE ofGojaraSessions(
+  username			varchar(max) NOT NULL,
+  transport			varchar(max) NOT NULL,
+  lastActivity		bigint(20) NOT NULL,
+  PRIMARY KEY (username, transport)
+);
+CREATE INDEX ofGojara_lastActivity_idx  ON ofGojaraSessions(lastActivity);
 
 CREATE TABLE ofGojaraStatistics (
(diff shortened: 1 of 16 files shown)


src/plugins/gojara/src/database/gojara_sqlserver.sql
```diff
@@ -1,12 +1,19 @@
...
+);
+CREATE INDEX ofGojara_lastActivity_idx  ON ofGojaraSessions(lastActivity);
 
 CREATE TABLE ofGojaraStatistics (
   logID 			bigint(20) NOT NULL AUTO_INCREMENT,
   messageDate		bigint(20) NOT NULL,
   messageType 		tinytext NOT NULL,
-  fromJID 			text NOT NULL,
-  toJID 			text NOT NULL,
-  component			text NOT NULL,
```
Fank: `text` is deprecated and needed to be replaced with `varchar(max)`

akrherz:
@Fank thanks, please bump the plugin's `plugin.xml` so that this change can be immediately released as well.

Fank:
@akrherz done

akrherz:
@Fank thanks, but you need to bump the <date> in `plugin.xml` and update the `changelog.html`

akrherz:
hopefully, you can easily rebase this to current master too now, if not, I can push this over the finish line for you, if you like.

akrherz:
OK, I attempted to rebase this and update the changelog, did I get it right?

PR #891, opened 2017-10-06
OF-35 Node config and Service config pages

ian-green:
Adding pages to configure the PubSub service and configure leafNodes.

Node configuration is based on existing Configuration forms mechanism, Data form is requested and a UI is generated from the form, the form is tailored slightly and there is some validation (on fields with users and groups). On submission the completed form goes back to the existing Node.configure() method.

Service configuration uses a similar mechanism in order to re-use form generation and processing code.

Note that these pages are (currently) only for PubSub and do not show up when viewing PEP nodes.

Second commit fixes a couple of bugs in previous OF-35 work including OF-1386.

### DIFF
--- a/src/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
+++ b/src/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
@@ -149,40 +149,37 @@ public class PubSubServiceInfo {
         for (FormField field : form.getFields()) {
 
             if (excludedFields == null || !excludedFields.contains(field.getVariable())) {
+
+                FormField completedField = completedForm.addField(field.getVariable(), field.getLabel(), field.getType());
+
                 switch (field.getType()) {
                 case boolean_type:
-                    FormField boolField = completedForm.addField(field.getVariable(), field.getLabel(),
-                            field.getType());
-                    boolField.addValue(ParamUtils.getBooleanParameter(request, field.getVariable()));
+                    completedField.addValue(ParamUtils.getBooleanParameter(request, field.getVariable()));
                     break;
                 case jid_multi:
-                    FormField multiJidField = completedForm.addField(field.getVariable(), field.getLabel(),
-                            field.getType());
                     for (String param : ParamUtils.getParameters(request, field.getVariable())) {
-                        multiJidField.addValue(param);
+                        completedField.addValue(param);
                     }
                     break;
                 case list_multi:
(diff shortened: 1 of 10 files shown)


src/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
```diff
@@ -1,19 +1,47 @@
 package org.jivesoftware.openfire.pubsub;
 
+import org.jivesoftware.openfire.XMPPServer;
```
guusdk: Something is off with the indentation in this file. Could you make it consistent please?

src/web/pubsub-form-table.jsp
```diff
@@ -0,0 +1,151 @@
...
+<script>
+function clearSelected(id){
+    var elements = document.getElementById(id).options;
+
+    for(var i = 0; i < elements.length; i++){
+      elements[i].selected = false;
+    }
+  }
+
+  function delteTableRow(rowId) {
```
guusdk: At least you consistently mistype 'delete' :)

src/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
ian-green: Done, also re formatted pubsub-form-table.jsp

src/web/pubsub-form-table.jsp
ian-green: Done.

guusdk:
Is this what it intends to look like. Beauty is in the eye of the beholder, but perhaps you could push the "Allowed to create" and "System Administrators" in their own blocks (with the text that's now to the right, on top)?
 
image

ian-green:
Yeah... this was a bit function over form, putting them in separate boxes might be OK but might look weird on the Node config page (which uses the same form layout).

how does this grab you?
screen shot 2017-10-06 at 16 15 49

GregDThomas:
Why not place the Add in each table? So you'll end up with (excuse the ASCII)

```
User       Action
------     -------
alice      [Remove]
bob        [Remove]
charlie    [Remove]
[        ] [Add]
```

ian-green:
Service Config:

screen shot 2017-10-06 at 17 29 26

Node Config: 

screen shot 2017-10-06 at 17 29 55

An improvement?

GregDThomas:
Looks much better to me.

guusdk:
Ah yes!

deleolajide:
molto bene

PR #895, opened 2017-10-16
OF-1412 Add automated setup support on first run

latentflip:
Note: this doesn't yet support every piece of configuration that the UI does, but does support most of it.

I was unsure exactly where the code should go, maybe a standalone class that the web UI uses too? But this is a start, and works for our purposes currently.

If `<setup>true</setup>` does not exist in openfire.xml (i.e. setup is not complete, and a block does exist called `<autosetup>` with a `<run>true</run>` property inside it, that looks like this:

```xml
    <autosetup>
        <run>true</run>

        <locale>en</locale>
        <xmpp>
            <domain>localhost</domain>
            <fqdn>localhost</fqdn>
        </xmpp>
        <encryption>
            <algorithm>AES</algorithm>
            <key>some-key</key>
        </encryption>
        <database>
            <mode>standard</mode>
            <defaultProvider>
                <driver>org.postgresql.Driver</driver>
                <serverURL>jdbc:postgresql://localhost:5432/a-database</serverURL>
                <username>a-database</username>
                <password>a-password</password>
            </defaultProvider>
        </database>
        <admin>
            <email>admin@example.com</email>
            <password>admin</password>
        </admin>
    </autosetup>
```

Then setup will be run on first launch, and the `<autosetup />` section will be deleted from openfire.xml

### DIFF
--- a/src/java/org/jivesoftware/openfire/XMPPServer.java
+++ b/src/java/org/jivesoftware/openfire/XMPPServer.java
@@ -50,6 +50,7 @@ import org.jivesoftware.openfire.session.RemoteSessionLocator;
 import org.jivesoftware.openfire.spi.*;
 import org.jivesoftware.openfire.transport.TransportHandler;
 import org.jivesoftware.openfire.update.UpdateManager;
+import org.jivesoftware.openfire.user.User;
 import org.jivesoftware.openfire.user.UserManager;
 import org.jivesoftware.openfire.vcard.VCardManager;
 import org.jivesoftware.util.*;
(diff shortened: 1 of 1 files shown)


dwd:
Looks good to me.

GregDThomas:
This is an existing "feature", but can't help wonder if lines 445+, currently
```
if (JiveGlobals.getProperty(propName) == null) {
    JiveGlobals.setProperty(propName, JiveGlobals.getXMLProperty(propName));
}
```
shouldn't be

```
final String xmlPropertyValue = JiveGlobals.getXMLProperty(propName);
if (!xmlPropertyValue.equals(JiveGlobals.getProperty(propName))) {
    JiveGlobals.setProperty(propName, xmlPropertyValue);
}
```

i.e. ensure the database property matches the XML property, not just there is a property.

GregDThomas:
Also, line 450;
```
JiveGlobals.setProperty("sasl.scram-sha-1.iteration-count", Integer.toString(ScramUtils.DEFAULT_ITERATION_COUNT));
```

dwd:
Pushing as is, because while I think this could go further it is doing quite a bit and doesn't appear to do any harm.

PR #896, opened 2017-10-17
OF-1408 Show the cache expiry time on the caches page

GregDThomas:
This PR simply displays the cache expiry time (in milliseconds, if any) on the Cache Summary page.

image


GregDThomas:
A couple of things 
a) joda-time has a nice PeriodFormat class which will prettily format a duration (e.g.  21600000 -> "6 hours", but it seems overkill to add yet another dependency just for this one page.
b) I've added an english entry to all the i18n files for the new column header - I'm not sure if that's strictly necessary - does it default to english if a country specific version is not present?

akrherz:
@GregDThomas Thanks, please consider opening Jira issues ahead of the PR so that your code change can be associated with an issue.  Currently, the issue workflow is how we generate the changelog :/

GregDThomas:
Thanks - I hadn't realised I had JIRA creation rights. Now i do - raised at https://issues.igniterealtime.org/browse/OF-1408

yuan24zz:
How to store online chat records into  table ofoffline column “stanza” like offline chat

guusdk:
@yuan24zz Please do not comment on PRs with unrelated questions. Please ask your questions in https://discourse.igniterealtime.org/

guusdk:
This is a nice addition. I agree that adding a new library is overkill, but we might have something for this. Is something like this re-usable? https://github.com/igniterealtime/Openfire/blob/master/src/web/index.jsp#L211

GregDThomas:
`StringUtils.getElapsedTime(long)` is certainly a good starting point; I think I might need to create a new version, say `StringUtils.getFullElapsedTime(long)` , to make it more useful in this case as "Less than 1 minute" is perhaps less than ideal.

guusdk:
Whatever suits you - I'd not invest to much time though :)

GregDThomas:
Here you go - a definite improvement, IMHO.

image

PR #899, opened 2017-10-23
OF-1408: Display count of entries/hits/misses on Cache summary screen

GregDThomas:
Screen shot of extra information:

image


guusdk:
Nice! For extra glitter, perhaps format the hit and miss numbers with the locale-appropriate thousands separator? Those numbers are likely to grow very fast and 16,253,353,233 is more readable than 16253353233.

GregDThomas:
Noted; I've added a NumberFormatter to the output; I'm hoping it's correct, it will take a while before it's ticked over!

GregDThomas:
Certainly OK in the UK; 
image

PR #902, opened 2017-10-24
OF-1335: Don't strip namespaces everywhere.

guusdk:
Openfire prevents the need to re-scope XML (as described in RFC 6120 section 4.8.3) by
pre-emptively remove certain namespaces (such as 'jabber:client' and 'jabber:server').
It should not, however, remove these namespaces everywhere. This commit prevents such
removal in places other than the first and second level elements (typically: stream
and stanza elements).

### DIFF
--- a/src/java/org/dom4j/io/XMPPPacketReader.java
+++ b/src/java/org/dom4j/io/XMPPPacketReader.java
@@ -16,6 +16,8 @@ import org.xmlpull.v1.XmlPullParserFactory;
 
 import java.io.*;
 import java.net.URL;
+import java.util.Arrays;
+import java.util.Collection;
 
 /**
  * <p><code>XMPPPacketReader</code> is a Reader of DOM4J documents that
@@ -54,6 +56,15 @@ public class XMPPPacketReader {
      */
     private long lastActive = System.currentTimeMillis();
 
+    /**
+     * Stream of various endpoints (eg: s2s, c2s) use different default namespaces. To be able to use a stanza that's
+     * parsed on one type of endpoint in the context of another endpoint, we explicitly ignore these namespaced. This
+     * allows us to forward, for instance, a stanza received via C2S (which has the "jabber:client" default namespace)
+     * on a S2S stream (which has the "jabber:server" default namespace).
+     *
+     * @see <a href="https://xmpp.org/rfcs/rfc6120.html#streams-ns-xmpp">RFC 6120, 4.8.3. XMPP Content Namespaces</a>
+     */
+    public static final Collection<String> IGNORED_NAMESPACE_ON_STANZA = Arrays.asList( "jabber:client", "jabber:server", "jabber:connectionmanager", "jabber:component:accept", "http://jabber.org/protocol/httpbind" );
 
     public XMPPPacketReader() {
     }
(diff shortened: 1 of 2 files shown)


guusdk:
This is an alternative to https://github.com/igniterealtime/Openfire/pull/883

dwd:
This looks much better than #883. Tests and everything!

src/java/org/dom4j/io/XMPPPacketReader.java
```diff
@@ -364,15 +375,13 @@ public Document parseDocument() throws DocumentException, IOException, XmlPullPa
...
                     QName qname = (pp.getPrefix() == null) ? df.createQName(pp.getName(), pp.getNamespace()) : df.createQName(pp.getName(), pp.getPrefix(), pp.getNamespace());
-                    Element newElement = null;
-                    // Do not include the namespace if this is the start tag of a new packet
-                    // This avoids including "jabber:client", "jabber:server" or
-                    // "jabber:component:accept"
-                    if ("jabber:client".equals(qname.getNamespaceURI()) ||
-                            "jabber:server".equals(qname.getNamespaceURI()) ||
-                            "jabber:connectionmanager".equals(qname.getNamespaceURI()) ||
-                            "jabber:component:accept".equals(qname.getNamespaceURI()) ||
-                            "http://jabber.org/protocol/httpbind".equals(qname.getNamespaceURI())) {
```
dwd: Absolutely prefer to push as-is, but it might be worth checking the content namespace is consistent and correct, given that otherwise we won't detect when someone sends jabber:server over S2S, for example.

src/java/org/dom4j/io/XMPPPacketReader.java
```diff
@@ -381,9 +390,13 @@ public Document parseDocument() throws DocumentException, IOException, XmlPullPa
...
-                        if (pp.getNamespacePrefix(i) != null) {
-                            newElement
-                                    .addNamespace(pp.getNamespacePrefix(i), pp.getNamespaceUri(i));
+                        final String namespacePrefix = pp.getNamespacePrefix( i );
+                        final String namespaceUri = pp.getNamespaceUri( i );
+                        if ( namespacePrefix != null ) {
+                            newElement.addNamespace( namespacePrefix, namespaceUri );
+                        } else if ( !( pp.getDepth() <= 2 && IGNORED_NAMESPACE_ON_STANZA.contains(namespaceUri) ) ) {
+                            // Do not include certain default namespace on the root-element ('stream') or stanza level. This makes stanzas re-usable between, for example, c2s and s2s.
+                            newElement.addNamespace( "", namespaceUri );
```
dwd: So I think this won't catch:

`<message xmlns='jabber:client'><body xmlns='jabber:client'>Blah</body></message>`

This probably doesn't matter for now.

guusdk: Should we even try to normalize XML that passes through?

src/java/org/dom4j/io/XMPPPacketReader.java
guusdk: (aren't you supposed to send jabber:server over s2s? You ment jabber:client, right?)

I'm not sure what you mean here. Yes, we might want to check, inbound, that the namespace matches the connection type (although I think that RFC6120 doesn't require that and even allows for other namespaces to be used - which would all be terribly confusing). On the outbound stream, we define the default namespace in our code. As all stanzas are stripped of their default namespace _on the stanza level_ (not in their child elements), things should be OK?

dwd:
So, somewhere, this isn't working. Candy sessions to open_chat are passing stanzas through to S2S sessions under the jabber:client namespace, which is problematic as many servers won't handle that.

C2S session seem fine, S2S sessions seem fine. So it's possibly Websocket only that's affected?

PR #903, opened 2017-10-26
OF-1418: Reduce noisy logs from LDAPManager

GregDThomas:
Logging that a UserNotFoundException / GroupNotFoundException is about to be thrown, and then logging that exception, can be quite noisy on some sites. This PR ensures that these exceptions are only logged at trace level, keeping the log file cleaner.


guusdk:
Thanks!

PR #905, opened 2017-10-26
OF-1134 - Update copied users nickname on other users rosters

nerino1:
Still works as intended if you are just renaming a user.

Now, if you are using it to copy a user, the new users nickname, on other people rosters, will reflect the new name and not the old users nickname.

### DIFF
--- a/src/plugins/justmarried/changelog.html
+++ b/src/plugins/justmarried/changelog.html
@@ -44,6 +44,11 @@
 Just married Plugin Changelog
 </h1>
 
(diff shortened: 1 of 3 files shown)


akrherz:
Thank you very much.  Please consider including a commit that bumps the version in `plugin.xml` and updates the `changelog.html`, such that we could release an updated plugin once/if merged.

nerino1:
Thanks! this is my first time ever trying this :)

PR #906, opened 2017-10-27
OF-1373: Allow certificate changes to be applied without a restart.

guusdk:
This commit grew a little larger than anticipated - my apologies for the reviewer. The original intend
for this was to allow changes in the java Keystore to be applied, without requiring a restart of Openfire.
This is what allows outside processes to update the Keystores - think Let's Encrypt - but could also be
useful for other purposes.

In the process of writing this code, I've found that a restart of the Connection Acceptor would still cause
all existing connections to be terminated. This is why a 'reload configuration' method was added.

I've found that the old Certificate Event Listener methods were unuseful. They were pretty detailled, but
all implementations used them in the same way: "reload everything". I've replaced those with one
'something changed' event. Also, the event listeners are no longer triggered in various places in the code.
Instead, the event listeners will now be triggered by the filesystem-based change of the keystore - the same
event that's used to reload configuration when Let's Encrypt updates pop up.

I've removed various bits of lengthy, unused code in the old CertificateManager (primarily code that
interacts directly with KeyStores, as that caused timing issues during the reload).

One functional change (that was marked as 'unsure why we do this' in code) is that after this commit, the
content of the identity store is no longer merged with the content of the trust store, while determining if
a particular end-entity certificate is to be trusted.

### DIFF
--- a/src/java/org/jivesoftware/openfire/container/AdminConsolePlugin.java
+++ b/src/java/org/jivesoftware/openfire/container/AdminConsolePlugin.java
@@ -44,6 +44,7 @@ import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.eclipse.jetty.webapp.WebAppContext;
 import org.jivesoftware.openfire.JMXManager;
 import org.jivesoftware.openfire.XMPPServer;
+import org.jivesoftware.openfire.keystore.CertificateStore;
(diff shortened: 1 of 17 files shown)


guusdk:
Oh, as a plus, this commit will also allow other connection configuration changes to be applied without a restart.

PR #908, opened 2017-10-27
OF-1422 Handle rejoins with nick-sharing

dwd:
This checks to see if any occupant exists with
the same full jid, and if so treats this as a
clientOnlyJoin (ie, rejoin).


akrherz:
I attempted to grok this, but failed.  It seems like the occupant check should be against the nickname and not the full JID as what happens if I attempt to join the room twice and pick a nickname already used by another user.  It seems this change would let me in.

PR #911, opened 2017-11-02
OF-1423 Honour configuration for message size

dwd:
Openfire already uses xmpp.parser.buffer.size as
the maximum stanza size in NIO client sessions.

This change ensures that this is also used when
configuring Websocket sessions.


akrherz:
My only comment here is that `XMLLightweightParser.java` implements a property listener to change this "on the fly", it would be nice if this code did it as well, or is that possible?

akrherz:
Oh, this is created for each new websocket connection?  If so, no worries then about a listener...

dwd:
Oh, except I see a stupid typo.

akrherz:
@dwd What typo are you referencing?

PR #916, opened 2017-11-09
OF-1309 Route based on DomainPairs

dwd:
Because we cannot always multiplex outgoing sessions, we may have multiple outgoing routes to the same domain.

This means that when we have a route A->B, and try to send traffic for conference.A->B, the traffic is rejected and the session closed, and we can end up creating a new session which becomes the new route to B, unsuited to traffic from A itself. This has some interesting (read, confusing) side effects, beyond simple loss of communication.

For example, if a user at B has a user at A in their roster, the session B-A is typically open before the user can join a room at conference.B. When joining, the stanzas associated with the join are all sent "at once", and promptly rejected. Other users then see both an unavailable (ie, leaving) presence from the user at B, followed by a successful join - the user at B however sees absolutely nothing.

Meanwhile, the session A->B has now gone, and traffic from user@A will be initially lost to user@B before the session "flips" again.

This reworks routing such that the route tracked is not simply "to B", but "from A to B", allowing us to track multiple outgoing routes.

### DIFF
--- a/src/plugins/hazelcast/src/java/org/jivesoftware/openfire/plugin/util/cache/ClusterListener.java
+++ b/src/plugins/hazelcast/src/java/org/jivesoftware/openfire/plugin/util/cache/ClusterListener.java
@@ -34,6 +34,7 @@ import org.jivesoftware.openfire.handler.DirectedPresence;
 import org.jivesoftware.openfire.handler.PresenceUpdateHandler;
 import org.jivesoftware.openfire.plugin.util.cluster.HazelcastClusterNodeInfo;
 import org.jivesoftware.openfire.session.ClientSessionInfo;
+import org.jivesoftware.openfire.session.DomainPair;
 import org.jivesoftware.openfire.session.IncomingServerSession;
 import org.jivesoftware.openfire.session.RemoteSessionLocator;
 import org.jivesoftware.openfire.spi.BasicStreamIDFactory;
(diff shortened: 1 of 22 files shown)


dwd:
I'm still testing this (ie, I haven't even run it yet), so don't merge - but I'd welcome any early comments.

dwd:
I can't break this, now. But some of the older (non-Mavenized) plugins don't build, like Kraken. I'll work on fixing those now.

dwd:
Kraken was the only one (now fixed).

akrherz:
@dwd won't kraken's release and minimum openfire version need bumped as well?

dwd:
You're quite right. I'll sort that now.

src/plugins/hazelcast/src/java/org/jivesoftware/openfire/plugin/util/cache/ClusterListener.java
```diff
@@ -71,20 +72,19 @@
...
     private static final int ANONYMOUS_C2S_CACHE_IDX = 1;
-    private static final int S2S_CACHE_NAME_IDX= 2;
-    private static final int COMPONENT_CACHE_IDX= 3;
+    private static final int COMPONENT_CACHE_IDX= 2;
 
-    private static final int SESSION_INFO_CACHE_IDX = 4;
-    private static final int COMPONENT_SESSION_CACHE_IDX = 5;
-    private static final int CM_CACHE_IDX = 6;
-    private static final int ISS_CACHE_IDX = 7;
+    private static final int SESSION_INFO_CACHE_IDX = 3;
```
guusdk: How will this behave in clusters that are being upgraded? Would it be safer to not re-number the indexes?

PR #925, opened 2017-11-14
Fix for time assurance test

Alameyo:
Test was looking trying to assure that after some time a number of seconds will pass, but with time units like "30 seconds". That was causing problems while I was building Openfire in Polish version of OS and units were named in Polish (e.g. second - sekunda). Credit for this to Guus but I was asked to PR this.

### DIFF
--- a/src/test/java/org/jivesoftware/util/StringUtilsTest.java
+++ b/src/test/java/org/jivesoftware/util/StringUtilsTest.java
@@ -5,9 +5,17 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
 
+import java.util.Locale;
+
+import org.junit.Before;
 import org.junit.Test;
 
 public class StringUtilsTest {
+    
+    @Before
+    public void setUp() {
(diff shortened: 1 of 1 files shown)


src/test/java/org/jivesoftware/util/StringUtilsTest.java
```diff
@@ -5,10 +5,18 @@
...
 import static org.junit.Assert.fail;
 
+import java.util.Locale;
+
+import org.junit.Before;
 import org.junit.Test;
 
 public class StringUtilsTest {
 	
+	@Before
```
guusdk: Install Editor Config and don't have weird indentation like this any more! :)

GregDThomas:
Apologies for this, entirely my fault - a problem I introduced with https://github.com/igniterealtime/Openfire/commit/57944fa8af506ad2bcbfdf37ae34227fe787c05e

src/test/java/org/jivesoftware/util/StringUtilsTest.java
Alameyo: Done.

PR #926, opened 2017-11-15
Fix typo in MUCEventListener

mityada:


guusdk:
Thanks for your contribution!

This change will break the API, that is likely in use by third-party developers. I don't like doing that for a non-functional change.

Perhaps you could restore the old methods, and mark them 'deprecated'?

mityada:
MUCEventListener is an interface, so just adding a new method will break third-party plugins. So maybe leave it as is until a new major release?

PR #927, opened 2017-11-15
Set correct sequence number in the <resumed/> element

mityada:
According to [XEP-0198](https://xmpp.org/extensions/xep-0198.html#resumption):
> If the server can resume the former stream, it MUST return a <resumed/> element, which MUST include a 'previd' attribute set to the SM-ID of the former stream and MUST also include an 'h' attribute set to the sequence number of the last handled stanza sent over the former stream **from the client to the server** (in the unlikely case that the server never received any stanzas, it would set 'h' to zero).

I'm using Smack client. Stream resumption was failing with StreamManagementCounterError because the number of stanzas sent **from the server to the client** was set in the <resumed/> element.


dwd:
Thanks for this. This is really early code, so it's great to see people are testing it (and finding bugs!)

akrherz:
cherry-picked as 56ac5211782bcbf5feb3853e64b1698bcea550fb

PR #928, opened 2017-11-15
Denote Openfire 4.2.0-beta

akrherz:
Note, I didn't do anything with the Maven version, unsure what should be done there.


dwd:
What needs to happen to this on the Maven front?

GregDThomas:
At a guess, edit the <version> in xmppserver\pom.xml - though it's currently set to 4.2.0-SNAPSHOT, so perhaps nothing needs to be done.

PR #930, opened 2017-11-16
OF-1429 Fix client route removal

dwd:
Cut/paste error in removeClientRoute() caused old sessions to be
kept around.

In order to help debug this, I also added some detailing to
session-details.jsp which turned out not to be needed. But it
might prove useful/interesting.


akrherz:
Tested this and was not able to reproduce, great!

Regarding the admin console addition, looks good to me.

PR #932, opened 2017-11-18
OF-1431 Add null safety on packet.getType()

glennular:
OpenFire was throwing a NPE when receiving a Ping request with the type attribute missing.

This PR will just make it return null, and not throw an exception.


akrherz:
Thank you!  I have filed [OF-1431](https://issues.igniterealtime.org/browse/OF-1431) , if you want to update your commit message to use that issue, please feel free.  No biggie though.

dwd:
This patch looks great, but suggests an underlying issue that Openfire shouldn't be accepting such stanzas. IQ stanzas MUST have a type, so we can, and should,  reject them earlier.

PR #933, opened 2017-11-18
replace old community links with discourse equiv

akrherz:


dwd:
Pinko-subversive branch name, though.

PR #934, opened 2017-11-21
OF-1434 Check if sender is a component first in PubSubEngine::createNodeHelper

mityada:
There is no point in calling `isRegisteredUser` (and thus making a `disco#info` query) if the sender is a component.

There is a bigger [issue](https://discourse.igniterealtime.org/t/calling-usermanager-isregistereduser-jid-from-i-o-handler-thread/79779) related to calling `isRegisteredUser` from the I/O handler thread. Beside optimization this fix avoids it for the components.


dwd:
This looks about right. Also the title uses the C++ scope operator, so wins.

GregDThomas:
(or Java 8 method reference)

dwd:
C++ got there first, so it wins. ;-)

PR #936, opened 2017-11-23
OF-546: Exclude MINA bundles

guusdk:
By default, the MINA artifacts pull in 'bundle' instead of 'jar' dependencies. This confuses Intellij,
and likely, other tooling. This commit excludes the 'bundle' dependencies, instead relying on the
fact that all excluded dependency-dependencies are present (as first-level dependencies of the project
itself).

This commit should not introduce any functional changes, but should facilitate developers.


guusdk:
Background: https://stackoverflow.com/questions/11802507/use-mina-jar-dependencies-instead-of-bundles

PR #937, opened 2017-11-23
OF-1437 Try to stop sessions having their routes removed

dwd:
This patch adds a lot of debug logging, plus stops the
console NPE'ing when you try to view a detached session.
    
It also detaches a session before trying to reattach it, to
avoid it being unrouted when the connection is closed.


dwd:
Wrong upstream branch. Having a bad day, obviously.

PR #942, opened 2017-11-27
OF-1435: Ensure that we start a new thread before calling UserManager…

GregDThomas:
…::isRegisteredUser(JID)

### DIFF
--- a/src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
+++ b/src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
@@ -29,6 +29,7 @@ import org.jivesoftware.openfire.pubsub.cluster.RefreshNodeTask;
 import org.jivesoftware.openfire.pubsub.models.AccessModel;
 import org.jivesoftware.openfire.user.UserManager;
 import org.jivesoftware.util.StringUtils;
+import org.jivesoftware.util.TaskEngine;
 import org.jivesoftware.util.cache.CacheFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


GregDThomas:
A general note; I've checked everywhere that UserManager::isRegisteredUser(JID) is called, and I think that the only place that this can cause grief is in the PubSubEngine - other uses appear to be mainly from the UI.

src/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
```diff
@@ -115,7 +123,14 @@ public boolean process(PubSubService service, IQ iq) {
...
-                createNode(service, iq, childElement, action);
+                final Element finalAction = action;
+                // Complete this asynchronously, as UserManager::isRegisteredUser(JID) blocks, waiting for a result which may come in on this thread
+                TaskEngine.getInstance().submit(new Runnable() {
+                    @Override
+                    public void run() {
+                        createNode(service, iq, childElement, finalAction);
+                    }
+                });
                 return true;
```
guusdk: This return statement indicates that the IQ stanza was handled, which, now that things are asynchronous, is no longer the case. I'm not sure how this will hold up in a larger chain of handlers. Even when there are no issues here, I'm not comfortable by setting a precedent here. Perhaps we could make things more explicit, and return a Future-like promise instead?

GregDThomas: I like this idea; I think (in the absence of returning a Java 8 `Optional<Future>`) we'll have to return `null` (to indicate the packet was not handled) or the `Future` itself where it is.

GregDThomas: PS When I wrote the original changes, I did check if the existing uses were likely to be thrown off by the async processing, and it all seems to be OK, but making it more explicit is definitely a plus point.

PR #943, opened 2017-11-28
OF-1441: Ensure only the necessary JAR files are included in the dist…

GregDThomas:
…ribution

A couple of additional notes; 

I removed a duplicate dependency on commons-pool2, which IntelliJ correctly picked up as being included twice. 

The scope of concurrentlinkedhashmap-lru was changed from test to runtime, as it is actually used. 

The resulting difference is just 6 jar files;

byte-buddy-1.7.4.jar
byte-buddy-agent-1.7.4.jar
hamcrest-core-1.3.jar
junit-4.11.jar
mockito-core-2.10.0.jar
objenesis-2.6.jar


guusdk:
Thanks!

PR #944, opened 2017-11-28
OF-1442: Exclude the old version of dom4j

GregDThomas:
Caused by the groupId changing from dom4j to org.dom4j between the two versions - so the older version needs to be specifically excluded.


guusdk:
Thanks!

PR #947, opened 2017-11-30
OF-1444: Ensure that mvn packaged log4j.xml file does not expand vars

GregDThomas:


guusdk:
Odd that it was deliberately added as a filtered resource. Someone had a very specific thought there. I'd be interested in knowing what that was.

Instead of toggling it to 'false', we might as well remove the line. Filtering is sufficiently uncommon that no-one will wonder about the missing filtering statement (keeping it in, using the default, might even confuse some).

GregDThomas:
A fair point. I'm too often guilty of trying to the minimum required :( 

Anyway, I've updated the PR

PR #951, opened 2017-12-05
OF-1254 drop ofMucConvLog_msg_id index prior to creating it

akrherz:
This is an attempt to workaround an issue introduced whereby the database schema version is incorrect.  More info is here
https://discourse.igniterealtime.org/t/79916

I did not test this with mysql yet, as it has the most complex change to try dropping an index.


akrherz:
Note, maven builds are currently busted due to unrelated issues to this change.

akrherz:
I modified the mysql change and actually tested it on mariadb, so maybe it is good?

PR #953, opened 2017-12-08
OF-1447: Improve Maven structure

guusdk:
This commit aims to make the Maven structure more compatible with Sonatypes guidelines for artifact distribution, as defined on http://central.sonatype.org/pages/ossrh-guide.html

A new Maven profile 'release' is added, that intends to produce all required artifacts. Usage: `mvn -P release clean verify`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,6 +8,8 @@
     <name>Parent</name>
     <packaging>pom</packaging>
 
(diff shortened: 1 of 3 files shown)


guusdk:
Note that `mvn -P release clean verify` on this state of the code will result in an error for the very last module ("Web-based Administration Interface"), when it appears to gpg-sign an artifact that does not exist.

However, before the error, all artifacts required for uploading to Sonatype have been generated.

PR #955, opened 2017-12-08
OF-1448 no i18n encoded sources

guusdk:


guusdk:
I've deliberately left the Russian file out, as there's an ongoing PR for that.

PR #980, opened 2017-12-14
OF-1419: Replace Proxool with Apache Commons DBCP

GregDThomas:
[A second attempt to submit a PR, with a single commit]

Probably the biggest impact of this is a slight change to the stats that are shown on the Database admin page / exposed by the jmxweb plugin


akrherz:
Thanks, I was able to build this with ant and it appeared to run just fine on Postgresql and Mariadb.  Can you elaborate some on what will change with the database stats / jmxweb plugin?

GregDThomas:
On the admin screen:
The following information on active/idle connections were removed, as they are not exposed by commons-dbcp:

- `server.db.connection_details.id` - the connection id
- `server.db.connection_details.when_created` - the time the connection was created
- `server.db.connection_details.last_used` - the time the connection was last used
- `server.db.connection_details.thread` - The name of the thread that last used the connection

The following were added, as they could potentially be useful:
- `server.db.connection_idle_time` - how long before an idle connection is evicted from the pool
- `server.db.connection_max_wait_time` - the maximum time a client should wait for a connection before failing
- `server.db.test_timeout` - how long to wait before the connection test fails
- `server.db.connection_mean_borrow_time` - the average time a client waited for a connection
- `server.db.connection_max_borrow_time` - the maximum time a client waited for a connection

On the JMXWEB plugin, the following were removed as again they were no longer available;

- `getMaximumActiveTime()`
- `getOfflineConnectionCount()`

akrherz:
Tagging @deleolajide here since it touches jmxweb, but this seems ready to go.

deleolajide:
Thanks for the heads up :-)

PR #983, opened 2017-12-19
Update openfire_i18n_ru_RU.properties

Ivan4537:


src/i18n/openfire_i18n_ru_RU.properties
```diff
@@ -1825,6 +1825,15 @@ ssl.certificates.signature-algorithm=Алгоритм подписи
 ssl.certificates.signature-algorithm-parameters=Параметры алгоритма подписи
 ssl.certificates.identity-store=Хранилище удостоверений
 ssl.certificates.trust-store=Хранилище доверия
+ssl.certificates.alternative-name.other=(Прочие имена
```
guusdk: The last character is missing. Should be: (Прочие имена)

Ivan4537: Corrected

PR #986, opened 2017-12-21
OF-1404: add the MS SQL Server drivers to the distribution.

GregDThomas:
Add the MS JDBC drivers to the distribution, and make them selectable from the setup screens. 

Also update the supporting documentation.


GregDThomas:
This works well for me, but I'd be grateful if any other SQL Server users could check it too

PR #992, opened 2017-12-22
OF-546 Add install4j support to the Maven build.

guusdk:
Using the 'install4j' profile on a server that has a suitable install4j and JREs, Maven can now
generate install4j builds, using: mvn -P install4j clean package

Note that Maven, even without the install4j profile, will no longer output the build artifacts
in /target. Instead, these are now outputted into /distribution/target/distribution-base/


guusdk:
Please note that we should properly re-test both the 'base distribution' as well as the install4j generated artifacts.

akrherz:
This PR was discussed today at the weekly meeting and was decided to merge and continued to test and iterate forward on this task.

PR #995, opened 2018-01-03
OF-1461 - Making fastpath plugin compatible with openfire meetings

deleolajide:
Fixed minor issues blocking ofmeet working with fastpath.

See https://issues.igniterealtime.org/browse/OF-1461


wrooot:
Should be 2018.

akrherz:
Thanks @wrooot

PR #998, opened 2018-01-25
OF-1468 Group Chat History returns one message too many

annovanvliet:
When specifying “Show a Specific Number of Messages” for Group Chat
History Settings, the user get one message to many. For example, if 10
is specified. the user get the last eleven messages in the room.


annovanvliet:
There is a relation with an older Pull request : OF-144

wrooot:
Created https://issues.igniterealtime.org/browse/OF-1468 to track this.

sco0ter:
+1 I have seen this too, but it never really bothered me.

PR #999, opened 2018-01-26
OF-1467, Check nodeRoutes.containsKeys(key) to avoid NPE 

dbh:
Avoid NPE Exception in ClusterListener, nodeCleanup, when there are no sessions on the node to cleanup


dbh:
Adjusted as suggested. Please review

PR #1001, opened 2018-01-26
OF-1469 New plugin: XEP-0215 External Service Discovery

guusdk:


dwd:
Is this something that's better off in its own repository, or is this a possibility we'd want to ship this by default?

guusdk:
I'd not be opposed merging this into core, to combine this with the socks5 / stun / media and possibly jingle code that we have, at some point.

deleolajide:
The media proxy implemented by Jive in the core openfire code, is very dated and was done even before the Jingle specs were finalised. I am of the opinion that it should be replaced with an up-to-date TURN server which would be very useful for simple peer2peer webrtc calls using jingle. making this part of core gets my vote.

PR #1003, opened 2018-02-01
OF-1345 Can't create/reset passwords when using HybridAuthProvider

annovanvliet:
Although not in line with that what has been stated in this issue, this
fixes the possibility to change passwords in the GUI for those users
which are stored in the DefaultUserProvider.

### DIFF
--- a/src/java/org/jivesoftware/openfire/auth/HybridAuthProvider.java
+++ b/src/java/org/jivesoftware/openfire/auth/HybridAuthProvider.java
@@ -216,7 +216,42 @@ public class HybridAuthProvider implements AuthProvider {
     public void setPassword(String username, String password)
             throws UserNotFoundException, UnsupportedOperationException
     {
-        throw new UnsupportedOperationException();
+      // Check overrides first.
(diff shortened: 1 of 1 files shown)


guusdk:
I'm not a big fan of changing the functionality of the existing provider. I fear that it introduces issues in systems that somehow depend on the existing functionality.

As an alternative, would using the newly introduced https://github.com/igniterealtime/Openfire/blob/master/src/java/org/jivesoftware/openfire/user/PropertyBasedUserProviderMapper.java (and friends) be usable for your purpose? It allows you to define from which provider certain data originates, which will also give you all functionality from that provider, for that user.

annovanvliet:
My main reason for fixing this , was that you now can/cannot add local users in the admin console. You are allowed to add them but it fails with an error when actually do it. Inconsistent behaviour in the admin console.

sco0ter:
I can't imagine a system which depends on an `UnsupportedOperationException` being thrown (the existing functionality), so I don't understand Guus' reasoning here.

PR #1007, opened 2018-02-07
OF-1477: Sorting order should be based of bare JID

annovanvliet:
When comparing JID to determine the order to show, do not assume full JID's. A affiliation is linked to a bare JID.


sco0ter:
+1

If the JID's compareTo method is properly implemented, it should check for null resources anyway.
Comparing JIDs directly should be the preferred way instead of comparing string values.

PR #1008, opened 2018-02-08
OF-1475: Update to Hazelcast 3.9.2.

GregDThomas:
Essentially a straight-swap out from 3.5.1 to 3.9.2, with minor changes to the config file to remove use of deprecated configuration items.


GregDThomas:
This works fine for me. However, other people need to to try it (Note; I've deliberately left it as a SNAPSHOT version as there are other changes I want to make to the clustering plugin before it gets released).

src/plugins/hazelcast/plugin.xml
```diff
@@ -5,7 +5,7 @@
     <name>Hazelcast plugin</name>
     <description>Adds clustering support</description>
     <author>Tom Evans</author>
-    <version>2.2.4</version>
-    <date>10/26/2017</date>
-    <minServerVersion>4.2.0</minServerVersion>
+    <version>2.3.0-SNAPSHOT</version>
+    <date>mm/dd/2018</date>
```
akrherz: valid date please :)

GregDThomas: Fixed :)

PR #1011, opened 2018-02-09
OF-1474_hazelcast_rest_api_create_room_npe

dbh:
Upon External Serialization, do a null check on modificationDate. It is possible that the modificationDate is null, if the room was just created through REST API and Hazelcast enabled with more than 1 node in the cluster.  That results in an NPE, partial room creation, and failure in communicating the room updated event via hazelcast.

If modification date is null, send the creationDate value instead for the hazelcast room updated event

see https://discourse.igniterealtime.org/t/openfire-hazelcast-clustering-rest-api-room-creation-npe-when-no-modificaitondate-sent/80619/2

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -2723,7 +2723,12 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
         ExternalizableUtil.getInstance().writeSafeUTF(out, subject);
         ExternalizableUtil.getInstance().writeLong(out, roomID);
         ExternalizableUtil.getInstance().writeLong(out, creationDate.getTime());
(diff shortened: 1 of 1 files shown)


GregDThomas:
This does mean the same room on different nodes of the cluster will have different modificationDates, which seems wrong. Instead, could you you do something like ...

`ExternalizableUtil.getInstance().writeLong(out, modificationDate == null ? null : modificationDate.getTime());`

and

`Long modificationDateLong = ExternalizableUtil.getInstance().readLong(in);`
`modificationDate = modificationDateLong == null ? null : new Date(modificationDateLong);`

dbh:
I'll play with this a bit and get back to you

dbh:
This appears to now long be needed. Greg Thomas pushed a commit for this particular issue.

GregDThomas:
(Yes, this PR still needs to be merged, though I did tweak it to ensure a null modificationDate was preserved)

PR #1015, opened 2018-02-12
OF-1480: Ensure that the cached LDAP information is serializable.

GregDThomas:


GregDThomas:
Guus - I think you introduced this with OF-1306 - I'm unable to test this locally, are you able to take a look?

PR #1016, opened 2018-02-13
Denote Openfire 4.2.2 Release

akrherz:


GregDThomas:
I  worry that the plugins parent version is changed with no other changes. It does mean that there (for example) multiple versions of candy plugin v. 2.2.0 Release 2. One built on Openfire 4.2.2, one built on Openfire 4.2.1 etc..

I think the parent versions all be frozen, and only changed when necessary, at the same time as the plugin version. 

[Given this is regular practice, probably too late at this point, but going forward ...]

akrherz:
@GregDThomas your comment is duly noted!

PR #1017, opened 2018-02-13
OF-1481: Don't attempt to send over a non-existing connection.

guusdk:
I've found this after a round of testing by Daniel (of Conversations fame). I'm assuming, but was unable to verify, that this condition is caused by our recent addition of Stream Management - which leaves a session open, outliving its connection, perhaps.

### DIFF
--- a/src/java/org/jivesoftware/openfire/session/LocalSession.java
+++ b/src/java/org/jivesoftware/openfire/session/LocalSession.java
@@ -410,6 +410,11 @@ public abstract class LocalSession implements Session {
 
     @Override
     public void deliverRawText(String text) {
+        if ( conn == null )
+        {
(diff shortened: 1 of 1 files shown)


guusdk:
@dwd Is this a symptom fix, or a proper one?

sco0ter:
Just as note: Now the caller of this method has no idea whether the delivery was successful or not, in both cases the method returns normally. I'd generally be careful about such design, but maybe it's appropriate here.

Furthermore a race condition could happen here. If `setDetached()` is called after the null check, the `conn` will be `null`. The `conn` variable lacks some proper synchronization.

dwd:
That was more or less why I didn't do this guard originally - I decided a hard fail was preferable, and tried to audit the code instead.

FWIW, I still think auditing the code is the "right" solution, and I think you're right in both cases, but this is thought to make the sessions markedly more stable.

PR #1022, opened 2018-02-16
OF-1200 OF-1486 OF-1487 OF-1489: Monitoring plugin fixes

guusdk:
I'm making this PR as a call for feedback. 

The fix for OF-1200 is not complete, and will starve the client thread pools under even moderate amounts of clients that concurrently perform MAM requests - which is dangerous. The fix does hint at what direction I plan to take this though.

The other commits should be workable, I think, but can perhaps be improved. Feedback, please!

### DIFF
--- a/src/plugins/monitoring/changelog.html
+++ b/src/plugins/monitoring/changelog.html
@@ -48,6 +48,7 @@ Monitoring Plugin Changelog
 <ul>
     <li>[<a href='https://issues.igniterealtime.org/browse/OF-1486'>OF-1486</a>] - MAM RSM queries for MUC should allow for 'backwards-paging'.</li>
     <li>[<a href='https://issues.igniterealtime.org/browse/OF-1487'>OF-1487</a>] - Archived messages should become available in the database instantly.</li>
+    <li>[<a href='https://issues.igniterealtime.org/browse/OF-1200'>OF-1200</a>] - Don't answer MAM queries before relevant archives have been stored.</li>
 </ul>
 
 <p><b>1.5.9</b> -- Feb 13, 2018</p>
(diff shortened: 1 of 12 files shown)


guusdk:
I feel that I've now addressed the OF-1200 issue that I described earlier.

@GregDThomas I've taken a different approach for your a and b concerns, but it should address your concerns, I think? As for c and d: those would be nice, but if we go down that road, I'd rather apply that structurally throughout the code, instead of just here. Finally, the clustering issue: I'm worried that polling all of the cluster nodes here introduces a lot of overhead. I do expect that most to-be-stored-data will live on the local node (where the user is connected) anyways, so can we get away with _not_ querying the cluster?

GregDThomas:
re a) & b), yes, I think that's a good step. I'm still not sure about clustering, though. Doesn't the plugin return archived messages from all users?

So if userA is connected to nodeA and sends a message, and userB connects to nodeB and asks for the recent messages, ideally it should wait until nodeA has written userAs message before continuing?

I should highlight that I'm not familiar with the plugin, so it's quite possible I've missed something ...

guusdk:
I think you're right about the clustering bit. I've now added a cluster-aware check. Thoughts?

GregDThomas:
It looks like what you've got will work - but now you're polling across the network.

Would it be better to encapsulate the polling inside ConversationManager, with a 

    public void waitUntilLocalNodeHasWrittenAllData(final Date targetEndDate) throws InterruptedException, TimeoutException {

        while ( System.currentTimeMillis() < targetEndDate.getTime() + 2000 ) // TODO Use the value in org.jivesoftware.openfire.archive.ConversationManager.ArchivingRunnable.maxPurgeInterval but also add time to allow for query execution time.
            if (conversationManager.hasWrittenAllDataBefore(targetEndDate)) {
                break;
            }
            Log.debug("Not all data that is being requested has been written to the database yet. Delaying request processing. ");
            Thread.sleep(100);
            if (!conversationManager.hasWrittenAllDataBefore(targetEndDate)) {
                Log.warn("Retrieving data from the database to formulate a response to a MAM query, while data is still waiting to be written there. The response might be incomplete.");
                throw new TimeoutException();
            }
        }

    }

My thinking is that you'll only ever need to call across the network once per node - and that call will block until complete/times out.

guusdk:
That's an improvement, but not one that I have the energy for to put in right now. If what we have now is good enough, I'm tempted to push this as-is, and let others take improve upon what we have here.

GregDThomas:
+1

PR #1025, opened 2018-02-20
OF-1492 remove brackets from default JDBC connection strings

akrherz:
comments welcome


GregDThomas:
Looks Ok to me

GregDThomas:
Although - and I can’t check this right now- would hostname instead of host-name be better (similar for database-name) as it would be easier to select using the mouse. A double click may well select the whole string “hostname” which it won’t for host-name.

guusdk:
One more step towards foolproofness! Thanks man.

PR #1032, opened 2018-03-01
HZ-9: Ensure clustering works properly when enabled through the admin UI

GregDThomas:
This turned out to be because the clustering didn't like being enabled as part of the Jetty thread - so it was necessary to start it in another thread, and wait for it to complete. 

I suggest it's a candidate for 4.2.3 as well as 4.3.0


guusdk:
This now needs rebasing.

GregDThomas:
Have done!

PR #1037, opened 2018-03-07
Denote Openfire 4.2.3 Release

akrherz:


GregDThomas:
(again, the parents of otherwise unchanged plugins have changed ...)

akrherz:
rebased and updated, anybody with thoughts other than the ongoing maven versioning :)

guusdk:
Hurgh. I'm somewhat tempted to revert https://github.com/igniterealtime/Openfire/pull/1017 - which was a hackish fix that does not appear to solve the issue that it was attempting to solve - but reverting that would require additional fixes. I don't think it'd hurt to do a release with that change in it. I'm cautiously in favor of doing a release based on the code as-is.

akrherz:
updated PR now and am smoke testing current 4.2 branch build on Ignite.  @dwd may we proceed?

PR #1039, opened 2018-03-08
Updating the REST API Plugin to 1.3.0

Redor:
Added: Security Audit endpoint to get the security logs
Added: More details by error
Improvement: Better JSON mapping (e.g. single element in an array bug)


guusdk:
I think you're missing a library used by the Ant build. Probably one of the transitive dependencies that you're pulling in with Maven?

src/plugins/restAPI/plugin.xml
```diff
@@ -5,8 +5,8 @@
     <name>REST API</name>
     <description>Allows administration over a RESTful API.</description>
     <author>Roman Soldatow</author>
-    <version>1.2.6</version>
-    <date>05/31/2017</date>
+    <version>1.3.0</version>
+    <date>03/07/2017</date>
```
akrherz: 2018 please :)

Redor: Good catch!

PR #1041, opened 2018-03-08
OF-1505: Prevent admin console stacktrace on TLS config page.

guusdk:


guusdk:
This was caused by me going back and forth on using a different trust store for client certs. I now feel we should, as that's what we've been doing historically.

PR #1044, opened 2018-03-09
When closing a session, log why.

guusdk:


src/java/org/jivesoftware/openfire/net/MultiplexerStanzaHandler.java
```diff
@@ -38,6 +41,8 @@
  */
 public class MultiplexerStanzaHandler extends StanzaHandler {
 
+    private static final Logger Log = LoggerFactory.getLogger( MultiUserChatService.class );
```
dwd: Really? :-)

guusdk: Mu<tab> & commit, obviously.

PR #1049, opened 2018-03-16
OF-1511 OF-1513 OF-1514: Pubsub / PEP changes

guusdk:
This PR includes a new feature, as well as a bugfixes.


dwd:
What @sco0ter said - can we refactor the boolean checks to the form object? ("we" - heh)

guusdk:
Something like this? I've copied the utility method that aught to go in Tinder to one of the classes here, marking it deprecated. That saves me from having to do a Tinder release just for this PR. I'll do that later, obviously.

dwd:
Oh, Kay.

PR #1050, opened 2018-03-16
OF-1516 convert to unicode

wrooot:
Unicode conversion provided by Alex in the forums


wrooot:
I haven't asked if this is just a conversion or something has changed. Looks like conversion. I have checked how Registration plugin's page looks in Openfire and it shows garbled text when Russian language is in use.

At least in the past there was no other way to make it look fine in Spark (and Openfire) other than use \u for non latin symbols. Not sure if something has changed recently.

wrooot:
For main translation files for Openfire and Spark you have already get rid of  the need to encode them. But plugins have their own i18n files that still have to be encoded. I have created tickets for this.
https://issues.igniterealtime.org/browse/OF-1517
https://issues.igniterealtime.org/browse/SPARK-2054
Such change will make this PR unneeded.

akrherz:
Anybody with thoughts on what to do with this PR prior to a potential 4.3 beta release?

wrooot:
OF-1517 is the preferable fix than this PR. Or at least to do something about Monitoring plugin's Russian translations. There are a dozen or so reports in the forums from Russian users. https://issues.igniterealtime.org/browse/OF-1500

wrooot:
Closed in favor of https://github.com/igniterealtime/Openfire/pull/1174

PR #1052, opened 2018-03-18
OF-827: Add support for XEP-0191: Blocking Command.

guusdk:


guusdk:
The build breaks for the Ant build only. This is caused by a change that's required in the Hazelcast plugin. The Maven builds are fixed on a version of Openfire that does not have this change, which is why they succeed.

@GregDThomas how would you like to proceed? Merge this PR (assuming that nothing else is wrong) and take the required changes for Hazelcast as a separate PR, or make the changes in this PR? Also, before releasing a version of Hazelcast with the changes required, would you like to have an earlier release that does nothing but to set a 'max server version', allowing users to gracefully upgrade to a certain version of the hazelcast plugin when they use older versions of Openfire?

GregDThomas:
a) To keep things simple, I'd suggest fixing the HZ plugin with this PR. 
b) It is probably possible to keep this in a backward compatible way. So long as the new version of HZ implements the has/setHasRequestedBlocklist(boolean) methods, but does not tag it with @Override, I don't see why the min server version/parent POM of HZ needs to change.

guusdk:
Would that be safe?

GregDThomas:
I don't see why not. The @Override tag is advisory only. 

 On older versions of OF, the methods will never be used. 
On newer versions of OF, the methods will happily be called.

guusdk:
And interop wise, between cluster nodes? Wouldn't you need different class
signatures to detect properly that they'd not be able to exchange data?

GregDThomas:
If you're running different versions of OF in the cluster with the same version of the HZ plugin, there won't be a problem, as the right methods will be called, as all plugins have the same signature of RemoteClientSession

If you're running different versions of the HZ plugin, then a 2.3.1 calling a 2.3.0 will probably fail with some sort of run time introspection exception, but you've got that problem. anyway.

Neustradamus:
For the 4 years birthday, it is the best gift :)
- https://issues.igniterealtime.org/browse/OF-827

GregDThomas:
No comment on the XEP work itself (my knowledge of the details of XMPP are scarily limited!) but the HZ plugin changes seem fine; I'm glad you left it as a SNAPSHOT as that allows me to make a couple of other API changes that should be done (i.e. making sure that the Cache only contains serialisable and non-null objects.

PR #1053, opened 2018-03-20
Plugin assembly: Allow .html, .xml and .properties files to be filter…

sco0ter:
…ed by Maven

Motivation:
Give third-party developers more flexibility to build their plugins. Currently it's not possible to filter anything besides the plugin.xml.

Further improvements:
- Include lib folder in the assembly (for non-Maven artifacts)
- Update namespace "http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2" -> "http://maven.apache.org/ASSEMBLY/2.0.0"


akrherz:
ping @guusdk on this one, probably good to have it go in before all the maven/install4j work gets done :)

PR #1056, opened 2018-03-24
OF-1491: log4j to log4j2

guusdk:
This commit replaces Log4j with the more up-to-date Log4j2 (which, amongst others, allows one to update the XML config without requiring an Openfire restart).

As most logging functionality is used through the facade that is being provided by the SLF4J library, the amount of changes needed is limited.

I failed to get the logging portion of the server-to-server test tool to work properly. That probably needs some additional work.

Note that when merged, these plugins are updated (and should be re-released):

- gojara 
- justmarried
- restapi
- userservice

No functional changes were applied to those plugins, but they still used the log4j API, instead of the SLF4j API.


GregDThomas:
+1 on the migration to log4j2.

However, next time it's worth highlighting that log4j-api is just a facade over log4j-core. There's also a log4j-1.2-api facade (and log4j-slf4j-impl and log4j-jcl and log4j-jul) that maps the legacy APIs to the new  log4j core keeping code changes to the minimum.

guusdk:
Although I didn't know that, I'm not terribly against updating the implementation itself to match the new API. As we were using the SLF4J api, changes were light (with the exception of some plugins, that still used Log4j directly - I now moved those to SLF4J too).

guusdk:
I've rebased the PR to the current master, and started to work on adding that dependency that you asked, @GregDThomas only to find that it was already in.

Next, I started a clean build of Openfire, and loaded the Hazelcast plugin from the website (which obviously doesn't have any log4j2 changes.

This was logged:

```
2018.07.16 13:56:55 INFO  [pool-61-thread-1]: org.jivesoftware.openfire.plugin.HazelcastPlugin - Waiting for other plugins to initialize before initializing clu
2018.07.16 13:56:55 INFO  [pool-61-thread-1]: org.jivesoftware.openfire.plugin.HazelcastPlugin - All plugins have initialized; initializing clustering

```
Seems to be OK to me.

GregDThomas:
Yup, looks good to me

akrherz:
I'm a day late and multiple dollars short here, but why are there now new jar of `log4j-api-2.11.0.jar` and   `log4j-core-2.11.0.jar` files in `build/lib/merge` ?

guusdk:
Hargh. I think I started this before we ditched Ant. I'll remove them.

PR #1058, opened 2018-03-31
OF-1499 display text for no max users instead of zero.

704nate:
This should fix the referenced ticket: OF-1499. Instead of zero being displayed when there are is no max number of users, "unlimited" will display.

### DIFF
--- a/src/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
@@ -219,7 +219,7 @@ public class IQMUCSearchHandler
         {
             boolean find = false;
 
-            if (names.isEmpty())
+            if (!names.isEmpty())
             {
                 for (final String name : names)
                 {
--- a/src/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
+++ b/src/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
@@ -28,6 +28,7 @@ import org.xmpp.packet.PacketError.Condition;
 import org.xmpp.resultsetmanagement.ResultSet;
 import org.xmpp.resultsetmanagement.ResultSetImpl;
 
+import javax.validation.constraints.NotNull;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


guusdk:
Hi @704nate  - thanks for this! Your last commit seems to have broken the build. Care to give it another go? :)

704nate:
Hi @guusdk -- my apologies. I should have stuck around until the build competed but signed off. I guess the package containing @NotNull isn't available which is why it broke. It wasn't critical so I removed the annotation and it's all better now :)

src/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
```diff
@@ -219,7 +219,7 @@ public IQ handleIQ(IQ iq)
         {
             boolean find = false;
 
-            if (names.isEmpty())
+            if (!names.isEmpty())
```
akrherz: Is this change fixing an unrelated BUG in Openfire?

src/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
```diff
@@ -151,41 +155,41 @@ public IQ handleIQ(IQ iq)
             }
         }
 
-        final FormField matchFF = df.getField("name_is_exact_match");
+        final FormField matchFF = df.getField("nameIsExactMatch");
```
GregDThomas: Did you really mean to change the name of the field, `"name_is_exact_match"` -> `"nameIsExactMatch"` or did you actually mean to introduce a constant `nameIsExactMatch` ?

704nate: Hi Greg. This is just a Boolean variable declared on line 143. I just renamed it to the camel case standard instead of separating w/underscores. So yes I just meant to change the name of the field.

flipflapslap:
Is this issue fixed? Saw that it was open but it looks like it's been taken care of.

akrherz:
@flipflapslap the code has not been merged yet

PR #1062, opened 2018-04-17
OF-546: Drop Apache Ant in favor of Apache Maven.

guusdk:
This commit removes support for Apache Ant, and replaces (some, not all)
references with references to Apache Maven.


guusdk:
This should not end up on the 4.2 branch!

PR #1068, opened 2018-04-23
REST API update to 1.3.1

Redor:
Fixed: That created rooms were not propagated to other nodes OF-1535
Fixed: Missing documentation for "new" json mapping

### DIFF
--- a/src/plugins/restAPI/changelog.html
+++ b/src/plugins/restAPI/changelog.html
@@ -44,6 +44,12 @@
 REST API Plugin Changelog
 </h1>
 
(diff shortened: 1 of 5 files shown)


akrherz:
@Redor thanks, is there an actual code change here associated with OF-1535 ?

Redor:
Good catch :) added it

akrherz:
Just to verify that last commit is the change you wanted? Looks like a massive whitespace change

Redor:
@akrherz yes it looks right. (Formatter kicked in and did a big whitespace change)

lagix:
In Openfire 4.2.3 the REST API page shows blank

PR #1069, opened 2018-04-23
OF-1539: Negative caching for groups.

guusdk:
Openfire uses various caches. Typically, some value is looked up from the database (say, a Group) when it's not already in a cache, and is stored in the cache. The next time that that group is used, it is retrieved from the cache.

This works well, but only for data that exists. In the above example, a query for a non-existent Group will cause a database lookup each time (as the Group is not cached).

Openfire should cache the negative ("group does not exist"), to prevent unneeded look-ups.

Note that this PR currently applies negative caching to only one type of object: groups. If accepted, this should be done for other types of cached objects too.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
@@ -30,6 +30,7 @@ import org.jivesoftware.openfire.event.GroupEventListener;
 import org.jivesoftware.openfire.event.UserEventDispatcher;
 import org.jivesoftware.openfire.event.UserEventListener;
 import org.jivesoftware.openfire.user.User;
+import org.jivesoftware.util.CacheableOptional;
 import org.jivesoftware.util.SystemProperty;
 import org.jivesoftware.util.cache.Cache;
 import org.jivesoftware.util.cache.CacheFactory;
@@ -79,7 +80,7 @@ public class GroupManager {
         return GroupManagerContainer.instance;
     }
 
-    private Cache<String, Group> groupCache;
+    private Cache<String, CacheableOptional<Group>> groupCache;
     private Cache<String, Serializable> groupMetaCache;
     private static GroupProvider provider;
 
@@ -106,7 +107,7 @@ public class GroupManager {
                 }
                 
                 // Since the group could be created by the provider, add it possible again
-                groupCache.put(group.getName(), group);
+                groupCache.put(group.getName(), CacheableOptional.of(group));
 
                 // Evict only the information related to Groups.
                 // Do not evict groups with 'user' as keys.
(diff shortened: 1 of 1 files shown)


GregDThomas:
This will fail if clustering is enabled; a ClusteredCache is backed by an IMap, which does not allow null keys or values. 

A while ago, I did submit a PR that added similar non-null checks to the non-clustered DefaultCache, but it got backed out after it uncovered a case where a null was added. I may look at trying again, now that that (hopefully only) case has been fixed.

Coming back to your immediate problem, I think you'll need to introduce a new object, CachedGroup, which has two fields, boolean isPresent and Group group - the latter only populated if isPresent is set to true.

dwd:
What @GregDThomas says. But we could put in a group with a null name, or some other kind of marker.

GregDThomas:
[Aside; I've just raised https://issues.igniterealtime.org/browse/OF-1589 with the aim of preventing this in the future. I'll try and get a PR in for the 4.3 release]

akrherz:
Can we please get a verdict on this PR prior to our beta release?

GregDThomas:
As it stands, the PR should not be merged (once rebased). It places null values in to a cache, which will fail when clustered. One way to resolve this is to update it to make use of the CacheableOptional class introduced for just this sort of reason.

akrherz:
Everybody agree to close this PR? @guusdk @GregDThomas

GregDThomas:
Up to Guus, really; it might be easier to re-work this PR than start from scratch, or vice-versa.

guusdk:
I've reworked (but not tested) the PR, to make use of the new "CacheableOptional" interface that @GregDThomas introduced a while ago. That should prevent having null values in the cache.

xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
```diff
@@ -339,15 +340,26 @@ public Group getGroup(String name, boolean forceLookup) throws GroupNotFoundExce
...
         } else {
-            group = groupCache.get(name);
+            group = groupCache.get(name).get();
         }
-        // If ID wan't found in cache, load it up and put it there.
+
         if (group == null) {
             synchronized ((name + MUTEX_SUFFIX_GROUP).intern()) {
-                group = groupCache.get(name);
+                group = groupCache.get(name).get();
```
GregDThomas: NPE alert!

groupCache.get(name) may return null. You need something like ...

```
CacheableOptional<Group> optionalGroup = groupCache.get(name);
goup = optionalGroup == null ? null : optionalGroup.get();
```

xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
```diff
@@ -339,15 +340,26 @@ public Group getGroup(String name, boolean forceLookup) throws GroupNotFoundExce
...
                 if (group == null) {
-                    group = provider.getGroup(name);
-                    groupCache.put(name, group);
+                    if (groupCache.containsKey(name) && !forceLookup) {
```
GregDThomas: `if( optionalGroup == null && !forceLookup) {`

PR #1071, opened 2018-04-27
OF-1540 Rest API, delete room should send RoomRemovedEvent to cluster

dbh:
Rest API and Multi-node Hazelcast cluster - Delete Room not propagated to other nodes
With this change, all nodes in the cluster are aware of room deletion.


Redor:
Thy dbh for the pull request

akrherz:
I had trouble cherry-picking this change back to the 4.2 branch, but haven't allocated much time into resolving why.

PR #1073, opened 2018-05-09
OF-1542: Apply filtering to the user summary page

GregDThomas:


src/web/user-summary.jsp
```diff
@@ -48,23 +48,69 @@
...
+            }
+        });
     }
 
-    // Get the user manager
-    int userCount = webManager.getUserManager().getUserCount();
-
-    // Get the presence manager
-    PresenceManager presenceManager = webManager.getPresenceManager();
+    final ListPager<User> listPager = new ListPager<>(request, response, new ArrayList<>(webManager.getUserManager().getUsers()), filter,
```
guusdk: This, sadly, won't work on a server with lots of users. Getting them all from database (and keeping them in memory) is simply to much of a resource hog. Ignite Realtime (250,000 users, last I checked) has been running with 100% CPU since it started, 40 minutes ago. I believe this to be the culprit.

GregDThomas: Some comments added to https://issues.igniterealtime.org/browse/OF-1542 for a new approach - any further suggestions appreciated. 

Though 250,000 is not a big number today. If the core user details (username, name, email) are ~100 bytes, that's less than 25MB. The real issue is that this 25MB is fetched over I/O each and every time user-summary is accessed with the original implementation. I would guess that it was necessary to cache only recently accessed users some years back when less RAM was generally available.

PR #1080, opened 2018-06-18
OF-1528: Allow Java 8 to be used when compiling JSP pages

GregDThomas:


src/web/session-summary.jsp
```diff
@@ -95,99 +95,72 @@
     sessionResultFilter.setSortOrder(order);
     // Filter out the dodgy looking sessions
     // Note; dodgy looking cast as we're stuck using Java 7 to compile JSPs - see OF-1528
```
guusdk: This comment is no longer valid now, is it?

GregDThomas: You're right, yes.

GregDThomas:
I was just about to merge this when I realised that plugins could not take advantage of Java 8 - so I updated the parent pom.xml for the plugins, too.  As far as I can tell, there's no way to inherit the same jetty version from the core xmppserver pom.xml, though happy to be corrected. 

Also, does this mean we should bump the minJavaVersion to 1.8 for all the plugin.xml files, now?

guusdk:
> I was just about to merge this when I realised that plugins could not take advantage of Java 8 - so I updated the parent pom.xml for the plugins, too.  As far as I can tell, there's no way to inherit the same jetty version from the core xmppserver pom.xml, though happy to be corrected. 

I think you're right

> Also, does this mean we should bump the minJavaVersion to 1.8 for all the plugin.xml files, now?

Unsure - that's property defines a _runtime_ restriction. Does the compiled JSP run on java 7? If not, I guess we need `minJavaVersion` on all plugins that have admin console pages (I'm no sure if all plugins do).

GregDThomas:
A JSP compiled with a target of 1.8 will not run on Java 8.

On reflection, I think the correct route is
a) Remove the <source> / <target> blocks from the parent pom.xml - so by default plugins will compile to Java 7
b) If a plugin requires Java 8 JSP, then can set the <source> / <target> appropriately along with their minJavaVersion.

guusdk:
Maybe this discussion can be short-circuited. I just remembered that on master, we already dropped Java 7 support: https://github.com/igniterealtime/Openfire/pull/1028

GregDThomas:
Good point, well made. I'll leave it as it is.

guusdk:
... do we need to bump the version number for all plugins?

GregDThomas:
Another good point, yes. I'll bump them to SNAPSHOT versions, though - the existing released versions (compiled with Java 7) will continue to work on Openfire 4.3 with Java 8 so there's no need to update them, but next time they are released they will be Java 8 only.

GregDThomas:
(but the good news is it's not all plugins; only those that depend on Openfire 4.3.0-SNAPSHOT)

GregDThomas:
Right, so anything using Openfire 4.3.0 as a parent was bumped to "next-minor-version".SNAPSHOT. Let's hope that's end end of it!

akrherz:
Hi @GregDThomas what's the status of this PR?  Attempting to get things triaged

GregDThomas:
I'll see if I can get it re-based tomorrow

GregDThomas:
Closed in favour of https://github.com/igniterealtime/Openfire/pull/1133

PR #1083, opened 2018-06-18
OF-1552: Allow the XML debugger plugin to log to file as well as stdout

GregDThomas:


guusdk:
It probably has a good reason, but why are you tying this to the Log4j API directly, instead of through SLF4j?

GregDThomas:
No sure it's a good reason, but Log4J2 is my default for two reasons;

a) The ability to cut'n'paste `Logger LOGGER = LogManager.getLogger();` between classes without having to try and remember to change the class in `Logger LOGGER = LoggerFactory.getLogger(TheClassIJustCopiedFrom.class);`

b) I'm a big fan of the log4j2 CloseableTheadContext (https://logging.apache.org/log4j/2.x/manual/thread-context.html#CloseableThreadContext) which I find very useful in multithreaded applications. Our code will generally add the Stanza ID, and possible the requesting user, to a CloseableTheadContext during the start of processing, so making it much easier to work out which log message applies to which processing thread.

That said, the change was done subconsciously, I realise that the rest of Openfire uses the SLF4J API, so if general consensus is that we should keep using the  SLF4J API I'm more than happy to update the PR.

guusdk:
I'd prefer to go with one API, unless there are strong reasons to deviate. Not being able to copy/paste is not one of them in my book. :dagger: 

As for log4j2: https://issues.igniterealtime.org/browse/OF-1491 I did some work on that, but I don't appear to have pushed that for some reason. I'll check to figure out why I didn't.

guusdk:
@GregDThomas The log4j to log4j2 migration is in this branch: https://github.com/guusdk/Openfire/tree/OF-1491_log4j-to-log4j2

Currently, it compiles, and Openfire starts and appears to do logging - but I also see exceptions in std-out. I've not spent the time yet to look into that.

GregDThomas:
OK, so I switched from log4j-api to log4j-slf4j-impl and now I'm getting no logging output :S There's obviously something more needed, somewhere ...

GregDThomas:
(The SLF4J API is still a work in progress, need to ponder why it's not working)

GregDThomas:
I gave this some more thought over the weekend and some more testing this morning.

As it stands, this PR will allow the plugin to work with Openfire 4.0, 4.2 and the current master (I've not tried 4.1, but I'm guessing it will, too). 

Relying on Openfire 4.3 as a parent, and/or using the SLF4J->Log 4J2 API bridge, restricts that to just Openfire 4.3. 

For that reason, I'd like to leave it as it is; is that a compelling enough reason for you, @guusdk ?

guusdk:
I've grudgingly have to admit that this doesn't sound barking mad. Can we, however, follow up with a second release for the plugin, that _does_ have the log4j2 API bridge, and defines a minServerVersion?

GregDThomas:
Sure thing - https://issues.igniterealtime.org/browse/OF-1570 raised.

PR #1085, opened 2018-06-19
OF-1554: Remove MySQL Connector/J from distribution

GregDThomas:
If an admin selects the MySQL database option during setup, they are advised to download the JDBC driver it from the MySQL web site.


GregDThomas:
Cancelled, as we can use a newer MySQL driver with a FLOSS exemption

PR #1095, opened 2018-06-20
OF-1561: Update Apache Ant to 1.10.3

GregDThomas:


GregDThomas:
Note; Ant is required for the JspC compilation that's supported for development JSPs in PluginServlet.jsp

PR #1098, opened 2018-06-26
OF-1433: Presence updates should be reflected to sender.

guusdk:
RFC-6121, section 4.4.2 states:
    The user's server MUST also send the presence stanza to all of the
    user's available resources (including the resource that generated the
    presence notification in the first place).


guusdk:
This is not behaving in the way I intended. I'll retry this later.

PR #1099, opened 2018-06-27
OF-1563: Disable stream management by default

GregDThomas:
Note; I've also moved the detection of enabled/disabled to a single place, rather than in two places across the code (hopefully we'll be able to re-enable it RSN),


guusdk:
Thanks Greg!

PR #1102, opened 2018-07-02
OF-1566: Fix for archive search with password protected chatrooms

NielsKr:
Fix for preventing users to access message history (archive search) of a chatroom when they are no occupant in the chatroom at the moment of search.
The fix results in a Forbidden response being send when the requestor of the archive search is not an occupant in the chatroom.

See [Openfire 4.2.3 monitoring service 1.6.0 does not check user has the right to enter the archive] (https://discourse.igniterealtime.org/t/openfire-4-2-3-monitoring-service-1-6-0-does-not-check-user-has-the-right-to-enter-the-archive/82037)

### DIFF
--- a/src/plugins/monitoring/plugin.xml
+++ b/src/plugins/monitoring/plugin.xml
@@ -6,7 +6,7 @@
     <description>Monitors conversations and statistics of the server.</description>
     <author>IgniteRealtime // Jive Software</author>
     <version>1.6.1</version>
-    <date>03/05/2018</date>
+    <date>28/06/2018</date>
     <minServerVersion>4.1.0</minServerVersion>
     <databaseKey>monitoring</databaseKey>
     <databaseVersion>4</databaseVersion>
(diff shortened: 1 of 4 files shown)


src/plugins/monitoring/plugin.xml
```diff
@@ -5,7 +5,7 @@
     <name>Monitoring Service</name>
     <description>Monitors conversations and statistics of the server.</description>
     <author>IgniteRealtime // Jive Software</author>
-    <version>1.6.0</version>
+    <version>1.6.1</version>
     <date>03/05/2018</date>
```
guusdk: Probably should change the date here.

guusdk:
Thanks for this! Could you update the release date please? Otherwise, the changes look good to me.

src/plugins/monitoring/plugin.xml
NielsKr: Updated the release date and included issue number in changelog.

guusdk:
@akrherz would you mind pushing a new plugin version to the website?

akrherz:
@guusdk thanks for your bamboo hint, this is done.

PR #1106, opened 2018-07-09
OF-1533: Use a random IV for each new encrypted property

GregDThomas:
Note 1; this only applies to the properties in the database. Encrypted properties in the `openfire.xml` all share the same, default IV. But as there are only two, and the only place they are used is in the same folder where we could save the IV (`security.xml`), there's no practical security advantage.

Note 2; If the Blowfish algorithm is used, an random IV is generated and saved, but it's not actually used during the decryption process.

### DIFF
--- a/src/database/openfire_db2.sql
+++ b/src/database/openfire_db2.sql
@@ -115,6 +115,7 @@ CREATE TABLE ofProperty (
   name        VARCHAR(100) NOT NULL,
   propValue   VARCHAR(3000) NOT NULL,
   encrypted   INTEGER,
+  iv          CHAR(24),
   CONSTRAINT ofProperty_pk PRIMARY KEY (name)
 );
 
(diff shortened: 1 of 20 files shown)


guusdk:
Note to self: Read "IV" as "Initialization Vector", not "four." This PR makes a lot more sense that way.

I don't see anything wrong with this. As it feels a bit like a point of no return, I though I'd be nitpicky on some details that I _think_ you've already covered.

Properties can be read from XML as well as from the database. Reading from the database is only possible after some properties have been read from XML (notably: the database connection details). Does this PR introduce the possibility of a race situation, where reading the same property has different results depending on how fast after initial boot-up the property is being read?

Also, although I assume that as the IV is stored in the same table as the property value, there are no issues with restoring backups - but have you given that any thought?

Something similar for upgrading from older versions (as opposed to using a fresh install of Openfire). I think I see that upon load, the IV is used only when it's there, and being added/changed upon store. That ought to be fine, but just to check: have you explicitly considered both scenarios (fresh installs / upgrades)?

GregDThomas:
You only /think/ I mean Initialsation Vector. I may have meant random 4. https://xkcd.com/221/

(I’ll give more serious consideration to your comments tomorrow)

guusdk:
@dwd do yo care to have a look at this?

dwd:
I'm possibly old fashioned, but running `new SecureRandom().nextBytes(...)` every update/insert of properties isn't going to use up all the entropy or anything nasty, is it? ISTR such things are now well-protected, but I thought I'd best ask. Meanwhile I think it's fine, and I'll just merge on the assumption this is safe - if it isn't, we should probably review all cases anyway.

GregDThomas:
It's only on those properties which are encrypted, so not a huge number of calls. I did debate about re-using an existing SecureRandom, but at the time found https://stackoverflow.com/questions/1461568/is-securerandom-thread-safe/2726868 and decided to leave it as is.

PR #1107, opened 2018-07-11
OF-1383: Changes to prepare for Java 9 onwards.

guusdk:
This PR contains a couple of smaller changes that allow the code to be compiled with Java 10, while remaining compatible with Java 8.

Note that even with these changes, Openfire still will not run with Java 10. At the very least, a [runtime issue with Apache MINA needs to be resolved](https://issues.apache.org/jira/projects/DIRMINA/issues/DIRMINA-1088). There likely are other problems too.

### DIFF
--- a/src/java/org/jivesoftware/openfire/ldap/LdapManager.java
+++ b/src/java/org/jivesoftware/openfire/ldap/LdapManager.java
@@ -16,7 +16,6 @@
 
 package org.jivesoftware.openfire.ldap;
 
-import com.sun.jndi.ldap.LdapCtxFactory;
 import org.jivesoftware.openfire.group.GroupNotFoundException;
 import org.jivesoftware.openfire.user.UserNotFoundException;
 import org.jivesoftware.util.JiveGlobals;
@@ -97,11 +96,7 @@ import java.util.regex.Pattern;
 public class LdapManager {
 
     private static final Logger Log = LoggerFactory.getLogger(LdapManager.class);
-    // Determine the name of the default LDAP context factory.
-    // NOTE: Extracting the name from the class rather than hard coding it allows the compiler to detect the use of this
-    // internal class and emit an appropriate warning ("warning: LdapCtxFactory is internal proprietary API and may be removed in a future release")
-    // This is deliberate, to highlight use of classes that may be removed in the future.
-    private static final String DEFAULT_LDAP_CONTEXT_FACTORY = LdapCtxFactory.class.getName();
+    private static final String DEFAULT_LDAP_CONTEXT_FACTORY = "com.sun.jndi.ldap.LdapCtxFactory";
 
     private static LdapManager instance;
     static {
--- a/src/web/security-certificate-details.jsp
+++ b/src/web/security-certificate-details.jsp
@@ -5,14 +5,12 @@
 <%@ page import="org.jivesoftware.openfire.spi.ConnectionType"%>
 <%@ page import="org.jivesoftware.util.ParamUtils"%>
 <%@ page import="org.jivesoftware.util.StringUtils"%>
-<%@ page import="javax.xml.bind.DatatypeConverter" %>
 <%@ page import="java.security.AlgorithmParameters" %>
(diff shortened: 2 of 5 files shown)


src/java/org/jivesoftware/openfire/ldap/LdapManager.java
```diff
@@ -97,11 +96,7 @@
 public class LdapManager {
 
     private static final Logger Log = LoggerFactory.getLogger(LdapManager.class);
-    // Determine the name of the default LDAP context factory.
-    // NOTE: Extracting the name from the class rather than hard coding it allows the compiler to detect the use of this
-    // internal class and emit an appropriate warning ("warning: LdapCtxFactory is internal proprietary API and may be removed in a future release")
-    // This is deliberate, to highlight use of classes that may be removed in the future.
-    private static final String DEFAULT_LDAP_CONTEXT_FACTORY = LdapCtxFactory.class.getName();
+    private static final String DEFAULT_LDAP_CONTEXT_FACTORY = "com.sun.jndi.ldap.LdapCtxFactory";
```
GregDThomas: It's all very well obfuscating the use of the com.sun.jndi.ldap class, but are you sure that that this will actually get around the problem? I suspect you're just delaying the problem until the class is instantiated via reflection. It may require a `-XaddExports:java.naming/com.sun.jndi.ldap=ALL-UNNAMED` flag, according to my Google-foo. 

Though of course the real fix is to not use com.sun, perhaps http://directory.apache.org/api/

guusdk: I must admit I did not test this. I am under the impression that, although the class is not exposed by the module, it's still available. Assuming that the instantiation (and usage) occurs within the same module, I don't expect problems.

GregDThomas: OK, fair enough. Another Google brought up  https://bugs.openjdk.java.net/browse/JDK-8186019 which suggests it /should/ work.

GregDThomas: PS. I've raised https://issues.igniterealtime.org/browse/OF-1573 - if I ever get time I may look at it. Though I am slightly nervous about being the last person to touch LDAP - I'll be branded with that forever.

guusdk: I'm sure that @speedy01, the current holder of that brand, would welcome your attempt. :)

speedy01: @GregDThomas Id very much welcome any attempt to make the ldap stuff better! I've held the brand long enough, its time to pass the torch! :)

GregDThomas: @speed01 - I'm still not entirely convinced it's a good idea to throw away the existing LDAP code that is essentially working just fine, and replace it with a whole new bunch of code that introduces a brand new set of bugs. But I worry that at some point the com.sun libraries will disappear, so I guess it probably really does need to be done.

GregDThomas: Everything you say here is right, but I've still got this feeling that at some point in the future, support for it, in one way or the other, will be dropped. Nothing I've seen on `com.sun.jndi.ldap.LdapCtxFactory` says anything on the matter either way, unfortunately.

But at least, for now, we're clear.

But at some point (probably not soon) I still aim to tackle OF-1573 as the Apache Directory LDAP API library provides a bunch of additional security features not present in JNDI. I'm also aware of a JNDI bug we've tripped up on which I have no realistic hope of being fixed (under certain circumstances, pooled connections that were no longer valid were being re-used. The default timeout (never) meant the code locked up, setting a sensible timeout simply meant that request, and all subsequent request using that pooled connection failed. The work-around was to disable pooling).

guusdk: Oh, I'm not against migrating at all. Go for it, by all means. I just ment to say that there's a chance that we won't be _forced_ to migrated

PR #1108, opened 2018-07-11
OF-1571: Ensure RECVd messages are logged

GregDThomas:
NB. Also removed logged of empty messages, and made the Interpreted logger be consistent with the name of `RECV`d messages.

### DIFF
--- a/src/plugins/xmldebugger/changelog.html
+++ b/src/plugins/xmldebugger/changelog.html
@@ -44,6 +44,14 @@
 XML Debugger Plugin Changelog
 </h1>
 
+<p><b>1.7.1</b> -- July 11, 2018</p>
+<ul>
+    <li>[<a href='http://issues.igniterealtime.org/browse/OF-1550'>OF-1550</a>] - XML Debugger plugin fails to display admin console page</li>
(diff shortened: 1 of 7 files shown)


akrherz:
@GregDThomas Does this mean white-space pings would not be logged?

GregDThomas:
Ah, is that what all those blank lines are for? Yes, it would. Let me know if that's an issue ...

akrherz:
@GregDThomas some XMPP clients do not do IQ-ping, so they do whitespace pings.  I am unsure if it is important to check those in the generated logs or not.

guusdk:
Maybe make it configurable?

GregDThomas:
OK, done :)

GregDThomas:
@akrherz - this bumped the Debugger plugin to 1.7.1 - any chance you could update the list of plugins at https://www.igniterealtime.org/projects/openfire/plugins.jsp with it? Thanks

akrherz:
@GregDThomas done.  FYI, the size of the plugin went from 15,855 bytes to 1,675,349 bytes.

GregDThomas:
Thanks Daryl. The size increase is because of some extra dependencies needed for pre-Openfire 4.3 servers, it will shrink again when https://issues.igniterealtime.org/browse/OF-1570 is implemented.

PR #1110, opened 2018-07-12
Add enforcer:enforce goal to check for SNAPSHOT dependencies

GregDThomas:
This adds a new enforcer:enforce goal that prevents anything being released with SNAPSHOT dependencies. Note - this PR alone will do nothing; it's necessary to add the enforcer:enforce goal to the CI server that does the builds.

And /then/ the builds will start failing ;)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -285,6 +285,28 @@
                 </configuration>
             </plugin>
 
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <version>3.0.0-M2</version>
+                <executions>
+                    <execution>
+                        <id>enforce-no-snapshots</id>
+                        <goals>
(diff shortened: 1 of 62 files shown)


guusdk:
Given that the build started failing, I guess we don't need to modify the build for this to become active? :)

GregDThomas:
So it did! I didn't realise the enforcer plugin was automatically triggered, though it makes sense. I'll update the PR to get it compiling again ...

GregDThomas:
The enforcer plugin highlighted many, many plugins with a release version + SNAPSHOT dependency.

For all of these plugins, I've updated the version to a SNAPSHOT, and set the release date to "tbc"

For those plugins where it was possible, I've also rolled back the parent.pom version to 4.2.0 - many had a minServerVersion of 4.0.0 or similar, so it made sense to build using an earlier (non-SNAPSHOT!) version of Openfire.

PR #1111, opened 2018-07-12
OF-1578 Monitoring small Graphs Not Being Generated.

annovanvliet:
Commented setter to null which is according to latest javadoc not allowed.

### DIFF
--- a/src/plugins/monitoring/plugin.xml
+++ b/src/plugins/monitoring/plugin.xml
@@ -6,7 +6,7 @@
     <description>Monitors conversations and statistics of the server.</description>
     <author>IgniteRealtime // Jive Software</author>
     <version>${project.version}</version>
-    <date>06/28/2018</date>
+    <date>07/19/2018</date>
     <minServerVersion>4.1.0</minServerVersion>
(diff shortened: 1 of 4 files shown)


guusdk:
Ah, that bugged me! Thanks for finding this! Could I tempt you to update the changelog et. al?

annovanvliet:
Is there a issue-nr to which  could refer? There is a topic in discource https://discourse.igniterealtime.org/t/monitoring-small-graphs-not-being-generated/82194

src/plugins/monitoring/plugin.xml
```diff
@@ -5,7 +5,7 @@
     <name>Monitoring Service</name>
     <description>Monitors conversations and statistics of the server.</description>
     <author>IgniteRealtime // Jive Software</author>
-    <version>1.6.1</version>
+    <version>1.6.2</version>
     <date>06/28/2018</date>
```
akrherz: please bump the release date too, thanks!

akrherz:
whoa, I did something bad here with git.  Sorry Boggling.

akrherz:
whew, awaiting CI before merging

PR #1113, opened 2018-07-18
OF-1576: MAM support advertisement

guusdk:


GregDThomas:
You'll have to make the plugin version SNAPSHOT if you want it to compile with a SNAPSHOT dependency ...

src/plugins/monitoring/pom.xml
```diff
@@ -8,7 +8,7 @@
     </parent>
     <groupId>org.igniterealtime.openfire.plugins</groupId>
     <artifactId>monitoring</artifactId>
-    <version>1.7.0</version>
+    <version>1.7.0-SNAPSHOT</version>
```
akrherz: @GregDThomas did I do this right then?

GregDThomas: Yup :) 1.7.0 has yet to be released, so it should still be a SNAPSHOT

PR #1115, opened 2018-07-23
OF-1579: Add support for XEP-0368 "SRV records for XMPP over TLS"

guusdk:
This adds support for doing 'direct' TLS (as opposed to StartTLS) when performing server-to-server communication (federation).

With this PR, Openfire can establish direct TLS s2s connections, and also can accept such connections.

### DIFF
--- a/src/i18n/openfire_i18n_en.properties
+++ b/src/i18n/openfire_i18n_en.properties
@@ -1597,6 +1597,7 @@ server.props.name=Server Host Name (FQDN):
 server.props.valid_hostname=Please enter a valid server host name or
 server.props.valid_hostname1=restore the default
 server.props.server_port=Server-to-Server Port:
+server.props.server_ssl_port=Server-to-Server SSL Port:
 server.props.component_port=Component Port:
 server.props.port=Client Port:
 server.props.valid_port=Please enter a valid port number or
@@ -1613,7 +1614,8 @@ server.props.admin_secure_port=Secure Admin Console Port:
 server.props.jmx_enabled=JMX Enabled:
 server.props.jmx_secure=JMX Require Admin User:
 server.props.jmx_port=JMX Connector Port:
-
+server.props.jmx_valid=Please enter a valid port number or
+server.props.jmx_valid1=restore the default
 # Server stopped Page
 
 server.stopped.title_restarting=Restarting Server
@@ -2301,6 +2303,9 @@ ssl.settings.server.label_notrequired=Optional
 ssl.settings.server.label_notrequired_info=Connections between servers may use secured connections.
 ssl.settings.server.label_custom=Custom
 ssl.settings.server.label_custom_info=Advanced configuration
+ssl.settings.server.legacymode.boxtitle=Encrypted (legacy-mode) connections
+ssl.settings.server.legacymode.info=Connections of this type are established using encryption immediately (as opposed to using STARTTLS). This type of connectivity is commonly referred to as the "legacy" method of establishing encrypted communications.
+ssl.settings.server.legacymode.label_enable=Enabled
 ssl.settings.server.dialback=Server Dialback:
 ssl.settings.server.customTLS=TLS method:
 
--- a/src/java/org/jivesoftware/openfire/Connection.java
+++ b/src/java/org/jivesoftware/openfire/Connection.java
@@ -331,39 +331,16 @@ public interface Connection extends Closeable {
      * connection the server requesting the TLS negotiation will be the client and the other server
      * will be the server during the TLS negotiation. Therefore, the server requesting the TLS
      * negotiation must pass <code>true</code> in the <tt>clientMode</tt> parameter and the server
-     * receiving the TLS request must pass <code>false</code> in the <tt>clientMode</tt> parameter.
-     * Both servers should specify the xmpp domain of the other server in the <tt>remoteServer</tt>
-     * parameter.<p>
-     *
-     * In the case of client-2-server the XMPP server must pass <code>false</code> in the
-     * <tt>clientMode</tt> parameter since it will behave as the server in the TLS negotiation. The
-     * <tt>remoteServer</tt> parameter will always be <tt>null</tt>.
-     *
-     * @param clientMode boolean indicating if this entity is a client or a server in the TLS negotiation.
-     * @param remoteServer xmpp domain of the remote server or <tt>null</tt>. When null a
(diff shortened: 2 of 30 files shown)


guusdk:
I've done some interop testing with another instance of Openfire, as well as with a number of other server implementations (ejabberd, metre). I didn't find any issues.

src/i18n/openfire_i18n_en.properties
```diff
@@ -2301,6 +2303,9 @@ ssl.settings.server.label_notrequired=Optional
 ssl.settings.server.label_notrequired_info=Connections between servers may use secured connections.
 ssl.settings.server.label_custom=Custom
 ssl.settings.server.label_custom_info=Advanced configuration
+ssl.settings.server.legacymode.boxtitle=Encrypted (legacy-mode) connections
```
dwd: We probably want to talk about "Direct TLS" rather than "Legacy-mode".

src/java/org/jivesoftware/openfire/Connection.java
```diff
@@ -361,9 +337,10 @@
      * <tt>clientMode</tt> parameter since it will behave as the server in the TLS negotiation.
      *
      * @param clientMode boolean indicating if this entity is a client or a server in the TLS negotiation.
+     * @param directTLS boolean indicating if the negotiation is directTLS (true) or startTLS (false).
      * @throws Exception if an error occured while securing the connection.
      */
-    void startTLS(boolean clientMode) throws Exception;
+    void startTLS(boolean clientMode, boolean directTLS) throws Exception;
```
dwd: Pretty sure that because of XEP-0368 mandating SNI, you'll need to be able to pass in a remote domain here, too.

src/java/org/jivesoftware/openfire/net/DNSUtil.java
```diff
@@ -63,21 +63,29 @@
...
      * @param domain the domain.
      * @param defaultPort default port to return if the DNS look up fails.
-     * @return a list of  HostAddresses, which encompasses the hostname and port that the XMPP
-     *      server can be reached at for the specified domain.
+     * @return a list of  HostAddresses, which encompasses the hostname and port
+     *         that the XMPP server can be reached at for the specified domain.
+     * @see <a href="https://tools.ietf.org/html/rfc6120#section-3.2">XMPP CORE</a>
+     * @see <a href="https://xmpp.org/extensions/xep-0368.html">XEP-0368</a>
      */
     public static List<HostAddress> resolveXMPPDomain(String domain, int defaultPort) {
```
dwd: This looks generally good, but if the server has TLS disabled we should probably not perform the "_xmpps-server" lookup.

src/java/org/jivesoftware/openfire/server/ServerDialback.java
```diff
@@ -207,14 +204,18 @@ public LocalOutgoingServerSession createOutgoingSession(String localDomain, Stri
...
+            if ( socketToXmppDomain  == null ) {
                 log.info( "Unable to create new outgoing session: Cannot create a plain socket connection with any applicable remote host." );
                 return null;
             }
 
+            final Socket socket = socketToXmppDomain.getKey();
+            final boolean directTLS = socketToXmppDomain.getValue();
             connection = new SocketConnection(XMPPServer.getInstance().getPacketDeliverer(), socket, false);
+            if (directTLS) {
+                connection.startTLS( false, directTLS );
```
dwd: You'd want to pass in the remote domain here for SNI.

src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
```diff
@@ -242,16 +244,21 @@ private static LocalOutgoingServerSession createOutgoingSession(String localDoma
...
             return null;
         }
+        final Socket socket = socketToXmppDomain.getKey();
+        final boolean directTLS = socketToXmppDomain.getValue();
 
         SocketConnection connection = null;
         try {
             connection = new SocketConnection(XMPPServer.getInstance().getPacketDeliverer(), socket, false);
+            if (directTLS) {
+                connection.startTLS(true, directTLS);
```
dwd: Set SNI here?

guusdk:
Ah, yes, I ment to check up on SNI, but forgot. 

When I read that Java 8 adds support for SNI, I had hoped that the SNI extension would be set automatically. It does not (I checked with tcpdump just now). I'm now trying to find how this is done in Java in the first place. Although I can find references that tell me to set this in an instance of SSLParameters, I fail to see how that's actually used. From what I can tell, only protocol versions and cipher suites are read from those parameters. Thoughts?

dwd:
Yikes: http://javabreaks.blogspot.com/2015/12/java-ssl-handshake-with-server-name.html

Yeah, this looks surprisingly complicated.

guusdk:
I've seen that, but it appears to be http specific. Our TLS wrapping does
not make use of SSLSockets either, which would be an alternative approach.
I'll try to set the parameters in the TLSEngine that we use, bit the
decompiled version shown by intellij suggests that the SNI part is not used
there at all.

dwd:
OK - so skip SNI for now, but do get the TLS disabled mode working (or at least checked).

src/i18n/openfire_i18n_en.properties
guusdk: Definitely. We got a an issue in jira to change all that. Until then, I thought to keep things consistent.

guusdk:
What 'tls disabled mode' are you referring to, exactly?

We've got a setting that will shut down our ports on which we listen for TLS, but that doesn't necessarily prevent us from making TLS connections ourselves, right?

There's an "SSL enabled" setting buried in http://localhost:9090/server-props.jsp somewhere, but that affects client listening ports only (and should arguably be renamed).

dwd:
So we have stuff like this:

https://github.com/igniterealtime/Openfire/blob/master/src/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java#L294

Where is TLS is disabled it won't try to negotiate it on outbound connections.

guusdk:
Like so?
I think it'll be pretty confusing if we make the same property also modify the state of the listening port for the acceptor - that means that two different settings affect its availability.

dwd:
Yeah... That'll do. I think we might want to ensure you can't disable TLS but still listen at some point, but that can be done in the future.

PR #1118, opened 2018-07-23
Use the version from the pom.xml to set the version in plugin.xml

GregDThomas:
During recent changes to various plugin.xml files for the plugins, it became clear that the plugin versions in plugin.xml did not necessarily match the plugin version in the pom.xml

Maven to the rescue - we already had the hooks in place to edit the pom.xml when the .jar file was created, it was simply a case of editing the plugin.xml appropriately. Now the plugin.xml and pom.xml have identical versions.

### DIFF
--- a/src/plugins/bookmarks/plugin.xml
+++ b/src/plugins/bookmarks/plugin.xml
@@ -3,10 +3,10 @@
 
     <!-- Main plugin class -->
     <class>org.igniterealtime.openfire.plugin.BookmarksPlugin</class>
-    
+
     <!-- Plugin meta-data -->
-    <name>Bookmarks</name>
-    <description>Allows clients to store URL and group chat bookmarks (XEP-0048)</description>
+    <name>${project.name}</name>
+    <description>${project.description}</description>
     <author>Ignite Realtime</author>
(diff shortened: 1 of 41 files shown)


guusdk:
Yup, I was meaning to do this after we dropped Ant support (but forgot). We can do something very similar for the name and description, by the way!

GregDThomas:
Yes ... that probably makes sense!

GregDThomas:
I did check; they are different. Which didn't concern me in the least, as the pom.xml names seemed more consistent with each other (.e.g `Callback on Offline Plugin` vs `CallbackOnOffline`). 

However, now you've raised it, I wonder if that's not such a good idea. I think the plugin names are used as keys, in the code, etc., so changing it is probably a big mistake. I suspect I should probably back that last change out.

guusdk:
I have the same concern, but I'm not quite sure if it's valid. If memory
serves, it's the directory/jar file me that's used. I'm not completely sure
though. Perhaps it's better to retain the original names, just to be sure.

GregDThomas:
Done; https://github.com/igniterealtime/Openfire/pull/1120

PR #1121, opened 2018-07-24
OF-1582: Ensure that only the newly added listener is fired

GregDThomas:
If a plugin added a listener when the plugins were already loaded, then existing listeners were re-notified as well as the new listener.

This ensures that only the new listener is notified.


guusdk:
I'm not sure if I agree with this being an issue. The javadoc for the listener defines: 

> Event indicating that the PluginManager has finished an attemp (sic) to load new plugins and unload old plugins.

That's exactly what it appears to be doing prior to the change. 

I have written in the past logic that depended on this ("do something when new plugins are available"), which I think would be broken by this change.

guusdk:
Aaah, no, I get it now. Disregard my last comment please.

PR #1122, opened 2018-07-24
OF-1491: Drop libs that are needed only for the Ant build.

guusdk:
This should address the issue raised by @akrherz in https://github.com/igniterealtime/Openfire/pull/1056#issuecomment-407268220


akrherz:
Thank you @guusdk

PR #1124, opened 2018-07-25
OF-1497: Stream management should count keepalive pings

guusdk:


src/java/org/jivesoftware/openfire/nio/ClientConnectionHandler.java
```diff
@@ -105,7 +108,14 @@ public void sessionIdle(IoSession session, IdleStatus status) throws Exception {
                     Log.debug("ConnectionHandler: Pinging connection that has been idle: " + connection);
                 }
 
-                connection.deliver(pingRequest);
+                // OF-1497: Ensure that data sent to the client is processed through LocalClientSession, to avoid
+                // synchronisation issues with stanza counts related to Stream Management (XEP-0198)!
+                LocalClientSession ofSession = (LocalClientSession) SessionManager.getInstance().getSession( entity );
+                if (ofSession == null) {
+                    Log.warn( "Trying to ping a MINA connection that's idle, but has no corresponding Openfire session. MINA Connection: " + connection );
```
GregDThomas: Is it worth attempting to close the connection that's not paired to prevent it handing around indefinitely?

I'm not fussed either way, but just throwing it out as an idea.

Other than that, looks fine to me.

guusdk:
I honestly don't know. Maybe, but I'm unsure if that has side effects. I
logged the occurrence on warn level, to see if someone reports this
actually happening. If it does, I'd like to investigate the cause.

PR #1125, opened 2018-07-26
OF-1563: Ensure that stream management settings are not enabled on a new install

GregDThomas:
I recently noticed that whilst the "enabled if no property is present" is correct set to false, on new installs the property is set to true.

I also removed the requestFrequency (which defaults to 5 anyway) and the now redundant, commented out SPDY settings.

### DIFF
--- a/src/conf/openfire.xml
+++ b/src/conf/openfire.xml
@@ -34,25 +34,4 @@
     </network>
     -->
 
-    <!-- SPDY  Protocol is npn.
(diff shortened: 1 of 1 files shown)


guusdk:
Woah. I'm surprised that it's even in there.

guusdk:
I don't think that we'd want this, even if we do want to have stream management enabled by default again, right?

PR #1127, opened 2018-08-02
OF-1532 continue work on Linux RPM and DEB install4j media files

akrherz:
WIP, will make some commentary here as I make progress.


akrherz:
Taking the PR to move things along, more is coming.

PR #1128, opened 2018-08-13
OF-1585 Add MUCRoom Get Chat History endpoint

dbh:
Add MUC Room Get Chat History endpoint
Add logging for MUCRoomController and UserServiceController

### DIFF
--- a/src/plugins/restAPI/pom.xml
+++ b/src/plugins/restAPI/pom.xml
@@ -8,7 +8,7 @@
     </parent>
     <groupId>org.igniterealtime.openfire.plugins</groupId>
     <artifactId>restAPI</artifactId>
-    <version>1.3.5-SNAPSHOT</version>
+    <version>1.3.5</version>
     <name>Rest API Plugin</name>
     <description>Allows administration over a RESTful API.</description>
     
--- a/src/plugins/restAPI/pom.xml
+++ b/src/plugins/restAPI/pom.xml
@@ -8,7 +8,7 @@
     </parent>
     <groupId>org.igniterealtime.openfire.plugins</groupId>
     <artifactId>restAPI</artifactId>
-    <version>1.3.4-SNAPSHOT</version>
+    <version>1.3.5-SNAPSHOT</version>
     <name>Rest API Plugin</name>
     <description>Allows administration over a RESTful API.</description>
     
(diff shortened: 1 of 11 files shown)


Redor:
It looks good for me, but I'm unsure about the the logging in each function. 
I would see them more on the debug level rather instead of info. 
Or even better: that it would be possible to disable/enable logging on the rest api plugin page. 

What do you think?

akrherz:
Thanks @dbh , please also bump the `plugin.xml` and `changelog.html` so that this change could be released.

dbh:
Sure , happy to make the above changes.

dbh:
Yes needs to be tweaked . That name crept in from one of my custom plugins. I’ll update  tonight

akrherz:
@dbh sorry to be a pain, but did you bump `plugin.xml` as well?  note it uses US datestyle

GregDThomas:
Some time back I changed the release date to "tbc" for plugins with a SNAPSHOT version - to encourage people to update the release date as part of the release process. Part of the work to stop plugins being released with SNAPSHOT dependencies.

dbh:
Thanks for the explanation!

src/plugins/restAPI/pom.xml
```diff
@@ -8,7 +8,7 @@
     </parent>
     <groupId>org.igniterealtime.openfire.plugins</groupId>
     <artifactId>restAPI</artifactId>
-    <version>1.3.4-SNAPSHOT</version>
+    <version>1.3.5-SNAPSHOT</version>
```
akrherz: @GregDThomas should the version here then be 1.3.5 so that we can release this plugin or ?

dbh: Yes, I can make this change — makes sense . Thanks for the patience while I learn how the project is run and processes involved .

dbh:
Hi Wrooot.  Thanks for the heads-up.  I had not previously checked that. However, for my own testing, I dropped the newly built plugin into Openfire 4.2.3 and it worked fine, including additions to the plugin's admin page.  

After reviewing OF-1572,  I checked the REST API plugin's parent and it does refer to version 4.2.0.
```
    <parent>
        <artifactId>plugins</artifactId>
        <version>4.2.0</version>
    </parent>
```

PR #1130, opened 2018-08-13
Revert "OF-1542: Apply filtering to the user summary page"

guusdk:
This reverts (most of) commit e1ee7a3c75b8110f921ae23b038d32cdd086d757.

The changes in the original commit caused all users to be obtained from
the user provider when looking at the user-summary page. For instances
with many users, this requires to much resources and takes to long to
complete.

### DIFF
--- a/src/i18n/openfire_i18n_en.properties
+++ b/src/i18n/openfire_i18n_en.properties
@@ -2640,7 +2640,6 @@ user.search.search=Search
 user.summary.title=User Summary
 user.summary.deleted=User deleted successfully.
 user.summary.total_user=Total Users
-user.summary.filtered_user_count=Filtered Users
 user.summary.sorted=Sorted by Username
 user.summary.users_per_page=Users per page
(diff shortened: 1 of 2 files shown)


guusdk:
@GregDThomas do you see another solution to this? I'd love to retain the functionality, if possible.

If all else fails, maybe have some kind of hard-coded limit under which the filtering is applied, but above which it is not?

guusdk:
CPU's on Ignite are still burning, even with this revert. Something else must be up.

GregDThomas:
(I see the Group Chat is not responding; I wonder if it's the LDAP server itself which is causing the problem, and needs a swift kicking?)

akrherz:
@GregDThomas LDAP shouldn't be at play for ignite.  I see that groupchat is hanging on the second bosh request.  Erm

akrherz:
@GregDThomas the cpu spinning starts with the first BOSH connection

PR #1133, opened 2018-08-16
OF-1528: Allow Java 8 to be used when compiling JSP pages

GregDThomas:
This is a re-work of https://github.com/igniterealtime/Openfire/pull/1080

My poor git skills, coupled with a fairly significant amount of change in master, meant it was easier for me to do this in  a a new branch.

Note; where possible, I'm made sure that plugins depend on the 4.2.0 parent - where they depend on 4.3.0-SNAPSHOT, I've made sure that they also specify a minServerVersion of 4.3.0 as changes in the Jetty compiler mean that a plugin compiled against 4.3.0 will not work with 4.2.x (though a plugin compiled against 4.2.0 will work on 4.3.0 unless there are 4.3.0 specifics in play).

### DIFF
--- a/src/web/session-summary.jsp
+++ b/src/web/session-summary.jsp
@@ -94,100 +94,72 @@
     SessionResultFilter sessionResultFilter = SessionResultFilter.createDefaultSessionFilter();
     sessionResultFilter.setSortOrder(order);
     // Filter out the dodgy looking sessions
-    // Note; dodgy looking cast as we're stuck using Java 7 to compile JSPs - see OF-1528
-    List<ClientSession> sessions = (List<ClientSession>)(Object)sessionManager.getSessions(sessionResultFilter)
+    List<ClientSession> sessions = sessionManager.getSessions(sessionResultFilter)
         .stream()
-        .filter(new Predicate<ClientSession>() {
-            @Override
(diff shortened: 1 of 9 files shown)


akrherz:
Thanks @GregDThomas , just to confirm that the JSP changes you did in #1080 are no longer needed here?

src/web/session-summary.jsp
```diff
@@ -94,100 +94,72 @@
...
-        filter = filter.and(new Predicate<ClientSession>() {
-            @Override
-            public boolean test(final ClientSession session) {
-                return searchNode.equals("local") && session instanceof LocalClientSession;
-            }
-        });
+    if(searchNode.equals("local")) {
+        filter = filter.and(LocalClientSession.class::isInstance);
+    } else if (searchNode.equals("remote")) {
+        filter = filter.and(clientSession -> !LocalClientSession.class.isInstance(clientSession));
```
guusdk: This doesn't appear to be a one-on-one conversion from Predicate to Lambda. Is that intentional?

GregDThomas: Yes, it is. The same effect in the end, but a lot clearer as to the intent (IMHO)

PR #1136, opened 2018-08-17
OF-1592: Be less verbose when VCard updates fail expectedly.

guusdk:
When the VCard provider is read-only, there's no need to log a VCard update attempt
to the 'error' level. This commit reduces the log verbosity, and improves the error
sent back to the client.


akrherz:
rebased and awaiting CI

PR #1137, opened 2018-08-20
OF-1593 mention MAM in readme

wrooot:
This change doesn't warrant new version. So current jar can just be replaced with new one and readme will be replaced on the site for new users to see this info.


GregDThomas:
If you don’t bump the version number how do you know which version you have installed?

wrooot:
In this case it doesn't matter as only one line has changed in the readme. So everyone having the latest version won't get an update. They will get readme updated once version changes. New installs will have updated readme. And most importantly new readme will be shown on the web. I can bump the version of course. Or maybe we can wait for more changes before releasing this fix.

GregDThomas:
This is a good point; apologies.

PR #1138, opened 2018-08-22
OF-1337 OF-944: First attempt to standardize the String-intern-as-a-mutex hack.

Gugli:
For now, only applies to usernames.

It fixes some deadlocks between PrivacyListManager and RosterManager.

It is related to the two following issues
https://issues.igniterealtime.org/browse/OF-1337
https://issues.igniterealtime.org/browse/OF-944

### DIFF
--- a/src/java/org/jivesoftware/openfire/group/GroupManager.java
+++ b/src/java/org/jivesoftware/openfire/group/GroupManager.java
@@ -51,6 +51,10 @@ public class GroupManager {
         private static final GroupManager instance = new GroupManager();
     }
 
+    private static final String MUTEX_SUFFIX_GROUP = " grp";
+    private static final String MUTEX_SUFFIX_USER = " grpu";
+    private static final String MUTEX_SUFFIX_KEY = " grpk";
+    
     private static final String GROUP_COUNT_KEY = "GROUP_COUNT";
     private static final String SHARED_GROUPS_KEY = "SHARED_GROUPS";
     private static final String GROUP_NAMES_KEY = "GROUP_NAMES";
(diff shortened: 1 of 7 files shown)


GregDThomas:
+1 on what you've done so far, IMHO.

However, general comments on the code, not your changes
a) I wonder if we're ensuring synchronisation every where we should be
b) There are places where we may be fooling ourselves. For example, LockOutManager synchronises on a (possibly clustered) cache. That will only synchronise on the local node - so what's the side effect of this lack of synchronisation? 
c) Are there places we can simply use more modern collections - taking LockOutManager, for example, it may be possible to wrap the Cache with `Collections.synchronizedMap()` and the `.computeIfAbsent(...)` method.
d) I've a dislike of the `synchronised(string.intern()) {}` pattern in general (*). I wonder if we should look at introducing a new better mechanism - the answer at https://stackoverflow.com/a/28347825/359394 provides a good starting point, though I'd probably want to make changes.

(*) Much has been written on it, and I think what you've starting doing avoids much of the problems, it just leaves a bit of a nasty taste in the mouth.

guusdk:
@GregDThomas All true, and we should improve on the architecture if we can - but for now, this fix appears to prevent deadlocks that Gugli is suffering from almost daily. I'd be in favor of merging a fix like this instead of holding out for a better fix.

GregDThomas:
Yes, so am I, sorry should have been clearer!

guusdk:
@Gugli could you rebase / reword the commit to include the JIRA issue number please?

dwd:
I also think this should be merged as-is and ASAP.

Gugli:
@guusdk I changed it in the commit, but I forgot to update the PR. It is now done.

guusdk:
This change fixed the mutex that's used, but does not change the fact that the database is being queried while a mutex is locked. I don't think OF-1337 can be closed by merging this PR alone.

PR #1139, opened 2018-08-26
OF-1548: Store the FQDN in the local config file

GregDThomas:
As it's host specific, the FQDN should be stored in the local to the host config file, rather than shared across all config files. 

A couple of comments;
a) I've introduce a set (currently only one entry) of XML properties that should not be copied to the database.
b) I've deliberately changed the property name from `xmpp.fqdn` to `fqdn` because it makes more sense within the context of config file as well as making searching for legacy uses easier
c) When upgrading from the previous version, the database version is only used in non-clustered deployments, otherwise the local host name (the old mechanism) is used.

### DIFF
--- a/src/java/org/jivesoftware/database/bugfix/OF1515.java
+++ b/src/java/org/jivesoftware/database/bugfix/OF1515.java
@@ -115,7 +115,7 @@ public class OF1515
         String domain;
         try
         {
-            domain = JiveGlobals.getProperty( "xmpp.domain", JiveGlobals.getProperty( "xmpp.fqdn", InetAddress.getLocalHost().getCanonicalHostName() ) ).toLowerCase();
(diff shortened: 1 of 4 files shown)


akrherz:
I am lukewarm on the idea of changing the name of a property, but I don't write any code, so :)

GregDThomas:
... but it's not just changing the name, it's moving it's storage from the DB to the XML properties

akrherz:
@GregDThomas doesn't it change the property name from `xmpp.fqdn` to `fqdn` ?

GregDThomas:
Yes, hence my comment "not just changing the name". xmpp.fqdn would be stored in the XML as
```
<xmpp>
  <fqdn>value</fqdn>
</xmpp>
```
which seemed an unnecessary level of indirection - especially as the FQDN isn't really a property of XMPP.

PR #1140, opened 2018-08-26
OF-1589: Ensure Cache entries are non-null and serializable

GregDThomas:
Two core changes;

o The default value for `cache.allow.null` in DefaultCache is now set to `false`; this helps to ensure that NULL values are not placed in the DefaultCache, only for everything to fail once clustering is enabled. On a case-by-case basis it can be put back to true as a temporary work-around for non-clustered sites running buggy code (e.g. the original implementation of OF-1539). 
o The Cache interface is now Cache<K extends Serializable, V extends Serializable>. Again, this is to ensure that everything placed in a Cache can continue to be in that Cache once clustering is enabled (cf. OF-1480). 

A side effect of the latter has meant the code is more explicit as to the values placed in some Caches - e.g. ArrayList/HashSet (which is serializable) vs List/Set (which is not). This did turn up an interesting possible bug in RoutingTableImpl which had 
`private Cache<String, Collection<String>> usersSessions;`
containing both Sets and Lists and values.


akrherz:
@GregDThomas Sorry, will need your help rebasing this.

GregDThomas:
Sure thing, though I probably won't get a chance to look at this until late next week, at the earliest.

GregDThomas:
Rebased :)

akrherz:
Thanks, this code is above my head though to review, hopefully @guusdk can do it.

dwd:
Merge failure, but otherwise the code looks good. I believe I see how to fix the merge.

GregDThomas:
Thanks @dwd - it was next on my list to examine once I'd fixed the other PR of mine that broke the build ;)

PR #1141, opened 2018-08-26
OF-1479: Show the number of cache culls in the last three hours

GregDThomas:
This shows the number of cache culls in the last 3/6/12 hours on the cache summary screen. Note that no cull counts are shown for clustered caches as the culling is done behind the scenes.

image


akrherz:
rebased and awaiting CI

PR #1142, opened 2018-08-27
REST API Plugin update to 1.3.6

Redor:
Fixed: Important security issue, that allowed to use REST API without
authentication


dbh:
Tested on Openfire 4.2.3, singled node and clustered. Fixed issue

PR #1144, opened 2018-08-29
OF-1595 ext service disco adminconsole udp bug

guusdk:
Also updates version number to denote new release of this plugin.


guusdk:
After this is merged, and a release is done, we should update the version in the pom.xml file to be a SNAPSHOT release again (eg: 1.0.2-SNAPSHOT)

PR #1145, opened 2018-08-30
Safety checks around MUC Room Get ChatHistory.  Return 404 if room do…

dbh:
…es not exist. Check for null and empty string props on a message.


Redor:
Thy @dbh

PR #1147, opened 2018-09-04
OF-546: Move the source for "startup.jar" in to the starter module file system tree

GregDThomas:
In addition to moving the java files and the log4j file in to the right place under "starter", this also replaces the use of org.jivesoftware.util.Log with the SLF4J API

Note; it may be possible to also move the build/installer/images in to the starter/src/main/resources too, but as there are two openfire.install4j files referencing them I left it well alone for now.


akrherz:
worked on my local dev build with install4j.

PR #1149, opened 2018-09-04
OF-546: Move the source for the various JSP tags to webadmintld 

GregDThomas:


GregDThomas:
Aside: I'm not actually sure why this needs to be a separate module. Why not just bundle the whole lot in to xmppserver?

guusdk:
> Aside: I'm not actually sure why this needs to be a separate module. Why not just bundle the whole lot in to xmppserver?

If memory serves, this prevented some kind of issue when re-using the taglib in plugins. I'm not sure if that's still needed though.

PR #1159, opened 2018-09-20
Various Certificate Manager plugin changes

guusdk:


src/plugins/certificateManager/plugin.xml
```diff
@@ -5,8 +5,8 @@
     <description>Adds certificate management features.</description>
     <author>Guus der Kinderen</author>
     <version>${project.version}</version>
-    <date>11/24/2017</date>
-    <minServerVersion>4.0.0</minServerVersion>
+    <date>09/20/2018</date>
+    <minServerVersion>4.3.0 alpha</minServerVersion>
```
GregDThomas: A trivial comment, but leaving `<minServerVersion>` as `4.3.0` is sufficient, no need for the alpha/beta/etc. indicators

guusdk: That's what I thought - but it caused a big red line around the plugin in the admin console page.

GregDThomas: Oh - never noticed that before!

PR #1162, opened 2018-09-21
OF-1613: Correct links on PEP admin console page

GregDThomas:
(also required a minor tweak to the ListPager itself to protect against JavaScript NPE)


src/web/pubsub-node-summary.jsp
```diff
@@ -84,7 +84,7 @@
         filter = filter.and(node -> node.getDescription().toLowerCase().contains(searchCriteria));
     }
 
-    final ListPager<Node> listPager = new ListPager<>(request, response, nodes, filter, "searchNodeId", "searchNodeName", "searchNodeDescription");
+    final ListPager<Node> listPager = new ListPager<>(request, response, nodes, filter, "searchNodeId", "searchNodeName", "searchNodeDescription", "username");
```
guusdk: will this work if username is null (this JSP is re-used for non PEP, generic Pubsub, where no username is in play)?

GregDThomas: Yes. I did some testing on both pages, they’re working fine. 

All page transitions are handled by a JavaScript function that actually submits a hidden form. Parameters in that form that are empty are disabled to keep the URL as short as possible, so it all comes out in the wash.

guusdk:
Thanks Greg!

PR #1167, opened 2018-10-04
OF-1619 / OF-1620: Inverse 4.0.2

guusdk:


src/plugins/inverse/plugin.xml
```diff
@@ -4,7 +4,7 @@
     <name>inVerse</name>
     <description>Adds the (third-party) inVerse web client to Openfire.</description>
     <author>Guus der Kinderen</author>
-    <version>4.0.1 Release 1</version>
+    <version>4.0.2 Release 1</version>
     <date>09/20/2017</date>
```
akrherz: date need bumped here too?

PR #1171, opened 2018-10-12
OF-546: Make xmppserver more Maven-like

GregDThomas:
There are three separate commits here;

https://github.com/igniterealtime/Openfire/commit/e57ad45798a2e9452658805971ae8a4b6824230f -  moves the src/java and test/java files under the xmppserver folder structure, with related pom.xml changes

https://github.com/igniterealtime/Openfire/commit/cc5e4a06756216ba9d245dede0dbfb7fbf8128e5 - removes the webadmintld module, merging the source in to xmppserver

https://github.com/igniterealtime/Openfire/commit/e83d6867d10c206592b0e6754a2cb627d1982f71 - removes the webadmin module, merging the source in to xmppserver (essentially making it look like a standard webapp to IDEs, which should make it easier to import)


akrherz:
Sorry, this is above my pay grade for review :(  Maybe I'll see if @sco0ter has an opinion here!

guusdk:
I've updated my existing Intellij project to this. That didn't introduce any problems (after the Intellij auto-magic occurred. I was still able to build and run the sources from the command line. Debugging also worked. I later removed all Intellij project files, and re-imported the sources as a new project (based on the Maven model). Worked like a charm! 

Perhaps we could add some more changes eventually (from a Maven point of view, I dislike having the 'src' folder in the root) but those can come later.

I do have a concern regarding all of the installation scripts that we have (although I think those are primarily based on the 'distribution' module, which should be safe. The proof of the pudding is in the eating. I suggest we merge, and see if Bamboo copes.

GregDThomas:
FWIW, my current thinking of an end structure is;

build/ - unchanged - non Maven files for building OS specific packages
dbutil/ - unchanged (already in correct Maven layout) - if it's still required. Could/should it be merged with xmppserver?
distribution/ - move in to here the files from src/ that get packaged up in to the distribution - specifically assembly, bin, con, database, security, resources
documentation/ - unchanged (perhaps move documentation/dist to distribution?)
i18n/ - unchanged
starter/  - unchanged
xmppserver/ - unchanged

That just leaves the following in src:
src/javadoc/jdk18 - not sure what to do with this right now
src/spank - not sure what to do with this right now
src/plugins - move this to the top level

PR #1173, opened 2018-10-15
OF-546: Remove Ant SubDirInfoTask

guusdk:
This was used for/by the Ant-based build. Now that we've switched to Openfire,
there's no point in keeping this around.


guusdk:
That's what you get from hiding in London!

PR #1174, opened 2018-10-15
OF-1517: Don't require encoded i18n files for plugins.

guusdk:
This commit prevents the need for \u<hex><hex><hex><hex> encoded content in translation files. This change applies the fix that was done to the core in OF-1448 to the plugin structure.

The plugin assembly now copies all i18n files from the source folder to the build folder. There, the content is encoded using the native2ascii maven plugin. Finally, the encoded content is used in the plugin assembly.

All plugins to which this change applies (all plugins that have i18n files) will need to get a new release. This fix requires the new plugins to be build by Openfire 4.3.0 (which is as of yet unreleased). Runtime, there shouldn't be a need for 4.3.0.

Relevant plugins have been modified. Their version number has been increased, and the changelog modified. Their pom files have been modified to inherit from 4.3.0-SNAPSHOT (which introduces this fix). This needs to be modified to 4.3.0 (sans -SNAPSHOT) before the plugin can be released (we should not release code that has -SNAPSHOT dependencies). For this to be possible, Openfire 4.3.0
must be released first.


guusdk:
What remains to be done is to encode all i18n files from the plugins into UTF-8, replacing the unicode notation \u<hex> with a human readable character. I've done this before as part of OF-1448, but I've forgotten how I did that...

guusdk:
Insta-update:

`$ find src/plugins/ -name "*.properties" -exec native2ascii -reverse '{}' '{}' \;`

GregDThomas:
If the parent is Ipenfire 4.3.0 then the min server version also needs to be bumped to 4.3.0 in the plugin.xml too.

guusdk:
Well, I don't need 4.3.0 runtime, only during building.

GregDThomas:
Didn't the monitoring plugin have problems when built against 4.3 and running on 4.2? Ah well, if it's working for you ...

GregDThomas:
A bit late, but just realised that a 4.3.0 parent means a minJavaVersion of 1.8 ...

guusdk:
Is that strictly needed if the plugin code doesn't use any 1.8 API?

GregDThomas:
I think so, yes. The 4.3.0 parent sets the <target> for the maven-compiler-plugin to 1.8.

guusdk:
Ah, I see your point. I didn't think of that.

guusdk:
Added in #1181

PR #1176, opened 2018-10-19
OF-546: Move all the files related to distribution ...

GregDThomas:
... under the distribution folder.

Also:
Shuffle around documentation
Restore "lost" Openfire 4.2.1->4.2.3 changelog entries


GregDThomas:
@akrherz - I've tweaked the instal4j file as part of this commit as a couple of the files have moved about (and to stop shipping .bat files to UNIX and .sh files to Windows) - but I did it largely by eye and can't test it, so you will probably want to check I've not broken things too badly.

akrherz:
thanks @GregDThomas , with regards to the `changelog`, we only update it on the `4.2` branch and then once `4.3` is released, we copy over the entries to it.

Will look this all over

GregDThomas:
Ah, OK the Changelog stuff makes sense. I'd assumed I'd broken something with all the file movements!

PR #1180, opened 2018-10-23
OF-1628 / OF-1629: Update inVerse plugin to 4.0.3.

guusdk:
This commit updates the Openfire inVerse plugin to make use of the newly released version 4.0.3 of the Converse project. Simultaneously, a bug with loading resources has been fixed.


guusdk:
Note that we should not release this plugin until after it has been build with Openfire 4.3.0.

guusdk:
Replaced by #1198

PR #1182, opened 2018-10-23
OF-1630: Properly escape rank to allow support for MySQL 8.x

GregDThomas:
Tested on MySql 8, PostgreSQL 9.6

### DIFF
--- a/distribution/src/database/openfire_mysql.sql
+++ b/distribution/src/database/openfire_mysql.sql
@@ -63,9 +63,9 @@ CREATE TABLE ofRoster (
 
 CREATE TABLE ofRosterGroups (
   rosterID              BIGINT          NOT NULL,
-  rank                  TINYINT         NOT NULL,
+  `rank`                  TINYINT         NOT NULL,
   groupName             VARCHAR(255)    NOT NULL,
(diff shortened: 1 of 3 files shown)


GregDThomas:
Though worryingly, a quick test with HSQLDB is showing:

    2018.10.23 16:21:17 WARN  [Jetty-QTP-AdminConsole-39]: org.jivesoftware.openfire.roster.DefaultRosterItemProvider - Error trying to insert a new row in ofRoster
    java.sql.SQLSyntaxErrorException: user lacks privilege or object not found: rank in statement [INSERT INTO ofRosterGroups (rosterID, "rank", groupName) VALUES (?, ?, ?)]

GregDThomas:
Aha, I see the problem. As the column name is not escaped in the DDL scripts, it's created as "RANK". As it is now escaped in the query, it's looking for "rank" - which doesn't exist.

GregDThomas:
OK, this now works with PostgreSQL, MySQL and HSQLDB. The identifier "rank" is only escaped on those DBs that need it - i.e. MySQL.

GregDThomas:
Rebased

PR #1183, opened 2018-10-23
Fix embedded-db-viewer

GregDThomas:
The names of the JAR files have changed, so use a wildcard instead. 

Also use CR/LF as it's a BAT file


GregDThomas:
Re-based

PR #1184, opened 2018-10-23
OF-1631: Fix NPE when deleting roster items

GregDThomas:


xmppserver/src/main/java/org/jivesoftware/openfire/pep/PEPServiceManager.java
```diff
@@ -88,10 +88,10 @@ public PEPService getPEPService(String jid) {
             } else {
                 // lookup in database.
                 pepService = loadPEPServiceFromDB(jid);
-                
-                // always add to the cache, even if it doesn't exist. This will
-                // prevent future database lookups.
-                pepServices.put(jid, pepService);
+
+                if (pepService != null) {
```
guusdk: Now, if a PEP service isn't found for the JID, the corresponding database lookup will be done over and over again. I'm concerned that this might introduce a lot of overhead.

GregDThomas: I’ve an idea formulating in the back of my mind on how to cache negative results. I’ll see if I can get some time on Monday to put my thoughts in code ...

GregDThomas:
OK, this now allows "empty" values to be cached - may also be useful for https://github.com/igniterealtime/Openfire/pull/1069

PR #1187, opened 2018-10-24
Prevent the compilation of JSPs throwing stack traces.

GregDThomas:
a) Allow provided dependencies to be included on the CP when compiling
b) Remove xmppserver has a dependency of the maven plugin


GregDThomas:
Tested with one plugin with a JSP, so confident that this is good to go given the mess the builds are currently in.

PR #1194, opened 2018-10-26
OF-1634: Remove Jive from admin console footer.

guusdk:


wrooot:
There is also a mention in the launcher splash screen. https://issues.igniterealtime.org/browse/OF-1523

PR #1198, opened 2018-10-30
OF-1628 / OF-1629 / OF-1635 / OF-1636: Update inVerse plugin to Converse 4.0.4.

guusdk:
Updates the inVerse plugin to use Converse 4.0.4.

The Converse authors are spewing out new releases faster than we release plugins. This commit includes the changes in #1180 

Note that we should not release this plugin until after it has been build with Openfire 4.3.0.


akrherz:
I attempted to rebase this.

PR #1206, opened 2018-11-02
Allow the JingleNodes Plugin to be compiled by Maven.

GregDThomas:
~~Note 1: The assembly process does not currently include the jnsapi.jar
file because it's not a proper dependency. It's therefore necessary to
copy this manually in to the expanded lib folder for anything to run~~
Note 2: I've not actually tested this in anyway whatsoever. It does
however at least compile and install as a plugin, now.


akrherz:
@deleolajide you OK with this?

PR #1207, opened 2018-11-02
Allow the SIP Phone Plugin to be compiled by Maven.

GregDThomas:
~~Note 1: The assembly process does not currently include the JainSipAPI1.1.jar
file because it's not a proper dependency. It's therefore necessary to
copy this manually in to the expanded lib folder for anything to run~~
Note 2: I've not actually tested this in anyway whatsoever. It does
however at least compile and install as a plugin, now.


GregDThomas:
Dele; I think you said you use this plugin - I've no real idea of what it does, but it at least now appears to install as a plugin properly - can you check it out?

deleolajide:
>can you check it out?

Sure. I will run it on my dev server.

> I've no real idea of what it does

image
It manages the mapping between JIDs and SIP URIs. I use it for client SIP endpoints as well as server trunk based SIP endpoints

GregDThomas:
Ah, OK, I've got it, now. You define a mapping in the admin UI, and the plugin responds with that mapping in disco#info responses. 

Thanks!

PR #1208, opened 2018-11-02
OF-1637: Ensure unique values for full-to-bare JID mapping

guusdk:
The routing table uses a cache that maps a bare JID to all full JIDs for a user.
The collection of full JIDs should not contain duplicates. Before the change for
OF-1589, this collection was defined as a Set, but changed to a List. This commit
changes it back to a Set, guaranteeing uniqueness again.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -89,11 +89,11 @@ public class RoutingTableImpl extends BasicModule implements RoutingTable, Clust
      */
     private Cache<String, ClientRoute> anonymousUsersCache;
     /**
-     * Cache (unlimited, never expire) that holds list of connected resources of authenticated users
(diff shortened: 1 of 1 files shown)


GregDThomas:
OK, looks good, should solve the problem. I think what threw me originally was that if there was only one entry in the cache, it was (originally) stored as Arrays.asList() (https://github.com/igniterealtime/Openfire/pull/1140/files#diff-42211d9b12f65e2f7e539cc97c0308afL170) which is just odd - seems the Cache was containing both lists and sets.

Early next week I plan to re-review the whole of OF-1589 in case I messed up anything else.

Also - one question. When a session ends, why isn't it removed from the collection (be it a Set or List) - should it be?

guusdk:
It was removed, but not after another bit of code added another entry: https://github.com/igniterealtime/Openfire/blob/3b9b8d1bd63de116c28d39a39724103e2f98c68b/xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java#L705. That's how one gets left in the List, while working with Sets does give the proper end result.

PR #1213, opened 2018-11-13
Fix Copy&Paste comment error

Zyndoras:
Fixed a confusing comment in ```DefaultVCardProvider``` that was probably just copy&pasted from the method above.


guusdk:
Thanks!

PR #1221, opened 2018-11-15
Stop embedding log4j2.xml in the startup.jar

GregDThomas:
This allows admin's to configure logging by editing log4j2.xml directly
rather than edit the file embedded in startup.jar. It does require an
additional parameter to be passed when starting the server, though.


akrherz:
thanks @GregDThomas I need to review this some more to make sure the file gets marked as a config file in the distribution artifacts.

GregDThomas:
Thanks; I'm all a bit hazy as to what install4j does/needs which is why I tagged you in this PR

akrherz:
It is not install4j for the debian, mac, and rpm files. those each have specific changes needed

GregDThomas:
eh, OK thanks.

FWIW, the log4j2.xml file isn't "new" - it's been in the distribution a while, it's just that it was ignored in favour of the one inside starter.jar. Cue much head scratching & gnashing of teeth when I was editing the visible file and seeing no changes :S

guusdk:
log4j2.xml was never supposed to be embedded in any jar in the first place. Instead, it was supposed to go in the library folder, which is where it was, last time I looked. Am I missing something?

guusdk:
Ah.

GregDThomas:
A bit more background; it seemed that Log4j1 "found" the lib/log4j.xml config file by itself. 

Log4J2 will use log4j2.xml on the classpath (i.e. the one in startup.jar) unless pointed to lib/log4j2.xml explicitly.

guusdk:
Hmm, I was probably tricked by the one in startup.jar then. But, regardless, log4j (1) found it in ``lib`` because all files in there are supposed to be on the classpath. Did that change?

GregDThomas:
> all files in there are supposed to be on the classpath

I don't believe that is true. I think only starter.jar is on the classpath, and that adds all the other JAR files on to the classpath in a well defined order. https://github.com/igniterealtime/Openfire/blob/master/starter/src/main/java/org/jivesoftware/openfire/starter/ServerStarter.java#L87

openfire.bat for example doesn't define a classpath and simply specifies startup.jar - https://github.com/igniterealtime/Openfire/blob/master/distribution/src/bin/openfire.bat#L19

guusdk:
Weird. I wonder how log4j (1) was able to find its configuration then.

akrherz:
Ok, I think we are good to go here as I reviewed the installers.

PR #1223, opened 2018-11-18
Added a status function to init.d/openfire

ychaouche:
Might be useful.

```
root@app-server[192.168.100.83] ~ # /etc/init.d/openfire status
best java alternative in: /usr/lib/jvm/java-8-openjdk-amd64/jre
openfire is not running
root@app-server[192.168.100.83] ~ # /etc/init.d/openfire start
best java alternative in: /usr/lib/jvm/java-8-openjdk-amd64/jre
Starting openfire: openfire.
root@app-server[192.168.100.83] ~ # /etc/init.d/openfire status
best java alternative in: /usr/lib/jvm/java-8-openjdk-amd64/jre
openfire is running with pid 21048
root@app-server[192.168.100.83] ~ # 

```

### DIFF
--- a/build/debian/openfire.init.d
+++ b/build/debian/openfire.init.d
@@ -39,6 +39,8 @@ NAME=openfire
 DESC=openfire
 DAEMON_DIR=/usr/share/openfire
 DAEMON_LIB=${DAEMON_DIR}/lib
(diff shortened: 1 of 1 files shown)


akrherz:
thanks @ychaouche

PR #1226, opened 2018-11-20
OF-1570: Use the SLF4J->Log 4J2 bridge for logging

GregDThomas:
(for consistency with other plugins).
Also allow a normal (ish) resources folder to be used.

I say "ish" because it really should be src/main/resources (along with src/main/java, etc), but that's a change for another day ...


GregDThomas:
(needs shifting to the XML debugger plugin repo)

GregDThomas:
Closed in favour of 
https://github.com/igniterealtime/Openfire/pull/1246
https://github.com/igniterealtime/openfire-xmldebugger-plugin/pull/3

PR #1227, opened 2018-11-20
OF-1640: Better protect against rogue responses

GregDThomas:


guusdk:
The stanza that causes this appears to be invalid (missing a child element).

PR #1232, opened 2018-11-28
Add MUC invitation endpoint to restAPI

Zyndoras:
This endpoint for the restAPI plugin allows to send MUC invitations on behalf of the chatroom.
An optional reason may be provided. It does _not_ change any memberships or room affiliations.

Tested (and works) on Psi+.


Redor:
Thank you for the pull request. It would be great if you could adjust the date and increase the version, so that we could build a new version.

In the **plugins.xml** from:
`<date>08/30/2018</date>`
to:
`<date>11/28/2018</date>`

and the version in **pom.xml**
from:  `<version>1.3.7</version>`
to:  `<version>1.3.8</version>`

PR #1235, opened 2018-12-03
Avoid using Map keySet().removeIf when evict cache

aberenguel:
The DefaultCache class returns a copy of internal map.keySet() when the keySet() method is called https://github.com/igniterealtime/Openfire/blob/57bcd0f8fa7cb44dead637c79a027cf7bd07aedf/xmppserver/src/main/java/org/jivesoftware/util/cache/DefaultCache.java#L465

So, the use of keySet().removeIf does not remove the entry of the internal map of DefaultCache.

The simple solution would be to return the actual *map.keySet()* instead of *new HashSet<>(map.keySet());*, but it could cause some unpredictable side effects in other caches.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
@@ -156,6 +156,7 @@ public class GroupManager {
 
                                 if ("everybody".equals(originalValue) || "everybody".equals(newValue)) {
                                     evictCachedUserSharedGroups();
+                                    evictCachedPaginatedGroupNames();
                                 }
                             }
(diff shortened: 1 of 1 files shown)


GregDThomas:
Have you tried this with Clustering enabled? I worry that an eviction may cause a concurrent modification exception as the ClusteredCache may have different behaviour.

aberenguel:
I just commented in [OF-867](https://issues.igniterealtime.org/browse/OF-867?focusedCommentId=27402&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27402). This issue causes an inconsistent cache after a group is created.

GregDThomas:
This was covered by https://github.com/igniterealtime/Openfire/pull/1271

PR #1238, opened 2018-12-07
use right search jar file for bundling into deb/rpm

akrherz:
This is a bandaid until we fully fix how plugin jars are included.


akrherz:
more boggling needed here :(

PR #1239, opened 2018-12-07
OF-1650: Add a seperate plugin.jar when deploying artifacts

GregDThomas:
This is a work in progress, but at least it's some progress;

Changes;
1. As 4.3.0-beta is released, update the built versions to 4.3.0-beta2-SNAPSHOT
2. For core artefacts, in the master `pom.xml` define a `<distributionManagement/>` block. 

This allows (e.g.) 
`mvn deploy -pl :distribution -am`
to populate Maven repos with meta-data and the built artefacts (check `Openfire/target/repo`).

3. For plugin artefacts, in the plugins parent `pom.xml`, define a similar `<distributionManagement/>` block for similar reasons.
4. In the bookmarks plugin pom.xml (first in the list!) update the parent pom to the beta2-SNAPSHOT and deploy an additional plugin jar file.

This allows (e.g.)

`mvn deploy -pl :bookmarks -am`

Which will result in;
* Core Openfire artefacts being deployed in `Openfire/target/repo/snapshots/(etc)/openfire.`
* Standard jar and plugin jar artefacts for the bookmarks plugin to be deployed in `Openfire/target/repo/snapshots/(etc)/openfire/plugins/bookmarks`
~~* The plugin jar file for the bookmarks plugin to be deployed in `Openfire/target/repo/plugins/(etc)/openfire/plugins/(version)bookmarks` with a plugin classifier~~

Improvements probably still necessary;
~~a) The name of the plugins repo artefacts is off - `1.0.1-SNAPSHOT-bookmarks-plugin.jar` should be `bookmarks-1.0.1-SNAPSHOT-plugin.jar`~~
~~b) Move the plugin block in bookmarks/pom.xml to the plugins parent pom.xml~~
~~c) Avoid having a separate repo for plugins - use the standard releases/snapshots repo (though not sure if the extra artefact will over-write the carefully created metadata of the regular jar file)~~
~~d) Bump the parent `pom.xml` for the other plugins to beta2-SNAPSHOT so `mvn deploy` works for everything.~~


GregDThomas:
Fixed elsewhere

PR #1241, opened 2018-12-10
OF-1649: Solve race condition when adding/housekeeping users

GregDThomas:
This introduces a new class, AutoCloseableReentrantLock - that provides an AutoCloseable locking mechanism that allows the code to lock a resource (e.g. a user, room) at a class specific level.

This is now used in MultiUserChatServiceImpl to prevent a user from being created/added to a room at the same time as they are housekept.


GregDThomas:
NB. This still requires a bit more work. I'm not 100% convinced that the GC won't throw things off, as the key ("ClassName#part") is re-generated each time. Thinking of a performant way around this.

PR #1245, opened 2019-01-04
OF-1660: Generate javadoc artifacts as part of the verify phase

GregDThomas:
`mvn verify` will now generate javadoc and sources jar files too.

`mvn package` does not - as the generation can be slow which is a PITA during the dev/test cycle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ script: mvn verify -P ci
 install: true
 deploy:
   provider: script
-  script: mvn deploy --settings target/ci-tooling/maven-settings-for-openfire-plugins.xml
+  script: mvn deploy --settings target/ci-tooling/maven-settings-for-openfire-plugins.xml  -P ci
   skip_cleanup: true
   on:
     all_branches: true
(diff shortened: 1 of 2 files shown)


GregDThomas:
Note; still requires the artifact to be moved to the right place.

guusdk:
I'm not a big fan to attaching this to a Maven phase where this action is unexpected. Can't we instead use either a profile, or an explicit additional argument in the maven command?

GregDThomas:
I hear what you're saying; the default phase for packaging javadoc/source files is, well, "package" - I pushed it back to verify to cut down the dev/test cycle time. 

Does the CI build things with a specific profile? Could it do so?

guusdk:
Thank you! Moving the artifact to the right place will first require us to move it into the Maven repository. If you have time left, maybe look at [line 17](https://github.com/igniterealtime/Openfire/blob/b6c34a27bd006ac3e8a31061e316e5cabc92ad43/.travis.yml#L17)?

GregDThomas:
Well, I changed that line too, I think it's right, but may need to be merged before anything happens?

guusdk:
It's probably a good thing that Travis does not deploy anything to our Maven repo based on an unmerged PR. :)

PR #1247, opened 2019-01-09
Add ClusterManager#waitForClusteringToStart()

GregDThomas:
and ClusterManager#isSeniorClusterMemberOrNotClustered()
which are hopefully largely self-explanatory.


GregDThomas:
@guusdk you may find this useful, though it does require https://github.com/igniterealtime/openfire-hazelcast-plugin/pull/2 in order for isSeniorClusterMemberOrNotClustered() to return the correct result if clustering is still starting.

PR #1249, opened 2019-01-11
#OF-1285 Create a simple option of resetting admin password.

ma1uta:
https://issues.igniterealtime.org/browse/OF-1285

There is a new option in the openfire.xml:
```
<oneTimeAccessToken>
  <username>admin</username>
  <token>123</token>
</oneTimeAccessToken>
```
to provide temporary access to admin console with privileges of the specified in the `username` tag user.
After successful login this option automatically will be removed from the openfire.xml.
It needs enter only the token to login instead of the username and password.


akrherz:
Woohoo!  Awesome work, do you think this could get documented within the documentation found in your git repo?

guusdk:
This looks great! Can we do it without a username, and just a token? A
username implies that a particular user is used, but that's not the case.
If we internally do require a username (for the AuthToken perhaps), then
I'd prefer to use a randomly generated username that is not shown to the
end user, and deleted immediately after use, if that's possible.

Op vr 11 jan. 2019 19:37 schreef daryl herzmann <notifications@github.com:

GregDThomas:
A couple of questions, I would test but I’m offline ATM. 

1. Is an audit entry created when someone logs in using this mechanism? It should. 
2. If you create an entry with a token for a user that doesn’t exist, does it work?
3. If it does, do subsequent audit actions get correctly logged? Not sure if that requires a user, or just something that looks like a user id. 

One suggestion; Pre-populate the initial openfire.xml config file with empty values (and a comment as to their purpose), and rather than delete the XML values, set them to empty strings. That way it’s far easier to set this up.

guusdk:
I'd prefer if this is explicitly not tied to an Openfire username. It is
uncertain, or even unlikely that the natural person performing the recovery
is the same person represented by the username. By using a particular
username, it is implied that the corresponding natural person performed the
action of recovery, which is undesirable.

Op vr 11 jan. 2019 23:52 schreef GregDThomas <notifications@github.com:

ma1uta:
> Woohoo! Awesome work, do you think this could get documented within the documentation found in our git repo?

I can write the documentation of course but I don't know where :) If you point me I will do it.

GregDThomas:
Re auditing, Have a look at the JSP that edits system properties - that creates an audit entry when something is changed using an existing framework. 

Re. not specifying a user, I worry that this will break other things (e.g. who gets audited when a sys prop is modified?), or will a random user be picked instead?

ma1uta:
I excluded the specifying user and use the random username when passing with the one time token with clearing the token after first successful login.

xmppserver/src/main/java/org/jivesoftware/util/WebManager.java
```diff
@@ -163,6 +165,9 @@ public User getUser() {
                 Log.debug( "Unable to get user: no auth token on session." );
                 return null;
             }
+            if (authToken instanceof OneTimeAuthToken) {
+                return new User(authToken.getUsername(), "one time user", null, new Date(), new Date());
```
guusdk: Instead of 'one time user', I'd use "Recovery via One Time Auth Token"

guusdk:
@SimonRWaters, would you mind reviewing this change, given that you have a strong background in security? This change attempts to introduce a feature where one can gain access to the Openfire admin console, by making a modification to the configuration file (that's on disc). The idea is to use this for password recovery purposes.

GregDThomas:
OK, had some more time to look at this - overall, this is really good, thank you. 

Two small things;
1. The prompt when logging in seems a little brief (i.e. login.token, set to just "token")- how about "Please supply the time login token".
2. The user in a "regular" Security Audit Message is different from that from when the OTP is first used;
image
How about using something like OTP-<uuid> for all, so it's clear it's a OTP thing on both login and subsequent actions?

One trivial thing, more a style thing, but feel free to ignore; I'd be tempted to make OneTimeAuthToken a static inner class of AuthToken, to better how closely the two are related.

guusdk:
I played with this just now. It is great, thanks again! I have a couple of small suggestions for improvement:

- "Please supply the time login token" needs a 'dot' in the end: "Please supply the time login token."
- the logintoken.jsp should have a very clear warning that Openfire is running in password recovery mode. Maybe put exactly that text on top of the input field, in bright orange letters: "Openfire is running in password recover mode!"
- Although it is good to have an example in openfire.xml, I prefer that this example is put inside XML comment tags. This prevents the element from being up completely, which seems safer.

PR #1250, opened 2019-01-14
OF-1667: expose servlet context through API

guusdk:


xmppserver/src/main/java/org/jivesoftware/openfire/http/SessionListener.java
```diff
@@ -26,23 +28,61 @@
...
      * A connection was opened.
      *
-     * @param session the session.
-     * @param connection the connection.
+     * @param context The servlet servlet context of the BOSH request that triggered this event.
+     * @param session the session for which a new connection was opened.
+     * @param connection the connection that was just opened.
+     */
+    void connectionOpened( AsyncContext context, HttpSession session, HttpConnection connection );
+
```
GregDThomas: You've broken backward compatibility for all existing implementations here - a fairly significant problem, I's suggest. I'd strongly recommend make this a default, null, implementation, e.g. something like ...

```
default void connectionOpened( AsyncContext context, HttpSession session, HttpConnection connection ) {
  // Do nothing, override if required
}
```
Similarly for the other new methods added. 

This would be a Java 8 thing, so may not make it in to 4.2.x, but would want to see it in 4.3.x

PR #1252, opened 2019-01-15
OF-1669 don't use exact versions in db guide

wrooot:
Documentation is suggesting to use super old MySQL versions, which are not even working with the currently included connector. So maybe we should suggest using recent and supported versions of MySQL. Also removed chapter about varchar as it is related to even older version of MySQL, probably not relevant for a long time. Updated reference link to the currently used version of connector.


guusdk:
That makes sense, thank you!

PR #1253, opened 2019-01-16
OF-1671: Ensure Openfire can copy with a non-snapshot four digit version

GregDThomas:
(e.g. 1.2.3.4)


GregDThomas:
NB; I think this code is in good need of tidy up; I took a pragmatic solution to fixing this problem (!)

guusdk:
Agreed :)

PR #1257, opened 2019-01-17
OF-1670: Allow MUC sysadmins to join a password protected room (4.3 backport)

guusdk:
When a MUC room is configured to require a password upon entry, a sysadmin should be able
to join the room (without providing the password).

The above will now be the default configuration. It can be configured, through the admin
console, on a per conference service basis (which is the same context in which sysadmins
are defined).


guusdk:
This is a backport of #1255.

PR #1264, opened 2019-01-17
OF-1672: Consistently log BOSH debug messages.

guusdk:
Make sure that 'SENT' and 'RECV' are recorded in the same place.


guusdk:
Right you are. Fixed.

PR #1266, opened 2019-01-17
OF-1674: Enable MUC archive logging by default.

guusdk:
Group chat logging (message archiving) is currently disabled by default. Nowadays, the typical user will expect that it'll be possible to retrieve history from the server. We should enable logging, by default.


akrherz:
How does this change impact the "persistence" of rooms?  Will any created room now be auto persisted to the database?

guusdk:
I've looked a bit further. I can think of two types of 'history' that we store:

- The "historystrategy" cache that is in Openfire itself. Messages in here are not persisted (only kept in memory). I can't tell without further testing if the history survives a room - maybe it'd be good to explicitly delete it when a room is destroyed.
- The monitoring-plugin provided archive. This appears to add a 'tombstone' to the archive, when a room is deleted (that's from reading the code - I've not verified if it actually works). That suggests that when serving up the history of a room, we could serve up everything up until the last tombstone.

akrherz:
> The persistence of a room is something that can be toggled itself,

How is that toggled? :)

guusdk:
When a MUC owner, through a client, requests the configuration of an
existing room, there's an option to specify "room is persistent". The same
option is present when you create a new room, via XMPP.

PR #1268, opened 2019-01-18
OF-1675 clarify tar.gz zip upgrade

wrooot:


guusdk:
That's a sensible addition, thanks!

PR #1272, opened 2019-01-24
slf4j to 1.8.0-beta2 (CVE-2018-8088)

Neustradamus:
slf4j to 1.8.0-beta2 (CVE-2018-8088)
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-8088

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
         <standard-taglib.version>1.2.5</standard-taglib.version>
         <mina.version>2.0.7</mina.version>
         <bouncycastle.version>1.60</bouncycastle.version>
-        <slf4j.version>1.7.25</slf4j.version>
+        <slf4j.version>1.8.0-beta2</slf4j.version>
         <log4j.version>2.11.1</log4j.version>
     </properties>
 


GregDThomas:
The details of the CVE do not align with the title of the CVE.

The facts are:
org.slf4j.ext.EventData has a security flaw.
slf4j 1.8.0-beta2 deprecated this class
slf4j 1.8.0-beta3 will remove this class, when it is released 

IgniteRealtime does not use this class, so it's all pretty irrelevant anyway.

Neustradamus:
But for be up-to-date from maven?
I am not sure that one day there will be a final 1.8.0 version...

guusdk:
Although I really appreciate the effort you're putting into this, I believe that the risk of introducing an issue with a beta release is greater than using an older version that has a known issue that does not affect our code.

Let's upgrade to version 1.8.0 whenever it is released.

Neustradamus:
An example: https://github.com/diegonehab/luasocket/releases
But used...

guusdk:
The fact that other projects use a beta version of some library (that Openfire does not even use) is not relevant for this discussion.

I can imagine that there are times where we would prefer to use a beta instead of waiting for a proper release, for instance when it has a solution for a specific problem that we are suffering from. This will always be evaluated on a case-by-case base. For the issue at hand, the evaluation is that updating slf4j to the latest beta is not worth the risk.

PR #1278, opened 2019-01-27
OF-1687: Introduce a new SystemProperty class

GregDThomas:
~~This is still a work in progress, but though I'd see if there was general support for this idea before I progressed further.~~

The idea is that a `SystemProperty` can represent the system properties currently accessed via `JiveGlobals`, stored in `ofProperty`. By providing a single place where they are referenced (the static method `SystemProperty.getProperties()`) it should be possible to incorporate their current/default values and description on the existing System Properties page. 

I'm not proposing to replace every access of JiveGlobals with this class, instead gradually migrate over time, and as more properties are migrated the System Properties page becomes more complete. 

Creating a sysprop can be done so;
```
        final SystemProperty<Long> property = SystemProperty.Builder.ofType(Long.class)
            .setKey("this-is-a-constrained-long-key")
            .setDefaultValue(42L)
            .setMinValue(0L)
            .setMaxValue(100L)
            .build();

        assertThat(longProperty.getValue(), is(42L));
        longProperty.setValue(84L);
        assertThat(longProperty.getValue(), is(84L));
```

See the tests for examples of how to use, but as it stands the properties can be a String, Integer, Long or Duration, with a default value, description, and optional min/max values. A listener can also be supplied, which will be called whenever the property changes.

### DIFF
--- a/i18n/src/main/resources/openfire_i18n_en.properties
+++ b/i18n/src/main/resources/openfire_i18n_en.properties
@@ -1582,6 +1582,7 @@ system_property.provider.auth.className=The class to use to authenticate users
 system_property.hybridAuthProvider.primaryProvider.className=The first class the HybridAuthProvider should to use to authenticate users
 system_property.hybridAuthProvider.secondaryProvider.className=The second class the HybridAuthProvider should to use to authenticate users
 system_property.hybridAuthProvider.tertiaryProvider.className=The third class the HybridAuthProvider should to use to authenticate users
+system_property.admin.authorizedJIDs=The bare JID of every admin user for the DefaultAdminProvider
 
 # Server properties Page
 
--- a/i18n/src/main/resources/openfire_i18n_en.properties
+++ b/i18n/src/main/resources/openfire_i18n_en.properties
@@ -1562,6 +1562,27 @@ server.db_stats.time=Total Time
 server.db_stats.average_time=Avg. Time
 server.db_stats.no_queries=No queries
 
+
+#
+# System property descriptions
+#
+system_property.log.debug.enabled=Controls the output of DEBUG level logs
+system_property.log.trace.enabled=Controls the output of TRACE level logs
+system_property.passwordKey=Used by the DefaultAuthProvider to encrypt passwords. If this property is changed, it will \
+  not be possible to authenticate users until their passwords are reset.
+system_property.update.lastCheck=The last time a check was made for available updates
+system_property.update.frequency=How often to check for available updates
+system_property.update.service-enabled=Determines if the automatic check for updates is performed
+system_property.update.notify-admins=Determines if the administrators receive a message when an update is found
+system_property.update.proxy.host=The proxy to use, if any, when checking for updates
+system_property.update.proxy.port=The port on the proxy to use, or -1 if no proxy
+system_property.xmpp.session.conflict-limit=-1 to never kick off existing sessions when another session with the same \
+  full JID joins, otherwise the number of login attempts before the existing session is kicked
+system_property.provider.auth.className=The class to use to authenticate users
+system_property.hybridAuthProvider.primaryProvider.className=The first class the HybridAuthProvider should to use to authenticate users
+system_property.hybridAuthProvider.secondaryProvider.className=The second class the HybridAuthProvider should to use to authenticate users
+system_property.hybridAuthProvider.tertiaryProvider.className=The third class the HybridAuthProvider should to use to authenticate users
+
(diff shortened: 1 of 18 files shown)


akrherz:
Hi @GregDThomas , thanks for undertaking this effort.  My main concern is how this change supports the auto-generation of documentation of all properties?  Or is the thought that the admin console would house the complete listing?

On the admin console, I would love UI that showed which settings are different than the default and which settings require a restart for application.

GregDThomas:
OK, so this latest update adds a (read-only) UI. You can access it via http://localhost:9090/SystemProperties.jsp (slightly different URL, so the existing one can be left in place for now for comparison).

Example of this is as follows;
image

Is there general consensus that this a "Good Thing" (TM) and worth completing?

akrherz:
@GregDThomas I certainly think it is a good thing.  Could there be an option to filter the listing to only show values that differ from their default?  The usage case is then an admin could know which options they have changed during the configuration of Openfire.

GregDThomas:
The ‘Default Value’ drop down actually does this; the options are All/Changed/Unchanged/Unknown

akrherz:
> The ‘Default Value’ drop down actually does this

Fancy, but my US two cents would be that this functionality should be for the 'Property Value' column, as you want to know if the property value differs from the default.

GregDThomas:
Well, my thought process was along the lines of;
I can see a case (albeit contrived) where I may want to search by current value (e.g. 'which of the hybrid auth providers, primary, secondary or tertiary, is the DefaultAuthProvider). For that reason, I felt a free text box to search on current value for that column would be useful. I couldn't contrive an example where I may want to search by the default value, so use for that column for the 'it's changed from default' filter.

That said, I don't feel that strongly either way, so happy to change it if that's the general feeling. Would you propose that the 'default value' column is a free-text search instead?

akrherz:
@GregDThomas Ah, I don't think I know much with regards to UI, so as long as the functionality is there, that works for me.

GregDThomas:
OK, I think this may be more or less completed. OK, still need to actually define some more SystemProperty fields, but it's now possible to properly edit/encrypt/delete these new style properties in a sane manner.

akrherz:
@GregDThomas What happens in the case of a single property being defined twice with different settings?  I am thinking of a property set in one class and then set again in another.

GregDThomas:
A SystemProperty is guaranteed to have a unique key (an attempt to create a second property with the same key will cause an exception). The intent being that a property can only have one definition; typically a ‘public static final’ field member. 

However this does not stop code from getting/setting the same value via the existing JiveGlobals class which the SystemProperty essentially is a wrapper for.

akrherz:
Thanks, I think I have exhausted my questions on this.  Is there a Jira issue associated for this change?

GregDThomas:
No , no Jira currently. It kind of grew out of a thought experiment, I’ll raise a Jira and tidy up a few loose ends probably Friday, now.

akrherz:
@GregDThomas I lied, I have another question.  What about i18n support for property descriptions?

GregDThomas:
Hmm, a good point. Currently none, I’ll think about how to fold them in ...

akrherz:
Thanks @GregDThomas , I think what you did for i18n is what I had in mind.

GregDThomas:
I /think/ (famous last words) this is done now. I need to make some final checks, will squash all those commits down to one, and then see if people are still willing to merge it in ;)

wrooot:
Don't forget the Jira ticket ;)

GregDThomas:
There you go @guusdk ; list types can now be defined;

```
    private static final SystemProperty<List> ADMIN_JIDS = SystemProperty.Builder.ofType(List.class)
        .setCollectionType(String.class)
        .setKey("admin.authorizedJIDs")
        .setDefaultValue(Collections.emptyList())
        .setDynamic(true)
        .addListener(jids -> AdminManager.getInstance().refreshAdminAccounts())
        .build();
```

GregDThomas:
OK, I think this does pretty much all it needs to. Lists, Sets, Enums, Classes, as well as a selection of basic types (String, Integer, Long, Boolean, JID, Duration, Instant). It's fairly easy to expand with new types (Collections and Enums probably being the hardest to implement!), so unless anyone objects in the next few days I plan to merge this in sometime next week.

guusdk:
This is awesome, thanks @GregDThomas !

As discussed, we'll merge this in Master (to be included in 4.4.0), but not in the 4.3 branch.

PR #1279, opened 2019-01-29
OF-1683: Fix for XEP-0359

guusdk:
This fixes a XMPP-spec violation, introduced in Openfire 4.3.1.


guusdk:
The only way that I can see that happen, is when the code is processing an IQ response (result) where the child element could be absent. I've now added a check for this.

PR #1286, opened 2019-02-08
OF-1774 CleanupTask did not remove empty rooms in clustered environment.

vincentwlau:
CleanupTask only removed empty rooms in senior member, but not in
non-senior members.  It caused MUC rooms out of sync among members.
Also make cleanup frequency configurable with system property
"xmpp.muc.cleanupFrequency.inMinutes" (default is 60 minutes.)


GregDThomas:
Now re-based.

PR #1292, opened 2019-02-14
Add Smack integration tests to Travis CI.

guusdk:
This commit adds steps to the Travis CI build script, that:
- start up Openfire after it finished building
- pulls in Smack
- executes the Smack integration tests

PR #1293, opened 2019-02-14
Add Smack integration tests

Flowdalic:
My premature, because untested, PR, based on the work by @guusdk. This is an alternative to #1292.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 sudo: false
 language: java
+addons:
+  hostname: example.org
+  hosts:
+    - example.org
 jdk:
 - oraclejdk8
 cache:
@@ -10,7 +14,9 @@ before_install:
 - cp target/ci-tooling/maven-settings-for-openfire-plugins.xml $HOME/.m2/settings.xml
 before_script:
 - gem install travis
-script: mvn verify -P ci
+script:
+  - mvn verify -P ci
+  - TERM="dumb" ./runIntegrationTests -g
 install: true
 deploy:
   provider: script
--- a/runIntegrationTests
+++ b/runIntegrationTests
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SMACK_COMMIT_ID="ae208d6a2f608a69c54aae7cb2b22dc52524e9ff"
+GRADLE_VERSION="5.2.1"
+FORCE_CUSTOM_GRADLE=false
+
+while getopts dgs: OPTION "$@"; do
+	case $OPTION in
+		d)
+			set -x
+			;;
+		g)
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -10,7 +14,8 @@ before_install:
 - cp target/ci-tooling/maven-settings-for-openfire-plugins.xml $HOME/.m2/settings.xml
 before_script:
 - gem install travis
-script: mvn verify -P ci
+script:
+  - TERM="dumb" ./runIntegrationTests -g
```
guusdk: I like the cleanliness of this solution. Can we do this without moving the actual build (`mvn verify -P ci`) into the runIntegrationTests script?

Flowdalic: Sure, but I deliberately did put it into the `runIntegrationTests` script, so that one could `git clone openfire && ./openfire/runIntegrationTests` (Which does not work currently because of various reasons). Isn't "mvn verify" a no-op if the tests where already run and the source didn't change (it's the case with `gradle` but I am not sure about `mvn`).

GregDThomas:
I think at best `mvn verify` will only re-run tests. In practice, for reasons I've yet to figure out, `mvn verify` has a tendency to re-compile everything for no apparent reason.

.travis.yml
GregDThomas: I think at best `mvn verify` will only re-run tests. In practice, for reasons I've yet to figure out, `mvn verify` has a tendency to re-compile everything for no apparent reason.

guusdk: `mvn verify` is shorthand for "run all Maven phases up to and including 'verify'." This is the command that does the build (I don't believe we actually have anything bound to the 'verify' phase currently, so we might as well have used `mvn package`). In any case, this command is responsible to do the actual building of the project. Your comment suggests that you think that by that stage, the project is already built - that's not the case.

We could experiment with _duplicating_ the command, which generally should result in no-ops (as you describe) - but as the build process is quite elaborated, some plugins might not play nice here. I'm unsure if it's worth trying.

Flowdalic:
I have pushed a new version, which, if it works as intended, hopefully makes all of us happy :)

akrherz:
Awesome @Flowdalic , could we iterate on this to get travis green before merging?  Looks like another test needs disabled.

guusdk:
The build is failing with an error that I don't get locally. I wonder if this is because we skip the Openfire setup. @GregDThomas just pointed me at how we can setup things: https://github.com/igniterealtime/Openfire/commit/8bf8d5cbb1589361266c9783d95ea9f38fb2a909

I'll try to add that (but later, time is slipping away from me today), to see if we can improve things.

One other improvement that I'd love to have is to get a clearer output of what tests are failing and succeeding. Do you have suggestions for that?

PR #1296, opened 2019-02-19
Integration test non shallow fetch fallback

Flowdalic:


guusdk:
Thanks for this. The Travis check that was triggered by this PR actually ran through the newly added IF statement, so that provided a good self-verification.

Why do shallow checkouts not consistently work?

Flowdalic:
> Why do shallow checkouts not consistently work?

Most likely a limitation by the service hosting the git (github in your case).

PR #1298, opened 2019-02-25
OF-1682: Don't double escape XML entities

GregDThomas:
It turns out the code has double escaped XML entities in `openfire.xml` for a long time. This PR fixes that, and also tidies up the code where the preset values are translated in to JavaScript (which is where I first thought the issue lay, incorrectly as it turns out).

Note; This also updates the preset MySQL JDBC driver name (the previous one was complaining about being deprecated in the console), and sets UTF-8 as the preset charset.


GregDThomas:
@guusdk - a candidate for the 4.3.x branch?

guusdk:
With changes to the driver used, I think putting this in a patch release might cause unexpected surprised with some users (although it should generally improve things, you never know). Let's keep this for 4.4 only.

PR #1299, opened 2019-03-01
Clean up the test output -  e.printStackTrace() is rarely good!

GregDThomas:


guusdk:
Sleep deprivation averted. Thank you!

PR #1302, opened 2019-03-01
OF-1697 Update to MINA 2.0.20

GregDThomas:
What could possible go wrong?
In addition to bumping the MINA version it also required the implementation of a new null-operation methods of an interface that's been updated.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,7 +120,7 @@
         <!-- Note; the following jetty.version should be identical to the jetty.version in plugins/pom.xml -->
         <jetty.version>9.4.12.v20180830</jetty.version>
         <standard-taglib.version>1.2.5</standard-taglib.version>
-        <mina.version>2.0.7</mina.version>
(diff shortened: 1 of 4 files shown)


akrherz:
@ma1uta this change is not meant to discourage your MINA rewrite work, but as a stop-gap to get us to support Java 11.  We'd still love to see MINA replaced :)

GregDThomas:
A brief comment on the (sole) failing Integration Test; clearly Smack isn't getting the </stream:stream> tag it's expecting. I'm guessing as nothing else has changed Openfire is still sending it, with the newer version do we need to flush before closing the connection? Or something else?

akrherz:
hey @tevans , you remember all the fun in this space when we last touched MINA? :)

akrherz:
this patch is currently running at igniterealtime's Openfire.  So far so good.  Awesome work @GregDThomas

dwd:
Well, what could possibly go wrong?

PR #1304, opened 2019-03-11
Improve concurrency and performance of conversation log history.

manuelrazola:
1. Use SequenceManager to assign the next messageID in ofMucConversationLog, instead of using select count(*) as part of the insert query, to solve 2 issues:
- In a clustered environment, with concurrent transactions taking place and default tx isolation level repeatable read, it was the cause of continuous deadlocks.
- If a process to remove old conversation logs is set up, having count(*) as messageID would lead to duplicates.

2. Instead of inserting conversation logs message by message, they're stored in batches as part of a single db transaction, to improve performance.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
@@ -68,6 +68,7 @@ public class SequenceManager {
         new SequenceManager(JiveConstants.ROSTER, 5);
         new SequenceManager(JiveConstants.OFFLINE, 5);
         new SequenceManager(JiveConstants.MUC_ROOM, 5);
+        new SequenceManager(JiveConstants.MUC_MESSAGE_ID, 50);
     }
 
     /**
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -594,7 +594,7 @@ public class MultiUserChatServiceImpl implements Component, MultiUserChatService
         @Override
         public void run() {
             try {
-                logConversation();
+                logConversationBatch();
             }
             catch (final Throwable e) {
                 Log.error(LocaleUtils.getLocalizedString("admin.error"), e);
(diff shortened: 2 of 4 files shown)


xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
```diff
@@ -68,6 +68,7 @@
         new SequenceManager(JiveConstants.ROSTER, 5);
         new SequenceManager(JiveConstants.OFFLINE, 5);
         new SequenceManager(JiveConstants.MUC_ROOM, 5);
+        new SequenceManager(JiveConstants.MUC_MESSAGE_ID, 50);
```
GregDThomas: This essentially sets the maximum number of (unique) messages that can be stored in a batch to 50.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -594,24 +594,33 @@ private void checkForTimedOutUsers() {
...
         }
     }
 
-    private void logConversation() {
+    private void logConversationBatch() {
         ConversationLogEntry entry;
         boolean success;
+
+        final LinkedList<ConversationLogEntry> batchList = new LinkedList<>();
+
```
GregDThomas: The next line limits the number of statements in a batch to log_batch_size

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
manuelrazola: 50 is the number of id's to "checkout" at a time, so each openfire node reserves 50 future IDs... but if more are requested, the SequenceManager performs another query under the hood, to reserve more IDs, without duplicates. So, every time we ask for:
SequenceManager.nextID(JiveConstants.MUC_MESSAGE_ID)
we get a unique ID, even if the batch size is greater than 50.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
manuelrazola: It has been tested in a clustered environment with a batch size of 300 per node, and there are no message IDs duplication. log_batch_size can be set to a value greater than SequenceManager#blockSize without problems (see previous comment)

GregDThomas: Ah, apologieds, mis-understood what was going on.

guusdk:
This change introduces a new counter for MUC messages to be stored in the database. I fear that it "starts counting" at 1 (or zero, I didn't check). This is a problem in environments that already have logged content.

manuelrazola:
@guusdk The message ID doesn't need to be unique: actually with the previous implementation, if you erased old messages, the IDs would repeat.
However, a query run along with the migration would solve it:
`insert into ofID (idType, id) values (27, (select max(messageID) from ofMucConversationLog));`

guusdk:
I've retroactively created https://issues.igniterealtime.org/browse/OF-1790 to track this PR.

Re-using message IDs introduces problems with clients, which use the message ID to paginate through archives. Even if this is a pre-existing bug, it's probably best to avoid the issue. I'll create database update scripts with that query to correctly initialize the new sequence.

On a side-note: are you using Openfire / XMPP professionally at Ubisoft? If so, would you mind having a short one-on-one with me?

guusdk:
As an aside: I've been working on a related change, an improvement in how data is written to the database: in https://github.com/igniterealtime/Openfire/pull/1385. I'd love your take on that approach.

PR #1306, opened 2019-03-14
OF-1713 Unit tests shouldn't break because of non English system locale

B-apt:
Openfire xmppserver unit test fails if the system locale isn't English. We can configure surefire plugin to start the JVM with a specific locale via the <argLine> option, so it only impact the unit tests.

[Forum thread](https://discourse.igniterealtime.org/t/openfire-xmppserver-unitary-tests-failing-because-of-locale/84549/2)
[JIRA](https://issues.igniterealtime.org/browse/OF-1713)

### DIFF
--- a/xmppserver/pom.xml
+++ b/xmppserver/pom.xml
@@ -88,6 +88,7 @@
                     <systemPropertyVariables>
                         <log4j.configurationFile>${project.build.directory}/test-classes/log4j2-test-mvn.xml</log4j.configurationFile>
                     </systemPropertyVariables>
+                    <argLine>-Duser.language=us -Duser.country=US</argLine>
                 </configuration>
             </plugin>
 


B-apt:
Some test on SMACK failed because of a timeout :
`SEVERE: EntityCapsTest.testEntityCaps (Normal) failed: java.util.concurrent.TimeoutException: Timeout waiting for condition to become true. Timeout was 60000 ms.
`
If you think it may be related to this change let me know.

guusdk:
Thanks for your contribution! The test failure is probably unrelated to your change, as I've seen it fail in different PRs too. I've restarted the tests.

GregDThomas:
That's a flickery test; I've just restarted the build so next time should be OK. 

Also, I'm surprised you're seeing that as I'm sure we did something sane for that a while back. Ah yes - https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/test/java/org/jivesoftware/util/StringUtilsTest.java#L17

But we have seen some issues more recently with i18n, perhaps this is one of them.

B-apt:
Thank you !
One of the test which failed for me is inside SystemPropertyTest.java so setting the Locale in StringUtilsTest may not be of any effect there ?
But for the others yes it's a bit strange...

guusdk:
(let the record show that I was a minute faster in restarting that test! :-) )

PR #1308, opened 2019-03-18
OF-1687: Retain mutability in returned list of admin JIDs.

guusdk:
Prior to the OF-1687 changes, the list returned by `org.jivesoftware.openfire.admin.DefaultAdminProvider#getAdmins` was mutable. This commit restores that characteristic, which is depended on in other parts of the code.


guusdk:
That'll require more changes, I think, and might still upset third-party code. Someone gave me sad faces last week for breaking API :)

Doesn't mean we can't _also_ do that, by the way, to prepare for a later -letsmakethisimmutable- type of change.

PR #1309, opened 2019-03-20
The encrypted properties cannot be decrypted correctly if your run multiple setups

promiselaoliu:
Symptom:
(1) Run setup;
(2) Change the setup property in the openfire.xml to false and run set up again;
(3) Set a different key;
Result:
In the setup-datasource-standard.jsp page, the db username is displayed as gibberish.
You can also see some error messages in the log file, something like "[ERROR] LWCrypto encryp or descrypt failed".

Root cause:
The original logic for encryption key update is unfinished.

Solution:
Keep an old key in JiveGlobals, so that when updating the encryption key both the new key (for updating) and the old key (for retrieving) are available.

For example, the method JiveProperties.put() has both updating (updateProperty(key, value, isEncrypted);) and retrieval (JiveGlobals.clearXMLPropertyEncryptionEntry(key) -> openfireProperties.getProperty(name)) operation.
In this case, if it is during the process of updating the property encryption key, both the new key and the old key will be required.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/XMPPServer.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/XMPPServer.java
@@ -440,6 +440,10 @@ public class XMPPServer {
     }
 
     void runAutoSetup() {
+        // Setup property encryptor as early as possible so that database related properties can use it
+        JiveGlobals.setupPropertyEncryptionAlgorithm(JiveGlobals.getXMLProperty("autosetup.encryption.algorithm", "Blowfish")); // or AES
+        JiveGlobals.setupPropertyEncryptionKey(JiveGlobals.getXMLProperty("autosetup.encryption.key", null));
+
         // steps from setup-datasource-standard.jsp
         // do this first so that other changes persist
         if ("standard".equals(JiveGlobals.getXMLProperty("autosetup.database.mode"))) {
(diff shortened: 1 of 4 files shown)


promiselaoliu:
Still has some bugs, will update soon

promiselaoliu:
still has a bug when updating the key in security.xml

GregDThomas:
Hi,

Can you confirm what the problem is that this PR should fix? We can then raise a ticket to keep it all tracked. Thanks.

promiselaoliu:
Hi,

This is just the solution for the encryption bugs from the embedded Openfire (4.3.1) in our product. I have not yet searched for the issues in Jira. I have tested this solution in our product covering the below scenarios:
(1) Complete a setup;
(2) Change the setup property in the openfire.xml to false and run set up again;
(3) Run auto-setup for multiple times;
(4) After setup (either above 1 or 2), change encryption key in security.xml;
Currently there's no more changes from me. 

Problem description (Symptom):
(1) Run setup;
(2) Change the setup property in the openfire.xml to false and run set up again;
(3) Set a different Property Encryption key and go to the next page;
Result:
In the setup-datasource-standard.jsp page, the db username is displayed as gibberish.
You can also see some error messages in the log file, something like "[ERROR] LWCrypto encryp or descrypt failed".

Root cause:
The original logic for encryption key update is far from finished.

promiselaoliu:
Just fixed the decryption error when you re-setup an openfire instance that has been previously autosetup.

GregDThomas:
Thanks once more!

PR #1313, opened 2019-03-25
OF-1714: Use the correct key store type

GregDThomas:


guusdk:
Good catch! Did you actually run into this issue?

GregDThomas:
Nope; it was reported on the forums

PR #1315, opened 2019-03-25
OF-1699: Add additional clustering information

GregDThomas:
a) Admin Console now highlights when it is the only member
image

b) Message now sent when a node leaves the cluster, e.g.
`<node> has left the cluster - there is now only 1 node in the cluster`

c) Message is sent when a multi-node cluster is reformed
`<node> has joined the cluster - resilience is restored`

d) Message is sent when this node leaves the cluster (e.g. if clustering is disabled)
`The local node ('<node>') has left the cluster - this node no longer has any resilience`


xmppserver/src/main/java/org/jivesoftware/openfire/XMPPServer.java
```diff
@@ -1754,4 +1757,17 @@ public void setRemoteSessionLocator(RemoteSessionLocator remoteSessionLocator) {
     public boolean isStarted() {
         return started;
     }
+
+    public void sendMessageToAdmins(final String message) {
```
guusdk: UpdateManager has similar code. We should unify that.

GregDThomas: You know what? I thought I had. My bad. https://github.com/igniterealtime/Openfire/pull/1332

PR #1316, opened 2019-03-26
PubSubPersistenceManager - Incorrect syntax near the keyword 'LEFT'

promiselaoliu:
Fix the SQL syntax error for sqlserver and oracle.


GregDThomas:
Hi; thanks for the contribution; this looks, to me, good to merge - can you confirm you've no more changes? 

(ref: https://issues.igniterealtime.org/browse/OF-1657)

promiselaoliu:
Yes, currently there's no more changes from me.
Thank you Greg.

promiselaoliu:
Note that my solution has only covered sqlserver and oracle, because DB2 is not supported in our product.

GregDThomas:
Thanks again!

PR #1319, opened 2019-03-26
OF-1698: Correct SQL Server syntax

GregDThomas:


akrherz:
In other upgrade scripts for sqlserver (ie 19), we use the nomenclature

```sql
DROP INDEX pubsubNodeGroups.pubsubNodeGroups_idx;
```

Should we just stick to that here or does the `IF EXISTS` cause a variance?

GregDThomas:
According to https://docs.microsoft.com/en-us/sql/t-sql/statements/drop-index-transact-sql?view=sql-server-2017 the syntax is essentially

DROP INDEX [IF EXISTS] <index name> ON <table name>

so it's probably wrong in the other upgrade scripts too - wonder if it's a more recent SQL Server thing, or just that no-one ever ran the upgrade on SQL Server?

akrherz:
My reading of the docs indicates `IF EXISTS` is a 2016+ feature and that what we currently have should work
```
<drop_backward_compatible_index> ::=  
    [ owner_name. ] table_or_view_name.index_name  
```

PR #1322, opened 2019-03-26
OF-1685: Add Group summary sorting/filtering/custom page sizing

GregDThomas:
(using ListPager). Page now looks like:
image

Note; the original code used the `GroupManager#getGroups(int, int)` method to return a subset of groups - for the specific page in question. This has been replaced with `GroupManager#getGroups()` - which returns every group on every page request. On the face of it, this appears to offer poorer performance, but ...

1. The number of groups is considerably smaller than (say) the number of users
2. The groups are already cached (for 15 mins), so they will only be fetched (from DB or LDAP or ...) the "first" time the page is fetched
3. `GroupManager#getGroups()` is already called throughout the code, so any performance issues are likely to already have been discovered, and the groups may already be cached anyway.


GregDThomas:
@guusdk I'm confident the code works, just wanted a second opinion on the caching issue highlighted in the PR comment.

wrooot:
I know some users have hundreds if not thousands of groups. I wonder if their performance won't be impacted somehow.

GregDThomas:
I don't think that sort of number is an issue. 

Firstly, assuming 10,000 groups, each group is 1,000 bytes. That's ~10MB of RAM consumed in the cache - and ~10MB transferred from the LDAP server at most every 15 minutes. Almost a rounding error in memory/network usage.

Secondly, the old page used to make a call to `LdapGroupProvider#getGroupCount()` - unless cached, the LdapProvider iterated over all the groups so it could count them. So the code /used/ to fetch all the groups from LDAP just to count them.

guusdk:
This change removes CSRF checks, even though the page is still reading request parameters. Was that removal intentional?

GregDThomas:
Yes. This page only displays information A CSRF check added complexity but did not prevent anything from being changed maliciously.

guusdk:
The page displays user input like searchGroupDescription. We probably could use CSRF checks to guard against people trying to inject stuff in this page.

GregDThomas:
That would be an injection attack. That’s prevented by properly escaping the user input fields (could be accidentally done by pasting malicious input in to one of the search fields)

guusdk:
Still unsure if it would hurt to retain the CSRF check (if only for consistency - I think it's on every page?) but not a hill for me to die on.

PR #1323, opened 2019-03-27
The old password always fails validation if the admin password is not the default one "admin"

promiselaoliu:
Symptom:
(1) Complete a setup;
(2) Change the setup property in the openfire.xml to false and run set up again;
Result:
In the page to set admin password, the old password always fails validation.

Root cause:
When validating the old password, the hardcoded default password is used.
AuthFactory.authenticate("admin", "admin");

Solution:
Use real password to authenticate


GregDThomas:
Many thanks (again!) for the PR.

PR #1325, opened 2019-03-28
OF-1665: Fix issue when adding "missing" remote users to a group

GregDThomas:
Note; there can still be a considerable delay when viewing groups with users of another domain that do not exist; this delay is the value of `usermanager.remote-disco-info-timeout-seconds` (default: 1 minute).

Once viewed, their lack of existence is cached in the `Remote Users Existence` cache (default: 30 minutes) so subsequent calls are much quicker until that cache expires.

Note; this caching/timeout behaviour is unchanged from before. It's just that the code no longer uses inappropriate thread synchronisation methods.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/SystemProperty.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/SystemProperty.java
@@ -231,6 +231,8 @@ public final class SystemProperty<T> {
     private final boolean sorted;
 
     private SystemProperty(final Builder<T> builder) {
+        // Before we do anything, convert XML based provider setup to Database based
+        JiveGlobals.migrateProperty(builder.key);
         this.clazz = builder.clazz;
         this.key = builder.key;
         this.description = LocaleUtils.getLocalizedString("system_property." + key);
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -1597,6 +1597,7 @@ system_property.provider.lockout.className=The class to use to lock-out Openfire
 system_property.provider.securityAudit.className=The class to use to audit actions performed by administrators
 system_property.provider.user.className=The class to use to provide the Openfire users
 system_property.provider.vcard.className=The class to use to provide vCard handling
+system_property.usermanager.remote-disco-info-timeout-seconds=The maximum time the UserManager should wait, in seconds, for the a remote server to respond to a disco#info request to confirm the presence of a user
 
 # Server properties Page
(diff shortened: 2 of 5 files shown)


xmppserver/src/main/java/org/jivesoftware/util/SystemProperty.java
```diff
@@ -231,6 +231,8 @@ public void xmlPropertyDeleted(final String property, final Map<String, Object>
     private final boolean sorted;
 
     private SystemProperty(final Builder<T> builder) {
+        // Before we do anything, convert XML based provider setup to Database based
+        JiveGlobals.migrateProperty(builder.key);
```
guusdk: This prevents SystemProperty to be used without incident for elements that need to remain in the XML, I think. I'm not saying that's bad perse, just wondering if that's intended.

GregDThomas: Yes, intended. I can’t see any reason for properties to remain in XML and have DB values too. Many places were calling the migrateProperty method, it made sense to me to do so in the one place. 

But yes, should have made that a bit more explicit.

guusdk: My only concern is that there might be properties that _need_ to remain in the XML (database connection strings, fqdn), that once being handled as a `SystemProperty`, get unexpectedly migrated.

GregDThomas: I struggle to see a circumstance where a property would need to be in both XML and database. Which value is correct? Placing DB connection strings in the DB is of little use (in fact I recently made a change to stop that happening, as IME too many people were editing the DB values instead of the XML - https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/main/java/org/jivesoftware/openfire/XMPPServer.java#L165-L184), placing the FQDN in the database is wrong.

Where might a property be in both?

guusdk: I can't think of a property that should belong in both. I was worried that properties belonging in XML would accidentally be migrated to the database, by using the SystemProperty constructor.

GregDThomas: The SystemProperty class only supports database properties, so not sure how/why that could happen ...

guusdk: My lack of thoroughly understanding/reading the code confused this conversation.

PR #1326, opened 2019-03-28
propagate saveToDb state change to other nodes in the cluster

barbarello:
following up on 
https://discourse.igniterealtime.org/t/localmucroom-savedtodb-state-change-not-propagated-to-other-nodes/84564


wrooot:
To fix https://issues.igniterealtime.org/browse/OF-1721

GregDThomas:
Many thanks for the PR, always appreciated!

PR #1327, opened 2019-04-11
OF-1728: Add the ability to view/delete the contents of caches

GregDThomas:
Clicking on the cache name in the summary page takes the user through to a new cache details screen looking something like ...

image

Usage is fairly self-explanatory.


xmppserver/src/main/webapp/WEB-INF/web.xml
```diff
@@ -155,6 +160,12 @@
         <url-pattern>/server-properties.jsp</url-pattern>
     </servlet-mapping>
 
+    <servlet-mapping>
+        <servlet-name>SystemCacheDetails</servlet-name>
+        <!-- NB. .jsp extension is only required so filters are applied -->
```
guusdk: That includes the Auth filter, right? This new servlet cannot be used by non-Admins?

PR #1329, opened 2019-04-11
OF-1731: http-bind bug fixes + config tweak

danielhams:
Got the OK from my employer to feed back some changes we maintained. This is the first of a few sets of changes, so if you want some other way to discuss or review, don't hesitate.

I realise you maybe want this split up with some discussion / rejection of each - that's fine - I realise your time is precious!

Issues these are to fix were seen by using an http-bind load tester (I can't release) along with using network parameters under linux to simulate 3G connections with out-of-order, duplication etc.

Contents:

(1) Unsynchronized use of packetsToSend - we've seen nullpointer exceptions in the server logs where the use of "packetsToSend.add" without a guard means that where they are being removed it's possible to get stale references

(2) sendPendingPackets - the logic there seems to only flush packets from the first collection removed - the tweak loops around all collections that are pending

(3)  onTimeout - we've seen cases of dangling threads + connections as the timeout handling code itself wasn't timing out - thus the tweak to use "deliverOnTimeout"

(4) overactivity bounds difference - we've had http-bind sessions closed prematurely due to a difference in the bounds checking - this change brings the bounds into line and resolves the premature closes we've seen

(5) A tweak to allow customising how often the http-session cleaner is run (we like to run it often)

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSessionManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSessionManager.java
@@ -100,6 +100,7 @@ public class HttpSessionManager {
         final int maxClientPoolSize = JiveGlobals.getIntProperty( "xmpp.client.processing.threads", 8 );
         final int maxPoolSize = JiveGlobals.getIntProperty("xmpp.httpbind.worker.threads", maxClientPoolSize );
         final int keepAlive = JiveGlobals.getIntProperty( "xmpp.httpbind.worker.timeout", 60 );
+        final int sessionCleanupCheck = JiveGlobals.getIntProperty("xmpp.httpbind.worker.cleanupcheck", 30);
 
         sendPacketPool = new ThreadPoolExecutor(getCorePoolSize(maxPoolSize), maxPoolSize, keepAlive, TimeUnit.SECONDS,
                 new LinkedBlockingQueue<Runnable>(), // unbounded task queue
@@ -110,7 +111,7 @@ public class HttpSessionManager {
 
         // Periodically check for Sessions that need a cleanup.
         inactivityTask = new HttpSessionReaper();
-        TaskEngine.getInstance().schedule( inactivityTask, 30 * JiveConstants.SECOND, 30 * JiveConstants.SECOND );
+        TaskEngine.getInstance().schedule( inactivityTask, 30 * JiveConstants.SECOND, sessionCleanupCheck * JiveConstants.SECOND);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


akrherz:
Thank you for this.  Just to clarify, you see these issues with 4.3.2 release and have a means to test these with current master builds of Openfire?

GregDThomas:
I’ve yet to check this, but my first thought is: Awesome, thank you!

danielhams:
I'm afraid the version these issues were discovered against is an unofficial branch of a snapshot we took to work against.

I don't currently have an environment with 4.3.2 - and our application wouldn't run on it without some tweaks anyway -> (11 VMs in the load testing lab).

Perhaps a way forward would be that I provide the negative - I find a way to get a load tester running against master HEAD and try and show some stack traces for (1) -> (4)?

Does that sound reasonable?

akrherz:
@danielhams I am not a hard code developer here, just have lots of history with this :)  The issue is that what sometimes appears to be bugs in code that does things like flushing packets are sometimes workarounds to unknown problems found with lower level libraries, like MINA.  So we fix a bug in openfire and then the strange behaviour comes back :)

Having said that, openfire now more closely uses current releases of low level libraries, so maybe they are all bug free now! 

testing master would be awesome, but we are certainly grateful for these types of PRs!

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSessionManager.java
```diff
@@ -110,7 +111,7 @@ public void start() {
 
         // Periodically check for Sessions that need a cleanup.
         inactivityTask = new HttpSessionReaper();
-        TaskEngine.getInstance().schedule( inactivityTask, 30 * JiveConstants.SECOND, 30 * JiveConstants.SECOND );
+        TaskEngine.getInstance().schedule( inactivityTask, 30 * JiveConstants.SECOND, sessionCleanupCheck * JiveConstants.SECOND);
```
GregDThomas: Trivial, but I think the second param should be `sessionCleanupCheck * JiveConstants.SECOND` too.

danielhams: Certainly could be. I think that's the initial delay before beginning the periodic looping - we left it at 30 to "give time for things to settle". No probs if you'd prefer it made the same.

guusdk:
Thanks for this, @danielhams ! I thank you and your employer for taking the time to feed back improvements upstream! If you'd like to explore more/other options to collaborate, we'd be open to discuss that.

I retroactively made an issue for these changes in our issue tracker, mostly because we use that to generate changelogs when a new version is released. You can find the issue at https://issues.igniterealtime.org/browse/OF-1731

Neustradamus:
@danielhams: Thanks a lot for your changes!
Linked to: https://issues.igniterealtime.org/browse/OF-626
- XEP-0124: Bidirectional-streams Over Synchronous HTTP (BOSH) : https://xmpp.org/extensions/xep-0124.html
- XEP-0206: XMPP Over BOSH : https://xmpp.org/extensions/xep-0206.html

There are other changes for be perfect?

guusdk:
@Neustradamus please do no abuse this contribution to push your own agenda.

PR #1330, opened 2019-04-12
DB Connection Pool Tweaks

danielhams:
General comment: The (new) connection pool updates for recent Openfire are a welcome addition! - thanks for those changes.

One thing we (my society) perhaps do a little differently to a stock install is that we don't use the "failure looping" of the DB connection pool - we found that this means unfair queueing occurs under burst load.

We use a configuration where maxRetries = 0 with a larger maxWaitTime (2.5 seconds).

To help us get there, we made a couple of tweaks - contents:

(1) getConnection logic change - ensure sleep on failure only occurs if we are about to go around the loop and ask for another connection (when maxRetries = 0, this change stops unnecessary sleeps)

(2) poolConfig - if the minimum connections asked for in the db config are set - tweak the poolConfig so that this is allowed (via maxIdleConnections)

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/database/DbConnectionManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/database/DbConnectionManager.java
@@ -124,6 +124,7 @@ public class DbConnectionManager {
         Integer maxRetries = JiveGlobals.getXMLProperty(SETTING_DATABASE_MAX_RETRIES, 10);
         Integer retryWait = JiveGlobals.getXMLProperty(SETTING_DATABASE_RETRY_DELAY, 250); // milliseconds
         SQLException lastException = null;
+        boolean loopIfNoConnection = false;
         do {
             try {
                 Connection con = connectionProvider.getConnection();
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/database/DbConnectionManager.java
```diff
@@ -145,13 +146,16 @@ public static Connection getConnection() throws SQLException {
...
-            } catch (Exception e) {
-                // Ignored, the thread was interrupted while waiting, so no need to log either
-            }
             currentRetryCount++;
-        } while (currentRetryCount <= maxRetries);
+            loopIfNoConnection = currentRetryCount <= maxRetries;
+            if (loopIfNoConnection) {
+                try {
+                    Thread.sleep(retryWait);
+                } catch (Exception e) {
```
GregDThomas: You're catching an InterruptedException here, which really means you should re-interrupt the thread. Something like

```
try {
  Thread.sleep(retryWait);
} catch (InterruptedException ex) {
  Log.info("Interrupted whilst waiting", e);
  Thread.currentThread().interrupt(); 
}
```

guusdk: Hmmm... shouldn't we consider that an interrupt here is likely an attempt to break the retry loop completely. Maybe we should throw an SQLException?

danielhams: Fair point - I'll have to read up a little to be honest and be sure what is good practice.

danielhams: So if I'm thinking along the right lines (@GregDThomas thanks for the link) - I'm leaning towards a combination of your two suggestions - maybe something like this:

```
try {
  Thread.sleep(retryWait);
} catch (InterruptedException ex) {
  String msg = "Interrupted waiting for DB Connection";
  Log.info(msg,ex);
  Thread.currentThread().interrupt(); 
  throw new SQLException(msg,ex);
}
```
We don't just swallow the interruption, but we fail fast from the point of view of the caller, too.

GregDThomas: Yup, concur with that :)

GregDThomas:
Thanks again @danielhams - the JIRA I've raised to track this is https://issues.igniterealtime.org/browse/OF-1733

PR #1331, opened 2019-04-12
Expire any admin console sessions if user admin rights revoked

danielhams:
How to reproduce on HEAD:

(1) Create a second administrator account "dan"
(2) Using web browser 1 log in as "dan" and open up the sessions list in the admin console
(3) User web browser 2 log in as other admin, and revoke the admin permission for "dan"
(4) Dan can still refresh and use the admin console

I've tested the regular paths with admin console users - but in all honesty I haven't tested the "auth token" path (although the code seems OK to my eyes).

How is the auth-token path exercised? (Clustering maybe?)


GregDThomas:
The one-time token is created by editing the `conf/openfire.xml` file; it provides a way for an administrator to login when they've forgotten the admin password.

GregDThomas:
Many thanks! I've tracked this as https://issues.igniterealtime.org/browse/OF-1734

PR #1333, opened 2019-04-14
OF-1718: Update to Mina 2.1.1 to resolve issues with MINA compression filter

GregDThomas:
Also take the opportunity to tidy up the pom.xml for the MINA artifacts

(need to make one final check before merging)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,7 +120,7 @@
         <!-- Note; the following jetty.version should be identical to the jetty.version in plugins/pom.xml -->
         <jetty.version>9.4.12.v20180830</jetty.version>
         <standard-taglib.version>1.2.5</standard-taglib.version>
-        <mina.version>2.0.20</mina.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -308,24 +308,16 @@
...
-            <!--<goals>-->
-            <!--<goal>findbugs</goal>-->
-            <!--</goals>-->
-            <!--</execution>-->
-            <!--</executions>-->
-            <!--</plugin>-->
-
+            <!--
+            The various MINA dependencies rely on artifacts of type 'bundle'
+            so add support to Maven
```
guusdk: Maybe keep a reference to https://issues.apache.org/jira/browse/DIRMINA-919 or some better link that explains what MINA does?

GregDThomas: Added a comment with a link to some information about OSGi bundles.

PR #1335, opened 2019-04-15
OF-1772 Enable configurable kick reasons for admin console + timeouts

danielhams:
Enable JiveGlobals kick reasons when kicking via admin console or via automated timeout.

We find these useful (audit reasons) - our application always fills in a kick reason, this change allows us to do the same for the Openfire server itself.

Not sure if the keys fit OK with existing JiveGlobals hierarchy

* admin.mucRoom.timeoutKickReason
* admin.mucRoom.consoleKickReason


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -564,14 +564,15 @@ private void checkForTimedOutUsers() {
                     continue;
                 }
                 if (user.getLastPacketTime() < deadline) {
+                    String timeoutKickReason = JiveGlobals.getProperty("admin.mucRoom.timeoutKickReason", null);
```
guusdk: It's probably fine to add a non-null default here.

PR #1336, opened 2019-04-15
OF-1770 Include nickname in MUCEventListener.occupantLeft

danielhams:
This one breaks API contracts unfortunately. As far as I can tell the OF "MonitoringPlugin" at least has impact from this change (GroupConversationInterceptor).

Background: We have a chunk of functionality we hang off join and leave events to MUC rooms. Having the nickname lets us skip some of the DB queries or "remembering" we otherwise would need to do in building our audit events.


danielhams:
Aha! great, someone has skills in the "keeping an API compatible" department :-)

I'll take a look.

GregDThomas:
PS. 
> someone has skills in the "keeping an API compatible" department

That includes you, now :)

akrherz:
Sorry @danielhams , could you look into rebasing this?  If you are busy, I can give it a swing.

danielhams:
@GregDThomas @akrherz 

Thank you gents!

PR #1337, opened 2019-04-17
httpbind: Improved Logging / Diagnostics

danielhams:
Didn't include this in the original httpbind pull request - it's a lot of noise for what are only logging changes (and I wanted to clean up our version so it was less verbose). Hope that's not too annoying.

Contents:

(1) Removes some unnecessary string concatenation on the hot path in HttpBindServlet.onDataAvailable/Read/WritePossible, HttpSession.checkOveractivity

(2) Improve lifecycle logging - use streamID where applicable to streamline issue diagnosis

Possible modification: One change to this approach might be to place the lifecycle logging ("session ## created etc") behind a different logging switch in JiveGlobals to the raw RECV and SEND. I leave this for the feedback cycle.

Thanks for your time, patience and help @guusdk and @GregDThomas - this is the last of my pull requests (for now at least .-) ).

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindServlet.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindServlet.java
@@ -340,7 +340,9 @@ public class HttpBindServlet extends HttpServlet {
 
         @Override
         public void onDataAvailable() throws IOException {
-            Log.trace("Data is available to be read from [" + remoteAddress + "]");
+            if( Log.isTraceEnabled() ) {
+                Log.trace("Data is available to be read from [" + remoteAddress + "]");
+            }
 
             final ServletInputStream inputStream = context.getRequest().getInputStream();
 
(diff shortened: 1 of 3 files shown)


akrherz:
Absolutely no need to apologize for these pull requests.  We are so thankful for them!

GregDThomas:
I'll echo what @akrherz says - PRs are always welcome!

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindServlet.java
```diff
@@ -340,7 +340,9 @@ protected static String getRemoteAddress(AsyncContext context)
 
         @Override
         public void onDataAvailable() throws IOException {
-            Log.trace("Data is available to be read from [" + remoteAddress + "]");
+            if( Log.isTraceEnabled() ) {
+                Log.trace("Data is available to be read from [" + remoteAddress + "]");
+            }
```
GregDThomas: For future reference, these three lines (and similar) can be done in one using SLF4J placeholders; 
`Log.trace("Data is available to be read from [{}]", remoteAddress);`
Won't ask you to change the PR on this basis, but just something to keep in mind for the future.

PR #1338, opened 2019-04-19
OF-1705: Update NodeID in ClientRoute

guusdk:
The ClientRoute class contains a reference to the server NodeID. This value must be updated when joining/leaving a cluster.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -42,6 +42,8 @@ import org.xmpp.packet.*;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
 /**
  * Routing table that stores routes to client sessions, outgoing server sessions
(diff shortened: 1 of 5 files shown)


xmppserver/src/main/java/org/jivesoftware/util/cache/CacheUtil.java
```diff
@@ -371,6 +372,49 @@
...
+            final K key = entry.getKey();
+
+            final Lock lock = CacheFactory.getLock( key, cache );
+            try
+            {
+                lock.lock();
+
+                final V modifiedValue = mapper.apply( entry.getValue() );
+                if ( modifiedValue.equals( entry.getValue() ) )
+                {
```
GregDThomas: Don't you want `if ( !modifiedValue.equals(...) )` ??

xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
```diff
@@ -1077,6 +1083,10 @@ public void leftCluster() {
             final NodeID nodeID = server.getNodeID();
```
GregDThomas: Recent changes made sure that the nodeId never changes. For performance reasons, you could wrap this whole block in `if(!nodeId.equals(defaultNodeID))`

xmppserver/src/main/java/org/jivesoftware/util/cache/CacheUtil.java
guusdk: :man_facepalming:

xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
guusdk: That's a great suggestion. I've added such a check.

PR #1339, opened 2019-04-19
OF-1737: Fix Java Spec version comparison.

guusdk:
Java 11 should be considered newer than Java 1.8. The change in this commit changes the comparison to only use the last numeric value.


akrherz:
rebased

akrherz:
this one got a different error, it seems

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project xmppserver: Fatal error compiling: error: option --add-modules not allowed with target 1.8

guusdk:
That's fixed by 6b079f4

akrherz:
rebased.

PR #1341, opened 2019-04-19
OF-1383: Have CI use both Java 8 and 11.

guusdk:
This commit causes Travis to compile the project not only with Java 8, but also with Java 11.


akrherz:
we should pull this in and then rebase the others, I think :)

PR #1342, opened 2019-04-19
OF-1738: Lose the nanosecond precision for Java 9+ compatibility

GregDThomas:


akrherz:
Thanks @GregDThomas , I am curious though about why this test needs the update, but not when the next test runs with the default value set?  Does setting via a default value mean that it is not round-tripped via a string?

GregDThomas:
Yes; the default value is never persisted in the database - it's the value that is used if there is nothing in the database (or the value in the database can't be converted to the appropriate type). 

On reflection, I'm slightly worried that this behaviour breaks the principle of least surprise. I think I'll re-work the PR to ensure that the SystemProperty class loses the precision, rather than the test.

akrherz:
thanks @GregDThomas , that was my concern as well.

PR #1344, opened 2019-04-20
Fix the IT by allowing Smack to compile with warnings

GregDThomas:


GregDThomas:
@guusdk @Flowdalic I'll be the first to admit that this is a bit hacky, but I'll merge it anyway. It does finally allow the the Openfire/Smack integration tests to run (and pass!) with Java 11. In the longer term I'd hope that the changes will be in Smack itself (PR submitted to cover off the finding of tests) but in the short term this is really just a work-around.

PR #1346, opened 2019-04-24
OF-1699 revisited

GregDThomas:
During testing of split brain recovery (https://github.com/igniterealtime/openfire-hazelcast-plugin/issues/23) it became clear that the cluster monitoring introduced in https://github.com/igniterealtime/Openfire/pull/1332 had two (minor) flaws;
* When the junior node loses network connectivity, it attempts to send a message as it is now senior; this message can be stored in the DB, which may also be inaccessible. This all takes some time (a few seconds) to fail. To avoid slowing down the (synchronous) raising of ClusterEventListener events the sending of messages is now done in a background thread.
* Ensure existing remote node names are cached when the local node joins the cluster, as well as when new remote nodes join.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/XMPPServer.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/XMPPServer.java
@@ -238,6 +238,17 @@ public class XMPPServer {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


GregDThomas:
@guusdk tagging you on this as it subtly changes the behaviour of UpdateManager too - this will make the admin messages it sends be sent asynchronously, but don't think that's any cause for worry..

GregDThomas:
Now re-based

PR #1347, opened 2019-04-24
 OF-188: disallow admin console to create longer subjects than 100.

mnsuccess:
The subject column of ofMucRoom table is limited to 100 characters, yet the admin console allows one to set a subject longer than that and does not indicate a failure to save a larger than 100 character subject to the database.


wrooot:
This pull request has history of your previous commits. It shouldn't have that.

PR #1348, opened 2019-04-24
OF-188: disallow admin console to create longer subjects than 100 characters.

mnsuccess:
The subject column of ofMucRoom table is limited to 100 characters, yet the admin console allows one to set a subject longer than that and does not indicate a failure to save a larger than 100 character subject to the database.


mnsuccess:
I created a new one here

wrooot:
See the review comment above. I suggest changing translate string to muc.room.edit.form.valid_hint_subject_too_long
I would change message to "Subject is too long. Maximum 100 characters."

wrooot:
I have also tried to build your fork and test this and it seems that it doesn't even show the warning and always saves only first 100 chars of the entered text (i think maxlength="100 does that). This is not very untuitive. One can enter Topic with 200 chars, hit Save and think that it was saved, when in fact it only saved first 100 chars without a warning.

wrooot:
Again. I would rename that translation string to say muc.room.edit.form.valid_hint_subject_too_long
longer is not informative for translators. Is it for longer subjects?

wrooot:
Ok, thanks for this patch. Merging now.

PR #1350, opened 2019-04-26
OF-1647: Rename the legacy WEB-INF/lib folder on startup

GregDThomas:
NB. If it can't be renamed (ten attempts are made) this will shut down Openfire


akrherz:
Could it be moved out of the way instead?  maybe somebody somewhere would have something valuable in that folder for some reason and not have backups of course :)

GregDThomas:
This is a very good point. Now renamed from `WEB-INF/lib` to `WEB-INF/lib-backup1' (or WEB-INF/lib-backup2` or ...)

PR #1375, opened 2019-04-30
Disable test that fails regularly

GregDThomas:


GregDThomas:
Thanks for merging. Tried re-building with this patch three times or so on Travis, all passed OK, hopefully this is the last one.

That said, am slightly worried about the increasing list of disabled tests, but not sure if Smack or Openfire is at fault.

PR #1376, opened 2019-05-01
OF-258: Allow nicknames of MUC members to be reserved in the admin console

mnsuccess:
Let configure the nickname of the new room member when adding a new member

### DIFF
--- a/xmppserver/src/main/webapp/muc-room-affiliations.jsp
+++ b/xmppserver/src/main/webapp/muc-room-affiliations.jsp
@@ -48,8 +48,10 @@
     JID roomJID = new JID(ParamUtils.getParameter(request,"roomJID"));
     String affiliation = ParamUtils.getParameter(request,"affiliation");
     String userJID = ParamUtils.getParameter(request,"userJID");
+    String nickName = ParamUtils.getParameter(request,"nickName");
     String roomName = roomJID.getNode();
     String[] groupNames = ParamUtils.getParameters(request, "groupNames");
+    final String affName = ParamUtils.getStringParameter(request, "affName", "");
 
     boolean add = request.getParameter("add") != null;
     boolean addsuccess = request.getParameter("addsuccess") != null;
@@ -115,6 +117,15 @@
                     Element item = frag.addElement("item");
                     item.addAttribute("affiliation", affiliation);
                     item.addAttribute("jid", memberJID);
+                    if(nickName != null){
+                        item.addAttribute("nick", nickName);
+                    }else{
+                        //set the name of the user as the nickname of the member
+                        if(userJID != null){
+                            String username = JID.escapeNode(userJID.substring(0, userJID.indexOf('@')));
+                            item.addAttribute("nick", username);
+                        }
+                    }
                 }
                 // Send the IQ packet that will modify the room's configuration
                 room.getIQAdminHandler().handleIQ(iq, room.getRole());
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/webapp/muc-room-affiliations.jsp
```diff
@@ -115,6 +117,15 @@
                     Element item = frag.addElement("item");
                     item.addAttribute("affiliation", affiliation);
                     item.addAttribute("jid", memberJID);
+                    if(nickName != null){
+                        item.addAttribute("nick", nickName);
+                    }else{
+                        //set the name of the user as the nickname of the member
+                        if(userJID != null){
+                            String username = JID.escapeNode(userJID.substring(0, userJID.indexOf('@')));
```
guusdk: Instead of parsing the string directly, it might be wiser to create a JID instance using the string, and get it's node.

wrooot:
I think it should appear on this page? Built Manasse's fork and get this: 
image

mnsuccess:
Yes
the entry for the nickname is only displayed if the Member option is selected

wrooot:
I see now. Also, i have noticed that Type selection moved to the top. Not sure if this is still intuitive. Well, i know what this means and how to add a user of a required type. But wonder how new users would see that. 
image

wrooot:
I guess this page is a bit of a mess anyway.

wrooot:
Maybe type should even go to the top of this segment. And also Add User maybe should be renamed to Select User. Because Add button is what actually does the add action. But that's a minor thing. Or we can just leave it as it is and wait for feedback :) @guusdk what do you think?

guusdk:
I think I should not be involved in UI/UX discussions :-D

If we can make the page a little nicer without to much effort, I'm all for it. If it takes to much effort (or discussion), then I suggest we accept this PR for the functionality it adds, and create a new issue for the desired user interface changes.

@wrooot did you happen to test if adding a nickname actually works?

wrooot:
Ok, i will create a separate ticket for UX.

I have tested this now. When i enter a nickname and press Add, it adds the user and you can see nickname besides it in parenthesis. But it doesn't preserve the case. I have put TTest and it produced: test@server (ttest).

Have tried to login with this user in Spark. Nickname is not showing in his vCard. In Gajim it also automatically sets test (same as username) for a nickname.

If you don't provide a nickname it uses username for it, not the actual Name value set for this user in Admin Console.

guusdk:
> Have tried to login with this user in Spark. Nickname is not showing in his vCard. In Gajim it also automatically sets test (same as username) for a nickname.

I think that is to be expected.

What @mnsuccess achieved in this PR is to make a reservation for a nickname: after reservation, only the member (as identified by its JID) is allowed to make use of that nickname, in that specific room.

Note that it does not require this value to be the global 'nickname' of the user. The user, when entering the room, can even use a completely different nickname.

Apart from adding it as a value in a bookmark (but I think creating/modifying a bookmark for this purpose is a bad idea), I don't think we can nudge/force the client to use this particular nickname in that room.

guusdk:
> When adding a single member, you can also assign a nickname for a user in this room. If not set, user's Name will automatically be assigned (or username, if Name is not set)

Adding a description is good, but I think your suggested text does not describe what the value means. It's not about _assigning_ a nickname to a user, but rather to _reserve_ one.

wrooot:
I understand now what this ticket was meant to accomplish. I think we should change its name and description to avoid same confusion like i had. Because i guess those who have voted for it or will read about it in the changelog will understand it the same way like me.

wrooot:
I have tested this again and nickname reservation works. Other user is not allowed to set same nickname as reserved. Case is not important. But it would be useful if it would use Name field instead of username for the default value.

wrooot:
Have proposed renaming here https://issues.igniterealtime.org/browse/OF-258 i have also edited my description comment above (reserve instead of assign).

guusdk:
> But it would be useful if it would use Name field instead of username for the default value.

I'm not sure if that's true, as many clients will use the username as the default nickname. Although it'd be _nicer_ to use the "name", I think there'll be less confusion if "username" is used.

guusdk:
Additionally: if we use the JID-node, instead of the human readable name of the user, we can apply it to _all_ members/JIDs, even the ones that are not local users.

wrooot:
I keep forgetting that this is only a reservation and actual use case of this is limited :) Say one would have users with usernames like FIN_lt_namelastname, but with a name Name Lastname. Reserving username as a nickname would be useful as nobody would want to enter such nickname anyway. This would help only against impersonation by using username of another person. But i also doubt it would be very helpful either. I feel this is not what ticket voters were hoping to get :) Anyway, i retract my comment about the name usage and will correct my description comment above.

wrooot:
I believe now the only thing missing to merge this is adding a description https://github.com/igniterealtime/Openfire/pull/1376#pullrequestreview-233016666

xmppserver/src/main/webapp/muc-room-affiliations.jsp
wrooot: @guusdk is this resolved?

guusdk: Yes.

PR #1379, opened 2019-05-09
OF-1771 Show software version of connection

mnsuccess:


mnsuccess:
I first built code for reading the response and displaying data in the admin console.

the code evaluates every stanza arrives on the Session StanzaHandle class.  if the stanza is an IQ and contains a child element that is <query xmls='jabber:iq:version'>, I saved it on Current Session and I can display it on Console admin...
 I tested it manually by sending a query result to the server with Spark .....
The Second step that I had trouble is where and how  I can send requests to clients....

guusdk:
I think you ment to create a "draft pull request" instead of a normal pull request, didn't you?

xmppserver/src/main/java/org/jivesoftware/openfire/net/StanzaHandler.java
```diff
@@ -325,6 +325,12 @@ private IQ getIQ(Element doc) {
         Element query = doc.element("query");
         if (query != null && "jabber:iq:roster".equals(query.getNamespaceURI())) {
             return new Roster(doc);
+        }else if (query != null && "jabber:iq:version".equals(query.getNamespaceURI())) {
```
guusdk: There are two issues here:

1. if a result does not include one of the fields, the 'getStringValue()' will cause a NullPointerException
2. If a result includes other fields, we're not registering them.

I suggest you replace this with a loop that iterates over all elements of query.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -96,6 +95,12 @@
      */
     private final Map<String, Object> sessionData = new HashMap<>();
 
+    /**
+     * SoftwareVersion (XEP-0092) temporary data. All data stored in this <code>Map</code> disapear when session
```
guusdk: Although this is correct, you're making this comment on a property of the session that disappears. I'd not mention this, as it's confusing. I suggest with replacing this with something like

> Software Version (XEP-0092) data as obtained from the peer on this connection.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -519,4 +524,35 @@ public String getCipherSuiteName() {
     public final Locale getLanguage() {
         return language;
     }
+
+    /**
+     * Retrieves SoftwareVersion data. This method gives access to temporary Software Version data only. You can
+     * Please see
+     * {@link #setSoftwareVersionData(String, String)}  description for more details.
```
guusdk: You're over-using javadoc annotations here. For simple getters and setters, there is no need to explicitly refer to it with \@link and \@see.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -519,4 +524,35 @@ public String getCipherSuiteName() {
...
+     *
+     * @param key a <code>String</code> value of stored data ID.
+     * @return a <code>Map<String,String></code> value of data for given key.
+     * @see #setSoftwareVersionData(String, String)
+     */
+    @Override
+    public Map<String, String> getSoftwareVersion() {
+        synchronized (softwareVersionData) {
```
guusdk: When a thread-safe Map implementation is used, then there's no need to synchronize! I suggest switching from HashMap to ConcurrentHashMap and remove synchronization!

xmppserver/src/main/java/org/jivesoftware/openfire/session/Session.java
```diff
@@ -47,7 +48,7 @@
     int STATUS_CLOSED = -1;
     int STATUS_CONNECTED = 1;
     int STATUS_AUTHENTICATED = 3;
-
+
```
guusdk: Try to avoid having white-space changes in a commit. There's a couple of these in this PR.

xmppserver/src/main/webapp/session-details.jsp
```diff
@@ -356,6 +356,21 @@
...
             } %>
+        </td>
+    </tr>
+    <tr>
+        <td class="c1">
+            <fmt:message key="session.details.software_version"/>
+        </td>
+        <td>
+                <%
+                try { %>
```
guusdk: I think that having a new row for each attribute looks a lot nicer. :)

xmppserver/src/main/webapp/session-details.jsp
```diff
@@ -356,6 +356,21 @@
...
+                <%= currentSess.getSoftwareVersion().get("name") %> / <%= currentSess.getSoftwareVersion().get("version") %> / <%= currentSess.getSoftwareVersion().get("os") %>
```
guusdk: It is important to properly escape the values that are being displayed, to avoid malicious user to be able to inject things in the admin console pages!

mnsuccess:
> I think you ment to create a "draft pull request" instead of a normal pull request, didn't you?

No, I made just a normal pull request

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -519,4 +524,35 @@ public String getCipherSuiteName() {
     public final Locale getLanguage() {
         return language;
     }
+
+    /**
+     * Retrieves SoftwareVersion data. This method gives access to temporary Software Version data only. You can
```
wrooot: I would add space between Software and Version. Also second sentence seems to end abruptly "You can..".

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -519,4 +524,35 @@ public String getCipherSuiteName() {
...
+     */
+    @Override
+    public Map<String, String> getSoftwareVersion() {
+        synchronized (softwareVersionData) {
+            return softwareVersionData;
+        }
+    }
+
+    /**
+     * Saves given session data. Data are saved to temporary storage only and are accessible during
```
wrooot: Data is saved to temporary storage only and is accessible during

xmppserver/src/main/java/org/jivesoftware/openfire/session/Session.java
```diff
@@ -216,4 +217,11 @@
      * @return The language for the session.
      */
     Locale getLanguage();
+    
+    /**
+     * Returns all software version as reported by the peer on this connection,
```
wrooot: Returns all Software Version data as..

guusdk:
> No, I made just a normal pull request

I know - but as the changes are not ready, I think you should have created a 'draft pr' :)

mnsuccess:
> > No, I made just a normal pull request
> 
> I know - but as the changes are not ready, I think you should have created a 'draft pr' :)

 Let me create another PR with draft

guusdk:
> Let me create another PR with draft

Let's not! That will cause all feedback to be lost!

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -96,6 +95,11 @@
      */
     private final Map<String, Object> sessionData = new HashMap<>();
 
+    /**
+     * Software Version (XEP-0092) data as obtained from the peer on this connection.
+     */
+    private Map<String, String> softwareVersionData = new HashMap<>();
+
```
GregDThomas: I dislike the idea of of maintaining this software version information as a map. XEP-0092 highlights three items; name, version, OS. I'd much rather see a POJO class containing these three items explicitly (perhaps with the IQ or parent Element passed in to the constructor) rather than a Map<String, String> which could hold just about anything.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -519,4 +523,25 @@ public String getCipherSuiteName() {
...
+    /**
+     * Retrieves Software Version data. This method gives access to temporary Software Version data only. 
+     * @param key a <code>String</code> value of stored data ID.
+     * @return a <code>Map<String,String></code> value of data .
+     */
+    @Override
+    public Map<String, String> getSoftwareVersion() {
+return softwareVersionData; 
+    }
```
GregDThomas: Exposing the Map exposes implementation details. It should be possible to swap out such implementation details with affecting the API (e.g. the XEP-0092 POJO I mentioned earlier). 

IMHO, as a rule of thumb, it's almost never appropriate to expose Maps via accessor methods.

xmppserver/src/main/webapp/session-details.jsp
```diff
@@ -356,7 +356,33 @@
...
+        try { 
+            if(currentSess.getSoftwareVersion()!= null ){ %>
+                <tr>
+                    <td class="c1">
+                        <fmt:message key="session.details.software_version"/>
+                    </td>
+                </tr>
+                <% 
+                for (Map.Entry<String, String> entry : currentSess.getSoftwareVersion().entrySet()){ %>
+                    <tr>
```
GregDThomas: The order of the entrySet is undefined, so different sessions may show the same information in a different order, which would be counter-intuitive.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
guusdk: I don't agree. Even though the XEP defines just three elements, there's no predicting what third-party implementations might put in there. This way we display everything advertised.

xmppserver/src/main/webapp/session-details.jsp
guusdk: I don't think it's to much of a problem, but maybe we can order them alphabetically?

guusdk:
I've tried running the code. When I connect a client and look at its session in the admin console, I see this:
image

This seems wrong: the last row in the table is malformed.

guusdk:
The current code logs every incoming stanza on the 'INFO' level. This adds _many_ lines like these in the log files. Although it's convenient for development purposes, the logging should not be included when we merge this PR.

```
2019.05.15 10:10:42 INFO  [socket_c2s-thread-3]: org.jivesoftware.openfire.net.StanzaHandler - STANZA INCOMMING <iq to='pubsub.laptop-guus' id='SdNQ2-22' type='get'><query xmlns='http://jabber.org/protoco
2019.05.15 10:10:42 INFO  [socket_c2s-thread-2]: org.jivesoftware.openfire.net.StanzaHandler - STANZA INCOMMING <iq to='laptop-guus' id='SdNQ2-24' type='get'><query xmlns='http://jabber.org/protocol/disco
2019.05.15 10:10:42 INFO  [socket_c2s-thread-2]: org.jivesoftware.openfire.net.StanzaHandler - STANZA INCOMMING <iq to='proxy.laptop-guus' id='SdNQ2-26' type='get'><query xmlns='http://jabber.org/protocol
2019.05.15 10:10:42 INFO  [socket_c2s-thread-2]: org.jivesoftware.openfire.net.StanzaHandler - STANZA INCOMMING <iq to='search.laptop-guus' id='SdNQ2-28' type='get'><query xmlns='http://jabbe
```

xmppserver/src/main/java/org/jivesoftware/openfire/session/SoftwareVersionManager.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public void anonymousSessionDestroyed(Session session) {
+
+    }
+
+    @Override
+    public void resourceBound(Session session) {
+        try {
+            IQ reply = new IQ(IQ.Type.get);
+            reply.setID(session.getStreamID().getID());
```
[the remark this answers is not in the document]
guusdk: the "ID" of a stanza must be a unique (random) number for every new stanza. Here, the value is set to the ID of the stream. Stream ID and stanza IDs are two unrelated concepts, and should not be mixed. 

As this event/method is invoked just once for a session, it will probably not cause duplicate IDs (so there's no issue) but, it's illogical to use the stream ID here. I suggest simply not explicitly setting the ID, which should cause the implementation to generate a random value.

guusdk:
It turns out that I gave you bad advise. The server cannot send requests to the client, immediately after the 'resource bound' event has occurred. Instead, it needs to wait for a 'session established' event, for which Openfire currently has no event listener.

As a work-around, I suggest we asynchronously wait a period of time (5 seconds) after the resource bound event has happened. By then, the session should have been established.

This is an ugly hack, that needs to be improved on, but that's something that we can do outside of the scope of this PR.

You can apply the ugly workaround by replacing your code in SoftwareVersionManager with this bit:

```java
@Override
public void resourceBound(Session session) {
    // The server should not send requests to the client before the client session
    // has been established (see IQSessionEstablishmentHandler). Sadly, Openfire
    // does not provide a hook for this. For now, the resource bound event is
    // used instead (which should be immediately followed by session establishment).
    TaskEngine.getInstance().submit( () -> {
        try {
            Thread.sleep(5000); // Let time pass for the session establishment to have occurred.

            IQ versionRequest = new IQ(IQ.Type.get);
            versionRequest.setTo(session.getAddress());
            versionRequest.setFrom(session.getServerName());
            versionRequest.setChildElement("query", "jabber:iq:version");
            session.process(versionRequest);
        } catch (Exception e) {
            Log.error("Exception while trying to query a client for its software version.", e);;
        }
    } );
}```

**Comment by guusdk** [comment]
With the work-around, I can see the software version as reported by the Swift client. It still does not look very nice. Maybe we should use a new table for "Software Version" (and give it the same nice header as "Session Details" table has):

image

File: xmppserver/src/main/webapp/session-details.jsp [hunk 15]
```diff
@@ -359,14 +359,20 @@
             } %>
         </td>
     </tr>
-    <% 
+</tbody>
+</table>
+<table cellpadding="0" cellspacing="0" border="0" width="100%">
```
guusdk: I'd only show this table when getSoftwareVersion != null.

mnsuccess: I already updated it on https://github.com/igniterealtime/Openfire/pull/1379/commits/76f25c6dc41d4730be9b554717b2080b5421b62a

```

File: xmppserver/src/main/java/org/jivesoftware/openfire/XMPPServer.java [hunk 16]
```diff
@@ -769,6 +770,7 @@ private void loadModules() {
         loadModule(OfflineMessageStrategy.class.getName());
         loadModule(OfflineMessageStore.class.getName());
         loadModule(VCardManager.class.getName());
+        loadModule(SoftwareVersionManager.class.getName());
```
guusdk: This new module is not a 'core' module. Please move it to the bottom of the list of the 'standard' modules.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -64,7 +63,7 @@
      * The Address this session is authenticated as.
      */
     private JID address;
-
+
```
guusdk: This is an odd whitespace to add. Let's remove it.

xmppserver/src/main/webapp/session-details.jsp
```diff
@@ -356,11 +357,50 @@
...
+                                <td class="c1">
+                                    <%= StringUtils.escapeHTMLTags(entry.getKey().substring(0, 1).toUpperCase()+""+entry.getKey().substring(1)) %>:
+                                </td>
+                                <td>
+                                    <%= StringUtils.escapeHTMLTags(entry.getValue())%>
+                                </td>
+                            </tr>
+                        <% 
+                        }
+                } catch (java.net.UnknownHostException e) { %>
```
guusdk: I don't think that `UnknownHostException` can be thrown in this block, can it? It should be safe to remove the `try/catch` here

xmppserver/src/main/webapp/session-details.jsp
```diff
@@ -365,7 +365,7 @@
 
 <br>
 <%  // Show Software Version if there is :
-    if (currentSess.getSoftwareVersion().size() > 1) {
+    if (currentSess.getSoftwareVersion().size() >=1) {
```
guusdk: Ah, good catch. Maybe use `!currentSess.getSoftwareVersion().isEmpty()`? That does the same, but is a bit clearer.

guusdk:
Excellent work, @mnsuccess ! Let's do the same for components and servers!

PR #1380, opened 2019-05-17
OF-1771: Make software version have a default value

guusdk:
This is intended to make the OF-1771 API change compatible with older/existing versions of the clustering plugin.


mightymop:
thx i will test it

guusdk:
Thanks for the confirmation!

PR #1383, opened 2019-05-22
OF-1771 Show SoftwareVersion for s2s connection and component

mnsuccess:


xmppserver/src/main/java/org/jivesoftware/openfire/event/ServerSessionEventDispatcher.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright (C) 2004-2008 Jive Software. All rights reserved.
```
guusdk: Please change the copyright to this year, and "Ignite Realtime Foundation"

xmppserver/src/main/java/org/jivesoftware/openfire/event/ServerSessionEventDispatcher.java
```diff
@@ -0,0 +1,133 @@
...
+
+        /**
+         * A session was destroyed
+         */
+        session_destroyed,
+        
+        /**
+         * An anonymous session was created.
+         */
+        anonymous_session_created,
```
guusdk: For S2S, I don't think we have anonymous sessions, nor resource binding. We might have some other events, but for now, lets just use "session_created" and "session_destroyed".

xmppserver/src/main/java/org/jivesoftware/openfire/event/ServerSessionEventListener.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (C) 2004-2008 Jive Software. All rights reserved.
```
guusdk: Change copyright to 2019 Ignite Realtime Foundation

xmppserver/src/main/java/org/jivesoftware/openfire/event/ServerSessionEventListener.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param session the authenticated session of a non anonymous user.
+     */
+    void sessionDestroyed( Session session );
+
+    /**
+     * Notification event indicating that an anonymous user has authenticated with the server.
+     *
+     * @param session the authenticated session of an anonymous user.
+     */
+    void anonymousSessionCreated( Session session );
```
guusdk: I'd only put in 'sessionCreated' and 'sessionDestroyed' for this first version.

xmppserver/src/main/java/org/jivesoftware/openfire/event/ServerSessionEventDispatcher.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.jivesoftware.openfire.session.Session;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Dispatches session events (s2s). Each event has a {@link EventType type}
+ *
+ * @author Manasse Ngudia
```
guusdk: If you want, you can add your e-mail address too. That makes it easier for people to contact you - assuming that this is something you're interested in. :)

guusdk:
Looks good! Now, adapt the Openfire code in such a way that these new events are dispatched when servers connect/disconnect!

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
```diff
@@ -395,7 +398,7 @@ public LocalIncomingServerSession createIncomingSession(XMPPPacketReader reader)
                 if ("db".equals(doc.getNamespacePrefix()) && "result".equals(doc.getName())) {
                     String hostname = doc.attributeValue("from");
                     String recipient = doc.attributeValue("to");
-                    Log.debug("ServerDialback: RS - Validating remote domain for incoming session from {} to {}", hostname, recipient);
+                    Log.debug("ServeDialback: RS - Validating remote domain for incoming session from {} to {}", hostname, recipient);
```
guusdk: This should remain "ServerDialback" (unsure why you removed the 'r'?)

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
```diff
@@ -266,6 +267,8 @@ public LocalOutgoingServerSession createOutgoingSession(String localDomain, Stri
                     // Set the hostname as the address of the session
                     session.setAddress(new JID(null, remoteDomain, null));
                     log.debug( "Successfully created new outgoing session!" );
+                    // After the session has been created, inform all listeners as well.
+                    ServerSessionEventDispatcher.dispatchEvent(session, ServerSessionEventDispatcher.EventType.session_created);
```
guusdk: This looks OK - but don't forget to also dispatch an event when the session is destroyed!

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
wrooot: This hasn't been fixed yet.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketReader.java
```diff
@@ -366,6 +367,18 @@ private IQ getIQ(Element doc) {
         Element query = doc.element("query");
         if (query != null && "jabber:iq:roster".equals(query.getNamespaceURI())) {
             return new Roster(doc);
+        }else if (query != null && "jabber:iq:version".equals(query.getNamespaceURI())) {
```
guusdk: I think (but have not tested) that this has an unexpected side-effect. It will not only process the software version response from the server (you might want to add a check to only process result stanzas, you're currently also processing 'request' stanzas), but you will also process the software version responses from any user on that server.

I suspect that, when user X on our server asks for the software version of the client of user Y on the other server, this bit of code will cause the software version of the other server to be reported as the client of user Y. That's probably not good. :)

Please test for this, and make changes if needed.

xmppserver/src/main/webapp/server-session-details.jsp
```diff
@@ -117,7 +122,44 @@
...
+                        <% 
+                        }
+                    %>
+                </tbody>
+            </table>
+        </div>
+    <%  } 
+    } catch (Exception e) { 
+       Log.error(e.getMessage(), e);
+    } %>
```
guusdk: It is good to log, but you probably want to show some kind of message to the user looking at the webpage too.

xmppserver/src/main/webapp/server-session-details.jsp
```diff
@@ -104,7 +109,7 @@
             if (!inSessions.isEmpty()) { %>
                 <%= inSessions.get(0).getHostAddress() %>
                 /
-                <%= inSessions.get(0).getHostName() %>
+                <%= inSessions.get(0).getHostName() %>
```
guusdk: There's a mysterious whitespace being added to the end of this line. Probably unneeded. :)

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketReader.java
mnsuccess: > I suspect that, when user X on our server asks for the software version of the client of user Y on the other server, this bit of code will cause the software version of the other server to be reported as the client of user Y. That's probably not good. :)

Yes, I tested and had that side-effect, the software version of the other server has reported as the client of the user :)
So what Should I do? maybe SocketReader it's not a good place for that code

guusdk: I think this is a good place for that code. You should, however, make the code smarter: it must check if the sender of the version result is the remote server (`from="example.org"`) and not a user on the remote server (`from="userY@example.org"`) or any other entity on that server (from="`mycomponent.example.org`").

guusdk: Your update is a good improvement, but still has one issue: the answer of an external component (`component.example.org`) on the remote server `example.org` would still be interpreted as the response of the server itself.

guusdk: I think that the current state of the code still has this problem: When an external component on a remote server responds with Software Version information, then Openfire, based on this code, probably will record the software version as being reported by the remote server, instead of the remote external component.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketReader.java
```diff
@@ -370,7 +370,7 @@ private IQ getIQ(Element doc) {
             return new Roster(doc);
         }else if (query != null && "jabber:iq:version".equals(query.getNamespaceURI())) {
             IQ iq = new IQ(doc);
-            if (iq.getType().equals(IQ.Type.result) && !iq.getFrom().toString().contains("@") && !(StringUtils.countMatches(iq.getFrom().toString(),".")>1)){
+            if (iq.getType().equals(IQ.Type.result) && !iq.getFrom().toString().contains("@") && iq.getFrom().equals(session.getAddress())){
```
mnsuccess: @guusdk I think the JID of the remote server. Maybe it's already available in the connection field for the current session connection. I can access it with` session.getAddress()`

guusdk: Did you test if it does? If so, then excellent!

guusdk:
This looks like it's pretty much finished, I think! Please remove the unused import statements, but other than that, you might want to consider unmarking this PR as a draft!

guusdk:
Excellent! Thank you!

guusdk:
After this was deployed, I have seen software versions in the admin console for clients, but not for servers :-(

image

PR #1385, opened 2019-05-31
OF-1788 / OF-1200: Record MUC messages in archive immediately.

guusdk:
This commit builds on OF-1200, which is now applied to the core Openfire MUC implementation (OF-1200 applied the mechanism to the Monitoring plugin only).

Additionally, the mechanism has been made generic, and has recieved additional features.


guusdk:
@GregDThomas I've got no idea what your availability / willingness to hurt your brain with my code is these days. :) Is there a chance of you reviewing this? If not, I'm tempted to push the big green button immediately and test this on Ignite.

GregDThomas:
I’ll have a look this weekend !

xmppserver/src/main/java/org/jivesoftware/openfire/archive/Archiver.java
```diff
@@ -0,0 +1,267 @@
...
+     * This method is intended to be used to determine if it's safe to construct an answer (based on database
+     * content) to a request for archived data. Such response should only be generated after all data that was
+     * queued before the request arrived has been written to the database.
+     *
+     * This method performs a cluster-wide check, unlike {@link #availabilityETAOnLocalNode(Instant)}.
+     *
+     * @param instant The timestamp of the data that should be available (cannot be null).
+     * @return A period of time, zero when the requested data is already available.
+     */
+    public Duration availabilityETA( Instant instant )
```
GregDThomas: This is possibly my mis-understanding of what's going on, but this method appears to return either 
a) The ETA on this node, if not zero, or
b) The max ETA on all the other nodes

Should it simply be the max ETA of all nodes? In a two-node cluster, if this node's ETA is 10 seconds, the other nodes ETA is 20 seconds, the overall ETA is 20 seconds, not 10? Should simply be a case of appending localNodeETA to the collection of other node ETAs

xmppserver/src/test/java/org/jivesoftware/openfire/archive/ArchiverTest.java
```diff
@@ -0,0 +1,343 @@
...
+        final Thread thread = new Thread( archiver );
+
+        try
+        {
+            // Execute system under test.
+            thread.start();
+            final Instant start = Instant.now();
+            archiver.archive( 1 );
+            archiver.archive( 2 );
+            if ( Duration.between( start, Instant.now() ).compareTo( gracePeriod ) >= 0 )
```
GregDThomas: I'd be very tempted to put a Thread.sleep(1) in your dummy archiver to ensure the clock ticks at least once between each saved item to help avoid this test becoming flickery.

xmppserver/src/main/java/org/jivesoftware/openfire/archive/Archiver.java
guusdk: Your description is correct. I did this on purpose: when it can be determined that the ETA is non-zero, the archive request can't be answered in any case. Adding the overhead of a cluster call, only to get the ETA potentially better, seems like overkill.

GregDThomas:
A couple of minor comments, not 100% sure if they're valid, but may be of use.

xmppserver/src/main/java/org/jivesoftware/openfire/archive/Archiver.java
GregDThomas: Hmm, depending on the usage of this method would it better be a boolean, isAllDataArchived() ? Currently a non-zero Duration could well be wrong, which will no doubt be misleading one day in the future, if not right now.

guusdk: That's what the initial signature of this method was. 😁 My concern was that that needed an arbitrary timeout for subsequent checks, which could lead to excessive amounts of checks, depending configuration and load. The ETA intends to replace that with a somewhat OKish estimation. Would putting more emphasis on the "this is an estimate, at best" nature of the return value put your mind at ease?

GregDThomas: 😁 
Yes, I think that should cover it 😄

guusdk: I've now added such a comment.

xmppserver/src/test/java/org/jivesoftware/openfire/archive/ArchiverTest.java
guusdk: I'm happy to do so, but I'm not sure how that would make it less flickery?

guusdk: I'm going to merge as-is. If tests turn out to be flickery, I do expect a surplus of I-told-you-so. :)

PR #1386, opened 2019-06-03
OF-1786 Update MINA to 2.1.3

akrherz:


akrherz:
mavencentral website does not show this artifact is available, but it worked locally and I assume the website has some caching at play.

PR #1387, opened 2019-06-04
OF-1790: Initialize ofID sequences.

guusdk:
An earlier change introduced a new ofID-based sequence used for messageID, when writing MUC messages to the database. This commit initializes the counter that's used to the highest value that has already been logged.


guusdk:
This PR intends to improve the change in https://github.com/igniterealtime/Openfire/pull/1304

GregDThomas:
Don't forget to bump the expected DB version to 30 so Openfire actually runs the script!

manuelrazola:
To avoid a "Column 'id' cannot be null" when there are no messsages in ofMucConversationLog, we could use coalesce function:

`insert into ofID (idType, id) values (27, (select coalesce(max(messageID), 0) from ofMucConversationLog));`

guusdk:
Ah, yes. Now to figure out if that's supported on all databases that Openfire supports...

PR #1393, opened 2019-06-16
OF-1796: Fix presence probe after subscription from remote user

guusdk:
Old code assumed that users were local users, while they need not be.


guusdk:
This needs a proper test, which is easier after a merge (because I can then deploy nightlies to a remote domain).

PR #1394, opened 2019-06-17
OF-1797 Accelerate loading MUC rooms

abe-ta:
Hi, I tested Openfire 4.2.3 with 200,000 MUC rooms. And I found out loading MUC rooms can be very slow in proportion to the number of rooms. In my case, it needed over 1 hour to start Openfire.
I identified the cause in MUCPersistenceManager. (This problem still exists in Openfire 4.3 or later)
- After fetching MUC rooms from `ofmucroom` table, each `LocalMUCRoom` loads its default properties from `ofmucserviceprop` table with MUCPersistenceManager. This process calls `SELECT` SQL about 200,000 * 12 times (the 12 is the number of LocalMUCRoom.xxx fields). In the end, the more MUC rooms the more time needed.

I tweaked MUCPersistenceManager.getProperty just a little bit. By this fix, Openfire startup time become about 90 seconds in 200,000 MUC room.

Please review.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
@@ -1113,11 +1113,12 @@ public class MUCPersistenceManager {
      * @return the property value specified by name.
      */
     public static String getProperty(String subdomain, String name) {    	
-        final MUCServiceProperties newProps = new MUCServiceProperties(subdomain);
-        final MUCServiceProperties oldProps = propertyMaps.putIfAbsent(subdomain, newProps);
-        if (oldProps != null) {
-            return oldProps.get(name);
+        final MUCServiceProperties cachedProps = propertyMaps.get(subdomain);
+        if (cachedProps != null) {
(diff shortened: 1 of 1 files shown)


akrherz:
Thanks, this seems good to me, but will let others weigh in.

guusdk:
Thank you for your suggestion. That, indeed, would dramatically improve things in large instances.

Can we further improve your change by applying thread safety, by replacing the entire method content with this?

```java
        final MUCServiceProperties props = propertyMaps.computeIfAbsent( subdomain, MUCServiceProperties::new );
        return props.get(name);
```

guusdk:
Upon closer inspection, I've found various other places in the same class that would benefit from your change. Would you mind applying it to all applicable places? I'd advice you to look at all invocations of the constructor of `MUCServiceProperties`

abe-ta:
Thanks for your great proposal! I fixed, and there was no problem launching Openfire and loading each settings.

guusdk:
Excellent! Did you verify that you still have the performance benefits from your original code?

abe-ta:
Openfire startup time remained 90 seconds. Other than `getProperty(String subdomain, String name)` seem to be not called from anywhere.

PR #1401, opened 2019-06-21
remove un-needed search.jar copy in build scripts

akrherz:
maven already provides `search.jar` thanks to some work @GregDThomas did.


akrherz:
Taking this over the unrelated travis failure and will test with ignite to verify it works.

PR #1406, opened 2019-06-25
OF-1806 -- Add support for XEP-0232

mnsuccess:
service discovery responses can include detailed information about the software application that powers to the remotes Entities.


guusdk:
Don't forget to add support for Openfire exposing this data, if it is being asked!

guusdk:
There is quite a bit of code duplication in this PR: there are many lines that do the same thing. This is generally considered bad practice. I'd suggest that you try and remove similar snippets of code, by replacing them with one method.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -513,6 +514,46 @@ public static DataForm getFirstDataForm(Set<DataForm> dataForms){
...
+     * responsed by the peer for the Software information request Service Discovery (XEP-0232)
+     * @param Element query represented on the response of the peer
+     * @param LocalSession session represented the LocalSession with peer
+     */
+    public static void setSoftwareVersionDataFormFromDiscoInfo(Element query ,LocalSession session){
+        Boolean containDisco = new Boolean("false");
+        Boolean typeformDataSoftwareInfo = new Boolean("false");
+        List<Element> elements = query.elements();
+        if (elements.isEmpty() && elements!= null && elements.size()>0 && session != null){
+            for (Element element : elements){
```
guusdk: You need to make sure that the `elements` collection is not null in a for-each loop like this:

    for (Object element : elements )

It is perfectly OK to use a collection that is empty (the loop will simply be skipped).

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -701,7 +740,33 @@ public DataForm getExtendedInfo(String name, String node, JID senderJID) {
...
+                        final FormField fieldSoftware = dataFormSoftwareVersion.addField();
+                        fieldSoftware.setVariable("software");
+                        fieldSoftware.addValue(AdminConsole.getAppName());
+
+                        final FormField fieldSoftwareVersion = dataFormSoftwareVersion.addField();
+                        fieldSoftwareVersion.setVariable("software_version");
+                        fieldSoftwareVersion.addValue(AdminConsole.getVersionString());
+
                         final Set<DataForm> dataForms = new HashSet<>();
+                        dataForms.add(dataFormSoftwareVersion);
```
guusdk: Whenever a new feature is built, it is good to include a "kill-switch": a property that, if changed, disables the functionality. This allows people to work around any issues that are introduced by the new feature (if any).

Please include a new `SystemProperty` of type Boolean, that by default is true, that is used to see if the new XEP-0232 dataform should be created/returned.

guusdk:
Looks good! Trivial things to change:

- let's not use upper-case letters for properties in the i18n files. I don't know if there's a reason for not doing that, but the convention is to not do it. Lets stick to that convention.
- there's an unintentional double space in the translated message, between the words 'to' and 'not' in: "Set to false to  not allow"

If you're happy with the PR, please remove the 'draft' status so that we can merge it!

guusdk:
Thanks @mnsuccess !

PR #1408, opened 2019-06-27
Updated spanish translation

drencrom:


wrooot:
Thanks. https://issues.igniterealtime.org/browse/OF-1807

akrherz:
Thank you, I don't see a problem with getting this into the 4.4.0 release.

i18n/src/main/resources/openfire_i18n_es.properties
```diff
@@ -2674,13 +3159,63 @@ client.connections.settings.ping.enable=Enviar una solicitud XMPP Ping a los cli
...
-connection-mode.unspecified=unspecified
+connection-type.webadmin=consola de administración
+connection-type.component=componente externo
+connection-type.connection-manager=administrador de conexiones
+connection-type.unspecified=no especificado
+connection-mode.plain=texto plano (con STARTSSL)
+connection-mode.legacy=cifrado (legacy-mode)
+connection-mode.unspecified=no especificado
+
+# DNS SRV Records check
```
guusdk: Everything under this line is in English - although it doesn't hurt to have this in, I'm assuming that the intention was to translate this into Spanish too?

i18n/src/main/resources/openfire_i18n_es.properties
```diff
@@ -68,8 +70,16 @@ tab.server.descr=Presione para administrar la configuración del servidor
         sidebar.transfer-proxy=Configuración de transferencia de archivos
         sidebar.transfer-proxy.descr=Presione para ver configuración de transferencia de archivos
     sidebar.sidebar-certificates=Certificados TLS/SSL
+        sidebar.certificate-stores=Almadenes de certificados
```
guusdk: "Almadenes" should be "Almacenes" ?

i18n/src/main/resources/openfire_i18n_es.properties
wrooot: Indeed, haven't noticed this. Maybe this is kind of a placeholder for future updates. Or a workaround for an issue where Openfire showed ??? instead of English if strings were missing in translation. Though, i think this issue is already fixed.

i18n/src/main/resources/openfire_i18n_es.properties
drencrom: Yes

i18n/src/main/resources/openfire_i18n_es.properties
drencrom: I did translate that. Maybe I messed up with the files when I commited. I'll check later.

PR #1409, opened 2019-06-28
skip travis deployment step for non-ignite builds

akrherz:
Developers that have Openfire forked and setup for Travis will always see failures due to `deploy` failing (rightfully so).  This change should skip the `deploy` script when travis is run on the forked repos.  This change [worked](https://travis-ci.org/akrherz/Openfire/builds/551854288) for me.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ script:
 install: true
 deploy:
   provider: script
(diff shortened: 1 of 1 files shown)


akrherz:
Note that Travis-CI will auto-skip the `deploy` on pull requests.

guusdk:
Thanks - that'll come in handy.

PR #1411, opened 2019-06-28
fix travis.yml loop for trigger plugins

akrherz:
quote the string and remove extraneous ";", maybe?


akrherz:
merging as it is about the easiest way to test this Travis-CI deploy-only change!  Crosses-fingers!

akrherz:
winner, winner, chicken dinner.  :zap:

PR #1412, opened 2019-07-01
OF-1809 Allow Openfire to include more then one XDataForm

mnsuccess:
https://xmpp.org/extensions/xep-0128.html#impl defines that a Service Discovery Info response can include more than one "extended info" dataform.

Openfire should allow for this. It currently allows for at most one form.

Openfire should be modified to allow for a set of dataforms to be returned. At the same time, it is desirable that the API change does not break existing (third party) code.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
@@ -16,9 +16,12 @@
 
 package org.jivesoftware.openfire.disco;
 
+import org.apache.juli.logging.Log;
 import org.dom4j.DocumentHelper;
 import org.dom4j.Element;
 import org.dom4j.QName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.jivesoftware.openfire.IQHandlerInfo;
 import org.jivesoftware.openfire.SessionManager;
 import org.jivesoftware.openfire.XMPPServer;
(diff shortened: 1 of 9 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -692,7 +696,19 @@ public boolean hasInfo(String name, String node, JID senderJID) {
 
             @Override
             public DataForm getExtendedInfo(String name, String node, JID senderJID) {
-                return null;
+                try {
```
mnsuccess: Do I have to update this method like this?

guusdk: Almost. I'm not sure why you wrap this in a try/catch. If it's needed, you must never, _never_ forget to log the exception message. You will learn this after you spend weeks tracking down a bug that was hidden from view because someone didn't log the message. :)

When you're logging the warning about there being more than one dataform, it is good to also log:
- how many dataforms there are
- how the code handles this situation (for example: "Only one of the forms will be returned.")

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -696,19 +696,16 @@ public boolean hasInfo(String name, String node, JID senderJID) {
...
+                if(dataForms != null && dataForms.size() == 1){
+                    return dataForms.iterator().next();
+                }else if (dataForms != null && dataForms.size() > 1){
+                    Log.warn("Set Data List contains "+dataForms.size()+" DataForms."+
+                    "Only the first one of the DataForms will be returned.");
+                    return dataForms.iterator().next();
+                } else {
                     return null;
-                } 
+                }
```
guusdk: Weird whitespace?

PR #1413, opened 2019-07-03
OF-1809 Allow Openfire to include more then one XDataForm

mnsuccess:
https://xmpp.org/extensions/xep-0128.html#impl defines that a Service Discovery Info response can include more than one "extended info" dataform.

Openfire should allow for this. It currently allows for at most one form.

Openfire should be modified to allow for a set of dataforms to be returned. At the same time, it is desirable that the API change does not break existing (third party) code.


xmppserver/src/main/java/org/jivesoftware/openfire/disco/DiscoInfoProvider.java
```diff
@@ -67,9 +69,25 @@
      * @param node the requested disco node.
      * @param senderJID the XMPPAddress of user that sent the disco info request.
      * @return an XDataForm with the extended information about the entity or null if none.
+     * @deprecated Replaced by {@link #getExtendedInfos(String, String, JID)}
      */
+    @Deprecated
     DataForm getExtendedInfo( String name, String node, JID senderJID );
 
+    /**
```
guusdk: The Javadoc (which is the "contract" for this method" defines that null is returned when there is no extended information dataform.

This does not correspond with the default implementation: when no form is available, I think the implementation returns a Collection that contains a single, null, element.

Two things:
- I think it's better to return an empty collection, instead of a collection that contains a null value. If a collection with a null value is needed, the reason for this must be clearly documented.
- In any case, the javadoc and the implementation must define the same implementation.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -191,12 +194,14 @@ public IQ handleIQ(IQ packet) {
                     // discovery broadcasts the disco#info feature.
                     queryElement.addElement("feature").addAttribute("var", NAMESPACE_DISCO_INFO);
                 }
-
-                // Add to the reply the extended info (XDataForm) provided by the DiscoInfoProvider
-                DataForm dataForm = infoProvider.getExtendedInfo(name, node, packet.getFrom());
-                if (dataForm != null) {
+                // Add to the reply the multiple extended info (XDataForm) provided by the DiscoInfoProvider
+                Iterator<DataForm> dataForms = infoProvider.getExtendedInfos(name, node, packet.getFrom()).iterator();
+                //Log.info("DATAFORM {}",dataForms.toString());
```
guusdk: I think you used this Log statement for debugging purposes. If we're not going to use it, let's remove it completely.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -493,6 +498,25 @@ private void restoreCacheContent() {
         }
     }
 
+    /**
```
guusdk: I think that the `getFirstDataForm` method will _always_ log a warning in this implementation, even if the provided input collection contains exactly one item. That's probably undesirable.

What happens if the input of the `getFirstDataForm` method is an empty set, or null? It is a good idea to write many small unit tests for scenarios like this. At first, this work seems silly, but you'd be surprised how many stupid bugs I've prevented in my own code changes by writing many simple tests (that does not mean that there are not a lot more stupid bugs in my code - but there are less :-) ). I suggest that you write a couple of very simple JUnit-based unit tests for this method:
- One that uses a null value as in input
- One that uses an empty collection as input
- One that uses a collection with exactly one item (that is null) as input
- One that uses a collection with exactly one item (that is not null) as input
- One that uses a collection that has two (non-null) items as input.

First, think of what _should_ happen in each of these scenarios, and then add a test implementation that verifies (asserts) that this indeed does happen. Please use `org.jivesoftware.openfire.keystore.CertificateUtilsTest` as an example for your unit tests. Create a new class (`org.jivesoftware.openfire.disco.IQDiscoInfoHandlerTest`) and add a method for each of the tests I described above (you can add more if you can think of other interesting scenarios to test). Each method should test exactly one thing (it's OK to have many very similar methods). Each test should have three blocks: "Setup test fixture", "Execute system under test", "Verify result". We do this, so that in 10 years, when someone else needs to maintain the code, can immediately see what the test intends to do.

Also, in the javadoc, please add a description for the parameter, and a better description of the return value (I can see that the return is a DataForm based on the method signature, so just writing `@return DataForm` does not add any value).

As a rule of thumb, it is good to define in the javadoc if a param is allowed to be null, and if the return value can expected to be null.

Suggestion:

    @param dataForms collection from which to return the first element (cannot be null, can be an empty collection).
    @return first dataform from the collection. Null if the collection was empty.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -505,15 +505,21 @@ private void restoreCacheContent() {
     public static DataForm getFirstDataForm(Set<DataForm> dataForms){
         if(dataForms != null && !dataForms.isEmpty()){
             Iterator<DataForm> iterator = dataForms.iterator();
-            if (iterator.hasNext()) {
-                Log.warn("Set Data List contains "+dataForms.size()+" DataForms."+
-                "Only the first one of the DataForms will be returned.");
+            DataForm dataForm = null;
```
guusdk: Although your test proves that the implementation is correct, it looks a bit. I think you could make it simpler, by splitting it in three parts:

```
if (dataForms.isEmpty() ) {
  return null;
}

if (dataForms.size() > 1) {
  Log.warn(" ... ");
}

return dataForms.iterator().next();
```

mnsuccess: I think it will not be complete when the collection has two elements the first null et the second not null !! it will be better to return the second instead to return null. the  function returns the first item that is not null

mnsuccess: imagine in another case if the collection has 4 elements and the no null form is in the fourth position

GregDThomas: Streams!

`return dataForms.stream().filter(Objects::nonNull).findAny().orElse(null);`

Also, it's a set, so un-order, so firndFirst/findLast would be non-deterministic anyway.

(Sorry, I'll go back in my hole now)

guusdk: Try what Greg said (although I think it's unlikely that there will be 'null' items in the collection in the first place.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -503,26 +503,14 @@ private void restoreCacheContent() {
...
-                    }else{
-                        Log.warn("Set Data List contains "+dataForms.size()+" DataForms."+
-                        "Only the first one of the DataForms will be returned.");
-                    }
-                    i++;
-                } 
-            }
-            return dataForm ;
-        }else{
+        if (dataForms == null || dataForms.isEmpty() ) {
```
guusdk: You can make this a little bit shorter, if you replace the first two if statements with just one:

```java
if (dataForms != null && dataForms.size() > 1) {
```

mnsuccess:
> You forgot to remove the first if-statement. :)
Unit Test failled 
testGetFirstDataFormNull()

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -503,14 +503,15 @@ private void restoreCacheContent() {
...
      */
     public static DataForm getFirstDataForm(Set<DataForm> dataForms){
-        if (dataForms == null || dataForms.isEmpty() ) {
+        try {
+            if (dataForms != null && dataForms.size() > 1) {
+                Log.warn("Set Data List contains "+dataForms.size()+" DataForms."+
+                "Only the first one of the DataForms will be returned.");
+            }
+            return  dataForms.stream().filter(Objects::nonNull).findAny().orElse(null);
+        } catch (Exception e) {
```
guusdk: Why do you wrap this in a try/catch?

Ohh, line 511 can nullpointer! I didn't see that before. Maybe your previous version was better after all. :)

If you ever catch an exception, always, always, always do something with the exception - if only logging it to 'debug'. I don't think we should need a try/catch block here, by the way - just use your previous version.

guusdk:
Can you please rebase this PR on the latest master? A change to the Travis-CI configuration was committed there, which should make the Travis tests less unhappy.

PR #1414, opened 2019-07-06
OF-1808: Be more restrictive over those caches we try to migrate 

GregDThomas:


xmppserver/src/main/java/org/jivesoftware/util/cache/CacheFactory.java
```diff
@@ -854,7 +854,9 @@ public static synchronized void joinedCluster() {
         // Loop through local caches and switch them to clustered cache (copy content)
         for (Cache cache : getAllCaches()) {
             // skip local-only caches
```
guusdk: Nitpicking, but shouldn't we change the comment that says 'skip local-only caches'? Also, what _is_ a "distributedWrappedCache" (I don't understand the method name). Let's either have javadoc to explain the meaning, or choose a different name for the method?

GregDThomas: What you say makes sense; I'll have a look ASAP.

GregDThomas: OK, latest commit makes the code seem a lot clearer to me; hopefully to others too.

PR #1416, opened 2019-07-08
OF-1811: Prevent NPE in GroupManager

guusdk:
This commit ensures that a previously cached negative result ('no such group') causes an exception to be thrown, rather than null to be returned.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
@@ -364,6 +364,10 @@ public class GroupManager {
                 }
             }
         }
+
+        if (coGroup.isAbsent() ) {
+            throw new GroupNotFoundException( "Group with name " + name + " not found (cached)." );
+        };
         return coGroup.get();
     }
 


xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
```diff
@@ -364,6 +364,10 @@ public Group getGroup(String name, boolean forceLookup) throws GroupNotFoundExce
                 }
             }
         }
+
+        if (coGroup.isAbsent() ) {
+            throw new GroupNotFoundException( "Group with name " + name + " not found (cached)." );
+        };
```
GregDThomas: I'm not sure this code is ever reachable. Line 351 or 361 will throw the exception if it's not found.

(though I think the code could do with refactoring to make it a bit more explicit!)

guusdk: I'd love to make this code more explicit, because I think you're now making the same mistake as I did. If coGroup is _not_ null in line 346, but if it is a CacheableOptional of null, it won't ever hit lines 351 or 361.

GregDThomas: A good point. As it stands, your code change is required.

But overall code could be simplified by using something like `groupCache.getOrDefault(Optional.absent())`

guusdk: does `computeIfAbsent` work with Caches?

GregDThomas: `getOrDefault(...)` will work because there are no thread safety issues.
`computeIfAbsent(...)` will certainly appear to work, I'm not entirely sure what will happen wrt. thread safety though.

guusdk: Bah, I can't refactor this without creating more of a cludge of it, instead of improve things. Feel free to have a stab at it, otherwise, let's merge this and at least fix the bug.

GregDThomas: Meh. Merge it. It's getting late here, I'm getting foul looks from people (well, person) I shouldn't be ...

guusdk: You're not the only recipient of looks like that...

PR #1419, opened 2019-07-09
Observed some errors in Openfire sybase db script and corrected.

balavihar:
In Sybase every column should be defined NONNULL/NULL constraint, observed some columns without NONNULL/NULL, fixed by adding NULL to columns.
Semicolons are removed because it stopping script to execute in Sybase.
Corrected some datatypes which are not supporting for Sybase.
Column name "serviceID" is corrected in index "ofMucRoom_serviceid_idx"

### DIFF
--- a/distribution/src/database/openfire_sybase.sql
+++ b/distribution/src/database/openfire_sybase.sql
@@ -1,10 +1,10 @@
 
 CREATE TABLE ofUser (
   username              NVARCHAR(64)    NOT NULL,
-  storedKey             VARCHAR(32),
-  serverKey             VARCHAR(32),
-  salt                  VARCHAR(32),
-  iterations            INTEGER,
+  storedKey             VARCHAR(32)     NULL,
+  serverKey             VARCHAR(32)     NULL,
+  salt                  VARCHAR(32)     NULL,
+  iterations            INTEGER         NULL,
   plainPassword         NVARCHAR(32)    NULL,
   encryptedPassword     NVARCHAR(255)   NULL,
   name                  NVARCHAR(100)   NULL,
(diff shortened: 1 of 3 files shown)


akrherz:
Do any of the upgrade scripts need fixing too?

GregDThomas:
Probably. 

But, I suspect the number of users of Sybase/SAP ASE with Openfire is vanishingly small. I could probably count them on the fingers of one hand. Wearing a mitten. And I know they are doing a new install not an upgrade (Hi, Bala!) so getting traction for someone to update the scripts and test them will be hard.

balavihar:
Ok, Ill test upgrade scripts, if there are any errors then I will update them too. But, I see 30 upgrade scripts, all should be tested and fixed? I think last one would be sufficient.

guusdk:
The update scripts are cumulative, so ideally, they're all to be tested.

Op wo 10 jul. 2019 10:44 schreef balavihar <notifications@github.com>:

balavihar:
I tested last upgrade script and corrected, I mean Openfire-4.4 install script and upgrade script.

wrooot:
Your last commit was in postgre script, not in sybase.

balavihar:
Sorry for that,  I have done in wrong file. Now I corrected both Postgresql and Sybase upgrade script files.

balavihar:
Could anyone please tell me, is this integration test got failed due to my commits in db scripts.

guusdk:
We have intermittent failures like that. I restarted the test

balavihar:
Closing this pull request as wrooot suggested, Ill try to remove wrong commits which i done and correct Sybase upgrade scripts.

PR #1422, opened 2019-07-14
Group manager improvements

GregDThomas:
@guusdk - for your consideration. This was bugging me. Even if you don't like the refactoring, it's probably worth keeping the tests, anyway.


guusdk:
Thanks! This bugged me to.

PR #1423, opened 2019-07-18
OF-1818: Some errors in Openfire sybase db script and last 5 upgrade scripts are corrected.

balavihar:
In Sybase every column should be defined NONNULL/NULL constraint, observed some columns without NONNULL/NULL, fixed by adding NULL to columns.
Semicolons are removed because it stopping script to execute in Sybase.
Corrected some datatypes which are not supporting for Sybase.
Column name "serviceID" is corrected in index "ofMucRoom_serviceid_idx".
Also, corrected last 5 upgrade scripts.

### DIFF
--- a/distribution/src/database/openfire_sybase.sql
+++ b/distribution/src/database/openfire_sybase.sql
@@ -1,10 +1,9 @@
-
 CREATE TABLE ofUser (
   username              NVARCHAR(64)    NOT NULL,
-  storedKey             VARCHAR(32),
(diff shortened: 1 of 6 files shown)


guusdk:
I have created https://issues.igniterealtime.org/browse/OF-1818 to track this issue.

guusdk:
Thanks for your efforts on this!

guusdk:
I've cherry-picked this change to the 4.4 branch too.

PR #1426, opened 2019-07-26
OF-1816: Close input streams in log.jsp correctly

cpetzka:
We had the problem that the logfiles sometimes cannot be loaded correctly. This pull request solves this problem.

### DIFF
--- a/xmppserver/src/main/webapp/log.jsp
+++ b/xmppserver/src/main/webapp/log.jsp
@@ -109,36 +109,17 @@
     try {
         String line;
         int totalNumLines = 0;
-        BufferedReader in = null;
-        InputStreamReader inputStreamReader= null;
-        FileInputStream fileInputStream=null;
-        try{
-        	fileInputStream = new FileInputStream(logFile);
-        	inputStreamReader = new InputStreamReader(fileInputStream, "UTF-8");
-        	in = new BufferedReader(inputStreamReader);
+        try(FileInputStream fileInputStream = new FileInputStream(logFile); InputStreamReader inputStreamReader = new InputStreamReader(fileInputStream, "UTF-8"); BufferedReader in = new BufferedReader(inputStreamReader);){
             while ((line=in.readLine()) != null) {
                 totalNumLines++;
         	}
-        }finally{
-            if (in!=null)
-            	in.close();
-            if (inputStreamReader!=null)
-            	inputStreamReader.close();
-            if (fileInputStream!=null)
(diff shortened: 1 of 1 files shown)


guusdk:
Ah, this seems like a sensible improvement. Can you tell me exactly what happened, as I never had the problem before?

Also, I suggest that you implement Java's try-with-resources, to do the same thing (ensure that streams are closed), but with a lot less code (which is easier to maintain).

guusdk:
I've created a new issue for this problem in our bugtracker: https://issues.igniterealtime.org/browse/OF-1816

cpetzka:
Sometimes we noticed that the log content in the admin console was empty even though the log files were not empty. Only restarts helped to display the log content again. After these changes were applied, the problem did not occur again.

I will adapt the code to your recommendations.

xmppserver/src/main/webapp/log.jsp
```diff
@@ -109,36 +109,17 @@
...
         String line;
         int totalNumLines = 0;
-        BufferedReader in = null;
-        InputStreamReader inputStreamReader= null;
-        FileInputStream fileInputStream=null;
-        try{
-        	fileInputStream = new FileInputStream(logFile);
-        	inputStreamReader = new InputStreamReader(fileInputStream, "UTF-8");
-        	in = new BufferedReader(inputStreamReader);
+        try(FileInputStream fileInputStream = new FileInputStream(logFile); InputStreamReader inputStreamReader = new InputStreamReader(fileInputStream, "UTF-8"); BufferedReader in = new BufferedReader(inputStreamReader);){
```
guusdk: Ugh. Mind adding some newlines? Something like this is perfectly valid.
```
try( FileInputStream fileInputStream = new FileInputStream(logFile);
     InputStreamReader inputStreamReader = new InputStreamReader(fileInputStream, "UTF-8");
     BufferedReader in = new BufferedReader(inputStreamReader); )
{
```

cpetzka: My mistake! I disabled automatic code formatting to keep whitespace changes at a minimum.

cpetzka: done.
Should I make a new pull request with only one commit?

guusdk: Don't bother, this is fine.

guusdk:
Thanks! I've cherry-picked this change to the 4.4 branch too.

PR #1427, opened 2019-07-26
OF-1953: fix an npe in the server sessions view on the admin console

cpetzka:
When viewing the server sessions on the admin console we have sometime NPE's:

```
	at java.util.Date.getMillisOf(Date.java:958)
	at java.util.Date.after(Date.java:931)
	at org.jivesoftware.openfire.admin.server_002dsession_002dsummary_jsp._jspService(server_002dsession_002dsummary_jsp.java:353)
	at org.apache.jasper.runtime.HttpJspBase.service(HttpJspBase.java:70)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:812)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1669)
	at com.opensymphony.module.sitemesh.filter.PageFilter.parsePage(PageFilter.java:118)
	at com.opensymphony.module.sitemesh.filter.PageFilter.doFilter(PageFilter.java:52)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.jivesoftware.util.LocaleFilter.doFilter(LocaleFilter.java:73)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.jivesoftware.util.SetCharacterEncodingFilter.doFilter(SetCharacterEncodingFilter.java:49)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.jivesoftware.admin.PluginFilter.doFilter(PluginFilter.java:226)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.jivesoftware.admin.AuthCheckFilter.doFilter(AuthCheckFilter.java:215)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:585)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:577)
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:223)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1127)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:515)
	at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:185)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1061)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
	at org.eclipse.jetty.server.handler.ContextHandlerCollection.handle(ContextHandlerCollection.java:215)
	at org.eclipse.jetty.server.handler.HandlerCollection.handle(HandlerCollection.java:110)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:97)
	at org.eclipse.jetty.server.Server.handle(Server.java:499)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:311)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:257)
	at org.eclipse.jetty.io.AbstractConnection$2.run(AbstractConnection.java:544)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:635)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:555)
	at java.lang.Thread.run(Thread.java:748)
```

This pull requests solves the NPE and fixes a releated multi threading issue in the LocalSessions.java file. 

For example:

```
 if (conn == null) return;
conn.close(); <-- At this point conn can be null
```


cpetzka:
changed to your recommendations

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -434,44 +467,89 @@ public void deliverRawText(String text) {
...
+            connection.close();
+        }finally {
+            lock.writeLock().unlock();
+        }
     }
 
     @Override
     public boolean validate() {
-        return conn.validate();
+        lock.writeLock().lock();
```
cpetzka: is here a write or read lock required?

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -135,16 +142,26 @@ public LocalSession(String serverName, Connection connection, StreamID streamID,
      * @return true if session detached
      */
     public boolean isDetached() {
-        return this.conn == null;
+        lock.readLock().lock();
```
GregDThomas: This lock achieves nothing. The conn is either null or not at the time of checking, so no inconsistency or other issues here.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -135,16 +142,26 @@ public LocalSession(String serverName, Connection connection, StreamID streamID,
...
     }
 
     /**
      * Set the session to detached mode, indicating that the underlying connection
      * has been closed.
      */
     public void setDetached() {
-        this.sessionManager.addDetached(this);
-        this.conn = null;
+        lock.writeLock().lock();
```
GregDThomas: Not sure this lock achieves much either. It does mean the object can be in state of detached with a conn, but don't see how that can have any side effects.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -198,14 +219,20 @@ public void setAddress(JID address){
...
-        if (conn == null) {
-            try {
-                conn.isClosed(); // This generates an NPE deliberately.
-            } catch (NullPointerException e) {
-                Log.error("Attempt to read connection of detached session: ", e);
+        lock.readLock().lock();
+        try {
+            if (conn == null)
+            {
+                try {
```
GregDThomas: Not a comment on your changes, but as you're here ...

replace this try/catch with a 

Log.error("Attempt to read connection of detached session", new IllegalStateException());

GregDThomas:
I tend to agree with Guus in that I don't think any of the locking is required. It certainly won't fix the NPE in the JSP.

As for the NPE in the JSP, I think you changes have somewhat obfuscated what's going on. At least, it's not obvious to me. Having look a bit closer, I don't see how it can happen at all.

Specifically; the NPE is in a call to `Date.after(Date)`. The only calls to this in the JSP pass in either an `IncomingServerSession.getCreationDate()` or an `OutgoingServerSession.getCreationDate()`. Both are these inherit from LocalSession.getCreationDate() which returns a new object every time (*) - and so can't be null.

So it looks to me like the NPE you've triggered no longer exists on master. Was it an off-the-shelf Openfire, and if so which version?

(*) Probably an opportunity for optimisation there, but that's not important right now.

cpetzka:
Hi,

for the JSP: 

> Specifically; the NPE is in a call to Date.after(Date). The only calls to this in the JSP pass in either an IncomingServerSession.getCreationDate() or an OutgoingServerSession.getCreationDate(). Both are these inherit from LocalSession.getCreationDate() which returns a new object every time (*) - and so can't be null.

I can follow your argumentation and also do not see how it can happen that a value is null at this point.

> So it looks to me like the NPE you've triggered no longer exists on master. Was it an off-the-shelf Openfire, and if so which version?

The Version was 4.1.4 self builded but not modified. Also i see that this lines for (creationDate) are modified in 2007 the last time. So there is no difference between 4.1.4 and the master branch.
So how can this happen? o.O

I would undo the changes and watch the logs if this problem occures again. Excuse me for taking up your time.

for the LocalSession.java

This was a result on the investigation of the NPE. We found out, that there might be possible cases where the connection can be null or modified by two threads at the same time. I think this is still relevant. So should i make a new merge request only for the LocalSession.java with your recommendations applied?

I think it would be enough to use local variables for the connection in all read methods and locks only in validate/reattach/setDetached. What do you think?

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: done

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: done

akrherz:
Do we want this change for Openfire 4.4.1 ?

cpetzka:
no not necessarily

GregDThomas:
This has been outstanding a while; is it possible related to https://discourse.igniterealtime.org/t/npe-in-localsession/86136 ?

cpetzka:
> This has been outstanding a while; is it possible related to https://discourse.igniterealtime.org/t/npe-in-localsession/86136 ?

Yes it is related.

The NPE is fixed on line https://github.com/igniterealtime/Openfire/pull/1427/files#diff-ea9a45ac8aa08fccbc303ffc0f705569R490 

If you want, I can create a new pull request that only changes the LocalSession.java.

cpetzka:
a) I really like the "Optional" approach, but technical its the same as in my first commit https://github.com/igniterealtime/Openfire/pull/1427/commits/da15b2079bb82db5c5774dd9a52f8ff74ee3f220#diff-ea9a45ac8aa08fccbc303ffc0f705569R452 (Save the connection in a local variable -> make a null check -> read it ... or have I forgotten something?)

@guusdk  mentioned: 
> If the changes to LocalSession are indeed fixing something, then we must be suffering from some kind of concurrency bug. If that's the case, then these changes will only furher hide that bug, and not fix it. I don't like that. Lets instead use some kind of mutex, that ensures that conn cannot be re-assigned while one of the methods is operating on it.

So the best way is to replace the read locks with optional or local variables and leave only the write locks or not?

b) That's a good question! I think the " safe default " will be "false". As a result, it will not always be displayed correctly. Could a new field help where we store if a attached connection was secure?

GregDThomas:
a) You're right, the Optional is the same as the local variable, but IMHO the Optional is a little more expressive as to what's happening.

a) ii) I don't think there's a need for the lock. From my reading the NPE occurs because the session gets set to null between checking its not null and accessing it. The local variable/Optional removes that possibility.

b) I think defaulting to insecure in the absence of a local session should suffice. As soon as the session is re-attached the correct value will be used, and as an insecure session can be promoted to secure it's difficult to know exactly when the cache the actual value.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: I think it can have side effects with the reattach method. So line 179 (https://github.com/igniterealtime/Openfire/pull/1427/files#diff-ea9a45ac8aa08fccbc303ffc0f705569R179) should also be in the lock.

The lock should ensure, that there is no other operation to which do a operation on the connection while it is set to null or another reference.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -416,12 +427,18 @@ public void process(Packet packet) {
 
     @Override
     public void deliverRawText(String text) {
-        if ( conn == null )
-        {
-            Log.debug( "Unable to deliver raw text in session, as its connection is null. Dropping: " + text );
-            return;
+        lock.lock();
```
GregDThomas: Don't think this lock/unlock is necessary

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -434,44 +451,64 @@ public void deliverRawText(String text) {
 
     @Override
     public void close() {
-        if (conn == null) return;
-        conn.close();
+        lock.lock();
```
GregDThomas: Don't think this lock/unlock is necessary

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -434,44 +451,64 @@ public void deliverRawText(String text) {
...
+              .ifPresent(connection ->  connection.close());
+        }finally {
+            lock.unlock();
+        }
     }
 
     @Override
     public boolean validate() {
-        return conn.validate();
+        lock.lock();
```
GregDThomas: Don't think this lock/unlock is necessary

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -434,44 +451,64 @@ public void deliverRawText(String text) {
...
-        return conn.isClosed();
+        return Optional.ofNullable(conn)
+                .map(Connection::isClosed)
+                .orElse(Boolean.TRUE);
     }
 
     @Override
     public String getHostAddress() throws UnknownHostException {
-        if (conn == null) {
+        Connection connection = conn;
```
GregDThomas: Could be simply

```
return Optional.ofNullable(conn)
    .map(Connection::getHostAddress())
    .orElseThrow(()->new UnknownHostException("Detached session");
```

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -434,44 +451,64 @@ public void deliverRawText(String text) {
...
             throw new UnknownHostException("Detached session");
         }
-        return conn.getHostAddress();
+        return connection.getHostAddress();
     }
 
     @Override
     public String getHostName() throws UnknownHostException {
-        if (conn == null) {
+        Connection connection = conn;
```
GregDThomas: See previous comment

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: It is not possible, because UnknownHostException is a checked Exception and Connection::getHostAddress() has declerad it. In lambda use it should be like:

```java 
return Optional.ofNullable(conn)
    .map(connection -> {
        try{
            return connection.getHostAddress();
         }catch(UnknownHostException e){
            throw new RuntimeException(e);
        }
     })
    .orElseThrow(()->new UnknownHostException("Detached session");
```

in my opinion this is not better.

GregDThomas: Good point, well made!

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: done

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: done

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: done

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
cpetzka: what about this?

GregDThomas:
Yup, LGTM. Thanks once more!

guusdk:
Thanks for all of the hard work that went into this issue!

I don't think we are tracking this in JIRA yet, are we? I've just now created https://issues.igniterealtime.org/browse/OF-1953 for this.

PR #1428, opened 2019-07-26
OF-1817: Fix a ClassCastException in LocalMUCRoom.java

cpetzka:
When using the Openfire with the hazelcast plugin we had some ClassCastExceptions in the LocalMucRoom at line 685 like: 

`java.lang.ClassCastException: org.jivesoftware.openfire.plugin.session.RemoteClientSession cannot be cast to org.jivesoftware.openfire.session.LocalClientSession`

For my bad, I don't have the full stack trace.

With this pull request this error does not occur anymore. It lets the three classes use the interface and not the implementing class.

This may only fix one symptom of another problem.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/HistoryRequest.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/HistoryRequest.java
@@ -131,7 +131,7 @@ public class HistoryRequest {
      * @param joinRole the user that will receive the history.
      * @param roomHistory the history of the room.
      */
-    public void sendHistory(LocalMUCRole joinRole, MUCRoomHistory roomHistory) {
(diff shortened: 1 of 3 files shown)


GregDThomas:
@cpetzka - was this running with Openfire 4.4 or something else?

cpetzka:
with 4.3

guusdk:
Is there a chance that this was already fixed in 4.4 / master?

GregDThomas:
Don’t think so, no. You simply won’t see it on 4.4 because of the broken clustering.

guusdk:
Ok, tx, I've created https://issues.igniterealtime.org/browse/OF-1817 to track this issue.

guusdk:
Thanks! I've cherry-picked this change to the 4.4 branch too.

PR #1429, opened 2019-07-26
OF-1930: Make LocalMUCRoom more thread safe

cpetzka:
The field `occupantsByNickname` is readed from and written to without a lock:

```java
List<MUCRole> occupants = occupantsByNickname.get(nickname.toLowerCase());
            // Do not add new occupant with one with same nickname already exists
            if (occupants == null) {
                occupants = new ArrayList<>();
                occupantsByNickname.put(nickname.toLowerCase(), occupants);
```

This pull request surround all code parts, in which this happens, with the existing write lock.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -830,17 +830,17 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
                 // Fire event that occupant joined the room
                 MUCEventDispatcher.occupantJoined(getRole().getRoleAddress(), event.getUserAddress(), joinRole.getNickname());
             }
-            // Check if we need to send presences of the new occupant to occupants hosted by this JVM
-            if (event.isSendPresence()) {
-                for (MUCRole occupant : occupantsByFullJID.values()) {
-                    if (occupant.isLocal()) {
-                        occupant.send(event.getPresence().createCopy());
-                    }
-                }
-            }
         }
         finally {
             lock.writeLock().unlock();
+        }
+        // Check if we need to send presences of the new occupant to occupants hosted by this JVM
+        if (event.isSendPresence()) {
+            for (MUCRole occupant : occupantsByFullJID.values()) {
+                if (occupant.isLocal()) {
+                    occupant.send(event.getPresence().createCopy());
+                }
+            }
         }
     }
 
@@ -937,15 +937,17 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
      * @param originator true if this JVM is the one that originated the event.
      */
     private void removeOccupantRole(MUCRole leaveRole, boolean originator) {
+        JID userAddress;
+        String nickname;
         lock.writeLock().lock();
         try {
-            JID userAddress = leaveRole.getUserAddress();
+            userAddress = leaveRole.getUserAddress();
             // Notify the user that he/she is no longer in the room
             leaveRole.destroy();
             // Update the tables of occupants based on the bare and full JID
             JID bareJID = userAddress.asBareJID();
 
-            String nickname = leaveRole.getNickname();
+            nickname = leaveRole.getNickname();
             List<MUCRole> occupants = occupantsByNickname.get(nickname.toLowerCase());
             if (occupants != null) {
                 occupants.remove(leaveRole);
@@ -961,13 +963,13 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
                 }
             }
             occupantsByFullJID.remove(userAddress);
-            if (originator) {
-                // Fire event that occupant left the room
-                MUCEventDispatcher.occupantLeft(getRole().getRoleAddress(), userAddress, nickname);
-            }
         }
         finally {
             lock.writeLock().unlock();
+        }
+        if (originator) {
+            // Fire event that occupant left the room
+            MUCEventDispatcher.occupantLeft(getRole().getRoleAddress(), userAddress, nickname);
         }
     }
 
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -497,11 +497,7 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
 
     @Override
     public MUCRole getOccupantByFullJID(JID jid) {
-        MUCRole role = occupantsByFullJID.get(jid);
-        if (role != null) {
-            return role;
-        }
-        return null;
+        return occupantsByFullJID.get(jid);
     }
 
     @Override
@@ -560,7 +556,7 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
                 throw new UnauthorizedException();
             }
         }
-        LocalMUCRole joinRole = null;
+        LocalMUCRole joinRole;
         lock.writeLock().lock();
         boolean clientOnlyJoin = false;
         // A "client only join" here is one where the client is already joined, but has re-joined.
@@ -598,7 +594,7 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
             if (isPasswordProtected()) {
                 final boolean isCorrectPassword = (password != null && password.equals(getPassword()));
                 final boolean isSysadmin = mucService.isSysadmin(bareJID);
-                final boolean requirePassword = isSysadmin ? mucService.isPasswordRequiredForSysadminsToJoinRoom() : true;
+                final boolean requirePassword = !isSysadmin || mucService.isPasswordRequiredForSysadminsToJoinRoom();
                 if (!isCorrectPassword && requirePassword ) {
                     throw new UnauthorizedException();
                 }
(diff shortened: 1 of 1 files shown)


GregDThomas:
Though this will work, I wonder if the lock is held too long in a couple of places? Could it be released earlier, e.g. after line 825, 963,

cpetzka:
@GregDThomas i agree with you that the lock can be released after line 963. 
But on line 833 there is a read from occupantsByFullJID.

Is it necessary that the read accesses are also synchronized by locks to avoid that read accesses within a method can return different results?

like in:
```java
    private void broadcast(Message message) {
        // Broadcast message to occupants hosted by other cluster nodes
        BroadcastMessageRequest request = new BroadcastMessageRequest(this, message, occupantsByFullJID.size());
        CacheFactory.doClusterTask(request);

        // Broadcast message to occupants connected to this JVM
        request = new BroadcastMessageRequest(this, message, occupantsByFullJID.size());
        request.setOriginator(true);
        request.run();
    }
```

GregDThomas:
The read around line 833 is telling others that a person has joined the room. There is no need to to be locked while that happens - if they leave just before the message is sent, they simply won't receive the message.

I also don't think there's a need to lock during reads.

cpetzka:
OK changed

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -792,41 +792,47 @@ private void sendInitialPresences(LocalMUCRole joinRole) {
...
-            // sanity check; make sure the nickname is owned by the same JID
-            if (occupants.size() > 0) {
-                JID existingJID = occupants.get(0).getUserAddress().asBareJID();
-                if (!bareJID.equals(existingJID)) {
-                    Log.warn(MessageFormat.format("Conflict detected; {0} requested nickname '{1}'; already being used by {2}", bareJID, nickname, existingJID));
-                    return;
+        lock.writeLock().lock();
+        try {
+            // Create a proxy for the occupant that joined the room from another cluster node
+            RemoteMUCRole joinRole = new RemoteMUCRole(mucService, event);
```
GregDThomas: 798/799/800 can be outside the lock

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -792,41 +792,47 @@ private void sendInitialPresences(LocalMUCRole joinRole) {
...
-        if (list == null) {
-            list = new ArrayList<>();
-            occupantsByBareJID.put(bareJID, list);
-        }
-        list.add(joinRole);
-        occupantsByFullJID.put(event.getUserAddress(), joinRole);
+            // Add the new user as an occupant of this room
+            occupants.add(joinRole);
+            // Update the tables of occupants based on the bare and full JID
+            List<MUCRole> list = occupantsByBareJID.get(bareJID);
```
GregDThomas: Not as a result of your changes, but as you're here, lines 819->825 can be replaced with simply

`occupantsByBareJID.computeIfAbsent(bareJID, jid->new ArrayList<>()).add(joinRole);`

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -792,41 +792,47 @@ private void sendInitialPresences(LocalMUCRole joinRole) {
...
-                if (!bareJID.equals(existingJID)) {
-                    Log.warn(MessageFormat.format("Conflict detected; {0} requested nickname '{1}'; already being used by {2}", bareJID, nickname, existingJID));
-                    return;
+        lock.writeLock().lock();
+        try {
+            // Create a proxy for the occupant that joined the room from another cluster node
+            RemoteMUCRole joinRole = new RemoteMUCRole(mucService, event);
+            JID bareJID = event.getUserAddress().asBareJID();
+            String nickname = event.getNickname();
+            List<MUCRole> occupants = occupantsByNickname.get(nickname.toLowerCase());
```
GregDThomas: Not as simple as line 819+, but this could be simplified with 

`List<MUCRole> existingOccupantsForNickname = occupantsByNickname.computeIfAbsent(nickname.toLowerCase, nickname->new ArrayList());
...`

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -931,28 +937,36 @@ public void leaveRoom(OccupantLeftEvent event) {
...
+        lock.writeLock().lock();
+        try {
+            userAddress = leaveRole.getUserAddress();
+            // Notify the user that he/she is no longer in the room
+            leaveRole.destroy();
+            // Update the tables of occupants based on the bare and full JID
+            JID bareJID = userAddress.asBareJID();
+
+            nickname = leaveRole.getNickname();
+            List<MUCRole> occupants = occupantsByNickname.get(nickname.toLowerCase());
```
GregDThomas: Lock doesn't need to be taken until this point.

GregDThomas:
A more general comment; is there a specific problem you are trying to fix here? Or is it just a feeling you have that the code should be less multi-threaded in this area? Having looked at it again, if there is a problem you're being overly aggressive with the locks. 

If you're seeing problems with two people joining at the same time, I think appropriate use of Map.computeIfAbsent() is sufficient. 

If you're seeing problems with the same person joining twice (different nicknames perhaps?) then at most you'd want to lock that person, not the whole room - in which case the snappily named AutoCloseableReentrantLock will help.

cpetzka:
We faced some synchronization errors in cluster environments or if a user who is online is kicked from the room. While looking for a solution we found out, that the LocalMUCRoom is only partly synchronized (joinRoom with lock, occupantAdded without lock) and added the missing locks.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
cpetzka: done

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
cpetzka: done

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
cpetzka: done

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
cpetzka: because of the return Statement at https://github.com/igniterealtime/Openfire/pull/1429/files/06bc632d69818756e603c229464c4c763af9e0e1#diff-e7b0858be35ba3ff73325d038d049ed5R812 it can not be done in a single "compute" without local variables or throwing a exception.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -2331,28 +2328,30 @@ public Presence kickOccupant(JID jid, JID actorJID, String actorNickname, String
...
-        for (MUCRole kickedRole : occupants) {
-            // Add the actor's JID that kicked this user from the room
-            if (actorJID != null && actorJID.toString().length() > 0) {
-                Element frag = kickPresence.getChildElement(
-                        "x", "http://jabber.org/protocol/muc#user");
-                Element actor = frag.element("item").addElement("actor");
-                actor.addAttribute("jid", actorJID.toBareJID());
-                if (nick != null) {
-                    actor.addAttribute("nick", nick);
+        List<MUCRole> occupants = occupantsByNickname.get(kickPresence.getFrom().getResource().toLowerCase());
```
cpetzka: fixes:

	at java.util.ArrayList.<init>(ArrayList.java:178)
	at org.jivesoftware.openfire.muc.spi.LocalMUCRoom.kickPresence(LocalMUCRoom.java:2327)
	at org.jivesoftware.openfire.muc.spi.LocalMUCRoom.applyAffiliationChange(LocalMUCRoom.java:1878)
	at org.jivesoftware.openfire.muc.spi.LocalMUCRoom.addNone(LocalMUCRoom.java:1781)
	at org.jivesoftware.openfire.muc.spi.IQAdminHandler.handleItemsElement(IQAdminHandler.java:340)
	at org.jivesoftware.openfire.muc.spi.IQAdminHandler.handleIQ(IQAdminHandler.java:100)

akrherz:
@GregDThomas thanks for your review, have your requested changes been addressed?

Do we want this change for Openfire 4.4.1 ?

cpetzka:
no not necessarily

cpetzka:
is there anything else to do here?

GregDThomas:
Apologies, now I no longer get paid to work on Openfire - albeit indirectly - it's taking me longer to look at this (and other) PRs thatn I'd like. Conscious that you're getting in the PRs much quicker than I can look at them! Sorry, hope to get back to you soon.

GregDThomas:
Yes, looks good to me. Thanks again!

cpetzka:
> Apologies, now I no longer get paid to work on Openfire - albeit indirectly - it's taking me longer to look at this (and other) PRs thatn I'd like. Conscious that you're getting in the PRs much quicker than I can look at them! Sorry, hope to get back to you soon.

That's a pity.

With my question I didn't want you to hurry, I just wanted to know if I forgot something and to get an overview again. Sorry for the confusion.

Thanks for the review!

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -2331,28 +2328,30 @@ public Presence kickOccupant(JID jid, JID actorJID, String actorNickname, String
...
+        if (occupants != null) {
+            for (MUCRole kickedRole : new ArrayList<>(occupants)) {
```
cpetzka: Line 2333 creates a copy of an ArrayList. At this point, an error may occur if the list is changed at the same time, e.g. if the list is changed in the constructor during copying.

I would change the type of the list to a CopyOnWriteList so that this case cannot occur. Is there anything against it?

GregDThomas: That's a good plan

cpetzka: done

guusdk:
Thanks for all of your work on this!

guusdk:
I've created https://issues.igniterealtime.org/browse/OF-1930 to track this change in our issue tracker.

PR #1430, opened 2019-07-30
OF-1819 update the german translation

cpetzka:
This pull requests changes the german translation:
* every key from the english resource file is translated to german
* every non existitng key from the english file is deleted
* a little bit reordering (intellij adds all new keys to the end. They have been moved to the respective group, but not sorted in this group.)


wrooot:
Thanks. Created https://issues.igniterealtime.org/browse/OF-1819

PR #1432, opened 2019-08-01
DOC: reduce size of VSCode PNG screenshots

akrherz:
scale the width and height by half and apply some PNG compression.


guusdk:
Reducing the file size without loosing to much details seems acceptable to me. @mnsuccess what do you think?

PR #1435, opened 2019-08-01
OF-1826: Prevent NPE in new DataForms implementation.

guusdk:
The implementation introduced by OF-1809 causes collections of DataForm instances to be processed.

The new contract defines that a set must be present - null values are not allowed.

It's easy to overlook this new contract. For example, our own code returned a null value, instead of an empty collection.

This commit modifies the return of a null value (it now returns an empty value). Additionally, this PR adds a guard for problems like this, by explicitly checking of a value is null.


guusdk:
@mnsuccess - please review this PR.

PR #1436, opened 2019-08-02
OF-1828 Fix that an empty string can not be inserted in ofgroupporp.propvalue

cpetzka:
Hi,

please see https://discourse.igniterealtime.org/t/cannot-insert-null-into-openfire2-ofgroupprop-propvalue/52384


akrherz:
restarted travis-ci build enough times to get passed transient test failure :(

PR #1438, opened 2019-08-04
OF-1821: Ensure that the ldap.pagedResultsSize is automatically set for ActiveDirectory

GregDThomas:
NB. Turns out my local corporate AD has ~750,000 user entries. Turns out to be a tad more than I expected.


akrherz:
This did not cleanly pick onto 4.4 branch.  So will leave it in master for now.

PR #1440, opened 2019-08-05
OF-1021: Additional checks on (uploaded) plugin files

guusdk:
This PR adds three new checks that are applied to to-be-installed plugin files:

1. A check if the browser-reported 'content-type' header matches that of a JAR file. This check is disabled by default, as I suspect it'd be trivial for a malicious attacker to circumvent this check
2. A validation of the first few bytes of the uploaded data (the magic bytes). Enabled by default. If these do not match a recognized JAR/ZIP pattern, the uploaded file is not even being written do disc (further reducing any attack vector that involves sending malicious files).
3. Verification of the presence of a 'plugin.xml' entry in the JAR file. This primarily is a UX improvement, as the current code happily accepts _any_ jar file, without giving a strong visual indicator that the plugin installation failed.


akrherz:
Thanks @guusdk , do you want this considered for the 4.4.1 release?

guusdk:
Yes please.

PR #1446, opened 2019-08-08
Fix and improve docker build

Fishbowler:
**Fixes entrypoint.sh**

The absolute path of java didn't match of that of the JRE image.
Fixed to be agnostic, by using `which java`

**Improved maven caching during Docker builds**

Previously would only cache if the entire source was unchanged.
Now, by only copying over items explicitly required to resolve maven deps, will cache if all pom files remain unchanged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,17 @@
 FROM maven:3.5-jdk-8
 WORKDIR /usr/src
-COPY . .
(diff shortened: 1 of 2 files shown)


Fishbowler:
I wanted that too, but start-stop-daemon won't do it.

PR #1447, opened 2019-08-09
OF-1829: Prevent NPE when removing from a non-cached collection.

guusdk:
When multi-valued caches are used, removing an element from a collection that is not (or no longer) in the cache should not throw a null pointer exception.

I suspect that in clustered environments, where caches are managed by all cluster nodes, there's a chance of the same removal to be executed by each node on the
same cache, which could trigger an issue like this.

### DIFF
--- a/xmppserver/src/test/java/org/jivesoftware/util/CacheUtilTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/util/CacheUtilTest.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright (C) 2019 Ignite Realtime Foundation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


xmppserver/src/test/java/org/jivesoftware/util/CacheUtilTest.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.ArrayList;
+
+import static org.junit.Assert.*;
+
+/**
+ * Unit tests that verify the functionality of {@link CacheUtil}
+ *
+ * @author Guus der Kinderen, guus.der.kinderen@gmail.com
+ */
+public class CacheUtilTest
```
GregDThomas: General comment: There's no need for the try/catch. Any exception thrown that isn't an expectedException will cause the test to fail.

guusdk: Yeah ... but it fails with an error, not with an assertion failure.

Meh.

akrherz:
Should this be cherry picked for 4.4.2?

akrherz:
> Should this be cherry picked for 4.4.2?

Done

PR #1452, opened 2019-08-15
Update OpenfireX509TrustManager.getAcceptedIssuers

nepozbremser:
Add new property xmpp.client.certificate.sendtrustedissuerlist (true, false) to manage, if 
the CertificateRequest message, sent to the client, has the list of CAs (distinguished names in Server Hello).

The property xmpp.client.certificate.sendtrustedissuerlist can be inserted as constant into ConnectionSettings.java
"...src\xmppserver\src\main\java\org\jivesoftware\openfire\session\ConnectionSettings.java"

CertRequest

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/keystore/OpenfireX509TrustManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/keystore/OpenfireX509TrustManager.java
@@ -112,6 +112,11 @@ public class OpenfireX509TrustManager implements X509TrustManager
     @Override
     public X509Certificate[] getAcceptedIssuers()
     {
+        if (JiveGlobals.getBooleanProperty("xmpp.client.certificate.sendtrustedissuerList", false)) {
+            // Answer an empty list 
+            return new X509Certificate[0];
+        }
+        
         final Set<X509Certificate> result;
         if ( checkValidity )
         {
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/keystore/OpenfireX509TrustManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/keystore/OpenfireX509TrustManager.java
@@ -3,7 +3,7 @@ package org.jivesoftware.openfire.keystore;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import org.jivesoftware.util.JiveGlobals;
 import javax.net.ssl.*;
 import java.security.*;
 import java.security.cert.*;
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks for your contribution! It is always great to see new people help us with improving the project!

I am having concerns:

- A minor concern is that you're using the 'old way' to read system properties. I'd prefer if instead you'd use the new `SystemProperty` class. It's a little harder to use, but it offers better functionality for the end user. An example can be found here: https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/main/java/org/jivesoftware/admin/AuthCheckFilter.java#L52-L58 - please use the `SystemProperty` class, instead of using `JiveGlobals.getBooleanProperty`
- Although I don't think existing functionality will break because of this change, that's almost by accident so. `javax.net.ssl.X509TrustManager#getAcceptedIssuers` by contract must return all accepted issuers, which is no longer true if your patch was applied and the boolean flipped. The only reason that things continue to work, is that the consumer of the value (`org.jivesoftware.openfire.keystore.OpenfireX509TrustManager#checkChainTrusted`) uses the value from a field, instead of the return value from the method. That's nothing more than coincidence, and should not be relied upon. Future developers are likely to 'fix' the code, which would cause a complete breakdown in TLS in servers that have your change enabled.
- On top of the above, even if the code change works with the existing Openfire code in our project, third-party development (plugins, etc) might depend on `#getAcceptedIssuers` to always return a list of issuers, as it is supposed to.
- I wonder if this change addresses the root cause of your problem. What is that exactly? You seem to want to prevent the server from sending a list of CAs that it trusts. The only reason I can think of that it does this in the first place, is to allow the client to determine what client certificate it can use in mutual authentication. You'd have to consult the TLS specification to be sure, but I'd not be surprised if the underlying problem that you're trying to solve goes away if you disable "mutual authentication" (I'm guessing it is not disabled on your server). See screenshot: 
image

nepozbremser:
thanks to akrherz, he showed me this possibility with PK

- I have not tested this yet.

- I deactivate Mutual Authentication in Openfire.
STARTTLS is activ.

The hole Certificate Request is not in the server hello now.
Some devices, that communicated before, cannot communicate now.

grafik

guusdk:
> One device type has internal buffer overflow because of "Distinguished Names"

This sounds as if the problem is with the software in that device, not with Openfire. If your clients depend on mutual authentication, then I fear that your suggested fix might not work.

It might be better to debug the device, and see exactly why it is having that buffer overflow problem.

nepozbremser:
Device update is the best option.
Client want to solve this on the portal side.

akrherz:
Hi @nepozbremser, do you have any current comments on this PR?  I don't think there is much appetite for merging it unless some more convincing details arrive here...

guusdk:
This PR seems stale. As-is, I'm not inclined to accept it for reasons described in earlier comments.

PR #1453, opened 2019-08-16
OF-1835: Prevent NPE when resuming faulty SM session

guusdk:
The error handler of SM sets the namespace to null, causing an NPE when the client tries to resume that session. This in turn causes the socket to be closed abruptly, which triggers a reconnect loop in Conversations.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
@@ -216,10 +216,12 @@ public class StreamManager {
         this.namespace = namespace;
         // Ensure that resource binding has NOT occurred.
         if (!allowResume() ) {
+            Log.debug("Unable to process session resumption attempt, as session {} is in a state where session resumption is not allowed.", session);
             sendUnexpectedError();
             return;
         }
         if (session.getStatus() == Session.STATUS_AUTHENTICATED) {
+            Log.debug("Unable to process session resumption attempt, as session {} is not authenticated.", session);
             sendUnexpectedError();
             return;
         }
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
```diff
@@ -269,6 +269,11 @@ private void startResume(String namespace, String previd, long h) {
             sendError(new PacketError(PacketError.Condition.unexpected_request));
             return;
         }
+        if (otherSession.getStreamManager().namespace == null) {
+            Log.debug("Not allowing a client to resume a session, the session to be resumed disabled SM functionality as a response to an earlier error." );
```
akrherz: could any details about the client be included in this message or does that automagically happen already with the logger?

guusdk: That's a good suggestion. I'll add that.

xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
```diff
@@ -262,41 +265,41 @@ private void startResume(String namespace, String previd, long h) {
             sendError(new PacketError(PacketError.Condition.item_not_found));
             return;
         }
-        Log.debug("Found existing session, checking status");
+        Log.debug("Found existing session for '{}, checking status", fullJid);
```
akrherz: missing a closing `'` here, sorry.

guusdk: force-pushed fix. No-one will ever find out.

PR #1454, opened 2019-08-19
OF-1251: Harden admin console for XSS and CSRF

guusdk:
This mainly applies to the setup pages: most (if not all) pages now use JSTL, which makes it easier to apply proper XML escaping when printing value. Additionally, CSRF checks have been added to pages where config changes are applied.


akrherz:
Wow @guusdk, this is thorough.  Thank you so much for efforting it!

PR #1455, opened 2019-08-20
OF-1837: JDBC Providers: allow JIDs to be (un)escaped

guusdk:


guusdk:
This commit adds new configuration options, but leaves the default behavior intact.

I'm considering toggling the default, as I suspect that it will help a lot of people that are now unaware of encoding issues. It can, however, cause issues for people that already have encoding in place, in their database. I'm not sure what outweighs what here.

guusdk:
FWIW, we've tested a backported version of this change on wikisuite.chat, that had a user with a space in the username. The change seems to prevent many stacktraces in the log files.

guusdk:
Auch - why is this failing in Travis, but not for me locally?

mnsuccess:
> Auch - why is this failing in Travis, but not for me locally?

failing on the test JDBCAuthProviderTest !! maybe something  wrong with encryption of the password according to the PasswordTypes of JDBCAuthProvider

guusdk:
> failing on the test JDBCAuthProviderTest !! maybe something wrong with encryption of the password according to the PasswordTypes of JDBCAuthProvider

Maybe - but it's strange that when I run the same tests at my computer, they all pass without problem. Can you reproduce the test failures on your computer?

GregDThomas:
What char encoding are you using on your computer?
What char encoding are you using Travis?

mnsuccess:
> > failing on the test JDBCAuthProviderTest !! maybe something wrong with encryption of the password according to the PasswordTypes of JDBCAuthProvider
> 
> Maybe - but it's strange that when I run the same tests at my computer, they all pass without problem. Can you reproduce the test failures on your computer?

okay !! Let me test

mnsuccess:
> > failing on the test JDBCAuthProviderTest !! maybe something wrong with encryption of the password according to the PasswordTypes of JDBCAuthProvider
> 
> Maybe - but it's strange that when I run the same tests at my computer, they all pass without problem. Can you reproduce the test failures on your computer?

strange the test does not fail when I built the source code with **mvn verify** and it fails when I run it directly ( failed 5)

guusdk:
I'm absolutely lost here.

guusdk:
I've rebased this PR just now.

akrherz:
The Travis-CI failures here look germane to the PR...

guusdk:
I agree - but I'm having trouble finding the cause. It is frustratingly error free on my local machine.

@GregDThomas help!

guusdk:
I've rebased this PR just now.

guusdk:
In an attempt to isolate the problem, I've replaced this on PR with three PRs, that each add one of the commits that were combined here.

These are the new PRs:
- #1643
- #1644 
- #1645

PR #1459, opened 2019-08-24
OF-1839 : add XEP-0411 support

mnsuccess:
Openfire already supports conversion between 'PEP' and 'Private Storage' type of bookmarks, but this support predates XEP-0411.

Openfire's support should match XEP-0411.


mnsuccess:
> This might very well be everything that's needed to complete the implementation. Did you check if conversion actually works as described by the XEP? Did you check if the feature can be disabled? If it can be disabled, then we should not add this feature when disabled.

the conversion is working, add feature only when the Private Storage is enabled

guusdk:
Excellent, thanks!

PR #1460, opened 2019-08-24
OF-1823 Admin console design improvements

mnsuccess:
The Openfire administration console (including the setup wizard) is one of the first things that administrators see of Openfire. It's user experience affects only a small group of people, but those are likely to include possibly decision makers. It would be good to offer a good user experience to those.

The design of the Openfire admin console is dated. It should be improved.

Important requirements:

The admin console must be usable by a very wide array of (desktop-based) browsers. Note that, in larger organisations, people might be forced to use specific (often old) browsers).
The admin console can be extended through plugins (see the [Plugin Developer Guide|http://download.igniterealtime.org/openfire/docs/latest/documentation/plugin-dev-guide.html])
It must remain possible to add admin console pages through plugins. If the way to do this changes, then the documentation linked above should be modified to reflect these changes.
The admin console pages of existing/old plugins (that have "old style" designed admin pages) must remain functional.
The administrative console functionality has important security aspects. The amount of attack vectors should be kept at a minimum. The implementation of the new design should not include unnecessary libraries or frameworks - neither home-grown or third-party provided. It is perfectly acceptable to include some libraries, but we should keep the code base concise.
Nice-to-haves:

Support for mobile phones


guusdk:
Can you include screenshots, for people like me that are lazy? 😄

guusdk:
What do we need jQuery for? I'd possible, is like to avoid using libraries like that. They require maintenance, and offer much functionality, some of which could be abused by malicious parties.

mnsuccess:
> What do we need jQuery for? I'd possible, is like to avoid using libraries like that. They require maintenance, and offer much functionality, some of which could be abused by malicious parties.

right !! let remove jQuery

jesuscristincarbajal:
fix it

mnsuccess:
Screenshot 2019-09-02 at 12 21 25
Screenshot 2019-09-02 at 12 25 26
Screenshot 2019-09-02 at 12 19 48
Screenshot 2019-09-02 at 12 20 53

akrherz:
Greetings, What is the status of this PR?  It would be good to get something like this in early for the 4.5.0 release process so there is more chance to find edge cases.  This is ready to leave draft status?

akrherz:
I built this branch now and fired up Openfire to look for bugs.  I did not find any so far.  What's left to get this out of draft status?  One thing I thought was discussed was to place the server's `xmpp.domain` on the login page.  Is that still on the TODO list here?

guusdk:
Let's add further improvements through new PRs.

Thanks for your hard work on this, @mnsuccess !

PR #1461, opened 2019-08-25
OF-1842 update SSL guide

wrooot:


guusdk:
This PR is 'done' right? JIRA reports that it is 'under review' - I'm unsure what that means.

wrooot:
Jira sometimes can't understand what is happening on GitHub side. I saw similar issues before. I guess this should be fine.

PR #1462, opened 2019-08-26
4.1.5-based Pubsub improvements

guusdk:
This PR should probably not be merged.

This includes improvements to the pubsub implementation, that was originally developed against v4.1.5 of Openfire (I'm providing this PR to compare the original changes with the code base it was originally based on).

The primary result of this PR is to introduce a pubsub implementation that does not use the database, but is in-memory only.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -28,8 +28,8 @@
     <property name="version.revision" value="5"/>
     <property name="version.extra" value=""/>    <!-- For 'beta' or 'alpha' -->
 
-	<property name="javac.source" value="1.7"/>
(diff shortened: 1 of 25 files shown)


akrherz:
@guusdk What's the status/future of this PR?  Thanks!

guusdk:
The feature that's added in this PR are now being used in 4.5.x in the proprietary solution that I've based the commits in this PR on. I'll see if I can either efficiently rebase this PR, or if I can replace altogether.

guusdk:
This PR is replaced by one that is based on the current HEAD of the master branch: #1596

PR #1464, opened 2019-08-30
[Fix] The interceptor could not intercept a domain that is null

eric-ampire:
https://issues.igniterealtime.org/browse/OF-1845


akrherz:
Thanks @eric-ampire for the contribution.  In the future, please consider putting the Jira issue in the commit message to make for an easy cross reference when reviewing the git history.

eric-ampire:
Okay great

guusdk:
Thanks!

PR #1468, opened 2019-09-02
OF-1838 XEP-0157: Improve support

mnsuccess:
XEP-0157 support in Openfire should be improved upon. Ideally, the admin console is used to determine what users are exposed as what kind of XEP-0157 contact.

### DIFF
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -2815,6 +2815,27 @@ user.groups.form.update=User groups edited successfully.
 # User Contact Adresses Page
 
 user.contact_addresses.title=Contact Addresses
+user.contact_addresses.info=Below is a list of the contact adresses.click the edit icon next to the contact to edit.
+user.contact_addresses.info_create=Use the form below to create a new contact address.
+user.contact_addresses.contact_type=Contact Type
+user.contact_addresses.address_type=Address Type
+user.contact_addresses.address_value=Address Value
+user.contact_addresses.description=Description
+user.contact_addresses.edit=Edit
+user.contact_addresses.search_all=All
+user.contact_addresses.search_admin=Admin
+user.contact_addresses.search_abuse=Abuse
+user.contact_addresses.search_feedback=Feedback
+user.contact_addresses.search_sales=Sales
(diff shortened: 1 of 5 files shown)


guusdk:
Do we need a servlet for this? It's not a bad approach, but we hardly use servlets, and do everything in JSP. If there's a reason for having a servlet, I'm not against it.

GregDThomas:
I've used servlets in a couple of places (e.g. the new System properties page). Possibly my bad for not highlighting that at the time, but ...

I'd encourage use of servlets. It does make things far neater, IMHO. A much cleaner seperation of logic and display, less likely to get CSRF problems, etc. etc.

deleolajide:
I use both in plugins and would like to keep both options open. Slightly different tools to achieve the same purpose.

guusdk:
I agree that using servlets is generally a good idea. In this instance, I'd like to keep the code somewhat consistent.

I'd be happy to introduce Servlets, but then I'd ask you to introduce them everywhere. :)

GregDThomas:
I'd like to see servlets everywhere, but practically speaking I think the only way you're going to get it done is a few at a time.

guusdk:
Three years in, we're still halfway through JSTL :(

guusdk:
@mnsuccess why prefix every column name with the letter 'c'?

mnsuccess:
> @mnsuccess why prefix every column name with the letter 'c'?
 prefix c as a contact. I wanted to prefix with the word contact it seemed long then I just prefixed it with c.  maybe we don't need to prefix column

guusdk:
Ah I see. It's a bit redundant as there will only be contacts in that table anyway, but hey :)

mnsuccess:
Should I keep using  servlet ?

guusdk:
Using servlets is fine, but please stop using the letter `c` as the first letter of each database column name.

akrherz:
@mnsuccess Are you still working on this PR?

mnsuccess:
@akrherz yes I'm still working on this !!

akrherz:
Hi @mnsuccess, are you able to rebase this PR onto current master?

Neustradamus:
@mnsuccess: Have you progressed?

PR #1470, opened 2019-09-03
OF-793: Fixes for 'unsupported record version' exception

guusdk:
This isn't the nicest of fixes, but my tests started to fail after attempting to improve on this. As a result, I'm providing this as-is, as I suspect it's a functional improvement.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/TLSWrapper.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/TLSWrapper.java
@@ -141,6 +141,8 @@ public class TLSWrapper {
         out = resizeApplicationBuffer(out);// guarantees enough room for unwrap
 
         // Record a hex dump of the buffer, but only when logging on level 'debug'.
+        // Create the dump before the buffer is being passed to tlsEngine, to ensure
+        // that the original content of the buffer is logged.
         String hexDump = null;
         if ( Log.isDebugEnabled() )
         {
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/TLSWrapper.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/TLSWrapper.java
@@ -17,7 +17,9 @@
 package org.jivesoftware.openfire.net;
 
 import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
 import java.security.*;
+import java.util.Arrays;
 
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLEngineResult;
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/net/TLSWrapper.java
```diff
@@ -136,10 +139,21 @@ public void enableLogging(boolean logging) {
     public ByteBuffer unwrap(ByteBuffer net, ByteBuffer app) throws SSLException {
         ByteBuffer out = app;
         out = resizeApplicationBuffer(out);// guarantees enough room for unwrap
+
+        // Record a hex dump of the buffer, but only when logging on level 'debug'.
+        String hexDump = null;
+        if ( Log.isDebugEnabled() )
```
GregDThomas: If debug is enabled, hexDump is calculated - but only ever used if the exception occurs. Is it worth moving this section to the exception handler?

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
```diff
@@ -680,6 +680,9 @@ private VerifyResult sendVerifyKey(String key, StreamID streamID, String recipie
                 doc = reader.parseDocument();
             } catch (DocumentException e) {
                 log.warn("Unable to verify key: XML Error!", e);
+                // Close the stream
+                writer.write("</stream:stream>");
+                writer.flush();
```
GregDThomas:
```
writer.write("</stream:stream>");
writer.flush();
```
Appears a number of times in this file. I wonder if it's worth creating a utility method, `closeStreamAndFlush(writer)` so this is all captured in one place?

xmppserver/src/main/java/org/jivesoftware/openfire/net/TLSWrapper.java
guusdk: The intend is to capture the buffer before it gets touched (possibly modified) by the TLS handling.

GregDThomas: A comment to that effect would probably help prevent this being refactored in error in the future ...

GregDThomas:
Nothing major, just a few style comments, but seems OK.

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
guusdk: That's not reducing complexity, that's just moving complexity around :)

PR #1471, opened 2019-09-03
OF-1849 allow startls on directtls port

guusdk:


guusdk:
The other way around (allowing directTLS on the startTLS port) might also make sense - but it's a bridge to far, effort-wise, for me today.

PR #1473, opened 2019-09-06
OF-1852: Improve the handling of unprocessable presences and messages

guusdk:


xmppserver/src/main/java/org/jivesoftware/openfire/MessageRouter.java
```diff
@@ -246,30 +246,76 @@ public void routingFailed( JID recipient, Packet packet )
...
 
         if ( msg.getType().equals( Message.Type.chat ) && serverName.equals( recipient.getDomain() ) && recipient.getResource() != null ) {
             // Find an existing AVAILABLE session with non-negative priority.
             for (JID address : routingTable.getRoutes(recipient.asBareJID(), packet.getFrom())) {
                 ClientSession session = routingTable.getClientRoute(address);
                 if (session != null && session.isInitialized()) {
                     if (session.getPresence().getPriority() >= 1) {
-                        storeOffline = false;
+                        // If message was sent to an unavailable full JID of a user then retry using the bare JID.
+                        routingTable.routePacket( recipient.asBareJID(), packet, false );
```
akrherz: This makes me nervous. what happens in the case of route-all-resources or is this code never hit in that case?

guusdk: The very specific line that you mention this on should not be functionally different than before. I intended to remove the `storeOffline` variable, without affecting (that part) of the handling. Did I miss something?

PR #1476, opened 2019-09-07
OF-1855: S2S: Allow Mutual Authentication by default when using signed certificate

guusdk:


guusdk:
Hmm, this change did not cause the s2s clientauth policy to switch to 'wanted' on igniterealtime.org (which does not have a property set, and which has a properly signed certificate. @GregDThomas can you spot my mistake perhaps?

GregDThomas:
Double checked. It looks good (other than previous comment). I can only suggest adding logging statements to see where it’s not behaving as expected.

guusdk:
> You’re turning a constant to a list of one and then seeing if that list contains something. What’s wrong with a simple equality check

Premature optimization: I expect more types to be included in that collection eventually

PR #1477, opened 2019-09-07
OF-1856: Fix isAnonymous detection

guusdk:
The isAnonymous check should not depend on the fact that the provided JID is a full JID. This will give incorrect results ('not anonymous') if the provided JID is bare.

This commit checks for any anonymous session that matches the bare JID - if there's one then that's enough (a user cannot be both anonymous and not anonymous).


guusdk:
The cache should only contain JIDs of the local domain. I'm not sure if the values (per full JID) are used for anything. Switching to a multimap would be a rather large change. These caches might be clustered, in which case retaining the structure would retain compatibility between cluster nodes that are updating from one version of Openfire to another.

PR #1479, opened 2019-09-08
OF-1858: Ensure that adding/deleting MUC services is cluster aware

GregDThomas:
This is an odd one; 
* Updating MUC services seems to be OK - handled by the ServiceUpdatedEvent cluster task.
* Cluster tasks exist for ServiceAddedEvent, ServiceRemovedEvent, but were never used.
* There was also a bug in the above two which meant that updating a MUC on node 1 would update the MUC on node 2 which would update the MUC on node 1 which would ... well, you get the idea. Perhaps this is why they were never used?

**NOTE:** I have been unable to test this in a cluster (!). I think it's OK, but it would probably be a useful exercise to do so.


guusdk:
The code change seems sensible. Is someone in a position to test this in a cluster?

PR #1481, opened 2019-09-11
OF-1865: Implement a GroupBasedAdminProvider

GregDThomas:
To use, simply set the property `provider.group.groupAdminProvider.groupName` to the name of the Openfire group (default ``), and set the System Property `provider.admin.className` to the value `org.jivesoftware.openfire.admin.GroupAdminProvider`

NB. Requires a restart


guusdk:
To shave a yak: I read this as something that made someone an admin of a group. Maybe rename it as 'GroupbasedAdminProvider'?

GregDThomas:
Sure thing, will look at renaming it.

GregDThomas:
Done. 
Class is now `org.jivesoftware.openfire.admin.GroupBasedAdminProvider`
Property is now `provider.group.groupBasedAdminProvider.groupName`

PR #1482, opened 2019-09-13
OF-1860: Check plugin archive structure.

guusdk:
Openfire should verify that the structure of the plugin archive does not include files that are expanded outside of the target directory.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginMonitor.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginMonitor.java
@@ -24,21 +24,7 @@ import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Comparator;
-import java.util.Deque;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.SortedSet;
-import java.util.StringTokenizer;
-import java.util.TreeSet;
+import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginMonitor.java
```diff
@@ -446,16 +435,32 @@ public Integer call() throws Exception
...
+                // Protect against zip-slip (before applying any file-system modifications).
+                if ( JiveGlobals.getBooleanProperty( "plugins.loading.zipslipDetection.enabled", true ) )
+                {
+                    for ( Enumeration e = zipFile.entries(); e.hasMoreElements(); )
+                    {
+                        JarEntry entry = (JarEntry) e.nextElement();
+                        Path entryFile = dir.resolve( entry.getName() );
+                        if ( !entryFile.normalize().toAbsolutePath().startsWith( dir.normalize().toAbsolutePath() ) )
+                        {
+                            throw new RuntimeException( "Plugin contains content that is outside of target plugin directory (possible zipslip attack)" );
```
akrherz: would this exception tell the user which plugin had content outside the target?

guusdk: It does so indirectly. This exception is caught a few lines down, and logged as:

> An exception occurred while trying to extract plugin '{}'

Fishbowler:
Worth doing PluginManager.deletePlugin since the jar isn't going to change?

What's the value difference in using RuntimeException when the other failures in PluginMonitor are Log.error('nope') and return false?

guusdk:
> Worth doing PluginManager.deletePlugin since the jar isn't going to change?

It'd remove a JAR, which likely is a deliberate attempt to cause mischief. I thought it'd be more interesting to keep it around for analysis, than to guard against spamming the logs. I'll be easily persuaded otherwise though.
 
> What's the value difference in using RuntimeException when the other failures in PluginMonitor are Log.error('nope') and return false?

Not much, other than more specific stack traces.

xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginMonitor.java
```diff
@@ -286,7 +272,10 @@ public boolean accept( final Path path ) throws IOException
                             // If the JAR needs to be exploded, do so.
                             if ( Files.notExists( dir ) )
                             {
-                                unzipPlugin( canonicalPluginName, jarFile, dir );
+                                if (!unzipPlugin( canonicalPluginName, jarFile, dir ) )
+                                {
+                                    return;
```
Fishbowler: Does this want to be a continue, so that later in the runnable plugins can still be unloaded, loaded, etc?

guusdk: oh, good catch!

PR #1485, opened 2019-09-18
Hello, tinder-2.0.0.jar Maven repository does not have this version

zbh924:
I didn't find tinder-2.0.0 dependency in maven. Where do tinder-2.0.0.jar need to download?


wrooot:
To ask questions use the forums. Or at least open an issue, not a pull request.

PR #1486, opened 2019-09-20
OF-1251: Expose time elapsed formatting as taglib function.

guusdk:
This allows for the formatting function to be used with JSTL.


guusdk:
Kind of weird that this change needs to be applied in three places, but this works. I'd like this to go into 4.4.x, to be able to use this new feature in a plugin.

GregDThomas:
I suspect at least one, if not two, these admin.tld files is superflous. They are probably a hang-over from the Great Maven Update (TM).

Greg

PR #1490, opened 2019-09-24
OF-1876: Upgrade Mysql Connector/J to 8.0.17

guusdk:


GregDThomas:
Only slightly related, but I recently spotted that Oracle are now (finally) publishing their JDBC driver; https://medium.com/@kuassimensah/oracle-jdbc-drivers-on-maven-central-64fcf724d8b

PR #1497, opened 2019-10-03
OF-1885: SSRF guard favicon

guusdk:


GregDThomas:
This is good as far as it goes; however I wonder if it could be locked down further? From memory, the favicon is used when displaying S2S sessions. How about restricting the FavIcon servlet to only fetch icons from servers from which a S2S session exists? And the favicon.ico at that?

guusdk:
I've considered that. I'm wondering if the added complexity outweighs the added security benefits. What do you think?

GregDThomas:
Having pondered, I worry that even if we're not displaying it, the admin console will happily fetch any file off any HTTP server it has access to. If it were me, I'd probably lean towards changing the way this works such that the servlet 
(a) Only fetches favicon's from S2S connected servers, and
(b) Only fetches the favicon

guusdk:
Yeah, you're right. That sounds better to me too. I'm not sure yet if I'll
have the time to do this right now. I might raise a ticket as a future
improvement to this.

akrherz:
> I might raise a ticket as a future improvement to this.

Filed [OF-1902](https://issues.igniterealtime.org/browse/OF-1902)

PR #1499, opened 2019-10-03
OF-1887 update postgresql JDBC to 42.2.8

akrherz:


akrherz:
version 42.2.6 added postgresql 12 support.

PR #1502, opened 2019-10-07
OF-1890 Added support for flatting nested groups in LDAP.

MarcelHeckel:
Nested / complex / hierarchic groups are flattened. This means: if
group A is member of group B, the members of group A will also be
members of group B.

Cyclic paths are detected and followed only once.

The features is disable by default.
Set property "ldap.flattenNestedGroups" to "true" to enable it.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
@@ -27,18 +27,43 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xmpp.packet.JID;
 
-import javax.naming.*;
-import javax.naming.directory.DirContext;
-import javax.naming.directory.InitialDirContext;
-import javax.naming.directory.SearchControls;
-import javax.naming.directory.SearchResult;
-import javax.naming.ldap.*;
-import javax.net.ssl.SSLSession;
 import java.io.Serializable;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Hashtable;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.StringTokenizer;
+
+import javax.naming.CompositeName;
+import javax.naming.Context;
+import javax.naming.InvalidNameException;
+import javax.naming.Name;
+import javax.naming.NamingEnumeration;
+import javax.naming.NamingException;
+import javax.naming.directory.Attribute;
+import javax.naming.directory.Attributes;
+import javax.naming.directory.DirContext;
+import javax.naming.directory.InitialDirContext;
+import javax.naming.directory.SearchControls;
+import javax.naming.directory.SearchResult;
+import javax.naming.ldap.Control;
+import javax.naming.ldap.LdapContext;
+import javax.naming.ldap.LdapName;
+import javax.naming.ldap.PagedResultsControl;
+import javax.naming.ldap.PagedResultsResponseControl;
+import javax.naming.ldap.Rdn;
+import javax.naming.ldap.SortControl;
+import javax.naming.ldap.StartTlsRequest;
+import javax.naming.ldap.StartTlsResponse;
+import javax.net.ssl.SSLSession;
 
 /**
  * Centralized administration of LDAP connections. The {@link #getInstance()} method
@@ -63,6 +88,7 @@ import java.util.*;
  *      <li>ldap.groupDescriptionField</li>
  *      <li>ldap.posixMode</li>
  *      <li>ldap.groupSearchFilter</li>
+ *      <li>ldap.flattenNestedGroups</li>
  *      <li>ldap.debugEnabled</li>
  *      <li>ldap.sslEnabled</li>
  *      <li>ldap.startTlsEnabled</li>
@@ -174,6 +200,10 @@ public class LdapManager {
         instance = new LdapManager(properties);
     }
 
+    /** Exposed for test use only */
+    public static void setInstance(LdapManager instance) {
+        LdapManager.instance = instance;
+    }
 
     private Collection<String> hosts = new ArrayList<>();
     private int port;
@@ -203,6 +233,7 @@ public class LdapManager {
     private String groupDescriptionField;
     private boolean posixMode;
     private String groupSearchFilter;
+    private boolean flattenNestedGroups;
 
     private final Map<String, String> properties;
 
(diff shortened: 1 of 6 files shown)


mightymop:
Nice, it is for shared roster right? did you test this scenarios:
-P1 and P2 are members of Group A, 
-P3 and P4 are members of Group B,
-Group B is also member of Group A.
> so we have a "little" tree.

Will P1 and P2 see P3 and P4?  
Will P3 and P4 see P1 and P2, 'cause P1 and P2 are in a "higher" group?

MarcelHeckel:
> Will P1 and P2 see P3 and P4?
Will P3 and P4 see P1 and P2, 'cause P1 and P2 are in a "higher" group?

P1 , P2, P3, P4 will be in Group A
P3, P4 will be in Group B

So when sharing Group A on the roster, all four will see each other.

I tested normal and posix mode. Also cyclic graphs. The test LDAP structure is attached 
[users.zip](https://github.com/igniterealtime/Openfire/files/3701622/users.zip)

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
```diff
@@ -377,6 +381,12 @@ public LdapManager(Map<String, String> properties) {
         }
         groupSearchFilter = properties.get("ldap.groupSearchFilter");
 
+        flattenNestedGroups = false;
+        String flattenNestedGroupsStr = properties.get("ldap.flattenNestedGroups");
+        if (flattenNestedGroupsStr != null) {
```
GregDThomas: Nitpicky, but no need for the null check as Boolean.valueOf does the right thing in this case.

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
```diff
@@ -2142,6 +2247,144 @@ public void setIsEnclosingDNs(boolean enable) {
...
+     * in the string before it is used as a string.
+     *
+     * @param attribute             LDAP attribute to be pulled from each result and placed in the return results.
+     *                              Typically pulled from this manager. Null means the the absolute DN is returned.
+     * @param searchFilter          Filter to use to perform the search.  Typically pulled from this manager.
+     * @param failOnMultipleResults It true, an {@link IllegalStateException} will be thrown, if the
+     *                              search result is not unique. If false, just the first result will be returned.
+     * @return A single string.
+     */
+    public String retrieveSingle(String attribute, String searchFilter, boolean failOnMultipleResults) {
```
GregDThomas: I /think/ this is returning a single attribute of a search result? If so, do you think you could rename it (and it's overloaded method) retrieveSingleAttribute.

Also, what's the (practical) use case for failOnMultipleResults = false? It looks to me like it's only ever set to true, not sure I can see a reason why false might be useful.

GregDThomas: (also, should these methods be private? not sure they need to be exposed outside this class)

GregDThomas:
Not tested it, but this broadly looks OK to me - but don't take that as a definitive answer!

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
MarcelHeckel: Yes you are right. But all other boolean properties are handled like this. E.g `"ldap.debugEnabled"` or `"ldap.sslEnabled"`. So it is just consistent when it is done like this.

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
MarcelHeckel: I used this name because the other function for multiple values of an attribute is called `retrieveList` and not `retrieveListAttribute`. I just want to be consistent. Should I still rename it?

Yes, `failOnMultipleResults` is always true until now. But this function is a "Generic routine for retrieving a single element from the LDAP server.". So maybe later someone may be interested in just one value of the attribute, even if there are multiple values.

It is `public` because it is a tool function like `retrieveList`. At minimum it could be package protected as it is used by the `LdapGroupProvider`.

GregDThomas:
No; leave it as it is; I"m just being overly picky,

guusdk:
Not realizing the overlap with this PR, I've merged a similar, but not identical, change in https://github.com/igniterealtime/Openfire/pull/1520 With that, is this PR still needed? If it is, kindly rebase it to the current master.

MarcelHeckel:
@guusdk No this MR is not about create hierarchies of groups, but flatting hierarchies existing in LDAP. So that there a no groups in groups.

@GregDThomas Will this feature be merged sometime in future? Because l don't like to handle conflicting again and again...

guusdk:
From my perspective, https://github.com/igniterealtime/Openfire/pull/1520 adds flattening to _all_ groups, not just the ones in LDAP. Does that overlap with this PR?

GregDThomas:
If this is still required (see also #1520), then I see no reason it coudn't be merged once it is rebased - but I've limited involvement now, so it's really Guus who'd probably be the arbiter of such things.

MarcelHeckel:
I think this MR handles something different. As I understand, this new property called "sharedRoster.groupList" is for sharing a roster group to another roster group? First, the property seams not to be filled with data from LDAP. Second: My use case is: I want to have a flattened group, not a flattened roster group, as we also use groups for rights management (so they are not shared on any roster).

Please correct me, if I'm wrong, but I don't see the following use case coved with "sharedRoster.groupList": A LDAP group A has a "memberOf" property that points to another LDAP group B. Will member of B be members of group A?

guusdk:
Ah, something starts to dawn on me. This PR will "flatten" nested groups based on their LDAP definition. Openfire has a similar but distinct definition of groups, that could also be nested, but there's probably no overlap here.

How does the code prior to this PR behave when it encounters LDAP nested groups?

MarcelHeckel:
It just ignored the entries that did not match to an user.

guusdk:
But the referenced groups are groups themselves, aren't they? Would they be imported as separate groups? And, if so, do we now get duplicated (sub)groups with this PR?

MarcelHeckel:
See start of discussion:

-P1 and P2 are members of Group A,
-P3 and P4 are members of Group B,
-Group B is also member of Group A.

Result:
- P1 , P2, P3, P4 will be in Group A
- P3, P4 will be in Group B

> But the referenced groups are groups themselves, aren't they? 
Yes

> Would they be imported as separate groups? 
Yes

> And, if so, do we now get duplicated (sub)groups with this PR?
Not sure what you mean.

guusdk:
Nevermind, it's late and my mind is seeing things that aren't there. I'd be happy to merge the functionality.

MarcelHeckel:
Fine. I'm currently writing some UnitTests....

guusdk:
Thanks for all the work that you've put into this one, and your continued following up!

PR #1504, opened 2019-10-13
OF-1894: Allow the contents of Cache's to be kept private

GregDThomas:
Note; this PR deliberately does not use system properties to determine if the contents of a cache should be displayed on there admin UI. If the developer does not wish a cache to be displayed on the admin UI, then it should not be possible for an administrator to override their wishes by changing a system property.

An implication of this is that the code must explicitly set the key and/or value to be secret.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/admin/servlet/SystemCacheDetailsServlet.java
+++ b/xmppserver/src/main/java/org/jivesoftware/admin/servlet/SystemCacheDetailsServlet.java
@@ -46,20 +46,23 @@ public class SystemCacheDetailsServlet extends HttpServlet {
             request.setAttribute("warningMessage", LocaleUtils.getLocalizedString("system.cache-details.cache_not_found", Collections.singletonList(cacheName)));
         }
 
+        final boolean secretKey = optionalCache.map(Cache::isKeySecret).orElse(Boolean.FALSE);
+        final boolean secretValue = optionalCache.map(Cache::isValueSecret).orElse(Boolean.FALSE);
+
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/admin/servlet/SystemCacheDetailsServlet.java
```diff
@@ -46,20 +46,23 @@ protected void doGet(final HttpServletRequest request, final HttpServletResponse
...
         }
 
+        final boolean secretKey = optionalCache.map(Cache::isKeySecret).orElse(Boolean.FALSE);
+        final boolean secretValue = optionalCache.map(Cache::isValueSecret).orElse(Boolean.FALSE);
+
         final List<Map.Entry<String, String>> cacheEntries = optionalCache.map(Cache::entrySet)
             .map(Collection::stream)
             .orElseGet(Stream::empty)
-            .map(entry -> new AbstractMap.SimpleEntry<>(entry.getKey().toString(), entry.getValue().toString()))
+            .map(entry -> new AbstractMap.SimpleEntry<>(secretKey ? "************" : entry.getKey().toString(), secretValue ? "************" : entry.getValue().toString()))
```
guusdk: I have no experience with swapping out SimpleEntry instances like this. How does this behave if a map has more than one secret key? I'm worried that both keys are replaced by `************`, which then is a duplicate key, causing map entries to disappear.

GregDThomas: What's being worked on is a list It's a list of map entries, so duplicate key's should not be a problem.

PR #1505, opened 2019-10-25
OF-1897 fix that the maxLifetime settings of a cache take effect on runtime

cpetzka:
It was just a typo. Also introduced constants for the property
parameters.

By testing this change I noticed that it is impossible to clear the cache properties because they are immediately rewritten with the default value. Is this intended?

(see https://discourse.igniterealtime.org/t/bug-in-cache-konfiguration/86452)


GregDThomas:
LGTM. re your comment about clearing the properties, I suspect it's not a deliberate intention, but it doesn't really have any adverse side affects AFAICT.

cpetzka:
> I suspect it's not a deliberate intention, but it doesn't really have any adverse side affects AFAICT.

I agree with you. A ticket should be opened for this.

guusdk:
Thank you for this, and all of the other contributions you've been putting in lately!

PR #1506, opened 2019-10-25
OF-1898 make the max cache size consistent to long

cpetzka:
(see https://discourse.igniterealtime.org/t/bug-in-cache-konfiguration/86452/)

this may be a breaking change!

I'll also make a pull request in the Hazelcast plugin.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/auth/POP3AuthProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/auth/POP3AuthProvider.java
@@ -120,7 +120,7 @@ public class POP3AuthProvider implements AuthProvider {
             Log.debug("\t authRequiresDomain: " + authRequiresDomain);
             Log.debug("\t authCacheEnabled: " + (authCache != null));
             if (authCache != null) {
-                Log.debug("\t authCacheSize: " + authCache.getCacheSize());
+                Log.debug("\t authCacheSize: " + authCache.getLongCacheSize());
                 Log.debug("\t authCacheMaxLifetime: " + authCache.getMaxLifetime());
             }
         }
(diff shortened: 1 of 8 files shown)


GregDThomas:
You know - I'm not entirely sure this is necessary;
Without this change, the maxCacheSize is set as an int, retrieved as a long. Yes, that's inconsistent (and perhaps worthy of a comment in the Javadoc), but doesn't actually break anything. 

This fix for this this inconsistency, however, changes an Interface in a non-compatible way which then becomes a PITA for consumers of that interface.

If you really want to offer a consistent API, can I suggest doing it in a backward compatible way? e.g. simply adding a new method that returns the cache size as an integer

```
default int getIntMaxCacheSize() {
  return Math.Min(Integer.MAX_VALUE, getMaxCacheSize());
}
```

cpetzka:
> If you really want to offer a consistent API, can I suggest doing it in a backward compatible way? e.g. simply adding a new method that returns the cache size as an integer
> 
> ```
> default int getIntMaxCacheSize() {
>   return Math.Min(Integer.MAX_VALUE, getMaxCacheSize());
> }
> ```

I don't quite see how your suggestion is better for consumers. They don't have to implement the standard method getIntMaxCacheSize (i think you mean getIntCacheSize) but still everything has to be changed to "long getCacheSize" or did I forget something?

Another way would be to extend the old API with additional "long" methods. But even then the consumers have to implement them or we implement them as default limited to "int" like:

```
default long getLongCacheSize() {
  return getMaxCacheSize();
}
```

GregDThomas:
a) the code example may have been a bit awry, for which apologies. However, I hope the intent was clear - to avoid breaking compatibility.

b) Yes, both /implementations/ of Cache (DefaultCache and the one in the HZ plugin) will need to override the default methods. However, by using appropriate default methods it then means that other plugins are no longer tied to particular Openfire versions; a plugin compiled against the Cache interface in Openfire 4.0 can work quite happily with 4.4 today, and by not removing/changing signature of existing methods could continue to work with Openfire 4.5. Yes, to take advantage of the new features of 4.5 it will need to compile against 4.5, but that's only normal.

cpetzka:
Hi,

i changed the api so that it is backward compatible, i hope this variant meets your intention.
if you want me to make any changes, let me know.

guusdk:
This looks good to me, but I'd like @GregDThomas to OK this before merging.

Do I understand correctly that with @cpetzka 's latest update, this is no longer a "breaking change" (can we include it in Openfire, without worrying about having an updated Hazelcast plugin)? If so, I'd like to include it in the upcoming patch release.

GregDThomas:
The proof is in the pudding (a UK saying). Run up a single node cluster with the released HZ plugin and see what happens. Something I hope to get to this weekend.

GregDThomas:
Yup, this look goods to me. Thanks for being patient with me!

cpetzka:
@guusdk I tested it with the latest version of the Hazelcast plugin from https://igniterealtime.org/projects/openfire/plugins/2.4.2/hazelcast.jar and there were no problems.

@GregDThomas  @guusdk Thanks for the review!

PR #1509, opened 2019-10-26
OF-1901: Pubsub unsubscribe permissions

guusdk:
This fixes two problems that occur while determining if a particular JID is allowed to unsubscribe a (potentially different) JID from a pubsub node.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
@@ -829,7 +829,7 @@ public class NodeSubscription {
      * @return true if the specified user is allowed to modify or cancel the subscription.
      */
     boolean canModify(JID user) {
-        return user.equals(getJID()) || user.equals(getOwner()) || node.getService().isServiceAdmin(user);
+        return user.equals(getJID()) || user.toBareJID().equals(getOwner().toBareJID()) || node.getService().isServiceAdmin(user);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


guusdk:
@GregDThomas I'd love your review on particularly the first commit. Am I missing something?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
```diff
@@ -829,7 +829,7 @@ void sendLastPublishedItem(PublishedItem publishedItem) {
      * @return true if the specified user is allowed to modify or cancel the subscription.
      */
     boolean canModify(JID user) {
-        return user.equals(getJID()) || user.equals(getOwner()) || node.getService().isServiceAdmin(user);
+        return user.equals(getJID()) || user.toBareJID().equals(getOwner().toBareJID()) || node.getService().isServiceAdmin(user);
```
GregDThomas: Good as far as it goes. Question; should there also be a `user.toBareJID().equals(getJID().toBareJID())` as well? It's quite possible that the modifying JID has a different resource from the subscribing JID, but be the same user.

GregDThomas:
One relatively minor comment, otherwise LGTM.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
guusdk: Agreed.

guusdk: On second thought: no, maybe not. The 'JID' is the target for notifications to be sent to. Although this would nearly always be equal to the owner, it might not be. If it's not, then I'm not sure if giving all resources of the receiving entity should be allowed access.

I'm thinking that this is largely a moot point, as Openfire will mostly decide based on 'owner' anyway. I'll leave the JID-handling as-is.

PR #1510, opened 2019-10-29
OF-1905: security audit page pagination

GregDThomas:
NB. During this I noticed that editing a property did not add an audit event. Fixed in this PR, may wish to merge this in to the 4.4 branch.


akrherz:
Thanks @GregDThomas , but what's up with the `-jsp.jsp` file name change?

GregDThomas:
The address of the page is `security-audit-viewer.jsp` but as this is routed to the Servlet first - `@WebServlet(value = "/security-audit-viewer.jsp")` - to handle the business logic before the requested is forwarding to the the JSP to render the page, the JSP itself needs to have a different name from the one associated with the Servlet. I chose to simply add a `-jsp` suffix to the original name. I could have left the JSP page as is, but changed the URL to the audit viewer, but chose not do do that in case it broke peoples bookmarks (*).

(*) Though currently the URL needs to end .jsp as there are a number of filters etc. that apply to `*.jsp` to handle things like ensuring the user is authenticated, applying the styling etc.

I'm not particularly wedded to any option, so happy to change if people have a strong feeling about it.

GregDThomas:
I do have concern that on older installations the size of the audit log may be massive enough to cause performance problems (the full log is fetched on each page view). Will cross that bridge if someone complains ;)

Of lesser importance,
(a) the from/to dates were not aligned with each other on the UI, and
(b) if you get the dates the wrong way around they should ideally be swapped,
but then I thought better of shaving that particular yak ;)

PR #1511, opened 2019-10-29
This fix is an interim solution to OF-1904

deleolajide:
This merge resets the roster sharing logic back to version 4.0.0 while the root issue is being investigated.
The changes are as follows:

* Remove the Log warns for **UserNotFoundException**  in Roster.java which were filling up the log file. 
* Reverted the optimized code that was using getRoster in RosterManager.java back to accessing rosterCache directly.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/roster/RosterManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/roster/RosterManager.java
@@ -59,7 +59,7 @@ public class RosterManager extends BasicModule implements GroupEventListener, Us
     private static final Logger Log = LoggerFactory.getLogger(RosterManager.class);
 
     private static final String MUTEX_SUFFIX = " ro";
-    
+
     private Cache<String, Roster> rosterCache = null;
     private XMPPServer server;
     private RoutingTable routingTable;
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/roster/RosterManager.java
```diff
@@ -361,16 +361,13 @@ else if ("sharedRoster.displayName".equals(keyChanged)) {
...
                 if (server.isLocal(updatedUser)) {
-                    try {
-                        roster = getRoster(updatedUser.getNode());
-
-                        // Update the roster with the new group display name
-                        roster.shareGroupRenamed(users);
-                    } catch (UserNotFoundException e) {
-                        Log.debug( "Unexpected exception while applying group modification for user '{}' .", updatedUser.getNode(), e );
-                    }
+                    roster = rosterCache.get(updatedUser.getNode());
```
guusdk: This only considers a roster that's already loaded in the cache, but not ones that are only in the database. Is this not a concern because we're processing a 'shared roster' change here (which perhaps does not affect data that's stored on the roster entry of individual users in the database)?

deleolajide: Maybe shared roster entries are constructed on demand until manual roster entries. I would need to dive much deeper into the code to confirm that. As this was the previous code, we have to consider that it has some merit.

PR #1512, opened 2019-10-30
OF-1906 & OF-1907

guusdk:


guusdk:
FWIW, This code was copied from an organization that had this running in production.

GregDThomas:
Well, if it's good code known to work in production, that's better than any review I could give ;)

PR #1516, opened 2019-11-01
denote Openfire 4.4.3 release

akrherz:


guusdk:
Might be good to follow up with another commit that marches on to 4.4.4-SNAPSHOT.

PR #1517, opened 2019-11-01
OF-1912: Stop using proprietary sun.security.x509.GeneralNameInterface

guusdk:
The implementation is used only for one constant definition. We can copy the constant value in our code, instead of depending on proprietary code.


guusdk:
This can go in the 4.4.x branch too.

PR #1519, opened 2019-11-04
OF-1903 OF-1916 Support for @WebServlet annotation

akrherz:
We need this support for the changes in #1510  to work.


GregDThomas:
Do! Of course, thank you for pointing out the obvious!

PR #1520, opened 2019-11-05
OF-1917: Work with cyclic dependencies in shared groups.

guusdk:
In order to avoid a stack overflow, this code first 'flattens' the set of dependent shared groups, before processing them.


deleolajide:
I wonder if there is any correlation to https://github.com/igniterealtime/Openfire/pull/1502 which is probably dealing with a similar issue, but from the perspective of LDAP.

PR #1521, opened 2019-11-07
When clustering is used, solves the problem of not joining the room

abdurrahmanekr:
Hello, when using bosh with Openfire cluster, could not join the MUC rooms.  This PR for controlling **occupantsByNickname** and **occupantsByFullJID** variables.

[openfire-cluster-behaves-unexpectedly-on-the-muc](https://discourse.igniterealtime.org/t/openfire-cluster-behaves-unexpectedly-on-the-muc/86520/6?u=abdurrahmanekr)


speedy01:
@GregDThomas @guusdk   please review! Thanks

GregDThomas:
I'm not sure I can add anything useful to this. 

But that doesn't normally stop me, so:

The Discourse thread seems to indicate that this only happens with BOSH (possibly also WebSocket?) connections, not TCP connections. 

Suggest to me that whilst the code fixes the underlying problem of a NPE, the fix is actually masking some issue with the BOSH connectivity.

akrherz:
@abdurrahmanekr , sorry this PR has languished.  Are you able to rebase it?

abdurrahmanekr:
Hi @akrherz, when I look at the last changes, I see that the same logic is used. That's why I tried to write the same code while solving the conflict. As mentioned in discouse, I had this problem only when using BOSH (http-bind). We are currently using this code in our system and I have to compile openfire's core accordingly. Because people can't join groups while using bosh (with cluster).

Fishbowler:
Given how long this has languished, and that the simple change no longer applies on the tip of main, should we close this?

I think perhaps implied by:
> when I look at the last changes, I see that the same logic is used
... this is already done?

abdurrahmanekr:
> Closing this — the underlying issue was resolved by the MUC refactor in 4.7.0 ([OF-2272](https://igniterealtime.atlassian.net/browse/OF-2272), commit 8483798), which replaced the separate `occupantsByNickname`/`occupantsByFullJID` maps with a single `occupants` collection, so the inconsistency this patch worked around can no longer occur. 

It felt good to get back into open-source projects after 7 years. Last week, there were mass layoffs because AI had created job vacancies. I’m currently unemployed. I’ll probably work on a few ideas I have related to Openfire—though first, let’s try out the latest version :) See you soon <3

[OF-2272]: https://igniterealtime.atlassian.net/browse/OF-2272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1522, opened 2019-11-11
Add coverage reports

Alameyo:
After build (like mvn verify) code coverage report is generated in xmppserver/target/site/jacoco/index.html (there are also xml and csv versions). It should be useful during planing which tests to write first and you will be able to see which lines are not covered yet.

Screenshot from 2019-11-11 15-19-39

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -207,9 +207,6 @@
 
         <profile>
             <id>coverage</id>
-            <properties>
-                <argLine>@{argLine} -Duser.language=us -Duser.country=US</argLine>
-            </properties>
             <build>
                 <plugins>
                     <plugin>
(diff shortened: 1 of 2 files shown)


guusdk:
This is pretty cool. How much time is added to the build process by having this report generated? If it's a lot, we might want to move it to an (optional) profile. That way, we won't add a lot of overhead for developers, while we can still generate the coverage reports automatically in Bamboo and/or Travis.

Alameyo:
On my PC it's jump from 17 to almost 19 seconds. 

Travis: it had run for 5 min 16 sec while other PR I see run for around 4 min 58 sec(+-3 secs).
LGTM 31 min 55 sec while some other 30m 43s.

Alameyo:
Now I have 2 profiles and removed jsp from coverage. We can run `mvn verify` for build without coverage and `mvn verify -P default,coverage` for coverage report.

One problem I am still facing is that I need to add @{argLine} as JVM argument for surefire. Jacoco set's argLine to provide it's runetime agent to surefire. But if I run without coverage profile it will be empty property, thus build will fail.

As workaround in a bit redundant way I am adding surefire plugin in booth profiles with different arguments. Anyway I am guessing I should be able to set up it as some kind of property depending on which profile is run?

akrherz:
@Alameyo you able to push this PR over the finish line so that it can be in the beta release?

Alameyo:
I can bring back simple version or I can try again this week with profiles though there was a detail with that I couldn't work around yet.

Alameyo:
Seems smack integration tests fail on OracleJDK while pass on OpenJDK which seems to me unrelated to changes.

PR #1524, opened 2019-11-13
Remove unused dependencies

cesarsotovalero:
Hello, I noticed that dependencies  org.igniterealtime.openfire`i18n`, `org.bouncycastle:bcpg-jdk15on`, `org.directwebremoting:dwr`, and `org.hsqldb:hsqldb` are declared in the `pom` of module `xmppserver`. However, these dependencies are not used and, therefore, they can be safely removed to make the `pom` more clear and the Maven dependency tree of the project less complex.

### DIFF
--- a/xmppserver/pom.xml
+++ b/xmppserver/pom.xml
@@ -90,11 +90,6 @@
     <dependencies>
         <!-- Ignite Realtime -->
         <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>i18n</artifactId>
-            <version>${project.version}</version>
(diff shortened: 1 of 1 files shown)


GregDThomas:
i18n, dwr and hsqldb are definitely used. May be not referred to directly in Java code, but referenced via the UI, for example. I also suspect bouncycastle is used, as it's a JCE provider.

cesarsotovalero:
I see. So these dependencies are not referenced directly from Java code but from other sources, e.g the webapp.  Thanks for the response :+1:

PR #1525, opened 2019-11-17
OF-1902: Restrict retrieve of favicon to connected servers

GregDThomas:
(also fixes the issue that prevents the load of the default icon where no favicon exists)

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
@@ -17,8 +17,10 @@
 package org.jivesoftware.util;
 
 import java.io.IOException;
-import java.net.MalformedURLException;
-import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Optional;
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
```diff
@@ -88,15 +92,15 @@ public void init(ServletConfig config) throws ServletException {
...
+           defaultBytes = Files.readAllBytes(Paths.get(JiveGlobals.getHomeDirectory(), "plugins/admin/webapp/images/server_16x16.gif"));
         }
-        catch (MalformedURLException e) {
+        catch (final IOException e) {
             LOGGER.warn("Unable to retrieve default favicon", e);
         }
         // Initialize caches.
         missesCache = CacheFactory.createCache("Favicon Misses");
         hitsCache = CacheFactory.createCache("Favicon Hits");
+        sessionManager = SessionManager.getInstance();
```
guusdk: Nit: maybe translate the sessionManager field to a local variable, to avoid keeping a reference? I'm guessing that the servlet is rarely called.

PR #1527, opened 2019-11-20
OF-1923 (and OF-1925, OF-1926, and trivial OF-1924) Fix for Stream Management 

guusdk:
The changes in this PR intend to resolve an issue where a connection that reconnects and binds using the same resource (full JID) gets disconnected.

The original problem is caused by the cleanup task for stream management cleaning up sessions by full JID, even if the original session has been replaced by a new session (that uses the same full JID).

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
@@ -23,6 +23,7 @@ import org.xmpp.packet.*;
 import java.math.BigInteger;
 import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.Date;
 import java.util.Deque;
 import java.util.LinkedList;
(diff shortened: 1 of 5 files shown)


guusdk:
I've slapped on a fix for OF-1925 (which should be benign).

xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
```diff
@@ -206,7 +222,11 @@ private void enable( String namespace, boolean resume )
         Element enabled = new DOMElement(QName.get("enabled", namespace));
         if (this.resume) {
             enabled.addAttribute("resume", "true");
-            enabled.addAttribute( "id", smId);
+            enabled.addAttribute("id", smId);
+            if ( !namespace.equals(NAMESPACE_V2) && LOCATION_ENABLED.getValue() ) {
+                // OF-1925: Hint clients to do resumes at the same cluster node.
+                enabled.addAttribute("location", XMPPServer.getInstance().getServerInfo().getHostname());
```
GregDThomas: (If this is a repeat, apologies, earlier comment along similar may have been lost)
https://xmpp.org/extensions/xep-0198.html#enable suggests that the location may optionally include a port. Might be worth including here too.

GregDThomas: PS. While you have the code open, is it worth adding the max="<max resumption time in seconds" attribute too.

guusdk: That's trickier to do, as you'd have to check if the current session was using Direct TLS or not. I'd rather not make this more complicated than needed.

PR #1531, opened 2019-11-23
OF-1219: Fix class cast exception

GregDThomas:


guusdk:
Makes sense. I'd be interested in finding out how the problem that this prevents could be triggered, but regardless, this change is an improvement.

GregDThomas:
If you change the property but don't restart Openfire, then the user provider currently active isn't what's in the property - so you get the class cast exception.

PR #1532, opened 2019-11-23
OF-1932: Ensure we use the correct password when running tests

GregDThomas:


GregDThomas:
Possibly a candidate for the 4.4 branch as this was introduced by https://issues.igniterealtime.org/browse/OF-1873

guusdk:
`computeIfAbsent` will not only return the 'if absent' value, but will also add that value to the map. Is that what is desired in this case?

GregDThomas:
Not sure at this point. Just copied the line you had used elsewhere. Will ponder the matter further ...

GregDThomas:
> `computeIfAbsent` will not only return the 'if absent' value, but will also add that value to the map. Is that what is desired in this case?

Yes, that is what is intended. The (existing) password needs to be in the settings being passed to the temporary LdapManager used to test the user/group so the test has a chance of passing.

guusdk:
Once again: thank you!

PR #1533, opened 2019-11-24
Add a Maven wrapper (mvnw.cmd / mvnw) to save installing Maven

GregDThomas:
This should make it easier to start developing - the Maven wrapper simply downloads the appropriate version of Maven the first time, and then uses it.


guusdk:
Does this play nice with IDEs?

GregDThomas:
It's fine with IntelliJ (which just ignores it). Can't comment on others from memory I don't think Eclipse used to require an Maven install, so should be OK with that too.

PR #1534, opened 2019-11-24
OF-1868: Share isMessageCarbonsEnabled and hasRequestedBlocklist state with other cluster nodes

mightymop:


xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalClientSession.java
```diff
@@ -945,6 +945,11 @@ public boolean isMessageCarbonsEnabled() {
     @Override
     public void setMessageCarbonsEnabled(boolean enabled) {
         messageCarbonsEnabled = enabled;
+        if (ClusterManager.isClusteringStarted()) {
+            // Track information about the session and share it with other cluster nodes
+            Cache<String,ClientSessionInfo> cache = SessionManager.getInstance().getSessionInfoCache();
+            cache.put(getAddress().toString(), new ClientSessionInfo(this));
```
guusdk: Why is it needed to modify the caches here (and on line 966)? It is strange to me that when these new properties get updated, a cache entry is added/updated, but when the existing properties get updated, this is not done.

mightymop:
maybe the changes in LocalClientSession.java are not needed?

guusdk:
I don't know. You tell me? :)

GregDThomas:
FWIW, I'd really like to avoid any changes to the HZ plugin - if it is possible to simply have a cluster task in core Openfire, that's a plus point for me. I've not looked too closely at this yet, though.

mightymop:
haha :D i thought you were the lead programmer here? ;D

> It is strange to me that when these new properties get updated, a cache entry is added/updated, but when the existing properties get updated, this is not done.

but by the way which existing properties do you mean?

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalClientSession.java
guusdk: My previous remark was incorrect. The exact same thing _does_ happen, for example in `org.jivesoftware.openfire.session.LocalClientSession#setOfflineFloodStopped`

guusdk:
I made (another) mistake: I didn't spot that this PR changed two files. I thought you added code to update caches `ClientSessionInfo`, where this is not done for any of the fields of that class. In reality, you added it to `LocalClientSession`, that indeed does the same for other fields (like in `org.jivesoftware.openfire.session.LocalClientSession#setOfflineFloodStopped`).

mightymop:
> I made (another) mistake: I didn't spot that this PR changed two files. I thought you added code to update caches `ClientSessionInfo`, where this is not done for any of the fields of that class. In reality, you added it to `LocalClientSession`, that indeed does the same for other fields (like in `org.jivesoftware.openfire.session.LocalClientSession#setOfflineFloodStopped`).

So anything left to change in this PR?

guusdk:
Is there reason to believe that https://github.com/igniterealtime/Openfire/pull/1534#pullrequestreview-322080315 is related to the change done in this PR?

GregDThomas:
Is that the right link?

guusdk:
The link looks good to me, but to avoid all confusion, I'm referring to this comment made earlier. Is there any reason to assume that the behavior that's described here is caused by the changes that are done in this PR? If so, we should investigate. If not: I'm confused why it is added as a comment to this PR.

> > FWIW, I'd really like to avoid any changes to the HZ plugin - if it is possible to simply have a cluster task in core Openfire, that's a plus point for me. I've not looked too closely at this yet, though.
> 
> I tested the changes with new code fore ClientSessionInfo on Openfire only and used the original Hazelcast Version 2.4.2 without code changes and the Message Carbons did not arrive on all clients! Tests were made 5min ago.
> example:
> 
> Node A:
> [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource1
> [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource2
> [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource3
> 
> Node B
> [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource4
> [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource5
> [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource6
> 
> Test A:
> [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource3 > MESSAGE > test1
> Result: message arrived on all clients of test1 but not on [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource6
> 
> Test B:
> [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource5 > MESSAGE > test2
> Result: message arrived on all clients of test2 but not on all clients of Node A ([test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource1and [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource2)
> 
> maybe it also has sth to do with OF-1888?
> https://issues.igniterealtime.org/browse/OF-1888

GregDThomas:
(Sorry, my bad was trying to do too many things at once and mis-read)

mightymop:
> The link looks good to me, but to avoid all confusion, I'm referring to this comment made earlier. Is there any reason to assume that the behavior that's described here is caused by the changes that are done in this PR? If so, we should investigate. If not: I'm confused why it is added as a comment to this PR.
> 
> > > FWIW, I'd really like to avoid any changes to the HZ plugin - if it is possible to simply have a cluster task in core Openfire, that's a plus point for me. I've not looked too closely at this yet, though.
> > 
> > 
> > I tested the changes with new code fore ClientSessionInfo on Openfire only and used the original Hazelcast Version 2.4.2 without code changes and the Message Carbons did not arrive on all clients! Tests were made 5min ago.
> > example:
> > Node A:
> > [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource1
> > [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource2
> > [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource3
> > Node B
> > [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource4
> > [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource5
> > [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource6
> > Test A:
> > [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource3 > MESSAGE > test1
> > Result: message arrived on all clients of test1 but not on [test2@xmmp.domain](mailto:test2@xmmp.domain)/ressource6
> > Test B:
> > [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource5 > MESSAGE > test2
> > Result: message arrived on all clients of test2 but not on all clients of Node A ([test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource1and [test1@xmmp.domain](mailto:test1@xmmp.domain)/ressource2)
> > maybe it also has sth to do with OF-1888?
> > https://issues.igniterealtime.org/browse/OF-1888

I've made the comment to OF-1888 to because it also has to do with message carbons not clustered correctly. I've debugged to this point with the localRoutingTable and indeed the remote jids are not in the routing table.

PR #1535, opened 2019-11-24
OF-1779: Fix SQL Statement in Pubsub

mightymop:
Fixes Statement for SQL SERVER purge function

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceManager.java
@@ -58,7 +58,7 @@ public class PubSubPersistenceManager {
 
     private static final Logger log = LoggerFactory.getLogger(PubSubPersistenceManager.class);
 
-    private static final String PERSISTENT_NODES = "SELECT serviceID, nodeID, maxItems " +
+    private static final String PERSISTENT_NODES = "SELECT DISTINCT serviceID, nodeID, maxItems " +
             "FROM ofPubsubNode WHERE leaf=1 AND persistItems=1 AND maxItems > 0";
     
     private static final String PURGE_FOR_SIZE =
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceManager.java
```diff
@@ -58,7 +58,7 @@
 
     private static final Logger log = LoggerFactory.getLogger(PubSubPersistenceManager.class);
 
-    private static final String PERSISTENT_NODES = "SELECT serviceID, nodeID, maxItems " +
+    private static final String PERSISTENT_NODES = "SELECT DISTINCT serviceID, nodeID, maxItems " +
```
guusdk: Although it makes sense that we'd like unique rows to be returned, I'm alarmed by the fact that there's a need to add the DISTINCT keyword. Is there a scenario where the original query returns a non-distinct result set?

guusdk:
Thanks! Exactly what issue does this fix? Can you give a description that we can use to create a new ticket in JIRA please? Afterwards, I'd love for you to re-word the commit message to include the JIRA ticket number.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceManager.java
mightymop: I've found doubled rows in database but dont know where they came from... 
this fix indeed would not fix the cause but i will prevent errors based on multiple results...

mightymop:
I think it was OF-1779 (https://issues.igniterealtime.org/browse/OF-1779?jql=text%20~%20%22sql%22)
i also tested all versions of the old statements (they were several commits to that bug)
With this pull request we had no errors anymore...

guusdk:
Awesome, thank you!

PR #1536, opened 2019-11-24
OF-1935: Feature keep muc from deletion

mightymop:
with option xmpp.muc.keep_from_cleanup you are now able to prevent mucs from deletion
Added options to service preferences tab on admin console

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
@@ -462,7 +462,7 @@ public class MUCPersistenceManager {
      * will be executed only when the service is starting up.
      *
      * @param chatserver the chat server that will hold the loaded rooms.
-     * @param emptyDate rooms that hadn't been used before this date won't be loaded.
+     * @param cleanupDate rooms that hadn't been used before this date won't be loaded.
      * @param packetRouter the PacketRouter that loaded rooms will use to send packets.
      * @return a collection with all the persistent rooms.
      */
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
@@ -70,13 +70,13 @@ public class MUCPersistenceManager {
     private static final String LOAD_HISTORY =
         "SELECT sender, nickname, logTime, subject, body, stanza FROM ofMucConversationLog " +
         "WHERE logTime>? AND roomID=? AND (nickname IS NOT NULL OR subject IS NOT NULL) ORDER BY logTime";
-    private static final String LOAD_ALL_ROOMS =
+    private static final String RELOAD_ALL_ROOMS_WITH_RECENT_ACTIVITY =
         "SELECT roomID, creationDate, modificationDate, name, naturalName, description, " +
         "lockedDate, emptyDate, canChangeSubject, maxUsers, publicRoom, moderated, membersOnly, " +
         "canInvite, roomPassword, canDiscoverJID, logEnabled, subject, rolesToBroadcast, " +
         "useReservedNick, canChangeNick, canRegister, allowpm " +
         "FROM ofMucRoom WHERE serviceID=? AND (emptyDate IS NULL or emptyDate > ?)";
-    private static final String LOAD_REALLY_ALL_ROOMS =
+    private static final String LOAD_ALL_ROOMS =
         "SELECT roomID, creationDate, modificationDate, name, naturalName, description, " +
         "lockedDate, emptyDate, canChangeSubject, maxUsers, publicRoom, moderated, membersOnly, " +
         "canInvite, roomPassword, canDiscoverJID, logEnabled, subject, rolesToBroadcast, " +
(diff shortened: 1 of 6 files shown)


mightymop:
Okay any suggestions? i did not find a flag in the database that would keep muc rooms

guusdk:
From looking at the code, I'd think that `emptyDate` is that flag.

mightymop:
1. the "CleanupDate" should be configurable via admin console... it is hard to search for options in the code. is it "chatServiceName.unload.empty_days" ?
2.  i think the problem sits here: 
`  Date emptyDate = room.getEmptyDate();
   if (emptyDate != null && emptyDate.before(cleanUpDate)) ...`
   i have several chat services with a lot of persistant mucs and they have to be kept in memory so that they are listed in the public dictionary. Most of the rooms are empty and had never seen any occupants for min. 2 month so all mucs have an "emptyDate" in the database! So they were allways deleted after reaching the "CleanupDate".

so it is better to change the code that emptyDate keeps null in all situations if my option is enabled?

guusdk:
I fear I have not provided you with the right flag. `emptyDate` is related, but not the correct field for this. The `emptyDate` field is described as this:

> The date when the last occupant left the room. A null value means that there are occupants in the room at the moment.

We cannot use it for another purpose, as that would break other things.

If we track the usages of the `emptyDate`, then we find something interesting. It is used by `org.jivesoftware.openfire.muc.spi.MultiUserChatServiceImpl.CleanupTask`, which is described as follows:

> Removes from memory rooms that have been without activity for a period of time. A room is considered without activity when no occupants are present in the room for a while.

This seems to be the implementation that is responsible for the behavior that you want to prevent.
 
The implementation is very straight-forward. Somewhat simplified:

    Date cleanUpDate = getCleanupDate();
    for ( LocalMUCRoom room : rooms ) {
        Date emptyDate = room.getEmptyDate();
        if (emptyDate != null && emptyDate.before(cleanUpDate)) {
            removeChatRoom(room.getName());
        }
   }

This code uses `getEmptyDate`, which is implemented as follows:

    // Returns the limit date after which rooms without activity will be removed from memory.
    return new Date(System.currentTimeMillis() - (emptyLimit * 3600000));

This used the `emptyLimit` field, which is described like this:

> Max number of hours that a persistent room may be empty before the service removes the room from memory. Unloaded rooms will exist in the database and may be loaded by a user request. Default time limit is: 30 days.

If you search for references, then you find (like you already did) by the `unload.empty_days` property of the MUC service.

I'd argue that the cleanest way to implement a solution that prevent rooms from being "cleaned up" is to work with this field, and to change its usages in such a way that the cleanup task is skipped when the value is either null, or -1 or something like that. That way, you don't need a second property (`skip`) either.

I like your idea of making this configurable on the admin console! That is a lot better then assuming that people magically know what property to set!

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
```diff
@@ -76,6 +76,12 @@
         "canInvite, roomPassword, canDiscoverJID, logEnabled, subject, rolesToBroadcast, " +
         "useReservedNick, canChangeNick, canRegister, allowpm " +
         "FROM ofMucRoom WHERE serviceID=? AND (emptyDate IS NULL or emptyDate > ?)";
+    private static final String LOAD_REALLY_ALL_ROOMS =
```
guusdk: Can you think of a better name for this constant? Maybe rename to `LOAD_ALL_ROOMS` and rename the original query to `RELOAD_ALL_ROOMS_WITH_RECENT_ACTIVITY` or something like that.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
```diff
@@ -499,9 +505,17 @@ public static void deleteFromDB(MUCRoom room) {
         ResultSet resultSet = null;
         try {
             connection = DbConnectionManager.getConnection();
-            statement = connection.prepareStatement(LOAD_ALL_ROOMS);
-            statement.setLong(1, serviceID);
-            statement.setString(2, StringUtils.dateToMillis(emptyDate));
+            if (emptyDate!=null)
```
guusdk: This was already in the code, but it confused me. Can you improve this? The argument that is named `emptyDate` really is not the "empty date" - it is the cleanup date. Can you rename the argument please (and do the same in the method that calls this?

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -631,13 +631,16 @@ public void run() {
...
+                if (cleanUpDate!=null)
+                {
+	                Iterator<LocalMUCRoom> it = localMUCRoomManager.getRooms().iterator();
+	                while (it.hasNext()) {
+	                    LocalMUCRoom room = it.next();
+	                    Date emptyDate = room.getEmptyDate();
+	                    if (emptyDate != null && emptyDate.before(cleanUpDate)) {
+	                        removeChatRoom(room.getName());
+	                    }
+	                }
```
guusdk: Please make sure that you use the same level of indentation. You're switching from 4 space characters to 2, for some reason.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
mightymop: i think its called emptyDate here because the flag in the DB is also called emptyDate which is asked in the SQL Query but i've changed it

guusdk:
Nice! Is there a JIRA issue for this that you know of? If not, I'll create one. Also, could you see why LGTM is unhappy?

mightymop:
I forgott to ghanfe javadoc 🙄

guusdk:
Thanks! I've created https://issues.igniterealtime.org/browse/OF-1935 to track this issue in our bugtracker.

guusdk:
To give you, as a relatively new contributor, a bit of feedback on this change. After releasing this as part of Openfire 4.5.0 beta, we've hit a snag: this PR causes problems when creating _new_ services. I guess, like me, you only tested this with existing services! See https://issues.igniterealtime.org/browse/OF-1954

Not a big deal, and I'm already fixing the code. Just wanted to ping back to you, to give you an opportunity to learn from the experience. :)

PR #1537, opened 2019-11-24
 Fixed a bug, because with containsAll you had to use all filters

mightymop:


guusdk:
If I understand this change correctly, it moves from "all fields that are requested must be a search field" to "we'll only return fields if they are requested _and_ are a search field"

To me, that would be confusing behavior, as it silently ignores fields that do not match.

mightymop:
I tested with gajim:
If i use ldap.searchFields, i got presented all the fields i set in this variable. But i allways got an exception on the server if i did not have checked all fields. But what if i only wanted to search a user with only a part of them? For example i only have a surname or only know the persons samaccountname... etc. furthermore searching over all fields was extremly slow. so i would prefer this solution...

guusdk:
Doesn't `a.containsAll(b)` return `true` for any b that is a subset of a?

mightymop:
No java docs say: (https://docs.oracle.com/javase/8/docs/api/java/util/List.html)

Returns:
    true if this list contains all of the elements of the specified collection

guusdk:
I think that you're misreading something. I have just ran this code in JShell:

```
import java.util.Arrays;
import java.util.Collection;
final Collection<String> a = Arrays.asList("blue", "red", "yellow");
final Collection<String> b = Arrays.asList("blue", "red");
final boolean result = a.containsAll(b);
```
Output:

```
Defined import java.util.Arrays
Defined import java.util.Collection

Defined field Collection<String> a = [blue, red, yellow]

Defined field Collection<String> b = [blue, red]

Defined field boolean result = true
```

mightymop:
Strange, ... i allways got an exception with containsAll unless i had a similiar list. So i changed the code to solve it

mightymop:
Will test it again

PR #1538, opened 2019-11-29
Fix Issue OF-1933 Step 1

deleolajide:
First step is to notify members only room that uses a shared group when group members join or leave.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -2955,7 +2955,7 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
         List<Presence> presences = applyAffiliationChange(getRole(), groupMember, null);
 
         if (presences.size() == 0 && isMembersOnly()) {
-            sendAffiliationChangeNotification(groupMember, affiliation);
+            sendOutOfRoomAffiliationChangeNotification(groupMember, affiliation);
         }
         else {      // member is in MUC, send presence stanzas
             for (Presence presence : presences) {
@@ -2964,7 +2964,7 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
         }
     }
 
-    private void sendAffiliationChangeNotification(JID jid, String affiliation) {
+    private void sendOutOfRoomAffiliationChangeNotification(JID jid, String affiliation) {
         // Announce affiliation change for a user that is NOT currently in the room,
         // XEP-0045 (v1.31.2) example 195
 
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -2933,35 +2933,56 @@ public void groupModified(Group group, Map params) {
...
 
+    private void sendAffiliationChangeNotification(JID jid, String affiliation) {
+        // Announce affiliation change for a user that is NOT currently in the room,
+        // XEP-0045 (v1.31.2) example 195
+
+        Message message = new Message();
+        Element fragment = message.addChildElement("x", "http://jabber.org/protocol/muc#user");
+        Element item = fragment.addElement("item");
+        item.addAttribute("affiliation", affiliation);
+        item.addAttribute("role", "none");
```
guusdk: I don't think it's correct to add the `role` attribute like this. This does not take into account whatever role the user had - it will always inform others that this user now has no role.

guusdk (commented):
This change will get complex fast, I think. It might be good to strictly limit the commit to non-white-line changes (maybe add the white-line changes to a different commit?). It'd also be good to add javadoc to the methods that you introduced.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
deleolajide: I agree, but the user is not in the room. His role will be updated when the join room event happens.

deleolajide: >It'd also be good to add javadoc to the methods that you introduced.

I agree, but there is no javadoc on any of the existing methods in that class. I will tackle that later for the entire class.

deleolajide: >I think. It might be good to strictly limit the commit to non-white-line changes (maybe add the white-line changes to a different commit?)

This is my third attempt at this commit. My editor was removing redundant white spaces introduced by a previous commit. I am not sure I have time to start a battle with my text editor.

guusdk: Tip: move all the white-space changes to the first commit, so that everything you touch afterwards does not need to change whitespace!

guusdk: > I agree, but the user is not in the room. His role will be updated when the join room event happens.

Even though that's not how you call the method now, there's nothing that prevents future developers from calling it when the person _is_ in the room.

deleolajide: Got it. I will rename the method accordingly

akrherz:
@deleolajide What's the status of this PR?

deleolajide:
> @deleolajide What's the status of this PR?

It is on hold for now. It has a much lower priority as the project that requested it no longer needs it. I hope to work on it sometime in the future and do a proper job of it then.

PR #1539, opened 2019-11-29
Better cross-platform integration tests

Fishbowler:
This started as "I'll fix that sed statement, then these will run on my Mac", but snowballed.

Improved the running script for readability
Added options for running tests against an already running host, even if it isn't local
Added ability to run tests in a local Docker container (because the Smack tests won't run on case-insensitive file systems)
Added option to set hosts file whilst the script is running

I believe this to be complete, but concede that in coping with lots more scenarios there are some combinations I haven't spotted or tested.

Feedback so very welcome!

### DIFF
--- a/runIntegrationTests
+++ b/runIntegrationTests
@@ -52,6 +52,8 @@ function cleanup {
 		echo "Resetting hosts file after local running. This may prompt for sudo."
 		sudo sed -i'.original' "/example.org/d" /etc/hosts
 	fi
+	echo "Stopping Openfire"
+	pkill -f openfire.lib #TODO: Can this be made more future-proof against changes in the start script?
 }
 trap cleanup EXIT
 
--- a/runIntegrationTests
+++ b/runIntegrationTests
@@ -5,11 +5,17 @@ SMACK_COMMIT_ID="9d626bf" # master 22 Nov 2019
 GRADLE_VERSION="6.0.1"
 FORCE_CUSTOM_GRADLE=false
 CURL_ARGS="--location --silent"
+DEBUG=false
+LOCAL_RUN=false
 
-while getopts dgs: OPTION "$@"; do
+while getopts dlgs: OPTION "$@"; do
 	case $OPTION in
 		d)
+			DEBUG=true
 			set -x
+			;;
+		l)
+			LOCAL_RUN=true
 			;;
 		g)
 			FORCE_CUSTOM_GRADLE=true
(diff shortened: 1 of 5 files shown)


Fishbowler:
Things I know that still need to be done:
* Readme, or similar
* Actually good networking on macOS - need to switch from docker host networking to "something else" (probably on $OSTYPE)
* Re-disable some tests that I optimistically re-enabled 😬

guusdk:
Thanks for this! It's nice how Travis immediately gave you some feedback (there's now a test failing - probably one of those that you re-enabled?)

The current use of the script is limited to being used by Travis, as far as I can tell. As long as it's still usable for that, I'm happy to accept any improvement. Yours seem like a very desirable set.

One thing that keeps nagging me: why are we recompiling Smack at all? It's a process that's complex, error prone, takes time, but yields a very predictable result.

Fishbowler:
Have ditched  Docker in favour of using built maven dependencies (since the whole reason for doing so was because of case insensitive file systems being unable to _build_, not unable to run - hat tip @guusdk)

This runs on macOS and on Ubuntu 18.04 LTS. Windows untested.

Some notable changes for anyone who used the old one:
* Now requires -l flag to run locally (added to Travis build config)
* Can temporarily rewrite your hosts file, alas with sudo (because macOS doesn't support HOSTALIASES)
* Will take a hostname if you want to test an existing installation (along with credentials)
* Uses a riff of the smack-integration-test's gradle file to run the tests, with no additional runner needed, since we already (optionally) grabbed Gradle anyway.

Fishbowler:
Hmmmm. Need to look into why this isn't happy on java 11...

akrherz:
> Hmmmm. Need to look into why this isn't happy on java 11...

bda2fc1c473e change needed here?

akrherz:
> Hmmmm. Need to look into why this isn't happy on java 11...

Sorry for my cryptic message yesterday, I meant was that I found it necessary to use `4.4.0-SNAPSHOT` of smack to work with Java 11.

Fishbowler:
Thanks for the help @akrherz & @guusdk.
Got it working, but there's some broken deps in Smack that mean a few more disabled integration tests :(
I've signed up to be notified via https://www.artifact-listener.org for when a new version of smack-integration-test gets published to see if we can re-enable some then.

runIntegrationTests
```diff
@@ -155,14 +155,15 @@ function runTestsInGradle {
 	echo "Starting Integration Tests (using Smack ${SMACK_VERSION})…"
 
 	DISABLED_INTEGRATION_TESTS=()
-	#DISABLED_INTEGRATION_TESTS+=(MoodIntegrationTest)
+	#Fails only in Java8, only on Travis
+	DISABLED_INTEGRATION_TESTS+=(MoodIntegrationTest)
 	#DISABLED_INTEGRATION_TESTS+=(MultiUserChatIntegrationTest)
```
Fishbowler: Do the tests that aren't disabled want deleting, or kept around as clues for future contributors for the time being?

akrherz: I really like what you currently have.

PR #1542, opened 2019-12-10
OF-1106: Allow for multiple namefields.

guusdk:
This commit allows one to configure more than one LDAP attribute to be used for the Name field, where the first non-empty attribute is used (note that this is different from combining multiple attribute values, which was already supported).

The new functionality is used by defining the mapping value as follows:

    (|({givenName})({displayName})({cn}))

In the example above, the attribute used will be the first non-empty attribute value from the list:

- givenName
- displayName
- cn

Obviously, other attribute names can be used. It's also possible to use fewer or more fields.

This particular format (which mimics LDAP filter syntax) was used to avoid clashes with mappings where it is desirable to join multiple fields (which is also supported in Openfire), for example:

    {familyName}, {givenName}


xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -230,19 +230,34 @@ public IQ handleIQ(IQ packet) {
      * @param node the node that the provider will handle.
      * @param provider the DiscoInfoProvider that will handle disco#info packets sent with the
      *        specified node.
+     * @return the provider that was previously associated to the node, if any.
```
GregDThomas: Changes to this file seem unrelated to the PR?

guusdk: Eek!

guusdk: should now be gone.

guusdk:
I have now added a fix for OF-1944 to this PR (as I suspect that this is what's causing @Fishbowler to have trouble testing this change). OF-1944 ensures that relevant caches are changed, after LDAP configruation is changed. This will make any changes immediately visible (instead of being hidden by stale cache entries).

guusdk:
Like so?

Fishbowler:
Setting ldap.nameField to `(|(displayName)(cn))` (no curly braces) gets the desired effect within Users/Groups.
Setting the ldap.vcard-mapping for GIVEN to `(|({displayName})({cn}))` (with curly braces - a simpler version of the example Guus gave) causes an error on clients requesting the vcard-temp

```xml
<iq type="error" id="o33vn-23" to="userone@example.org/Spark">
    <vCard xmlns="vcard-temp"></vCard>
    <error code="500" type="wait">
        <internal-server-error xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/>
    </error>
</iq>
```

Stack trace:

> 2019.12.11 12:13:22 DEBUG [socket_c2s-thread-2]: org.jivesoftware.openfire.ldap.LdapVCardProvider - LdapVCardProvider: Getting mapped vcard for userone
> 2019.12.11 12:13:22 ERROR [socket_c2s-thread-2]: org.jivesoftware.openfire.handler.IQHandler - Internal server error
> java.util.regex.PatternSyntaxException: Unclosed group near index 12
> (\{)((|()(})
> 	at java.util.regex.Pattern.error(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.accept(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.group0(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.sequence(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.expr(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.group0(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.sequence(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.expr(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.compile(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.<init>(Unknown Source) ~[?:1.8.0_231]
> 	at java.util.regex.Pattern.compile(Unknown Source) ~[?:1.8.0_231]
> 	at java.lang.String.replaceAll(Unknown Source) ~[?:1.8.0_231]
> 	at org.jivesoftware.openfire.ldap.LdapVCardProvider$VCard.treeWalk(LdapVCardProvider.java:551) ~[xmppserver-4.5.0-SNAPSHOT.jar:4.5.0-SNAPSHOT]
> 	at org.jivesoftware.openfire.ldap.LdapVCardProvider$VCard.treeWalk(LdapVCardProvider.java:557) ~[xmppserver-4.5.0-SNAPSHOT.jar:4.5.0-SNAPSHOT]
> 	at org.jivesoftware.openfire.ldap.LdapVCardProvider$VCard.getVCard(LdapVCardProvider.java:534) ~[xmppserver-4.5.0-SNAPSHOT.jar:4.5.0-SNAPSHOT]
> 	at org.jivesoftware.openfire.ldap.LdapVCardProvider.loadVCard(LdapVCardProvider.java:235) ~[xmppserver-4.5.0-SNAPSHOT.jar:4.5.0-SNAPSHOT]
> 	at org.jivesoftware.openfire.vcard.VCardManager.getOrLoadVCard(VCardManager.java:247) ~[xmppserver-4.5.0-SNAPSHOT.jar:4.5.0-SNAPSHOT]
> 	at org.jivesoftware.openfire.vcard.VCardManager.getVCard(VCardManager.java:240) ~[xmppserver-4.5.0-SNAPSHOT.jar:4.5.0-SNAPSHOT]
> 	at org.jivesoftware.openfire.handler.IQvCardHandler.handleIQ(IQvCardHandler.java:136) ~[xmppserver-4.5.0-SNAPSHOT.jar:4.5.0-SNAPSHOT]

Tracing code shows it's on the second pass through treeWalk, which puts it 3 elements deep (so on the GIVEN element) and failing with the regular expressions.

The expression on the Tokenizer ([here](https://github.com/guusdk/Openfire/blob/OF-1106_LDAP-multiple-namefields/xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapVCardProvider.java#L547)) shows that this is only expecting joined sets of LDAP properties (e.g. `{sn}, {givenName}`) rather than queries.

guusdk:
I believe @Fishbowler's concerns have now been addressed.

Fishbowler:
When an admin sets the LDAP VCard mapping for the Name field, either through setup or through Profile Settings (/profile-settings.jsp), this sets the appropriate field in ldap.vcard-mapping, but also affects the ldap.nameField property also.

The example given in [OF-1106](https://issues.igniterealtime.org/projects/OF/issues/OF-1106) was using "givenName sn", so using that...

Set the Name in Profile Settings to "{givenName} {sn}", and have LDAP users with these fields set.

image


Looking in Server Properties, this sees the ldap.vcard-mapping set as intended

image


..and also sees the ldap.nameField set to "givenName sn"

image


Logging in as a user from Spark sees the user's own name picked up when the client requests the vcard-temp (using the ldap.vcard-mapping to build the user's own name from LDAP)

image


But as an Admin looking at Users/Groups in the admin console, the Name field (that comes from ldap.nameField) isn't populated

image


Setting the Name mapping to a single token, like {cn}, {displayName} or even just {givenName} works fine.

guusdk:
That's a good find. I investigated, and found that this is a pre-existing issue (it was not introduced by the fix for OF-1106). The "combined" values that are expected do show in the 'test' popup box that's part of the profile setting / user mapping admin console page, but do not show up as the value for the 'name' property of users (although they do show in the VCard, as you've already found).

I've written this up as OF-1949.

Fishbowler:
👍

Works for joined values like “{givenName} {sn}” or “{givenName}_{sn}” (using space and underscore respectively)
This looks good in Admin, in the vcard-temp retrieved when you log into Spark, and in the roster / user search within Spark.

It’s also working for query values, like “(|({displayName})({cn}))”

PR #1543, opened 2019-12-10
OF-1888: Fix NPE

GregDThomas:
Also;
a) general code tidy up, based on IntellilJ inspections
b) Replace use of a deprecated method with new method
c) Remove unnecessary lock when a node leaves a cluster.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/carbons/Received.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/carbons/Received.java
@@ -10,8 +10,12 @@ import org.xmpp.packet.PacketExtension;
  * @author Christian Schudt
  */
 public final class Received extends PacketExtension {
+
+    public static final String NAME = "received";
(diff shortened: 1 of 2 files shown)


GregDThomas:
(possible candidate for 4.4?)

GregDThomas:
OK, I've had another go at this (with some more code tidy ups). 

Two words of caution;
a) This is entirely untested. The NPE seemed trivial - but then it grew. I need to work out how I can set up a cluster in my current environment.
b) I worry that a remote user session may receive the CC, and then forward that as a CC to its remote user sessions, which then forwards it to its remote user session, and so on until the whole cluster dies in a burning heap.

GregDThomas:
Latest commit further tidies up code, also ensures we don't CC a message that was already a CC. 

I don't think this was an issue before as the cc'd message was only ever routed directly to the (local) session. Now the CC could be routed to a remote session, the same code would be invoked on the remote session causing a CC of the CC to be sent back to the original session, and so on.

PR #1546, opened 2019-12-11
OF-1937: Reduce connection timeout for LDAP.

guusdk:
We've found that with a connection timeout of larger than 4 seconds, Spark fails to authenticate against LDAP, when LDAP integration is making use of more than one LDAP server with the first of the two unavailable.

This commit reduces the connection timeout to a value that allows Spark to authenticate. It does not, however, prevent that Openfire operates with degraded performance.


Fishbowler:
👍 Tested

PR #1551, opened 2019-12-17
Make better use of generics with Caches

GregDThomas:


GregDThomas:
This has been bugging me for ages, a hang-over from early days of Java. It does require a similar update to the Hazelcast plugin, but figured now is the time to do it as the HZ plugin now has a min server version of 4.5.

GregDThomas:
Note; I'm about to go offline for a week or so. Depending on when you plan  on starting 4.5 RCs, you may want to hold off on merging this due to the close tie in with the Hazelcast plugin.

akrherz:
@GregDThomas Thanks for your efforts here.  You want this in for the Openfire 4.5.0 release?

GregDThomas:
No; I think it’s too late for such a change with the beta already out. I’ll close this for now, perhaps look again next time around ...

PR #1555, opened 2019-12-18
OF-1951: Allow plugins to use i18n with SystemProperty descriptions

GregDThomas:


GregDThomas:
NB. Usage of SystemProperty in unchanged, simply requires a plugin to define an appropriate i18n file with definitions in. See https://github.com/igniterealtime/openfire-hazelcast-plugin/pull/38 for such an example.

guusdk:
Thanks once again, Greg!

GregDThomas:
Bah. There's a slight problem with this.
a) Naturally, a plugin is only (and can only be) added to the PluginManager /after/ it's class is instantiated
b) Any SystemProperty objects created by a plugin /during/ instantiation will attempt to get the plugin class loader from the PluginManager to retrieve the i18n definition of a property. Which fails, because at that point the plugin doesn't exist in the PluginManager. Will have to ponder that one some more ...

PR #1557, opened 2019-12-20
OF-1954: Fixes NPE when creating new MUC service

guusdk:
This fixes a problem introduced by OF-1935, which causes NullPointerExceptions to show in the admin console when creating a new MUC service.

@mightymop - as the original author, would you care to review this?


mightymop:
@guusdk: I have nothing to complain about. *thumbs up*

PR #1558, opened 2019-12-20
OF-2038: Fix Exception when search for shared Groups

mightymop:
It is not necessary to search for shared Groups for sharedRoster in LDAP.
Instead search the DB for them.
This fixes an exception when the LDAPGroupProvider would search for sharedRosterGroups in LDAP and can not find them, 'cause they are saved in DB (ofGroupProp).


guusdk:
Thank you!

PR #1562, opened 2019-12-31
OF-1956: Introduce a configurable pause when restarting connection listeners

guusdk:
To combat race conditions when restarting a (socket) listener, this commit introduces a configurable pause that occurs when restarting a listener (after 'stop', before 'start').

As the race condition is more likely to affect the older s2s code, a non-zero default has been configured for that listener. Client listeners have a default of zero.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java
@@ -9,10 +9,12 @@ import org.jivesoftware.openfire.keystore.IdentityStore;
 import org.jivesoftware.openfire.net.SocketConnection;
 import org.jivesoftware.util.CertificateManager;
 import org.jivesoftware.util.JiveGlobals;
+import org.jivesoftware.util.SystemProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java
```diff
@@ -322,7 +332,19 @@ public synchronized void restart()
         }
         finally
         {
-            start(); // won't actually start anything if not enabled.
+            // Protect against race conditions when restarting (OF-1956).
+            if ( pauseDurationDuringRestart != null && !pauseDurationDuringRestart.getValue().isNegative() )
+            {
+                try {
+                    Log.debug( "Sleeping {} before restarting...", pauseDurationDuringRestart.getValue() );
+                    Thread.sleep( pauseDurationDuringRestart.getValue().toMillis() );
```
akrherz: I am lukewarm on this approach.  I'd like to see `start()` catch the `BindException` and then only apply the sleep, then try once  more.  Otherwise, one is left wondering how long a duration one needs to tune for this bind exception to go away.  I'd then think the default should be longer than 500 millis.

guusdk:
Closed in favor of xmpp.server.startup.retry.delay

PR #1565, opened 2020-01-02
OF-1978 (feature) Added xmpp.muc.join.presence

abdurrahmanekr:
With this feature, "join presence" of participating users cannot be sent. It is very useful if you don't want to send join presence in all groups. This will be preparation for [MUC lite](https://xmpp.org/extensions/inbox/muc-light.html)


GregDThomas:
Thanks!

guusdk:
Thank you!

PR #1567, opened 2020-01-03
OF-1951 (plugin i18n for a SystemProperty) revisited

GregDThomas:
So, the original change for OF-1951 didn't work if a plugin attempted to create a SystemProperty during instantiation. This is because the SystemProperty attempts to retrieve the plugin class loader from the PluginManager to find the i18n file - but as the PluginManager has yet to fully load the plugin it can't find a class loader.

To fix this;
a) Default to the "current" class loader if a plugin-specific one cannot be found (instead of throwing an NPE), and
b) Instantiate the plugin with the plugin class loader as the current class loader.

The first of these I'm 100% comfortable with.
The second is more interesting. I don't think it would have any side effects as the same class loader is used to create the plugin either way - but other views welcome. 

I've tested this with a couple of plugins, including the Hazelcast plugin in a cluster, and it all seems to be OK.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginManager.java
@@ -16,6 +16,33 @@
 
 package org.jivesoftware.openfire.container;
 
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.DirectoryStream;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.attribute.FileTime;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.jar.JarFile;
+import java.util.zip.ZipException;
+
 import org.apache.commons.io.IOUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.spi.LoggerContext;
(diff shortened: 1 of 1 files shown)


guusdk:
Hmm, changes to the classloading mechanism almost always open a world of pain - even though I agree that what you did seems benign. A litmus test would be the OFMeet plugin, which makes extensive use of classloading shenanigans. If that plugin continues to load with this, I'm fairly confident that we'd not introduce to much issues by merging this PR. If it does not, I'd like to investigate further.

I've got a couple of things on my plate today - I'll try to look at it in the next few days (but feel free to beat me to the punch).

akrherz:
Per discussion, will await resolution of this issue prior to a GA release of Openfire 4.5.0

GregDThomas:
OK, I tried the ofmeet plugin 0.9.5. It seems to work as far as I went; it installs OK, I can create and join a (single person) meeting. Checking the logs and all seems OK, apart from a failure to parse an IQ message - which I think is entirely unrelated.

```
2020.01.06 20:10:41 WARN  [pool-48-thread-2]: org.jitsi.jicofo.xmpp.FocusComponent - (serving component 'Jitsi Meet Focus') Unexpected exception while processing IQ stanza: <iq type="get" id="107-13" from="component.btp191164.iuser.iroot.adidom.com" to="focus.btp191164.iuser.iroot.adidom.com"><query xmlns="jabber:iq:version"/></iq>
java.lang.IllegalArgumentException: name must not be null
	at org.jivesoftware.smack.util.StringUtils.requireNotNullOrEmpty(StringUtils.java:462) ~[smack-core-4.2.2-b1c107f.jar!/:?]
	at org.jivesoftware.smackx.iqversion.packet.Version.<init>(Version.java:71) ~[smack-extensions-4.2.2-b1c107f.jar!/:?]
	at org.jivesoftware.smackx.iqversion.packet.Version.<init>(Version.java:77) ~[smack-extensions-4.2.2-b1c107f.jar!/:?]
	at org.jitsi.jicofo.discovery.Version.<init>(Version.java:40) ~[jicofo-1.1-20190118.220443-57.jar!/:?]
	at org.jitsi.jicofo.discovery.VersionIqProvider.parse(VersionIqProvider.java:38) ~[jicofo-1.1-20190118.220443-57.jar!/:?]
	at org.jitsi.jicofo.discovery.VersionIqProvider.parse(VersionIqProvider.java:29) ~[jicofo-1.1-20190118.220443-57.jar!/:?]
	at org.jivesoftware.smack.provider.Provider.parse(Provider.java:32) ~[smack-core-4.2.2-b1c107f.jar!/:?]
	at org.jitsi.xmpp.util.IQUtils.convert(IQUtils.java:162) ~[jicoco-1.1-20180710.162527-5.jar!/:?]
	at org.jitsi.jicofo.xmpp.FocusComponent.handleIQGetImpl(FocusComponent.java:211) ~[jicofo-1.1-20190118.220443-57.jar!/:?]
	at org.jitsi.xmpp.component.ComponentBase.handleIQGet(ComponentBase.java:317) ~[jicoco-1.1-20180710.162527-5.jar!/:?]
```

guusdk:
Thanks!

PR #1570, opened 2020-01-11
OF-1946: Correct cache size causing incorrect values display on admin UI

GregDThomas:
Replace a Math.max with a Math.min


GregDThomas:
NB. Definitely a strong candidate for the 4.5.x branch

PR #1571, opened 2020-01-17
Fix Issue: Support for more than 500 users per LDAP-Group (Active Directory)

mightymop:
set ldap.groupMemberField to "member;range=0-*" and it will load all users of a group with more then 500 users with the underlying standard java ldap library we use in OF

testet it with a group of 2000 users...

without this patch OF will not load the users of that group!

MAYBE this depends also on ldap.pagedResultsSize > 1000 and the settings of your LDAP SERVER!!!


akrherz:
@mightymop I am unsure what to make of your requested change here.  Are you saying that the default paged request of users from the LDAP server is not working?

mightymop:
We have Active Directory with some Groups with more than 1500 Users and wanted to use them as shared rostet, but OF did not load them... although the ldap.pagedResultsSize set to -1 or 2000 die not help.... i debugged and found out that if you load such a big group the memberfield to load will not be only member but member;range=0-1499 or member;range=0-*

GregDThomas:
Have you tried simple setting ldap.pagedResultsSize to 1000? That matches the default AD value (and in new installs matches the default Openfire value when installed against AD).

mightymop:
Yes i tried of course... furthermore our ADs value for paging is set to 100000

mightymop:
ctx.getAttributes(relativeDN, standardAttributes);

... did not return member Attribute but member,range=... so the following code could not read the members and therefore groups were not read

mightymop:
could someone review this PR ?

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
```diff
@@ -124,7 +133,67 @@ private Group getGroupByDN(LdapName groupDN, Set<String> membersToIgnore) throws
...
+            NamingEnumeration<? extends Attribute> i = attrs.getAll();
+            String groupMemberField = manager.getGroupMemberField();
+            String searchString = groupMemberField.contains(";range=")?groupMemberField.substring(0,groupMemberField.indexOf(";range=")):groupMemberField;
+            while (i.hasMore())
+            {
+                Attribute attribute = i.next();
+                String key = attribute.getID().toLowerCase();
+                if (key.startsWith(searchString))
+                {
+                    paging=true;
```
guusdk, line 148: If I understand this correctly, then paging will only occur when an administrator explicitly sets the group member field to include `;range=`  I don't like that. Can't we make sure that paging happens, when needed, instead? That way, administrators need not be aware of a problem to be able to fix it.

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
```diff
@@ -124,7 +133,67 @@ private Group getGroupByDN(LdapName groupDN, Set<String> membersToIgnore) throws
...
+                }
+            }
+
+            if (!paging)
+            {
+                return processGroup(ctx, attrs, membersToIgnore);
+            }
+            else
+            {
+                //retrieve first 0-1499
```
guusdk, line 159: The size of the page should probably be configurable.

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
```diff
@@ -124,7 +133,67 @@ private Group getGroupByDN(LdapName groupDN, Set<String> membersToIgnore) throws
...
+                ArrayList<JID> members = new ArrayList<JID>();
+                members.addAll(tmpGroup.getMembers());
+                do
+                {
+                    try
+                    {
+                        ArrayList<String> stdAttr = new ArrayList<String>();
+                        for (int n=0;n<standardAttributes.length;n++)
+                        {
+                            if (!standardAttributes[n].contains(manager.getGroupMemberField()))
```
guusdk, line 173: Style: please always wrap if-blocks in `{` and `}` even if they're just one line. Mistakes can be made easily without them.

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
```diff
@@ -124,7 +133,67 @@ private Group getGroupByDN(LdapName groupDN, Set<String> membersToIgnore) throws
...
+                        stdAttr.add(genRangePart(countPage));
+                        attrs = ctx.getAttributes(relativeDN, stdAttr.toArray(new String[stdAttr.size()]));
+                        //retrieve next 1500
+                        tmpGroup=processGroup(ctx, attrs, membersToIgnore);
+                        if (tmpGroup!=null&&tmpGroup.getMembers().size()>0)
+                        {
+                            members.addAll(tmpGroup.getMembers());
+                        }
+                        else // no next found
+                            break;
```
guusdk, line 186: Style: please always wrap if-blocks in `{` and `}` even if they're just one line. Mistakes can be made easily without them.

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
```diff
@@ -124,7 +133,67 @@ private Group getGroupByDN(LdapName groupDN, Set<String> membersToIgnore) throws
...
+                    }
+                    catch (Exception e)
```
guusdk, line 188: Do not hide exceptions, by assuming they're caused by an expected condition. Always log them, even if it's just to debug level. I've spent to many weeks chasing a bug that in the end turned out to be hidden by things like `catch (Exception e} { /* can't happen. Ignore. */ }`

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
```diff
@@ -252,13 +321,24 @@ public int getGroupCount() {
...
+        if (ldapfilter.contains(searchRangeStr))
+        {
+            ldapfilter=ldapfilter.substring(0,ldapfilter.indexOf(searchRangeStr))+
+                    ldapfilter.substring(ldapfilter.indexOf("=", 
+                            ldapfilter.indexOf(searchRangeStr)+searchRangeStr.length()));
+        }
+        
         if (Log.isDebugEnabled()) {
-            Log.debug("Trying to find group names using query: " + filter.toString());
+            Log.debug("Trying to find group names using query: " + ldapfilter);
```
guusdk, line 335: While you're at it, you might as well use parameters:

    Log.debug("Trying to find group names using query: {}", ldapfilter);

With that, I don't think that there's a point in keeping the `if (Log.isDebugEnabled())` check.

mightymop: done...

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
mightymop: I'll write to debug log now...

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
mightymop: ok..

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
mightymop: it depends on the attribute that came back from ldap... AD automatically added the range 0-1499 ... and so on... so i think it was not configureable...

that is also the fundamental problem with such big groups... the range was added to the attribute by AD (ldap) so the OF paging fails.. because OF could not find the member attribute...

mightymop: ok i will take the value from ldap.pagedResultsSize

xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapGroupProvider.java
mightymop: i will look into it tomorrow... but i think there was a problem that we have to use "member;range=0-*" as value for ldap.groupMemberFeld otherwise AD wont five any results for that big group or max 500...

mightymop: there is a 1500 limit in AD so read this:
https://docs.microsoft.com/de-de/previous-versions/windows/desktop/ldap/searching-using-range-retrieval?redirectedfrom=MSDN

because of this i created this PR...

guusdk: I'm not saying that we should not add the range attribute to the request from Openfire to AD. I do wonder if we can hide this complexity from the administrator. Instead of requiring the administrator to add this parameter manually, it would be nice if Openfire could do it automatically, if needed.

mightymop: then we should always use the member;range=0-* as default attribute for AD

mightymop: but i dont know if this limit is only a problem with AD or a general LDAP issue

mightymop:
closing this because of a new rebased PR:
https://github.com/igniterealtime/Openfire/pull/1655

PR #1575, opened 2020-01-28
Exclude files not relevant to the docker image

Fishbowler:
Improves caching of build stages

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
 .git
+.gitignore
+Dockerfile


guusdk:
Did you mean to also exclude the Dockerfile itself?

Fishbowler:
I think so. The build process inside the docker container doesn't need to know about how to make another one.

Fishbowler:
Realistically the dockerignore file itself should be in the dockerignore!

PR #1577, opened 2020-02-05
OF-1963: Fix XSS issues identified by CSW Document No: C1055

GregDThomas:
Note that `StringEscapeUtils.escapeHtml4()` will not escape single/double quotes. However `StringEscapeUtils.escapeXml11()` will. Every day is a school day!

Plus a couple of other outputs that were simply un-escaped.


guusdk:
Thanks!

PR #1578, opened 2020-02-11
OF-1964 SystemProperty should load classes with Thread.getContextClassloader …

sco0ter:
…as fallback

OF-1964


GregDThomas:
This LGTM - so I'll merge it.

PR #1581, opened 2020-03-04
OF-532: Fixed by including sessions with non-negative priority=0

AndLuc:
To be compliant with https://tools.ietf.org/html/rfc6121#section-8.5.3.2.1,
for private messages(type="chat"), in case of no resource match,
server MUST deliver this message to any of available resource with non-negative priority


guusdk:
Thanks! I've created a new issue for this: OF-1970 (I don't think that OF-532 quite covers it).

PR #1583, opened 2020-03-19
OF-1974 :lock: Don't send Jetty server name & version in websockets response …

B-apt:
Currently Openfire send Jetty server name and version in response header :

image

This is a bad practice security-wise. Note that in the other connexions (like the admin panel) this option is already set to false.

This commit is for the 4.5 branch.


akrherz:
thank you, awaiting CI

PR #1587, opened 2020-03-21
OF-1976: Admin console: allow any positive number for max muc room occupants

guusdk:
Additionally, room listings no longer show a '0' when listing room occuptant count vs max occupants, when the room max occupant count is unlimited. For those rooms, only the occupant count is shown (eg: "28" instead of "28 / 0").


guusdk:
image

image

PR #1588, opened 2020-03-23
OF-2035 OF-2036 OF-2037: JDBC-based provider improvements

B-apt:
Hello everyone,
Here is my WIP to add the possibility to use an external database (like for Users or Groups) to retrieve the Groups properties.

It's still on testing but this first version is compiling and first tests are looking OK.
Please any comments are welcome and thank you for your time :)

### DIFF
--- a/documentation/db-integration-guide.html
+++ b/documentation/db-integration-guide.html
@@ -77,6 +77,26 @@
         in XML, so you should escape it using "&amp;amp;".
     </li>
 </ul>
+<p>
+  The next following properties are recommanded to finely tune the connection to the database.
+</p>
+<ul>
+  <li>jdbcProvider.username -- You can define the username to access the database with This
+    property if you don't want to set it in the connectionString.
+  </li>
+  <li>jdbcProvider.password -- You can define the password to access the database with This
+    property if you don't want to set it in the connectionString.
+  </li>
+  <li>jdbcProvider.poolMinConnection -- The minimum number of connections created in the
+    connectionPool (integer).
+  </li>
+  <li>jdbcProvider.poolMaxConnection -- The maximum number of connections that can be
+    created in the connectionPool (integer, must be more than poolMinConnection).
+  </li>
+  <li>jdbcProvider.connectionTimeout -- The time after which a connection in the connectionPool
+    will be renewed (double, in day. Example: 0,5 means half a day).
+  </li>
+</ul>
 
 <p>
     Below is a sample config file section (note: the "..." sections in the examples indicate areas
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -994,6 +994,7 @@ error.serverdown.login=Login to the admin console
 
 # General group settings.
 group.read_only=Not allowed: the group account system is read-only.
+group.properties.read_only=Not allowed: Group properties are read-only.
 
 # Group create Page
 
--- a/i18n/src/main/resources/openfire_i18n_cs_CZ.properties
+++ b/i18n/src/main/resources/openfire_i18n_cs_CZ.properties
@@ -450,6 +450,7 @@ error.serverdown.login=Přihlášení do konzole administrátora
 
 # General group settings.
 group.read_only=Není povoleno: systém účtů skupin je jen pro čtení.
+group.properties.read_only=Not allowed: Group properties are read-only.
 
 # Group create Page
 
@@ -2316,7 +2317,6 @@ ssl.import.certificate.truststore.error.alias-missing=Please provide an alias fo
 ssl.import.certificate.truststore.error.alias-exists=A certificate is already stored using this alias. Please provide a different alias, or remove the existing certificate.
 ssl.import.certificate.truststore.error.certificate=Please specify the content of the certificate to import.
 ssl.import.certificate.truststore.error.import=There was an error while trying to import the certificate.
-ssl.import.certificate.truststore.intro=Use the form below to import a certificate that represents a trusted party that can be used when verifying peer certificates while establishing encrypted communication.
 ssl.import.certificate.truststore.intro=Use the form below to import a certificate that represents a trusted party that can be used when verifying peer certificates while establishing encrypted communication.
 
 # Room Occupants Page
--- a/i18n/src/main/resources/openfire_i18n_de.properties
+++ b/i18n/src/main/resources/openfire_i18n_de.properties
@@ -574,6 +574,7 @@ error.serverdown.login=An der Admin-Konsole anmelden.
 
 # General group settings.
 group.read_only=Nicht erlaubt: Das Gruppenkontosystem ist Nur-Lesen.
+group.properties.read_only=Not allowed: Group properties are read-only.
 
 # Group create Page
 
--- a/i18n/src/main/resources/openfire_i18n_es.properties
+++ b/i18n/src/main/resources/openfire_i18n_es.properties
@@ -535,6 +535,7 @@ error.serverdown.login=Ingrese a la consola de administración
 
 # General group settings.
 group.read_only=No permitido: el sistema de grupos es de lectura solamente.
+group.properties.read_only=Not allowed: Group properties are read-only.
 
 # Group create Page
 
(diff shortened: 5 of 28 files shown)


B-apt:
After some work, I've discovered that classes `JDBCUserProvider/JDBCGroupProvider` were using stock DriverManager to get connections to the database and that is a major blocking point performances-wise.
So I'm currently in the process to adapt the current `DefaultConnectionProvider.java` to be able to use it when connecting to an external database (and not to Openfire's database as it is defined in openfire.xml). 

I'm beeing careful to not modify the default behaviours so as to be compliant with current installations of Openfire, but any advice is welcome.

B-apt:
To improve the performances of the JDBC providers classes:

```
- JDBCUserProvider
- JDBCUserPropertyProvider
- JDBCGroupProvider
- JDBCAdminProvider
- JDBCAuthProvider
```

I've added a new class: `org.jivesoftware.database.ExternalDbConnectionManager`  for two main goals:

1. Centralize and manage all the properties for the connection to the external database
2. Create a pool of connections to this external database

Then I refactored all the aforementioned JDBC classes to remove the properties from them and add a field of type `ExternalDbConnectionManager` to get SQL connections. This was mostly painless because all the JDBC providers used the same properties to setup their connection (which ultimately means that all those classes would use the same external database as the properties were the same from one to another).
 
**All but one**: `JDBCUserPropertyProvider` This class was using its own named properties ( `jdbcUserPropertyProvider.driver` and `jdbcUserPropertyProvider.connectionString` instead of `jdbcProvider.driver` and `jdbcProvider.connectionString` for all other classes). With my refactoring this class is now using the same properties than the other classes which potentially would break existing installation that is using yet another external database specific to `JDBCUserPropertyProvider`.

But seeing the current performances of the whole mechanism of JDBC providers, I'm willing to say that it's unlikely that anyone is using it in actual production environment.
@guusdk @akrherz What's your mind on that ?

GregDThomas:
Sorry, coming to this late I know. 

I've no real idea if anyone is using the JDBCUserPropertyProvider class today. However, it does strike me that it may be necessary to have different instances of ExternalDbConnectionManager - consider that you may wish to have basic user details (e.g. JDBCUserProvider / JDBCAuthProvider) from one DB (e.g. a phpBB or MediaWiki installation (*)) and more Openfire specific details (e.g. JDBCUserPropertyProvider) from another.

(*) Something I've done many years ago as both phpBB and MediaWiki provide handy registration mechanisms, but not with Openfire - so recognise that this is an entirely theoretical scenario.

So, in an ideal world, I'd like to see something like the following ...

`dbConnectionManager = ExternalDbConnectionManager.getInstance(JDBCUserProvider.class)`
vs.
`dbConnectionManager = ExternalDbConnectionManager.getInstance(JDBCAuthProvider.class)`
vs.
`dbConnectionManager = ExternalDbConnectionManager.getInstance(JDBCUserPropertyProvider.class)`

etc. The `ExternalDbConnectionManager` would get an existing (or create new) manager for the specified class. That would allow different classes to use different connection properties, with appropriate defaults if not set.

That said, I recognise I don't have bandwidth contribute much but feedback (criticism?) right now, so treat accordingly.

B-apt:
Hello @GregDThomas , no worries PR are for this!
So I've just commited a fix to have the possibility to use different external database depending of the JDBCproviders.

For now only the `JDBCUserPropertyProvider` has specific properties, but specific ones for each providers could be easily added. I haven't done this work for now as this is not already part of Openfire...

By the way, the Travis build is failing now, but before it was working (and tests are running OK on my local) ❓ 💭

akrherz:
> By the way, the Travis build is failing now,

We've had a long and painful history with non-deterministic tests failing for reasons that have been difficult to track down :(

GregDThomas:
Having had a quick look at the code, there's one thing I don't like - there's a circular dependency between `ExternalDbConnectionManager` and `JDBCUserPropertyProvider` - the latter uses the former, and the former knows which properties to use when called by the latter.

Could I suggest that instead of `ExternalDbConnectionManager.getInstance(className)` (and here I recognise you simply implemented my suggestion first time around, so apologies for that!) instead the caller passes in the properties to be used - that way the `ExternalDbConnectionManager` does not need to know about the properties used by JDBCUserPropertyProvider - it's simply told about them.

Additionally, given there are six string properties, rather than have a constructor with six parameters with nothing to distinguish them but parameter name, my preference would be to create a small immutable static class in `ExternalDbConnectionManager` with six properties and classes could create a static instance of that to use as a constructor parameter. This static class instance could also be used as the key to the map.

B-apt:
hello @GregDThomas , 
I've tried to fix my code as per your recommendations. I'm not really a fan of inner class, so I created a new one `ExternalDbConnectionProperties` which allow to define the properties to connect to a database. `ExternalDbConnectionManager` is consumming it to create the connections to the database.


I kept classname as keys in the map of DefaultConnectionProviders as it was clearer for me.
Let me know of what you think of this.

Note: documentation is not yet fully fix as per this new version.

And thanks for your time,

GregDThomas:
Thanks, I think that probably expresses intent better, and cuts out the circular dependencies.

guusdk:
Wow, a lot of work has gone into this! Thanks! I like the thoroughness that has gone into this!

One thing that concerns me is that this is a large and complex change. It's also _a lot_ more than just using an external database to retrieve group properties (as the name of the PR suggests)

It appears to me that a good portion of complexity is introduced by making `DefaultGroupPropertyMap` multi-purpose. It is now both the default, but also includes functionality to make the queries configurable. Would it make sense to move the new functionality to a distinctly new class (`JDBCGroupPropertyMap`) and introduce a way to switch implementations. That would also keep things more consistent with other implementations. The fact that the original code was named 'Default'something implies that the original implementator assumed that alternative implementations would exist.

I would like to isolate the changes into distinct parts, so that we can better track them individually. I don't think that there's a need for individual PRs though. Am I right to conclude that that the individual commits are already doing rather a good job isolating the distinct features? Maybe you can squash / fixup the commits a little, to have distinct commits for distinct features.

I have created these JIRA issues to track the distinct changes that are in this PR (please let me know if I need to create addition ones). Can you reword the commit messages to include the JIRA issue reference that describes the change in the commit please?

* [OF-2035](https://issues.igniterealtime.org/browse/OF-2035) Use connection manager for JDBC-based providers
* [OF-2036](https://issues.igniterealtime.org/browse/OF-2036) Introduce a JDBC provider for group properties
* [OF-2037](https://issues.igniterealtime.org/browse/OF-2037) Update commons-dbcp2

(Is there a specific reason to update commons-dbcp2 other than a generic newer-is-better? I'm OK with it either way, just interested)

guusdk:
A new feature for JDBC-based providers was introduced in https://github.com/igniterealtime/Openfire/pull/1644/commits/7004b6fd0196a55013347cb59e05abf7dfb54b01 - it is likely that this PR needs to reflect the same changes. Can you please rebase to master?

B-apt:
@guusdk 
To take it step by step:

- _Rebasing it to master_: it's done

- _Upgrade to commons-dbcp2_: If I remember correctly I upgraded to latest commons-dbcp2 to try to fix my performance issues. It didn't but I decided to keep the latest as "newer-is-better"... (sometimes)

- _JIRA tickets_: thank you

- _Renaming commit_: I will do the renaming and reorganize them so in the end we should have 3 commits corresponding to your 3 tickets, that's ok for you ?

- _JDBCGroupPropertyMap_: At first I tried to go this way but I had some problems with this solution because of the **AbstractGroupProvider** and the **JDBCGroupProvider** (the AbstractGroupProvider is as it's name indicates an abstract class while the other providers like user are based on one base Interface). So it was becoming a very complicated change. I will have another try at this anyway.

B-apt:
@guusdk 
So I took another way to improve the stuff around _DefaultGroupPropertyMap_:

I've moved out the default _SQL requests_ from **`DefaultGroupPropertyMap`** into the **`AbstractGroupProvider`** which will be using them to build its instance of the aforementioned.
It's consistent with the old code as we currently have some _default SQL requests_ managing **Group Properties** defined in **AbstractGroupProvider**.

Then in JDBCGroupProvider, I've left the possibility to customize the requests (if they are not present we will keep the default ofc).
If you're OK with those changes I will merge this last commit with previous one to keep only 3 commits as we said.

PR #1591, opened 2020-04-03
OF-158: Make room default settings use configurable service defaults.

guusdk:
MUC services have configurable default settings to be applied to new rooms. These defaults are already used when an end-user creates a room (through XMPP).

This commit takes applies the same settings to the room creation form that is in the Admin console, with one exemption: rooms created in the admin console are always persistent.

Note that it's possible to configure Openfire with more than one MUC service. If that's the case, selecting the MUC service in which the room is to be created is part of the room creation form. As at that point, the to-be-used service is still unknown, the (other) default values that are used on the same form are based on the first available service. That seems to be an acceptable trade-off, given the limited amount of installs that have more than one MUC service.

### DIFF
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -1206,6 +1206,15 @@ muc.default.settings.enable_logging=Log Room Conversations
 muc.default.settings.max_users=Maximum Room Occupants
 muc.default.settings.error=Error while saving default settings.
 muc.default.settings.update=Settings updated successfully.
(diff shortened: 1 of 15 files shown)


guusdk:
@Fishbowler would you like to review this please?

skipster2k2:
@dominic-surevine could you take a look and test please?

Dominic-surevine:
> @Dominic-surevine could you take a look and test please?
Looks good to me, tested on Chrome on mac osx and the MUC selection worked nicely.

PR #1594, opened 2020-04-07
OF-1984: Replace DummyExternalizableUtil

guusdk:
This introduces a new ExternalizableUtilStrategy that, unlike the existing DummyExternalizableUtil, is fully functional.

The new implementation is named DefaultExternalizableUtil. It's implementation is a near copy of ClusterExternalizableUtil from the Hazelcast plugin.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/cache/DefaultExternalizableUtil.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/cache/DefaultExternalizableUtil.java
@@ -1,0 +1,471 @@
+/*
+ * Copyright (C) 2020 Ignite Realtime Foundation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.util.cache;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.Externalizable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.ObjectStreamClass;
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Default serialization strategy.
+ *
+ * @author Tom Evans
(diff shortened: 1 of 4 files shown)


guusdk:
I didn't have enough time available today to add unit tests for all of the methods defined in ExternalizableUtilStrategy - I did all but four. I left placeholders. Feel free to complete my tests. :)

xmppserver/src/test/java/org/jivesoftware/util/cache/DefaultExternalizableUtilTest.java
```diff
@@ -0,0 +1,622 @@
...
+    }
+
+    /**
+     * Utility class that generates a DataOutput and DataInput that share a common resource, meaning that what is
+     * written to the DataOutput becomes available for reading on the DataInput.
+     *
+     * Instances are single-use. DataOutput should be written to before DataInput is obtained (this implementation was
+     * designed for unit testing only, not for re-use elsewhere).
+     */
+    public static class Bus
```
GregDThomas, line 610: Perhaps make the class/constructor private to re-enforce the "for testing only"

guusdk: Well, it's only available in the test scope anyway. This way, we can re-abuse it in other tests, if we need to.

xmppserver/src/main/java/org/jivesoftware/util/cache/DefaultExternalizableUtil.java
```diff
@@ -0,0 +1,471 @@
...
+
+/**
+ * Default serialization strategy.
+ *
+ * @author Tom Evans
+ * @author Gaston Dombiak
+ */
+// Note that this implementation is heavily based on the implementation provided by the 'clustering' plugin in
+// org.jivesoftware.openfire.plugin.util.cache.ClusterExternalizableUtil - if changes are made to either implementation,
+// it's likely that the other implementation needs a similar change. See https://issues.igniterealtime.org/browse/OF-1984
```
GregDThomas, line 43: Currently, the HZ implementation swaps out the default one with its own. 

I'd suggest that rather than maintain two identical classes (other than class name) that do the same thing, how about simply making ExternalizableUtil#setStrategy a deprecated, null operation and just use a single class (and ultimately remove the HZ one when support for older Openfire is dropped)

GregDThomas:
Overall - a couple of suggestions, but no real problems with this.

xmppserver/src/main/java/org/jivesoftware/util/cache/DefaultExternalizableUtil.java
guusdk: I'd be in favor of that, but, at least for now, both classes are not _exact_ duplicates. The Hazelcast one has one specific reference to Hazelcast implementation in it. I'm not sure what breaks if we effectively remove that.

GregDThomas: Ah, yes. Missed that. Would have to replace `HazelcastInstance.class.getClassLoader()` with introspection (as Openfire won't have access to the HZ classes).

guusdk: That seems like a bridge to far for now. Maybe we can make the implementation in the plugin extend this one to reduce code maintenance, after we bump the Openfire version dependency at some point?

PR #1596, opened 2020-04-08
Various pubsub improvements

guusdk:
This is a refactoring of the PubSub code. Please review individual commit messages for a listing of features.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
@@ -305,7 +305,7 @@ public class IQPEPHandler extends IQHandler implements ServerIdentitiesProvider,
             return reply;
         }
 
-        if (packet.getTo() == null)
+        if (packet.getTo() == null || packet.getTo().equals( new JID(XMPPServer.getInstance().getServerInfo().getXMPPDomain())) )
         {
             // packet addressed to service itself (not to a node/user)
             switch ( packet.getType() )
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
@@ -296,7 +296,7 @@ public class IQPEPHandler extends IQHandler implements ServerIdentitiesProvider,
      * org.jivesoftware.openfire.handler.IQHandler#handleIQ(org.xmpp.packet.IQ)
      */
     @Override
-    public IQ handleIQ(IQ packet) throws UnauthorizedException {
+    public IQ handleIQ(IQ packet) {
         // Do nothing if server is not enabled
         if (!isEnabled()) {
             IQ reply = IQ.createResultIQ(packet);
@@ -305,83 +305,123 @@ public class IQPEPHandler extends IQHandler implements ServerIdentitiesProvider,
             return reply;
         }
 
-        final JID senderJID = packet.getFrom();
-        if (packet.getTo() == null) {
+        if (packet.getTo() == null)
+        {
             // packet addressed to service itself (not to a node/user)
-            final String jidFrom = senderJID.toBareJID();
-            packet = packet.createCopy();
-            packet.setTo(jidFrom);
-
-            if (packet.getType() == IQ.Type.set) {
-
-                // Only service local, registered users.
-                if (!XMPPServer.getInstance().isLocal(senderJID) || !UserManager.getInstance().isRegisteredUser( senderJID.getNode()))
-                {
-                        final IQ reply = IQ.createResultIQ(packet);
-                        reply.setChildElement(packet.getChildElement().createCopy());
-                        reply.setError(PacketError.Condition.not_allowed);
-                        return reply;
-                    }
-
-                PEPService pepService = pepServiceManager.getPEPService(jidFrom);
-
-                // If publishing a node, and the node doesn't exist, create it.
-                final Element childElement = packet.getChildElement();
-                final Element publishElement = childElement.element("publish");
-                if (publishElement != null) {
-                    final String nodeID = publishElement.attributeValue("node");
-
-                    // Do not allow User Avatar nodes to be created.
-                    // TODO: Implement XEP-0084
-                    if (nodeID.startsWith("http://www.xmpp.org/extensions/xep-0084.html")) {
-                        IQ reply = IQ.createResultIQ(packet);
-                        reply.setChildElement(packet.getChildElement().createCopy());
-                        reply.setError(PacketError.Condition.feature_not_implemented);
-                        return reply;
-                    }
-
-                    if (pepService.getNode(nodeID) == null) {
-                        // Create the node
-                        final JID creator = new JID(jidFrom);
-                        final DefaultNodeConfiguration defaultConfiguration = pepService.getDefaultNodeConfiguration(true);
-                        final LeafNode newNode = new LeafNode(pepService.getUniqueIdentifier(), pepService.getRootCollectionNode(), nodeID, creator, defaultConfiguration);
-                        final DataForm publishOptions = PubSubEngine.getPublishOptions( packet );
-                        if ( publishOptions != null )
-                        {
-                            try
-                            {
-                                newNode.configure( publishOptions );
-                            }
-                            catch ( NotAcceptableException e )
-                            {
-                                Log.warn( "Unable to apply publish-options when creating a new PEP node {} for {}", nodeID, creator, e );
-                            }
-                        }
-                        newNode.addOwner(creator);
-                        newNode.saveToDB();
-                    }
-                }
-
-                // Process with PubSub as usual.
-                pepServiceManager.process(pepService, packet);
-            } else if (packet.getType() == IQ.Type.get) {
-                final PEPService pepService = pepServiceManager.getPEPService(jidFrom);
-                pepServiceManager.process(pepService, packet);
+            switch ( packet.getType() )
+            {
+                case set:
+                    return handleIQSetToService(packet );
+                case get:
+                    return handleIQGetToService(packet );
+                default:
+                    return null; // Ignore 'error' and 'result' stanzas.
             }
         }
-        else if (packet.getType() == IQ.Type.get || packet.getType() == IQ.Type.set) {
+        else
+        {
             // packet was addressed to a node.
+            if ( packet.isRequest() ) {
+                return handleIQRequestToUser( packet );
+            } else {
+                return null; // Ignore IQ packets of type 'error' or 'result'.
+            }
+        }
+    }
 
-            final String jidTo = packet.getTo().toBareJID();
+    /**
+     * Process an IQ get stanza that was addressed to the service (rather than to a node/user).
+     *
+     * @param packet The stanza to process.
+     * @return A response (can be null).
+     */
+    private IQ handleIQGetToService( IQ packet) {
+        final JID senderJID = packet.getFrom();
+        final String jidFrom = senderJID.toBareJID();
+        packet = packet.createCopy();
+        packet.setTo(jidFrom);
 
-            final PEPService pepService = pepServiceManager.getPEPService(jidTo);
-            pepServiceManager.process(pepService, packet);
-        } else {
-            // Ignore IQ packets of type 'error' or 'result'.
-            return null;
+        final PEPService pepService = pepServiceManager.getPEPService(jidFrom);
+        pepServiceManager.process(pepService, packet);
+        return null;
+    }
+
+    /**
+     * Process an IQ set stanza that was addressed to the service (rather than to a node/user).
+     *
+     * @param packet The stanza to process.
+     * @return A response (can be null).
+     */
+    private IQ handleIQSetToService( IQ packet) {
+        final JID senderJID = packet.getFrom();
+        final String jidFrom = senderJID.toBareJID();
+        packet = packet.createCopy();
+        packet.setTo(jidFrom);
+
+        // Only service local, registered users.
+        if (!XMPPServer.getInstance().isLocal(senderJID) || !UserManager.getInstance().isRegisteredUser( senderJID.getNode()))
+        {
+            final IQ reply = IQ.createResultIQ(packet);
+            reply.setChildElement(packet.getChildElement().createCopy());
+            reply.setError(PacketError.Condition.not_allowed);
+            return reply;
         }
 
-        // Other error flows were handled in pubSubEngine.process(...)
+        PEPService pepService = pepServiceManager.getPEPService(jidFrom);
+
+        // If publishing a node, and the node doesn't exist, create it.
+        final Element childElement = packet.getChildElement();
+        final Element publishElement = childElement.element("publish");
+        if (publishElement != null) {
+            final String nodeID = publishElement.attributeValue("node");
+
+            // Do not allow User Avatar nodes to be created.
+            // TODO: Implement XEP-0084
+            if (nodeID.startsWith("http://www.xmpp.org/extensions/xep-0084.html")) {
+                IQ reply = IQ.createResultIQ(packet);
+                reply.setChildElement(packet.getChildElement().createCopy());
+                reply.setError(PacketError.Condition.feature_not_implemented);
+                return reply;
+            }
+
+            if (pepService.getNode(nodeID) == null) {
+                // Create the node
+                final JID creator = new JID(jidFrom);
+                final DefaultNodeConfiguration defaultConfiguration = pepService.getDefaultNodeConfiguration(true);
+                final LeafNode newNode = new LeafNode(pepService.getUniqueIdentifier(), pepService.getRootCollectionNode(), nodeID, creator, defaultConfiguration);
+                final DataForm publishOptions = PubSubEngine.getPublishOptions( packet );
+                if ( publishOptions != null )
+                {
+                    try
+                    {
+                        newNode.configure( publishOptions );
+                    }
+                    catch ( NotAcceptableException e )
+                    {
+                        Log.warn( "Unable to apply publish-options when creating a new PEP node {} for {}", nodeID, creator, e );
+                    }
+                }
+                newNode.addOwner(creator);
+                newNode.saveToDB();
+            }
+        }
+
+        // Process with PubSub as usual.
+        pepServiceManager.process(pepService, packet);
+        return null;
+    }
+
+    /**
+     * Process an IQ set stanza that was addressed to the a node/user.
+     *
+     * @param packet The stanza to process.
+     * @return A response (can be null).
+     */
+    private IQ handleIQRequestToUser(IQ packet)
+    {
+        final String jidTo = packet.getTo().toBareJID();
+        final PEPService pepService = pepServiceManager.getPEPService(jidTo);
+        pepServiceManager.process(pepService, packet);
         return null;
     }
 
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
@@ -344,7 +344,7 @@ public class IQPEPHandler extends IQHandler implements ServerIdentitiesProvider,
                         // Create the node
                         final JID creator = new JID(jidFrom);
                         final DefaultNodeConfiguration defaultConfiguration = pepService.getDefaultNodeConfiguration(true);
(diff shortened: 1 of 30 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
```diff
@@ -314,40 +313,17 @@ public IQ handleIQ(IQ packet) throws UnauthorizedException {
             packet.setTo(jidFrom);
 
             if (packet.getType() == IQ.Type.set) {
```
GregDThomas, line 315: Probably a style preference more than anything else;
a) I'd like to see a `switch(packet.getType())` as is should make it clearer as to which types are being handled. I /think/ the `packet.getTo() == null` branch only handles a IQ Set or Get - what's the switch default condition? I think it's the final `return null` of the current method, but would be good to make that really explicit. 
b) Similarly, is it possible to break up the different branches into methods on their own - so we have small methods `handleIQSetToNull(...)`, `handleIQGetToNull(...)`, `handleIQRequest(...)` etc.

xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
```diff
@@ -428,6 +385,31 @@ else if (packet.getType() == IQ.Type.get || packet.getType() == IQ.Type.set) {
...
+     */
+    public void addSubscriptionForRosterItems( final PEPService pepService )
+    {
+        try {
+            final RosterManager rm = XMPPServer.getInstance().getRosterManager();
+            final Roster roster = rm.getRoster(pepService.getAddress().getNode());
+            for (final RosterItem item : roster.getRosterItems()) {
+                if (item.getSubStatus() == RosterItem.SUB_BOTH || item.getSubStatus() == RosterItem.SUB_FROM) {
+                    createSubscriptionToPEPService(pepService, item.getJid(), pepService.getAddress());
+                }
```
GregDThomas, line 406: What's the `else` block? If it should never happen, throw an IllegalStateException. If it's deliverately ignore, add a log statement so at least during debugging it's possible to figure it out.

xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
```diff
@@ -428,6 +385,31 @@ else if (packet.getType() == IQ.Type.get || packet.getType() == IQ.Type.set) {
...
+     * Populates the PEPService instance with subscriptions. The subscriptions that
+     * are added to the PEPService are based on the roster of the owner of the PEPService:
+     * every entity that's subscribed to the presence of the owner, is added as a
+     * subscriber of the PEPService.
+     *
+     * This method adds, but does not remove of update existing subscriptions.
+     *
+     * @param pepService The PEPService to be populated with subscriptions.
+     */
+    public void addSubscriptionForRosterItems( final PEPService pepService )
```
GregDThomas, line 398: (have yet to read where this method is called so take this comment with that in mind)
Given that this method can silently fail, surprised that there isn't some sort of return code (even if only `true`/`false`) to indicate success.

xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
```diff
@@ -499,7 +481,7 @@ private void createSubscriptionToPEPService(PEPService pepService, JID subscribe
      */
     private void cancelSubscriptionToPEPService(JID unsubscriber, JID serviceOwner) {
```
GregDThomas, line 482: See earlier comment about lack of return code

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProvider.java
```diff
@@ -0,0 +1,708 @@
...
+                }
+            } else {
+                break; // Operations that precede anything other than the last operations that are subscription changes shouldn't be replaced.
+            }
+        }
+        operations.add( NodeOperation.removeSubscription( subscription.getNode(), subscription ) );
+    }
+
+    private void process( final NodeOperation operation ) {
+        switch ( operation.action )
```
GregDThomas, line 343: Again, personal preference, but 
a) would make NodeOperation a param to each of the createNode/updateNode/etc/etc/ methods, and
b) Use a map of operation.action to Consumer<NodeOperation> to turn this method in to a couple of lines (albeit at the cost of more effort in class initialisation.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProvider.java
```diff
@@ -0,0 +1,708 @@
...
+            REMOVE_SUBSCRIPTION
+        }
+
+        final Node node;
+        final Action action;
+        final NodeAffiliate affiliate;
+        final NodeSubscription subscription;
+
+        private NodeOperation( final Node node, final Action action, final NodeAffiliate affiliate, final NodeSubscription subscription ) {
+            if ( node == null ) {
```
GregDThomas, line 662: `this.node = Objects.requireNonNull(node, "Argument 'node' cannot be null.");`
replaces four lines with one.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/CollectionNode.java
```diff
@@ -60,14 +68,23 @@
      */
     private int maxLeafNodes = -1;
 
-    public CollectionNode(PubSubService service, CollectionNode parentNode, String nodeID, JID creator) {
-        super(service, parentNode, nodeID, creator);
-        // Configure node with default values (get them from the pubsub service)
-        DefaultNodeConfiguration defaultConfiguration = service.getDefaultNodeConfiguration(false);
+    public CollectionNode( PubSubService.UniqueIdentifier serviceId, CollectionNode parentNode, String nodeID, JID creator, boolean subscriptionEnabled, boolean deliverPayloads, boolean notifyConfigChanges, boolean notifyDelete, boolean notifyRetract, boolean presenceBasedDelivery, AccessModel accessModel, PublisherModel publisherModel, String language, ItemReplyPolicy replyPolicy, LeafNodeAssociationPolicy associationPolicy, int maxLeafNodes)
```
GregDThomas, line 71: Ouch - that's a lot of constructor params and could rapidly become unmaintainable. I think this is possibly a good case for either a configuration POJO as a single constructor param or a builder pattern.

FWIW, I've become a great fan of Project lombok for this sort of thing.
```
@Builder
@NoArgsConstructor(access = AccessLevel.PRIVATE)
@AllArgsConstructor(access = AccessLevel.PRIVATE)
```
Will create a class that can only be constructed with a builder API, and create the builder too.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/Node.java
```diff
@@ -2325,4 +2342,267 @@ public boolean equals(Object obj) {
...
+     * The properties that uniquely identify a node are its service, and its nodeId.
+     */
+    public final static class UniqueIdentifier implements Serializable
+    {
+        private final String serviceId;
+        private final String nodeId;
+
+        public UniqueIdentifier( final String serviceId, final String nodeId )
+        {
+            if ( serviceId == null ) {
```
GregDThomas, line 2358: `Objects.requireNonNull(...,...)` again

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/Node.java
```diff
@@ -2325,4 +2342,267 @@ public boolean equals(Object obj) {
          */
         publisher;
     }
+
+    /**
+     * A unique identifier for a node, in context of all services in the system.
+     *
+     * The properties that uniquely identify a node are its service, and its nodeId.
+     */
+    public final static class UniqueIdentifier implements Serializable
```
GregDThomas, line 2351: Again lombok to the rescue?

Tagging this as `@Data` will create toString, hashCode, equals, methods for you, as well as field getters and setters (when not final).

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubService.java
```diff
@@ -239,4 +239,65 @@
...
+     * A unique identifier for an item, in context of all nodes of all services in the system.
+     *
+     * The property that uniquely identifies a service within the system is its serviceId.
+     */
+    final class UniqueIdentifier implements Serializable
+    {
+        private final String serviceId;
+
+        public UniqueIdentifier( final String serviceId ) {
+            if ( serviceId == null ) {
```
GregDThomas, line 262: Objects.requireNonNull ?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceProviderManager.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class PubSubPersistenceProviderManager
+{
+    private static final Logger Log = LoggerFactory.getLogger( PubSubPersistenceProviderManager.class );
+
+    private static PubSubPersistenceProviderManager instance;
+
+    private PubSubPersistenceProvider provider;
+
+    public synchronized static PubSubPersistenceProviderManager getInstance()
```
GregDThomas, line 36: If getInstance() is going to be called a lot, is it worth double-checking rather then using synchronized on the whole method?

i.e. 

```
public static PubSubPersistenceProviderManager getInstance() {
    if ( instance == null ) {
        synchronized(PubSubPersistenceProviderManager.class) {   
            if ( instance == null ) {
                instance = new PubSubPersistenceProviderManager();
            }
    }
    return instance;
}
```

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceProviderManager.java
```diff
@@ -0,0 +1,85 @@
...
+        initProvider();
+    }
+
+    public PubSubPersistenceProvider getProvider() {
+        return this.provider;
+    }
+
+    private void initProvider()
+    {
+        String className = JiveGlobals.getProperty( "provider.pubsub-persistence.className", CachingPubsubPersistenceProvider.class.getName() );
```
GregDThomas, line 55: I think there is a case for a `SystemProperty<>` here? I found it particularly useful for classes as typo's/incorrect classes end up as the default value rather than breaking things.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PublishedItem.java
```diff
@@ -300,8 +313,123 @@ public static String getItemKey(LeafNode node, String itemId) {
...
+     * The properties that uniquely identify an item are its node, and its itemId.
+     */
+    public final static class UniqueIdentifier implements Serializable
+    {
+        private final String serviceId;
+        private final String nodeId;
+        private final String itemId;
+
+        public UniqueIdentifier( final String serviceId, final String nodeId, final String itemId ) {
+            if ( serviceId == null ) {
```
GregDThomas, line 370: Objects.requireNonNull ...

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PublishedItem.java
```diff
@@ -300,8 +313,123 @@ public static String getItemKey(LeafNode node, String itemId) {
...
+
+        public Node.UniqueIdentifier getNodeIdentifier()
+        {
+            return new Node.UniqueIdentifier( serviceId, nodeId );
+        }
+
+        public String getItemId() { return itemId; };
+
+        @Override
+        public boolean equals( final Object o )
```
GregDThomas, line 409: Replace these with a lombok `@Data` ?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/cluster/FlushTask.java
```diff
@@ -4,21 +4,41 @@
...
-        PubSubPersistenceManager.flushPendingItems(false); // just this member
-    }
+		final PubSubPersistenceProvider provider = PubSubPersistenceProviderManager.getInstance().getProvider();
+		if ( provider instanceof CachingPubsubPersistenceProvider )
+		{
+			if ( uniqueIdentifier != null ) {
+				((CachingPubsubPersistenceProvider) provider).flushPendingItems( uniqueIdentifier, false ); // just this member
+			} else {
+				((CachingPubsubPersistenceProvider) provider).flushPendingItems( false ); // just this member
+			}
```
GregDThomas, line 39: The comment `just this member` appears in both branches - is that right?

xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
guusdk: I initially thought this to be a bit of overkill, but I actually found a small issue while doing this: https://issues.igniterealtime.org/browse/OF-1992

xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
guusdk: I think this is quite correct as-is? PEP subscription is based on presence subscription, so when a new PEP service is created, it's subscriptions should be based on the subscriptions in the roster. The states that are evaluated here are those that reflect an entity being subscribed to the owner-of-the-PEP-service presence (the others are not). Adding a log statement for those might confuse people more than that it prevents confusion, I'd think.

xmppserver/src/main/java/org/jivesoftware/openfire/pep/IQPEPHandler.java
guusdk: That's consistently done throughout Openfire - especially the older bits. 

The alternative here would probably be to prevent the PEP service from being created, which in itself would introduce other problems (every user of Openfire implicitly has a PEP service, even if the database structures do not yet exist).

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProvider.java
guusdk: The methods that are used now were largely pre-existing, and used by others. We could either overload them, or refactor the lot, but I don't think that that effort outweighs the benefit.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProvider.java
guusdk: Meh. I've been using this consistently. We might at one point want to consistently do a search/replace and replace the lot, but I'll leave things as-they-are for now.

Related: what are your thoughts on using `@javax.annotation.Nonnull` and `@javax.annotation.Nullable`? Intellij seems to work with them in a nice way.

GregDThomas: Yes! I use Nonnull / Nullable liberally throught my code, combined with Optional<> return values to help cut down on my mistakes.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/CollectionNode.java
guusdk: I've introduced this monstrosity to avoid the default config to be retrieved in the constructor (thus decoupling responsibilities).

There is an overloaded variant of this that takes a DefaultNodeConfiguration pojo - which basically does what you want. However, that pojo is re-used for both types of Nodes, which both require a different subset from its configuration.

We might want to split up that DefaultNodeConfiguration implementation in more specific variants.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceProviderManager.java
guusdk: I'm not particularly fond of these constructs. They make the code more complex (novice devs will make weird faces when looking at this). My experience with "things like this" in the broadest sense is that the JVM optimizes things in ways that make changes like this mostly ineffective anyways. Unless there's a strong indication that this has a measurable impact on performance, I'd prefer to not prematurely optimize.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/cluster/FlushTask.java
guusdk: Yes, it refers to the 'false' argument - it limits the action to the local cluster member.

guusdk:
Tests show that the implementation provided in `org.jivesoftware.openfire.pubsub.CachingPubsubPersistenceProvider` is not compatible with clustering - which makes sense, as it operates on non-clustered maps.

guusdk:
@GregDThomas My last commit disables a delegate that provides (local-node) caching when the code detects that clustering is enabled. This works around a problem that exists in that cached implementation. A better solution would be to fix that code to be properly cluster-compatible, but it's rather complex, and I don't have the time to invest into that now. Do you see a way to improve things furher?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceProviderManager.java
```diff
@@ -58,7 +58,16 @@ public PubSubPersistenceProvider getProvider() {
 
     private void initProvider()
     {
-        final Class clazz = PROVIDER.getValue();
+        Class clazz = PROVIDER.getValue();
```
GregDThomas, line 61: I probably would have left the default value in place, and do something like ...

```
clazz == ClusterManager.isClusteringEnabled()
    ? PROVIDER.getDefaultValue()
    : PROVIDER.getValue()
```
Though the ultimate end is essentially the same (enabling clustering /after/ the fact will of course cause problems!)

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceProviderManager.java
```diff
@@ -58,7 +58,16 @@ public PubSubPersistenceProvider getProvider() {
...
+            if ( ClusterManager.isClusteringEnabled() ) {
+                Log.debug("Clustering is enabled. Falling back to non-cached provider");
+                clazz = DefaultPubSubPersistenceProvider.class;
+            } else {
+                clazz = CachingPubsubPersistenceProvider.class;
+            }
+        }
+
         // Check if we need to reset the provider class
         if (provider == null || !clazz.equals(provider.getClass()) ) {
```
GregDThomas, line 72: Not sure this code branch is necessary, as the non-dynamic provider can't change, so no need to shut down the old provider.

Or make it dynamic, and add this method as an onChange method.

GregDThomas:
How hard would it be to make the Caching PubSub one cluster aware? Isn't it just a case of making the Map<> fields a Cache<>, and limiting flushes to the senior node?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceProviderManager.java
guusdk: I was meaning to express that this property has no (fixed) default. I thought that to be expressing intent a bit better.

guusdk:
It'd be "cluster aware" but it all caching operations (including reads) would depend on one cluster node. Might as well go straight to the database then, I assume.

Also, there are various collections that contain in-memory data that probably need rewriting. On top of that, they're making extensive use of the ConcurrentMap provided methods. I think those are not supported on Cache?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubPersistenceProviderManager.java
GregDThomas: Ah I see your point. So non-cluster default = `CachingPubsubPersistenceProvider`, cluster default = `DefaultPubSubPersistenceProvider`

Perhaps `DefaultPubSubPersistenceProvider` needs a rename if it's not actually the default?

guusdk: It kind of is - `CachingPubsubPersistenceProvider` will delegate. I could've named it JDBC- I guess, but that'd conflict with potential future ones that also operate on a database.

PR #1597, opened 2020-04-09
Sinttest: Use Smack 4.4.0-alpha3-20200409.010537-8

Flowdalic:


Flowdalic:
I put this on-hold until smack is build with "--release 8" which requires gradle 6.4 to be released.

PR #1598, opened 2020-04-09
Add aioxmpp-based tests to Travis

guusdk:
This commit adds a script, similar to the one for the Smack-based integration tests, that downloads and excutes the tests that are part of aioxmpp (https://github.com/horazont/aioxmpp)

aioxmpp requires anonymous authentication to be enabled. This commit achieves that by adding a corresponding autosetup configuration option to the demoboot config file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ dist: trusty
 sudo: false
 language: java
 addons:
-  apt:
-    packages:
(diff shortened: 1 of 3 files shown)


guusdk:
Thanks @akrherz! I've squashed the merge to hide the crimes.

PR #1606, opened 2020-04-11
OF-1997: Bouncy Castle 1.65 instead of 1.64

Neustradamus:
Bouncy Castle 1.65 instead of 1.64: https://www.bouncycastle.org/releasenotes.html
- http://www.bouncycastle.org/latest_releases.html


guusdk:
Thanks!

PR #1608, opened 2020-04-11
OF-2000: Stop using intern'ed strings as locks

guusdk:
This replaces the (often namespaced) interned strings used to synchronize on (which introduces a risk of deadlock) with a dedicated map of locks that mimics String interning for any other object type (that should be immutable).

Drawbacks of this solution:
- provided by Google Guava that is marked as 'Beta'
- probably more resource intensive than the native, String-based mutex.


GregDThomas:
LGTM. You may also want to look at replacing usages of org.jivesoftware.util.AutoCloseableReentrantLock and removing that class.

PR #1609, opened 2020-04-12
SLF4J 1.7.30 instead of 1.7.26

Neustradamus:
SLF4J 1.7.30 instead of 1.7.26: http://www.slf4j.org/news.html


Neustradamus:
@guusdk: Thanks!

PR #1617, opened 2020-04-14
OF-2004: Adding 'ping' ad-hoc command

guusdk:


akrherz:
taken already in #1614

PR #1620, opened 2020-04-15
OF-2033 Feature: Add automatic cleaning of outdated offline messages

mightymop:
Feature: 
- Add ability to auto clean outdated offline messages
- extended offline message preferences on admin console


i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -1426,6 +1426,13 @@ offline.messages.enter_store_size=Please enter a store size greater than 0 bytes
 offline.messages.bounce_option=Bounce
 offline.messages.drop_option=Drop
 offline.messages.store_option=Store
+offline.messages.clean.title=Clean Offlinemessages
```
guusdk, line 1429: Missing a space

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -79,11 +83,23 @@
...
 
     private static final int POOL_SIZE = 10;
-    
+
     private Cache<String, Integer> sizeCache;
 
+    /**
+     * Members for automatic offline message cleaning
+     * */
+    private int daystolive = 14; //days
```
guusdk, line 98: A default of 14 days is way to short for my feeling. Lets err on the side of caution, and use a year or something.

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -110,6 +126,16 @@ public static OfflineMessageStore getInstance() {
     public OfflineMessageStore() {
         super("Offline Message Store");
         sizeCache = CacheFactory.createCache("Offline Message Size");
+        try
+        {
+            this.daystolive = JiveGlobals.getIntProperty("xmpp.offline.autoclean.daystolive", 14);
+            this.checkinterval = JiveGlobals.getIntProperty("xmpp.offline.autoclean.checkinterval", 30);
+            this.enableAutoClean = JiveGlobals.getBooleanProperty("xmpp.offline.autoclean.enabled", false);
```
guusdk, line 133: Please use the newer construct using `SystemProperty` instead of `JiveGlobals.getProperty`, for example

    public static final SystemProperty<Boolean> PUBSUB_FOOBAR = SystemProperty.Builder.ofType(Boolean.class)
        .setKey("xmpp.pubsub.create.FOOBAR")
        .setDefaultValue(false)
        .setDynamic(true)
        .build();

And then use it as follows:

    if (PUBSUB_FOOBAR.getValue()) { // ... 

This construct replaces the old JiveGlobals. It adds type-safety, and, more importantly, adds a good deal of automatic documentation / configuration options in the admin console.

xmppserver/src/main/webapp/offline-messages.jsp
```diff
@@ -296,6 +308,67 @@
...
+                <tr valign="top">
+                    <td width="1%" nowrap>
+                        <b>
+                            <label for="offlinechecktimer">
+                                <fmt:message key="offline.messages.clean.timer.label" />
+                            </label>
+                        </b>
+                    </td>
+                    <td width="99%">
+                        <input type="text" name="offlinechecktimer" id="offlinechecktimer" size="5" value="<%=offlinechecktimer%>">
```
guusdk, line 349: Please use JSTL to escape values. You can add the value to the page context somewhere in a java snippet (typically on top of the page):

    pageContext.setAttribute( "offlinechecktimer", offlinechecktimer" );

and then use something like this to render it:

   <c:out value="${offlinechecktimer}">

xmppserver/src/main/webapp/offline-messages.jsp
```diff
@@ -296,6 +308,67 @@
...
+                <tr valign="top">
+                    <td width="1%" nowrap>
+                        <b>
+                            <label for="daystolive">
+                                <fmt:message key="offline.messages.clean.daystolive.label" />
+                            </label>
+                        </b>
+                    </td>
+                    <td width="99%">
+                        <input type="text" name="daystolive" id="daystolive" size="5" value="<%=daystolive%>">
```
guusdk, line 366: Please use JSTL instead of `<%=` - even if it is only a number, it's good practice to not inject raw data in the output that's displayed in the browser.

xmppserver/src/main/webapp/offline-messages.jsp
```diff
@@ -296,6 +308,67 @@
...
+                <tr valign="top">
+                    <td width="1%" nowrap>
+                        <b>
+                            <label for="offlinecleaner">
+                                <fmt:message key="offline.messages.clean.label" />
+                            </label>
+                        </b>
+                    </td>
+                    <td width="99%">
+                        <input type="checkbox" name="offlinecleaner" id="offlinecleaner" size="5" <%= offlinecleaner.equalsIgnoreCase("true") ? "checked" : "" %> >
```
guusdk, line 332: Please prevent usage of `<%=` - style snippets. This can probably be replaced by `${offlinecleaner ? "checked" : ""}` if you put that value to the page context.

i18n/src/main/resources/openfire_i18n.properties
mightymop: changed it to "clean offline messages"

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
mightymop: increased it to 365

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
mightymop: done...

xmppserver/src/main/webapp/offline-messages.jsp
mightymop: done...

xmppserver/src/main/webapp/offline-messages.jsp
mightymop: now using JSTL style

xmppserver/src/main/webapp/offline-messages.jsp
mightymop: changed to ${offlinecleaner ? "checked" : ""}

guusdk:
Thanks! I have created issue OF-2033 to track this new feature. I intend to eventually squash your commits into one commit that includes this JIRA identifier. If you want multiple commits, that's fine too, but then please include this identifier on the first line of one of your commits.

Would you mind looking at the issues that LGTM reported? They're insignificant, but it would be good to squash them.

Also consider using `Duration` instead of a numeric time for the SystemProperties that relate to a duration! The benefits of that is that of type safety, as well as not being limited to a specific time unit!

Example:

```
  public static final SystemProperty<Integer> OFFLINE_AUTOCLEAN_MAXAGE = SystemProperty.Builder.ofType(Duration.class)
    .setKey("xmpp.offline.autoclean.maxage")
    .setDefaultValue(Duration.ofYears(1))
    .setChronoUnit(ChronoUnit.DAYS)
    .setDynamic(true)
    .build();
```

Finally, please add a line for each SystemProperty that you added to the i18n files. The key name should be a combination of `system_property`, a period,  and the key of the system property, for example:

    system_property.xmpp.offline.autoclean.maxage=The time after which unread messages are removed from the offline message store

This message will appear in the system property page of the admin console, as highlighted below:

image

mightymop:
ok i've changed type of SystemProperty from Integer to Duration, added descriotions to i8n files and rebased the PR to one commit...

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -79,11 +85,44 @@
...
 
     private static final int POOL_SIZE = 10;
-    
+
     private Cache<String, Integer> sizeCache;
 
+    /**
+     * Members for automatic offline message cleaning
+     * */
+
```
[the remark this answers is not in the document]
mightymop: hmm ok you are semi right ;D
if you change the enable/disable switch in admin console the timer will be restarted with new values... but you are right in case of changing the values in the system values editor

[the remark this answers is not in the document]
mightymop: i have changed them to be not dynamic

guusdk:
Yey :) I think you'll need to rebase though. Github doesn't allow me to commit as-is.

The admin console page has some issues.

image

- When I enable the top checkbox and save, it reverts back to unselected again.
- The bottom two values should display the defaults, if no user data has been provided.

mightymop:
I will create a new PR for this... it is easier 
https://github.com/igniterealtime/Openfire/pull/1654

mightymop:
> 
> 
> Yey :) I think you'll need to rebase though. Github doesn't allow me to commit as-is.
> 
> The admin console page has some issues.
> 
> image
> 
>     * When I enable the top checkbox and save, it reverts back to unselected again.
> 
>     * The bottom two values should display the defaults, if no user data has been provided.

have solved / fixed this issue in the newer PR

PR #1621, opened 2020-04-16
OF-2008: Allow users to be provisioned through autosetup

guusdk:


guusdk:
Slightly awkward to have these 'user1' and 'user2' elements (same with roster items). That's a drawback of the XMLProperties implementation in Openfire, that can't navigate elements with duplicate names.

GregDThomas:
I'm /way/ too late to this party, for which I apologise, and I'm probably about to tell you something you already know, so double apologies, but just in case ...

Are you aware of the `XMLProperties.getProperties(String, boolean)` API - https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java#L247 - which returns a list of properties according to the Javadoc.

You may be able to leverage that to give a slightly nicer implementation of this.

guusdk:
I've tried using that, but failed to make it a success. That method returns the property _values_ which makes it hard to use in a nested setting.

If you can prove me wrong, I'd be actually very happy, as it'd improve things considerably!

GregDThomas:
Just had a quick look; the functionality you implemented went further than I expected.

To avoid the user1/user2/userN syntax, the a different way of doing it is ...

```
<users>
  <username>john</username>
  <username>jane</username>
  <john> <!-- tag name is the username to which these details apply -->
    <password>secret</password> <!-- Defaults to 'secret' if not present -->
    <name>John Doe</name> <!-- Defaults to ${username} if not present -->
    <email>john.doe@example.com</email> <!-- Defaults to ${username}@example.com if not present -->
    <roster>
      <jid>jane@example.org</jid>
      <jane> <!-- tag name is jid.getNode() -->
        <nickname>Jane</nickname> <!-- defaults to jid.getNode() if not present -->
      </jane>
    </roster>
  </john>
  <jane>
    <password>secret</password>
    <name>Jane Doe</name>
    <email>jane.doe@example.com</email>
    <roster>
      <roster>
        <jid>john@example.org</jid>
        <john>
          <nickname>John</nickname>
        </john>
      </roster>
    </roster>
  </jane>
</users>
```

You could then iterate over `getProperties("autosetup.users.username")` to find all the users, and

`getProperties("autosetup.users." + username +".password")` to find the password, etc.

Similarly for the roster entries. Do you think this is worth changing? I can't decide if I prefer it or not.

guusdk:
I actually had something like this, but was worried about issues with unicode names and having to put those in XML element names. At this point, I wonder if it's worth changing a compromise with another compromise. It doesn't really add anything new in terms of what we can do with it, does it?

GregDThomas:
Yeah, that was what I was thinking too, tbh, so I think it's best to leave as-is.

PR #1622, opened 2020-04-16
Messages with a type "normal" SHOULD be stored offline.

hamzaozturk:
According to XEP-0160:

normal -- Messages with a 'type' attribute whose value is "normal" (or messages with no 'type' attribute) SHOULD be stored offline.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStrategy.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStrategy.java
@@ -116,9 +116,9 @@ public class OfflineMessageStrategy extends BasicModule implements ServerFeature
                 // 8.5.3.  localpart@domainpart/resourcepart
                 // 8.5.3.2.1.  Message
 
-                // For a message stanza of type "normal", "groupchat", or "headline", the server MUST either (a) silently ignore the stanza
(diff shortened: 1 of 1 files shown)


akrherz:
Sorry that this PR did not reviewed in time for the next release of Openfire.

guusdk:
Thanks for your PR! Sadly, I think the change will break XMPP specifications.

https://xmpp.org/extensions/xep-0160.html defines what to do when a message, according to the rules in RFC 6121, is eligible for offline handling. RFC 6121 is still normative, as far as I can tell.

If a message of type 'normal' is eligible for offline handling seems to differ based on the 'to' address.

[RFC 6121 section 8.5.2.2.1](https://xmpp.org/rfcs/rfc6121.html#rules-localpart-barejid). applies to messages when the JID contained in the 'to' attribute is of the form `localpart@domainpart`. This section reads:

> For a message stanza of type "normal" or "chat", the server SHOULD either (a) add the message to offline storage (...)

[RFC 6121 section 8.5.3.2.1](https://xmpp.org/rfcs/rfc6121.html#rules-localpart-fulljid) applies to messages when the JID contained in the 'to' attribute  is of the form `localpart@domainpart/resourcepart`. This section reads:

> For a message stanza of type "normal", "groupchat", or "headline", the server MUST either (a) silently ignore the stanza or (b) return an error stanza to the sender (...)

The code that is modified in this PR matches is the code that is executed for the last scenario. I believe that the original code is correct.

PR #1623, opened 2020-04-24
sinttest: update smack to 4.4.0-alpha3-20200416.010506-6

Flowdalic:
and disable tests that fail with openfire.


guusdk:
Thanks! Do you feel like repeating the same thing in https://github.com/igniterealtime/Openfire/blob/master/.github/workflows/continuous-integration-workflow.yml ? :)

PR #1630, opened 2020-05-06
OF-963 - WebSocket keep-alive is not working as expected

deleolajide:
This fix is to remove the test for resuming a stream session in websockets before closing the xmpp session. XEP-0198 stream resumption is not yet implemented.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
@@ -162,8 +162,6 @@ public class XmppWebSocket {
             closeStream(null);
         }
         if (xmppSession != null) {
-            xmppSession.getStreamManager().formalClose();
-
             if (!xmppSession.getStreamManager().getResume()) {
                 Log.debug( "Closing session {}", xmppSession );
                 xmppSession.close();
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
```diff
@@ -162,6 +162,8 @@ void closeSession() {
             closeStream(null);
         }
         if (xmppSession != null) {
+            xmppSession.getStreamManager().formalClose();
+
```
guusdk, line 166: This does not seem right. Effectively, we've allowed resumption to be negotiated, but when the session is closed, we disregard that completely, and instead simply close the session.

It would be a lot better to prevent resumption from being negotiated for these kind of sessions in the first place. I think that can be done by modifying `org.jivesoftware.openfire.streammanagement.StreamManager#allowResume`

deleolajide: > It would be a lot better to prevent resumption from being negotiated for these kind of sessions in the first place. I think that can be done by modifying `org.jivesoftware.openfire.streammanagement.StreamManager#allowResume`

Isn't that a global setting for BOSH as well as TLS over 5222?

deleolajide: Thanks for the quick videoconf to clarify.I will modify allowResume to return false for websockets and then use it in XmppWebSocket#closeSession  with getResume to determine if the xmpp session should be closed or not

PR #1631, opened 2020-05-11
OF-2019 muc archiving delay start

guusdk:
Please refer to individual commit messages for a breakdown of the changes introduced by this PR.

### DIFF
--- a/xmppserver/changelog.html
+++ b/xmppserver/changelog.html
@@ -661,6 +661,8 @@ hr {
 </li>
 <li>[<a href='https://issues.igniterealtime.org/browse/OF-1786'>OF-1786</a>] -         Update MINA to 2.1.3 to avoid high CPU usage on Linux
 </li>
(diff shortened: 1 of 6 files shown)


guusdk:
The original code (prior to this PR) causes this behavior when a couple of thousand MUC services are used (having that amount of services is very uncommon)

image

Note the high CPU load. No users were logged in to Openfire. These CPU cycles are caused by the Archiver thread polling the work queue.

After my initial changes, things improved:

image

@evdherberg further improved the PR to avoid all threads from being instantiated before they are needed. That removes the short spike in CPU.

Note that, as soon as messages are actually being logged, all of these Archive instances (and corresponding CPU utilization) will still occur. The changes in this PR only have a (lasting) effect for environments where logging is disabled.

guusdk:
Mind Cherry picking this to 4.5?

Op zo 17 mei 2020 13:26 schreef daryl herzmann <notifications@github.com>:

akrherz:
> Mind Cherry picking this to 4.5?

Eh, it was not clean when I tried just now.

PR #1636, opened 2020-05-18
OF-1982: Additional JVM arguments to have better TLS compatibility

guusdk:


GregDThomas:
Could/should whatever is used to start the Windows service also have the JVM arguments updated?

guusdk:
That perhaps needs to be found in the install4j scripts?

akrherz:
> That perhaps needs to be found in the install4j scripts?

Attempted in #1637

PR #1638, opened 2020-05-24
OF-2027 - Enable PEP service for local users as well as registered users

pervezroy:
@deleolajide thanks to you and Guus for your time on this issue. I've created a pull request as discussed here: https://issues.igniterealtime.org/projects/OF/issues/OF-2027 and here: https://discourse.igniterealtime.org/t/major-service-issues-caused-by-migrate-private-xml-storage-data-to-pep/87952

### DIFF
--- a/.bash_profile
+++ b/.bash_profile
@@ -1,0 +1,1 @@
+export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/
--- a/.bash_profile
+++ b/.bash_profile
@@ -1,1 +1,0 @@
-export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pep/PEPServiceManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pep/PEPServiceManager.java
@@ -21,6 +21,7 @@ import org.jivesoftware.openfire.user.UserManager;
 import org.jivesoftware.util.CacheableOptional;
 import org.jivesoftware.util.cache.Cache;
 import org.jivesoftware.util.cache.CacheFactory;
+import org.jivesoftware.util.JiveGlobals;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xmpp.packet.IQ;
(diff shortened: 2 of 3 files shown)


deleolajide:
In principle, I am happy with this change request to allow the concept of a registered user to be extended to local JIDs that could be bots or virtual users with a component XMPP address. The use of a setting with default value of false makes it safe to use and available for backward compatibility for plugins that require this. However, we need @guusdk to review this before I can merge.

pervezroy:
Hi Guus

Thanks for reviewing the changes

I can make the changes to use SystemProperty instead of JiveGlobals.
The property xmpp.pubsub.create.localuser  can be named differently if you want ) when set to true manually will only allow users with a valid user JID of the form user@domain to create the Private storage nodes - we only need to read it but don’t need to set it in the code itself 
> With this, there's no predicting what kind of entity ends up creating things. Instead, I'd prefer to explicitly check for the allowable creators/owners of a node. I believe you intend to include these:
> 
> registered, local users
> unregistered (anonymous), local users
> local components

With my changes bullet points 1 and 2 are covered.. local components cannot create PEP nodes and Pubsub Items (for private storage)

And as far as i understand entities other than users with a valid JID will not “pass" through the  check in both the create node methods (in PubsubEngine and PEP service )

In PUBSUB, node creation the creator/requester can be different from the end-user, but for PEP nodes, the requester and end-user are one and the same – because there is only one jid parameter in that can be passed in.

> Unlike regular Pubsub, PEP provides functionality that are explicitly associated with a particular XMPP entity (the account owner). When allowing anonymous users to create PEP Service instances, these instances should be removed as soon as the anonymous user logs off (similar to the removal that takes place when a registered user is removed from the system). This prevents wildgrow of data and corresponding resource usage.
> 
This requirement will not meet our use case, because the the private data stored for local users who have a valid JID, but not a registered account.

We need both types of users: local end users (who have valid JID but don’t need or ability to login to openfire) and still have the use of Private Storage (as in previous versions of Openfire)…
as well as Openfire LDAP users (Authorised Agent users - who can make commands on behalf of such end users).

To allay any security concerns, in this use-case, it will be a valid component component that creates the Private Storage entries for all users (whether registered or un-registered local users.

Private storage is accessed by registered users (user-agent) only on behalf of the local-users.
Our Unregistered local users cannot access the private storage (as they can’t create a session)

Regards
Pervez

guusdk:
I'm probably misunderstanding something. You seem to define a user that is local, has a valid JID, but not an account, and did not authenticate anonymously.

That does not compute with me. Such a user definition doesn't exist in Openfire, as far as I can tell. All users are by definition registered, anonymous, or non-local.

guusdk:
As discussed, we' will replace this with a feature that will go into Openfire's 4.5 branch.

PR #1639, opened 2020-06-01
WEBADMIN should be of C2S connection type

promiselaoliu:
According to OF-1191(https://github.com/igniterealtime/Openfire/commit/83c47b30237bca298da62c04faee39cc940c156c), WEBADMIN should be of client-oriented connection types.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionType.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionType.java
@@ -33,7 +33,7 @@ public enum ConnectionType
     /**
      * Openfire web-admin console.
      */
-    WEBADMIN( "admin.web.ssl.", SOCKET_S2S ),
+    WEBADMIN( "admin.web.ssl.", SOCKET_C2S ),
 
     /**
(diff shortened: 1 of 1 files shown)


guusdk:
This is something that we can bike-shed endlessly over, I think. Is there a particular reason that you need the fallback to be C2S, instead of S2S? From a security perspective, S2S seems more appropriate than C2S, but my biggest concern is that I'd like to avoid changing the existing behavior (that others might depend on) unless there's a good reason to do so.

promiselaoliu:
I just want to make it consistent.
As of OF-1191(https://github.com/igniterealtime/Openfire/commit/83c47b30237bca298da62c04faee39cc940c156c), WEBADMIN also uses the default client.truststore. 

In this case, if S2S seems more appropriate than C2S, then ConnectionType.WEBADMIN should be removed from the condition there.
>>>>>>>>>>>>>>>
if ( Arrays.asList( ConnectionType.SOCKET_C2S, ConnectionType.BOSH_C2S, ConnectionType.WEBADMIN ).contains( type ) )

guusdk:
I think you're right. Thanks!

PR #1646, opened 2020-06-05
OF-1837 move config to system properties 2

guusdk:


guusdk:
I can't figure out why this keeps failing. The improvements here aren't that important, and merging them will cause more conflicts with https://github.com/igniterealtime/Openfire/pull/1588

I'll close this PR in favor of https://github.com/igniterealtime/Openfire/pull/1588. We can circle back to this later, when https://github.com/igniterealtime/Openfire/pull/1588 has been merged, if need be.

PR #1649, opened 2020-06-06
Enable PEP service for non-registered users as well as registered users

pervezroy:
Enable PEP service for non-registered users as well as registered users


i18n/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
guusdk, line 1: This change is probably not needed?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
```diff
@@ -48,6 +64,12 @@
 
     private static final Logger Log = LoggerFactory.getLogger(PubSubEngine.class);
 
+	public static final SystemProperty<Boolean> PUBSUB_NONREGUSER = SystemProperty.Builder.ofType(Boolean.class)
+		    .setKey("xmpp.pubsub.createnode.nonregistereduser")
```
guusdk, line 68: in `openfire_i18n.properties`, add a line like this:

`system_property.xmpp.pubsub.createnode.nonregistereduser=Allows PubSub nodes to be created for JIDs that do not represent an existing user.`

i18n/.gitignore
pervezroy: agree .. probably appeared automatically by the merge process

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubEngine.java
pervezroy: updated the i18.properties file with description

i18n/.gitignore
guusdk: It's still there though. :)

pervezroy: removed

guusdk:
We've agreed that this change will be pushed to the 4.5 branch (to be part of release 4.5.3), but will not be brought to 4.6 or any other future releases (other than those from the 4.5 branch).

PR #1650, opened 2020-06-07
Fix an issue in Fixtures.java and add some Unit Tests for UserManager

lishan89:
1. Fixed an issue in Fixtures.java createUser method that forgot to store the user to the Map.
2. Added some Unit tests for UserManager


guusdk:
Thanks for this! It is always useful to increase test coverage!

Did you create these tests because you're running into some kind of problem in the code?

I like how you kept each new test concise! 

Based on your Github profile, I believe that you are a student. Please allow me to give you a few pointers that will make your future contributions even more valuable: please consider adding javadoc, to clearly document what the purpose of the test is. I also prefer to separate the code in each test in thee distinct parts: "set up test fixture", "execute system under test" and "verify results". This will help future developers to maintain your code.

I've squashed the many commits (that seem to create, then remove, unrelated changes) into one for brevity.

PR #1653, opened 2020-06-08
OF-1836: Properly handle S2S stream errors

guusdk:
Up until now, receiving stream errors were processed in the chatch-all "we don't know what to do with this" code. Instead, Openfire should support proper error parsing.

Functionally, this won't introduce much of a change, as the behavior for processing unprocessable data is to close the connection, which is similar to what happens after processing a stream error.


guusdk:
@dominic-surevine, can you review this please?

PR #1654, opened 2020-06-08
OF-2033 Feature: Add automatic cleaning of outdated offline messages

mightymop:


mightymop:
This PR is a rebase of https://github.com/igniterealtime/Openfire/pull/1620

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -563,4 +619,154 @@ static boolean shouldStoreMessage(final Message message) {
...
+                    }
+                    else
+                    {
+                        if (deleteOldOfflineMessagesFromDB())
+                        {
+                            readSizeForAllUsers();
+                        }
+                    }
+                } catch (Exception e) {
+                    Log.error("Offline message cleaning - Could not set timer for check interval!", e.fillInStackTrace());
```
guusdk, line 649: Why do you use `Log.error("my message", e.fillInStackTrace());` here? Is that different from just doing `Log.error("my message", e);` ?

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -563,4 +619,154 @@ static boolean shouldStoreMessage(final Message message) {
...
+    }
+
+    private void cancelTimer() {
+        if (timer != null) {
+            Log.info("Offline message cleaning - Stop old timer if started");
+            try {
+                timer.cancel();
+                timer.purge();
+            } catch (Exception e) {
+                Log.warn("Offline message cleaning - Could not stop the timer!", e.fillInStackTrace());
```
guusdk, line 662: Same here: I don't think you need `fillInStackTrace()` ?

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -563,4 +619,154 @@ static boolean shouldStoreMessage(final Message message) {
...
+            Log.error(LocaleUtils.getLocalizedString("admin.error"), e);
+        }
+        finally {
+            DbConnectionManager.closeConnection(rs, pstmt, con);
+        }
+    }
+
+    private boolean deleteOldOfflineMessagesFromDB() {
+
+        Log.info("Offline message cleaning - Deleting offline messages older than "+String.valueOf(daystolive)+" days.");
```
guusdk, line 699: No need to change, but for future reference, something like this is easier to use and read:

    Log.info("Offline message cleaning - Deleting offline messages older than {} days", daystolive);

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -563,4 +619,154 @@ static boolean shouldStoreMessage(final Message message) {
...
+            pstmt = con.prepareStatement(DELETE_OFFLINE_MESSAGE_BEFORE);
+
+            long now = System.currentTimeMillis();
+            long delta = daystolive*24*60*60*1000;
+            long pastTime = now - delta;
+            String creationDatePast = String.format ("%015d", pastTime);
+
+            pstmt.setString(1,creationDatePast);
+
+            if (pstmt.execute())
```
guusdk, line 716: Please use `{` and `}` even for single-line blocks. Errors are easy to make without them.

xmppserver/src/main/java/org/jivesoftware/openfire/OfflineMessageStore.java
```diff
@@ -563,4 +619,154 @@ static boolean shouldStoreMessage(final Message message) {
...
+                Log.info("Offline message cleaning - Cleaning successful");
+            else
+                Log.info("Offline message cleaning - Error while execute the cleaining sql script!");
+
+            pstmt.close();
+
+            return true;
+        } catch (SQLException sqle) {
+            Log.error("Offline message cleaning - "+sqle.getMessage(), sqle);
```
guusdk, line 725: Including the message and then logging the stack trace is redundant, as the stack trace will also include the message.

Also, as this will not break functionality, let's log it at as a warning, instead of an error.

xmppserver/src/main/webapp/offline-messages.jsp
```diff
@@ -161,6 +173,14 @@
         if (quota < 0) {
             quota = 0;
         }
+
+        offlinecleaner=manager.getAutoCleanOfflineMessages()?"true":"";
+        offlinechecktimer=String.valueOf(manager.getAutoCleanOfflineMessagesTimer());
```
guusdk, line 178: I don't think you're required to use Strings - you could've just used then numeric values.

mightymop: ok next time ;)

guusdk:
While working on this PR, I found that it had a bug that prevented it from ever being used (one of the SQL queries has a syntax error). Is this code tested?

guusdk:
I've pushed various changes. This probably needs a good once-over before merging.

guusdk:
I've now also rebased to the current tip of Ignite's MASTER branch.

akrherz (approved):

PR #1657, opened 2020-06-20
Bring protocol support in-line with Openfire 3.6.0

laukner:
Scanned Openfire, Tinder and plugins for keywords RFC and XEP. Did not check maturity and completeness of implementation.


laukner:
Found some inconsistencies, also don't like table cell left alignment and will therefore submit another patch

PR #1658, opened 2020-06-20
Bring protocol support in-line with Openfire 4.6.0

laukner:
* Scanned Openfire, Tinder and plugins for keywords RFC and XEP. Did not check maturity and completeness of implementation.
* Aligned tables with XEP-0423: XMPP Compliance Suites 2020


guusdk:
This is great! Thank you so much for providing this!

PR #1660, opened 2020-06-22
OF-2024: Fix favicons on admin console

guusdk:
Openfire loads the resource from the remote server, then tries to determine if its binary content is an image. Sadly, natively, Java's ImageIO implementation does not support ICO content. This commit adds a third party plugin that adds such support.


GregDThomas:
It adds a library - but AFAICT it's never invoked outside of the unit tests?

guusdk:
It's not even directly invoked in the unit test. The library registers itself as a plugin for Java's ImageIO implementation. That's how support is added.

PR #1662, opened 2020-06-22
Many pubsub fixes

guusdk:
This builds on the feedback that we got from changes to PEP and PubSub in 4.6.0 beta.


GregDThomas:
It warms my heart to see all these Strings being replaced with real objects! I keep telling my team that if we want a weakly typed language we shouldn’t be using Java.

guusdk:
I've now added improvements to CAPS to this PR. These changes are related in so far that Pubsub needs to sent notifications when entity capabilities change. The failure to detect changing CAPS for an entity in a timely manner is what caused the Smack integration tests to fail. This should now be fixed. I've re-enabled these tests.

PR #1664, opened 2020-06-26
OF-2046: Ignore comments in sidebar-admin.xml

guusdk:
Under certain conditions (that I have been unable to reproduce) adding comments in the admin-sidebar.xml appears to influence the rendering of the admin console.

It has been reported that reconfiguring the SAX reader to ignore comments prevents the issue from occurring.

Although I've not been able to reproduce the issue, ignoring comments is sufficiently benign that it doesn't hurt to add that to the implementation.


akrherz:
Updated PR to reflect proper Jira issue.  I would speculate that comments are adding stray line feeds that can sometimes cause trouble?  Shrug, appears a harmless change to me, awaiting CI.

PR #1665, opened 2020-06-26
PEP fixes

guusdk:
This addresses issues that I've introduced in my previous PR. Those issues occurred when loading a PEP service that already existed (as opposed to creating a new, empty, service).


akrherz:
Taking this over believe to be known CI failures to support more testing.

PR #1673, opened 2020-07-20
OF-2054: Add form field type to XEP-0157 form field

adiaholic:
If form fields do no mention type, they are by default assumed to be of type text-single. This causes problem for clients in cases where multiple form field values are included without mentioning type. Check [here](https://xmpp.org/extensions/xep-0157.html#example-2).

However, field standardisation for the same xep-0157 [here](https://xmpp.org/extensions/xep-0157.html#registrar-formtype) suggests that the form field type is list-multi. And so, it is better if the server adds type to form fields as this saves the client from first checking `FORM_TYPE` and determine form field type as per standardisation .
Solves OF-2054


guusdk:
Thanks!

PR #1674, opened 2020-07-27
OF-2059: PubSub unsubscription could fail with “xmpp.pubsub.multiple-subscriptions=true”

kosky:
more info here:
https://discourse.igniterealtime.org/t/pubsub-unsubscription-could-fail-with-xmpp-pubsub-multiple-subscriptions-true/88415


guusdk:
Thank you!

PR #1677, opened 2020-08-13
OF-2058: LDAP group with not existing user not loaded

MarcelHeckel:
When 'flattenNestedGroups' was true and a group contained a not existing user, the whole group was not loaded.

(The previous version was also contributed by me https://github.com/igniterealtime/Openfire/pull/1501)


guusdk:
Thanks!

PR #1680, opened 2020-08-14
OF-2062 pubsub item payload size

guusdk:
I've only tested the SQL for HSQLDB, but this PR also contains scripts for all other databases. A review of those would be desirable.

### DIFF
--- a/distribution/src/database/upgrade/31/openfire_oracle.sql
+++ b/distribution/src/database/upgrade/31/openfire_oracle.sql
@@ -2,7 +2,7 @@
 -- Oracle doesn't allow for a data type change. We'll create a new column, then rename it to replace the existing one.
 ALTER TABLE ofPubsubItem ADD ( temp clob NULL );
 UPDATE ofPubsubItem SET temp=payload, payload=null;
-ALTER TABLE ofPubsubItem DROP COLUMN y;
+ALTER TABLE ofPubsubItem DROP COLUMN payload;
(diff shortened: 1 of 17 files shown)


guusdk:
This PR now changes more than what its title leads to believe.

Not only does it update the code that defines the default payload size, it also contains a database update script that will:
- for some databases, increase the size of the column that will hold the payload. This makes all databases roughly equal again in this sense (some were lagging behind - everything now allows for at least ~1GB);
- update the default configuration for each individual PEP service to match the new size;
- update the default configuration for each individual PEP service to now persist items (this relates to OF-2061, which was added in https://github.com/igniterealtime/Openfire/pull/1679 but that was already merged by the time I realized it was needed);

The same to changes are also applied to the configuration of each (leaf) node of all PEP services.

distribution/src/database/upgrade/31/openfire_oracle.sql
```diff
@@ -0,0 +1,17 @@
+-- OF-2062: Allow for plenty of data to be stored.
+-- Oracle doesn't allow for a data type change. We'll create a new column, then rename it to replace the existing one.
+ALTER TABLE ofPubsubItem ADD ( temp clob NULL );
+UPDATE ofPubsubItem SET temp=payload, payload=null;
+ALTER TABLE ofPubsubItem DROP COLUMN y;
```
akrherz, line 5: `DROP COLUMN y;` ?

PR #1684, opened 2020-08-17
Fix discoInfo features when the server is mounting server caps

aberenguel:
When the server is mounting its stream capabilities, the packet.from is not set and remains null.

https://github.com/igniterealtime/Openfire/blob/b8e6a666e9f04b1a443481c424bc096dd87d8221/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalClientSession.java#L922

https://github.com/igniterealtime/Openfire/blob/b8e6a666e9f04b1a443481c424bc096dd87d8221/xmppserver/src/main/java/org/jivesoftware/openfire/entitycaps/EntityCapabilitiesManager.java#L722

So the senderJid null should be considered the server itself.

It is impacting in the calculation of the caps hash version.


guusdk:
Interesting. How do you trigger this problem?

guusdk:
I believe that this change is no longer needed, if https://github.com/igniterealtime/Openfire/pull/1732 gets merged.

guusdk:
I'm closing this in favor of #1732 - if anyone feels that this is wrong, please speak up.

PR #1687, opened 2020-08-19
Fix: Purging ofPubsubItem table when configured with Oracle because of incorrect Oracle query

devillaccc:
Fix: Openfire if configured with Oracle, clearing ofPubsubItem table. On comparing with MySQL or PostgreSQL queries, oracle sub query is not returning expected rows. Its causes table getting cleared 

@guusdk @akrherz @deleolajide


akrherz:
Is [OF-1720](https://issues.igniterealtime.org/browse/OF-1720) at play here?

devillaccc:
OF-1720 raised for DB2. Description mentioned about Oracle too.

Actual issue is with condition "rownum < ?" in inner query "SELECT id FROM ofPubsubItem WHERE serviceID=? AND nodeID=? ORDER BY creationDate DESC) where rownum < ? order by rownum".

It should be rownum <= ? or with optimized query in this pull request.

PR #1688, opened 2020-08-19
OF-2064: Invoke IQResultListener on other cluster nodes

guusdk:
If an IQ result can't be processed locally, try to identify another cluster node that can.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
@@ -17,6 +17,9 @@
 package org.jivesoftware.openfire;
 
 import org.dom4j.Element;
+import org.jivesoftware.openfire.cluster.ClusterManager;
+import org.jivesoftware.openfire.cluster.IQResultListenerTask;
+import org.jivesoftware.openfire.cluster.NodeID;
 import org.jivesoftware.openfire.container.BasicModule;
 import org.jivesoftware.openfire.handler.IQHandler;
 import org.jivesoftware.openfire.interceptor.InterceptorManager;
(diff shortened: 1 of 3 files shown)


guusdk:
This is a bit of a puzzler. I've found this code in a copy of Openfire for one of my customers. That code base was originally based on Openfire 4.1.5. This code was added, using the commit message:

> backport of cluster fixes from Openfire 4.3.0

I can't, however, find any related code in Openfire 4.3.0 or later.

In any case, does this change ring a bell with anyone? /me eyes Greg

What I assume has happened is that I've identified a problem, fixed it in the fork, with every intention of feeding it back upstream, but never got around to the latter.

I've retroactively created a JIRA issue that describes the behavior of this patch.

xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
```diff
@@ -296,18 +303,27 @@ private void handle(IQ packet) {
         }
         if (packet.getID() != null && (IQ.Type.result == packet.getType() || IQ.Type.error == packet.getType())) {
             // The server got an answer to an IQ packet that was sent from the server
+
+            // If there's a listener for this result at all, then it's likely that that listener had been registered
+            // on this cluster node. For efficiency, try the local cluster node before triggering tasks in the rest
+            // of the cluster.
             IQResultListener iqResultListener = resultListeners.remove(packet.getID());
             if (iqResultListener != null) {
```
GregDThomas, line 311: Appreciate you didn’t change this, but would be very attempted to use the return value of the remove() method to obtain the resulListener to reduce potential concurrency issues.

GregDThomas (approved):

xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
GregDThomas: Other than the trivial comment above, this all looks good to me. I don’t recall anything like this, but it seems to make sense.

guusdk: I'm not sure what you mean.

GregDThomas: Yeah, misread something - ignore that comment!

PR #1691, opened 2020-08-26
OF-2060: revert OF-974

guusdk:
(the description of the PR is copied from the commit message of the second commit, which holds most relevant changes)

When a server joins or leaves a cluster, the implementation that's behind the Cache interface is swapped. When switching _to_ a clustered implementation (_from_ a local/non-clustered implementation), the replacement cache can be used to interact with the cache content that's shared in the cluster. When switching _from_ a clustered implementation (_to_ a local/non-clustered implementation) the cache is replaced with a new (empty) default cache.

OF-974 (now reverted) introduced a change that caused the content from the old implementation to be copied to the new one during the switch-over. This prevented data to be 'lost'. This approach has a considerable drawbacks: When data exists in both caches under the same key, one of the entries will be lost. Depending on the usage of a cache, different techniques to merge data can be desirable. The solution for OF-974 does not allow for that. Also, OF-974 assumes that it is desirable to retain data from the cluster after the node leaves the cluster. This is questionable. In cases, usage-specific code is needed to 'clean up' caches. This leads to a fragmentation of the responsibility of maintaining the cache content over various places: the utilizing code of the cache, as well as the CacheFactory. This adds code complexity.

The original strategy (which has been restored by reverting OF-974) was for code to anticipate that, during cluster join and leave action, the data from the local node was missing from the cache. This introduced the various `restoreCacheContent` method implementations. This has it's own challenges: there's the problem of 'missing data' (which lead to OF-974 in the first place), but also of not having access to the state of the cache prior to the change makes it hard to detect what changes are applied, which in turn makes it difficult to invoke corresponding event listeners, etc). It does, however, allow for far more granular control over the cache content, on a per-usage base. Additionally, cache content control can more easily be implemented in one central place, reducing complexity of the solution.

This commit restores the original strategy, mainly by restoring notion of 'caches will not contain local data directly after a switchover' and re-introducing various `restoreCacheContent` implementations, in places that currently inherit from `ClusterEventListener`.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java
@@ -1229,40 +1229,6 @@ public class SessionManager extends BasicModule implements ClusterEventListener
         return false;
     }
 
-    /**
-     * Removes a session that existed on a remote cluster node.
-     *
-     * This method was designed to be used only in case of cluster nodes disappearing.
-     *
-     * This method takes responsibility for cleaning up the state in RoutingTableImpl
-     * as well as SessionManager.
-     *
-     * @param fullJID the address of the session.
-     */
-    public void removeRemoteClientSession( JID fullJID )
-    {
-        // Remove route to the removed session.
-        routingTable.removeClientRoute(fullJID);
-
-        // Existing behavior when this implementation was introduced was to not dispatch SessionEvents
-        // for destroyed remote sessions.
-
-        // Non-local sessions cannot exist in the pre-Authenticated sessions list. No need to clean that up.
-
-        // TODO only sent an unavailable presence if the user was 'available'. Unsure if this can be checked when the remote node is shut down.
-        Presence offline = new Presence();
-        offline.setFrom(fullJID);
-        offline.setTo(new JID(null, serverName, null, true));
-        offline.setType(Presence.Type.unavailable);
-        router.route(offline);
-
-        // Stop tracking information about the session and share it with other cluster nodes.
-        // Note that, unlike other caches, this cache is populated only when clustering is enabled.
-        sessionInfoCache.remove(fullJID.toString());
-
-        // connectionsCounter tracks only local sessions. No need to decrement it here.
-    }
-
     public int getConflictKickLimit() {
         return conflictLimit;
     }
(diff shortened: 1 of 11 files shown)


guusdk:
This is a change that can have a considerable impact. I intend to build on this and test things further, before merging, but I'd love to have feedback on the general course of action. This is why I created the PR, but left the it in 'draft' for now.

:eyes: @GregDThomas

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
```diff
@@ -708,29 +708,35 @@ public void joinedCluster( final byte[] nodeID )
...
-        modified.get( false ).keySet().forEach( removedDomain -> {
-            Log.debug( "The local cluster node left the cluster. The component '{}' was living on one (or more) other cluster nodes, and is no longer available. Invoking the 'component unregistered' event.", removedDomain );
-            notifyComponentUnregistered( removedDomain );
-
-            // Also clean up the cache of cluster-wide service discovery results.
-            componentInfo.remove( removedDomain.toString() );
-        } );
+        final Iterator<Map.Entry<String, IQ>> iterator = componentInfo.entrySet().iterator();
+        while (iterator.hasNext()) {
+            // TODO: Iterating over the 'componentInfo' is not ideal, as there might be components for which there is no recorded data in 'componentInfo'. There currently is no other way to determine what components were active on the cluster though.
```
GregDThomas, line 729: It it worth raising a ticket so that this TODO isn't forgotten about ...

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
```diff
@@ -708,29 +708,35 @@ public void joinedCluster( final byte[] nodeID )
...
+        if (XMPPServer.getInstance().isShuttingDown()) {
+            // Do not put effort in restoring the correct state if we're shutting down anyway.
+            return;
         }
 
-        // The local cluster node left the cluster.
-        //
+        // Upon leaving a cluster, clustered caches are reset to their local equivalent (by the swap from the clustered
+        // cache implementation to the default cache implementation that's done in the implementation of
+        // org.jivesoftware.util.cache.CacheFactory.leftCluster). This means that they now hold no data (as a new cache
```
GregDThomas, line 719: I don't doubt what you've written here ("This means that they now hold no data") but this seems ... surprising. There's a fair bit of logic in `CacheFactory#leftCluster()` that seems to be redundant and/or straight up broken. A comment to that effect in there may be useful.

xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java
```diff
@@ -1613,98 +1579,110 @@ public int getSessionDetachTime() {
...
+        if (XMPPServer.getInstance().isShuttingDown()) {
+            // Do not put effort in restoring the correct state if we're shutting down anyway.
+            return;
         }
+
+        // Upon leaving a cluster, clustered caches are reset to their local equivalent (by the swap from the clustered
+        // cache implementation to the default cache implementation that's done in the implementation of
+        // org.jivesoftware.util.cache.CacheFactory.leftCluster). This means that they now hold no data (as a new cache
+        // has been created). Data that's available on the local node needs to be added again.
+        restoreCacheContent();
```
GregDThomas, line 1610: Is it necessary to tell any listeners that previously available 'remote' sessions are no longer available?

GregDThomas:
A couple of comments, but this looks like it will solve the problem, IMHO. 

I wonder what `CacheFactory#leftCluster()` is trying to achieve. There seems to be a lot of work there to simply create an empty bunch of Maps. 

I do wonder if it's worth providing an API to register a ClusteredCacheListener to a Cache. An interface with three methods ...
* This node has joined a cluster.  Adds local resources to add to the cache, and informs remote cluster nodes as appropriate.
* This node has left the cluster. Removes remote resources from the cache (though given behaviour of CacheFactory actually re-adds local resources) and informs any interested parties that remote resources are no longer available.
* Another node has left the cluster. Called on senior node only, removes all resources on that node, and tells interested parties that they are no longer available.

This is essentially what you've done, but by registering a listener 
a) Makes it nice and explicit, and
b) Allows a nice green 'cluster safe' tick to be applied on the Cache admin-ui page, so it's obvious which caches have been considered.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
guusdk: Yes - and there's a similar TODO elsewhere that should also be captured. Will do after we're confident that this PR actually gets merged.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
guusdk: I do not understand the point that you're trying to make.

As I read the code, this is the event of things when the local node leaves the cluster:

1. `CacheFactory#leftCluster()` gets invoked, which swaps the implementation of each Cache. This swaps the factory for caches for one that provided `DefaultCache` (which is non-clustered), and replaces the wrapped cache implemention, by removing the Hazelcast-backed version with a new `DefaultCache` implementation. As the cache instance is new, it is empty.
2. All `org.jivesoftware.openfire.cluster.ClusterEventListener#leftCluster()` event listeners are invoked. This invokes the method to which you added this comment. As the cache is now empty, it will need to be repopulated with (local) data. This is done by the `restoreCacheContent()` invocation. After the cache is repopulated, any and all event listeners that need to be triggered to reflect the fact that we've "lost" resources on the cluster are invoked

That last bit, triggering the event listeners, can be tricky, as it can be hard to figure out what data was 'lost' (as it's already gone).

xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java
guusdk: Yes, good catch, I've missed this here.

What I (tried to) do is look at the 'regular flow' that adds or removes entries in the cache. As that flow includes invoking event listeners, so should the corresponding actions that are done when entries are added or removed as a result of joining or leaving a cluster.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
GregDThomas: Sorry, was probably being a bit obtuse. 

My basic point was that I was surprised that open leaving a cluster, a clustered cache is completely emptied. A comment, somewhere, to that effect would be useful, perhaps in CacheFactory.

In more detail, looking in `CacheFactory#leftCluster()` - https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/main/java/org/jivesoftware/util/cache/CacheFactory.java#L904 - line 904 seems to attempt to put the old entries of the cache back in to the newly wrapped cache - `standaloneCache.putAll(cache);`. If the newly wrapped cache is always empty then I can only assume `cache` is always empty by this point. This looks like a bug to me, but not sure if it's even possible to fix, or if it's wise to do so.

guusdk:
Thanks for your feedback! As always, that's greatly appreciated!

> I wonder what `CacheFactory#leftCluster()` is trying to achieve. There seems to be a lot of work there to simply create an empty bunch of Maps.

As I mentioned in a response to an in-line comment above, I don't quite get this point. I don't feel that `CacheFactory#leftCluster()` is doing all that much? It swaps out the cache factory, then iterates over all caches that are, up to that point, clustered, and replaces them with a non-clustered implementation. By going through the factory, the configuration options (size, lifetime) of each cache is guaranteed to be applied.

I'm not sure about the `ClusteredCacheListener` that you suggest. It does pretty much the same thing as the pre-existing `ClusterEventListener`, although that doesn't act on a per-cache basis? In some points, that might be beneficial, as you'd might want to work on a couple of caches at the same time?

An alternative solution that I'd love to have is to be able to somehow retain data from the cluster briefly, when switching from the Hazelcast-backed cache to the local one. I wonder if that's feasible though:

- when the node has left the cluster, it can't access data on the cluster.
- this means that at all times, a copy of all entries must exist on all cluster nodes.

This will have scalability issues.

The inverse (keeping a reference to locally-provided data when joining a cluster might be more feasible.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
guusdk: Hah! That `putAll()` is the fix for OF-974, which is _exactly_ what is being reverted by the first commit in this PR!

I'm assuming a `putAll()` is used as a result of re-using the same construct that was used when _joining_ a cluster (in [CacheFactory#joinedCluster()](https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/main/java/org/jivesoftware/util/cache/CacheFactory.java#L882)). There, the cache that's being operated on is not empty: It "contains" all data that's available on all other cluster nodes (through Hazelcast).

GregDThomas:
> I'm not sure about the `ClusteredCacheListener` that you suggest. It does pretty much the same thing as the pre-existing `ClusterEventListener`, although that doesn't act on a per-cache basis? In some points, that might be beneficial, as you'd might want to work on a couple of caches at the same time?

That's a good point, a class may have many inter-related caches, so a single ClusterEventListener would be better, as you have now.
 
> An alternative solution that I'd love to have is to be able to somehow retain data from the cluster briefly, when switching from the Hazelcast-backed cache to the local one. I wonder if that's feasible though:

I would hope that a Hazelcast IMap continues to hold the existing (local) entries when it leaves the cluster. Though a quick Google isn't showing me much either way. 

> * when the node has left the cluster, it can't access data on the cluster.
> * this means that at all times, a copy of all entries must exist on all cluster nodes.
> 
> This will have scalability issues.

Currently, the backup count is 1 for all our clustered maps - `hazelcast-cache-config.xml`. That means in a two-node cluster, both nodes have a copy of all entries. Obviously the more nodes you get, the less each node has - but I'm not sure what the normal deployment is, if there is such a thing.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
guusdk: This basically is the crux of issue that I'm trying to solve. Those `putAll()`'s are effective, but to crude.

A notable problem is when there are cache entries both on the cluster as on the local node that both use the same key, but different values. With the `putAll()`, one of them will be lost. That is the cause of OF-2060.

GregDThomas: Doh! Missed that. Hmm, this might be a better case for implementing a new listener method somewhere - the method takes the existing set of data, and throws away the stuff it doesn't want, firing off any listeners. The default implementation throws away everything.

guusdk:
> as you have now.

Note that I didn't introduce/create this functionality just now. This has always been how clustering in Openfire works, as per Gato's implemention in ~2007.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
guusdk: Yeah, I was thinking of that. My only concern is that this would require the code to pull in all cached data to a cluster node that has already left the cluster - which won't work, as it's no longer in the cluster. The obvious alternative being a setup in wich all data is being stored on all cluster nodes, which probably is unacceptable from a scalability perspective.

Maybe it's feasible to define a smaller data object, that only contains enough data needed to invoke event listeners et al? Maaaaybe all _key_ values of each cache is always available on all cluster nodes, or something like that?

GregDThomas: Though on reflection - this code is only triggered when the local node leaves the cluster. This is probably an exceptional condition anyway (split brain, or cluster turned off manually are the only two cases I can think of).

So either accept (a) listeners aren't fired - i.e. as implemented in this PR, or (b) only 'some' listeners are fired if the number of nodes in the cluster before leaving was > 2.

But if this PR stands as-is, I would like to see somewhere a nice big comment highlighting that plugins that make use of  their own clustered caches will need to realise that the cache will be empty when the local node leaves a cluster.

guusdk: It's already being mentioned in https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/main/java/org/jivesoftware/openfire/cluster/ClusterEventListener.java#L34-L36 and https://github.com/igniterealtime/Openfire/blob/master/xmppserver/src/main/java/org/jivesoftware/openfire/cluster/ClusterEventListener.java#L58-L59

If we can make things more clear, I'm all for it though. Should we add it to the Javadoc of Cache itself?

xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java
guusdk: In hindsight, I don't think this is needed after all: There are six caches in play here, but only the content of one of them goes accompanied by firing off event listeners (sessionInfoCache). However, when already running in a clustered environment, those events are never broadcasted over the cluster, so there shouldn't be a need to do so for all sessions that were gained/lost when joining or leaving a cluster either.

guusdk: I've now added code comment to most implementations, to explicitly define what the thoughts are on the need to invoke event listeners.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
guusdk: I've added a bit of additional javadoc to the Cache interface.

xmppserver/src/main/java/org/jivesoftware/openfire/component/InternalComponentManager.java
guusdk: https://issues.igniterealtime.org/browse/OF-2066
https://issues.igniterealtime.org/browse/OF-2067

guusdk:
I'm thinking all concerns have been addressed?

GregDThomas:
Yes, all looks good to me.

PR #1693, opened 2020-08-27
OF-2068 Display the versions of various nodes in the cluster

GregDThomas:
This is to fix https://github.com/igniterealtime/openfire-hazelcast-plugin/issues/47 but it turned out to be easier to implement this in core Openfire rather than the clustering plugin itself. @guusdk is it worth raising an Openfire JIRA ticket to cover this off?

It was also fairly easy to display the various plugin versions too - probably useful as I've certainly tripped myself up when running subtly different versions on different nodes.

If any version is different, then a warning symbol is displayed.

What you see if you're running an old version of Openfire in the cluster;
With old server

What you see when that version has been upgraded;
With new server
(note the different version of the Hazelcast plugin and the missing XML Debugger plugin)

And what the other node sees:
From other side

### DIFF
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -3555,6 +3555,7 @@ system.clustering.using-embedded-db=Clustering is not available when using an em
 system.clustering.not-installed=Clustering support was not found on the system. Install a <a href="available-plugins.jsp">plugin</a> that adds clustering support.
 system.clustering.not-valid-license=Clustering license is not valid. You need to update your license to enable clustering.
 system.clustering.starting=Clustering is being started. It may take up to 30 seconds to complete. Click {0}here{1} to refresh.
+system.clustering.versions.label=Cluster Versions Overview
 
 # Security Auditor page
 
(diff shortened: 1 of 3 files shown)


guusdk (approved):
Ah, that's convenient! Yeah, if it doesn't use anything Hazelcast-specific, it might as well be moved into core. Having an issue is good for bookkeeping. I've created this one: https://issues.igniterealtime.org/browse/OF-2068
Can you reword the commit to include a reference to the issue ID?

GregDThomas:
Looking at it now. Also, after reflection, I think it's worth moving the Openfire version in to the "versions" table, rather than in the cluster overview. I put it in the overview when it was just the Openfire version, but as there are now many versions, I think it's worth keeping them all together.

GregDThomas:
Commit reworded, and code updated to a single table like:

image

guusdk:
With a significant amount of plugins, positioning them all isn't ideal. The table quickly becomes wider than the browser:

image

akrherz:
> The table quickly becomes wider than the browser:

I think this is a problem generally with the Openfire Console CSS, many of the console's pages do this.

GregDThomas:
What’s the ‘normal’ number of servers in a cluster? Would be easy to pivot the table (each plugin on a separate row instead of each server). 

Alternatively, any other suggestions welcome! Even if it’s a CSS tweak! (I hate HTML+CSS!)

guusdk:
> (I hate HTML+CSS!)

same same

I'd say the chances of having a large amount of plugins and a small amount of cluster nodes is larger than the other way around, so pivoting the table might be good.

GregDThomas:
Have a look at https://github.com/igniterealtime/Openfire/pull/1697 ...

PR #1698, opened 2020-08-31
OF-2072: Allow empty mechanisms tag to be suppressed

guusdk:
By setting sasl.client.suppressEmpty and/or sasl.server.suppressEmpty one can now prevent a <mechanisms> tag to be included when it would not contain any SASL mechanisms (eg: when they're all disabled in configuration).


guusdk:
Tested by community member: https://discourse.igniterealtime.org/t/non-sasl-obsolete-xep-0078/88679/7?u=guus

PR #1703, opened 2020-09-15
OF-2101: JWTAuthProvider for authenticating with Json Web Token 

mightymop:
- login with user & token
- token will be checked for
  > "subject" (same as user and will be used for user search)
  > expiration date
  > issuer must be same as configured in OF SystemProperties
- set JWT Secret String in OF SystemProperties (used vor validating jwt signature)


guusdk:
This is an exciting new feature to add! However, especially since it's security related, I'd prefer to see both:

- more (any!) code comments and javadoc
- unit tests
- end-user / admin oriented documentation on how to use this new feature (similar to what we have for the JDBC-providers.

Can you add those?

guusdk:
I've created this issue in JIRA to track this new feature: https://issues.igniterealtime.org/browse/OF-2101

There are a lot of commits in this PR. If you would not mind squashing them, can you add "OF-2101" to a commit message? That way, automation will link the JIRA issue to these commits.

mightymop:
> 
> 
> This is an exciting new feature to add! However, especially since it's security related, I'd prefer to see both:
> 
>     * more (any!) code comments and javadoc
> 
>     * unit tests
> 
>     * end-user / admin oriented documentation on how to use this new feature (similar to what we have for the JDBC-providers.
> 
> 
> Can you add those?

I have added comments and javadoc and cleaned up the code...
But i have never used unit tests...

xmppserver/pom.xml
```diff
@@ -378,6 +378,51 @@
...
+            <type>pom</type>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-root</artifactId>
+            <version>0.11.2</version>
+            <type>pom</type>
+        </dependency>
+        <dependency>
+            <groupId>org.scala-lang</groupId>
```
guusdk, line 422: Is scala-lang actually needed? I can't seem to find any imports that depend on it.

mightymop: Removed the unused dependency...

guusdk:
I'm very interested in having this in Openfire, but I'd like to go over this in more detail that what I am comfortable with prior to the upcoming 4.6.0 release. Lets aim to include this in 4.7.

gjaekel:
@deleolajide You know about this PR? I wonder if this Auth-Provider may be also used for OFmeetings/Pàdé? 
Then, in the next sprint, it might be married with it and enhanced to optionally "freeze" some other properties within the users JWT like the name of the room it's valid for.

I got this idea from a German school, which is using (at the moment) plain Jitsi and a self-made, web-based central class schedule application. After login to this, each pupil get an individual time table with lessons, task and tests with identification based on JWTs. For a common virtual classroom session, it has just to follow the link and will be dropped into the right room. And the JWT in this link isn't usable for other times, other rooms and the teacher can be comparable sure that the user of the link is "this" pupil.

If OpenFire/OpenMeetings is used in a business context, this might be also used to provide "foreign tickets" to any session with different restrictions as need (timeslot-based, room-based).

PR #1708, opened 2020-09-21
UserManager getUser method overloaded for JID

abdurrahmanekr:
There is a usage in the form of `webManager.userManager.getUser(member)` in the [group-edit.jsp](https://github.com/igniterealtime/Openfire/blob/c4cbe6564c9aa5fc6593bb1ebbcde1dbdf1e6764/xmppserver/src/main/webapp/group-edit.jsp#L672) file. The `member` is of type JID, but it is converted to string when called. And it comes to getUser method as "username@example.com". This is "java.lang.IllegalArgumentException: The input 'username@example.com' is not a valid JID node part: Contains prohibited code points." causing error.

This PR fixed that issue.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/user/UserManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/user/UserManager.java
@@ -292,6 +292,10 @@ public final class UserManager {
             throw new UserNotFoundException("user cannot be null");
         }
 
+        if (!xmppServer.isLocal(user)) {
+            throw new UserNotFoundException("Cannot get remote user");
+        }
+
         return getUser(user.getNode());
     }
 
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/user/UserManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/user/UserManager.java
@@ -281,6 +281,21 @@ public final class UserManager {
     }
 
     /**
+     * Returns the User specified by jid node.
+     *
+     * @param user the username of the user.
+     * @return the User that matches {@code username}.
+     * @throws UserNotFoundException if the user does not exist.
+     */
+    public User getUser(final JID user) throws UserNotFoundException {
+        if (user == null) {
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/user/UserManager.java
```diff
@@ -280,6 +280,21 @@ public User getUser(String username) throws UserNotFoundException {
...
+     * @param user the username of the user.
+     * @return the User that matches {@code username}.
+     * @throws UserNotFoundException if the user does not exist.
+     */
+    public User getUser(final JID user) throws UserNotFoundException {
+        if (user == null) {
+            throw new UserNotFoundException("user cannot be null");
+        }
+
+        return getUser(user.getNode());
```
guusdk, line 295: Thanks for your contribution! I don't think that there's a reason why we should not have an overloaded method like this. However, there is one improvement that should be made: the code should check if the domain of the provided JID matches the domain of the server.

guusdk (changes requested):
Thanks for your contribution! I think that one important additional check should be added. I've highlighted that in the code.

xmppserver/src/main/java/org/jivesoftware/openfire/user/UserManager.java
abdurrahmanekr: hi @guusdk, can I use the isRegisteredUser method to do what you say?

guusdk: I don't think so, as, if I recall correctly, that potentially queries a remote server to see if the user is a known/registered user on that domain (I think that this is an aweful, confusing API, but sadly, that's what we have). 

My suggestion would be to use `XMPPServer.getInstance().isLocal( jid );`

abdurrahmanekr:
@guusdk Is the last thing I did okay?

guusdk:
Looks fine to me! I've created issue [OF-2100](https://issues.igniterealtime.org/browse/OF-2100) to track this change. Thanks again!

PR #1710, opened 2020-09-22
Defend against flaky tests

Fishbowler:
Some PRs are failing due to 1 flaky test in the aioxmpp suite. For now, retry 3 times to get PRs with accurate info.

This isn't the real solution. This is an interim solution to help the CI surface accurate feedback.
(Next steps: Fix or exclude the test in aioxmpp)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ rpmbuild/
 debian/
 
 .gradle
+
+aioxmpp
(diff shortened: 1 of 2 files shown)


Fishbowler:
Looking at the first run, seems that the aioxmpp test runner isn't returning a 0 exit code for 0 failures. 🤔
Silly assumption. More research needed.

Fishbowler:
Turns out you don't exit a loop by magic. You actually need to write the codes. 🤦‍♂️

PR #1713, opened 2020-09-28
Ignore .factorypath files

Fishbowler:
I'm a monster who chooses to do Java work in VS Code on occasion, and these files are generated by the maven plugin.


guusdk:
image

PR #1719, opened 2020-10-01
WIP: Github Actions Otimisations

Fishbowler:


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -290,6 +290,14 @@ jobs:
           # Defend against another commit quickly following the first
           # We want the one that's been tested, rather than the head of master
           ref: ${{ github.event.push.after }}
+      - name: Cache Maven repository
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-java11-maven-${{ hashFiles('**/pom.xml') }}
```
akrherz, line 185: why is java11 hard coded here?

Fishbowler: Because we set up for Java11 just below. The build runs for 2 versions of Java, but the publish only runs for 1.

PR #1722, opened 2020-10-02
OF-2030 The author is always not null.

Alameyo:
The author is assigned through new at the beginning of the method, it always is not null (or the method will throw an exception) so "else" never happens.

I am playing a bit with SonarQube so you might expect few more such small patches :)


guusdk:
Looks good, thanks! Would you mind rewording the commit message to reference issue OF-2030? That way, JIRA automatically references this commit.

Alameyo:
Sure, I will do so in a moment.

PR #1723, opened 2020-10-03
OF-2093: Prevent JDK11 NoSuchMethodError by something else than an explicit cast because Greg pointed me to a much better solution.

guusdk:
When Openfire is compiled with Java 11, but executed in Java 8 a NoSuchMethodError can be thrown when a client tries to connect. This appears to be caused by Java 11 overriding the `Buffer.flip()` method in ByteBuffer and CharBuffer. To avoid issues, this commit casts explicitly to Buffer when the flip method is used, which is the suggested fix found in https://stackoverflow.com/questions/61267495/exception-in-thread-main-java-lang-nosuchmethoderror-java-nio-bytebuffer-flip

I'm not sure if this is the most optimal way to fix this problem, nor if a similar fix needs to be applied in other places. In any case, the client connection issue disappears with this fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,6 +127,18 @@
     </properties>
 
     <profiles>
+
+        <!-- OF-2093: Ensure proper cross-compilation, using a compiler flag that's introduced in Java 9. -->
+        <profile>
+            <id>jdk9+</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
+            </properties>
+        </profile>
+
         <profile>
             <id>release</id>
             <build>


guusdk:
@Fishbowler would this scenario (compile with one version of Java, run with another) make for an interesting addition to continuous integration?

GregDThomas:
a) If I were writing this code, I'd be tempted to tag the offending lines with
`//noinspection RedundantCast`
to avoid the IntelliJ warning about the cast. I'd hope that it's removal would be caught by a review process, but it's an extra added safeguard. I know I like to fix the low-hanging IntelliJ inspection fruit, that's the sort of thing I might do by accident without it.

b) However I probably wouldn't write that code. The `--release` flag on newer versions of javac will remove the need for the explicit cast (covered in the SO post linked in the PR).
 
There's a post at https://stackoverflow.com/a/57695729/359394 which shows how this can be implemented with Maven in a sensible manner. It also means any casts you may have missed will not be an issue, or any future peices of code where a cast is required but the author forgets (or is simpy unaware) will not be an issue.

guusdk:
I think you're right in that using `--release` is the way to go. @Flowdalic suggests something similar in https://issues.igniterealtime.org/browse/OF-2093

GregDThomas (approved):
LGTM. Might want to change the title of the MR, though.

PR #1724, opened 2020-10-04
OF-2094 Mitigate XXE attacks

Alameyo:
Not properly configured XML parsers are potentially vulnerable to XXE attacks. Adding these settings should prevent most of this attacks.
https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing
Solution:
https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#saxreader

Found with SonarQube


guusdk:
Thanks for this! It's good to apply hardening like this.

PR #1721, opened 2020-10-02
OF-2030: FMUC persist config

guusdk:
This adds configuration persistency, by storing config to the database. As a result of adding this, some rework was needed. Basic testing seems to indicate that functionality has been preserved, but this could use further testing.

PR #1725, opened 2020-10-04
Remove duplicated code

Alameyo:
Not functional changes, just refactor to use less code to do the same things. Perhaps I could work to reduce it a bit more.


guusdk:
The same Smack integration test that is (occassionally) failing in this PR now causes an unrelated PR (#1721) to fail too. This makes it less likely that the cause is introduced by these PRs.

guusdk:
A workaround has been merged to master, that should prevent this problem from causing problem. If you rebase your PR, it should go green.

guusdk:
I just found out that I have the power to rebase this branch. I just attempted to do so. :)

Alameyo:
Ok, so I guess it can be merged now or should I change anything else?

guusdk:
Nope, this should be fine. Thanks!

PR #1731, opened 2020-10-13
PEP notification rewrite

guusdk:
This primarily addresses the reported memory leak (OF-2092) but has also fixes for issues identified in the progress of performing the rewrite.


Fishbowler:
LGTM.  
PEP between users is working (using Gajim to test for Moods).
User deletions are triggering appropriate actions.
Can't find any unintended consequences of this change.

Fishbowler (approved):

PR #1732, opened 2020-10-15
OF-2106: UserManager.isRegisteredUser(JID address) API change

guusdk:
The existing method signature is very easy to 'get wrong', as indicated by the various places where I've fixed it (see follow-up commits in this PR). I've changed the signature to better express its desired operation, leaving the old method in place, marking it deprecated.

As a byproduct, I found OF-2107, which I think is a bug that is unlikely to have manifested itself.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
@@ -117,7 +117,7 @@ public class PubSubServiceInfo {
             try {
                 if (username.contains("@")) {
                     JID jid = new JID(username);
-                    if (userManager.isRegisteredUser(jid)) {
+                    if (userManager.isRegisteredUser(jid, true)) {
                         return jid;
                     }
                 } else if (userManager.isRegisteredUser(username)) {
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
@@ -120,8 +120,12 @@ public class PubSubServiceInfo {
                     if (userManager.isRegisteredUser(jid, true)) {
                         return jid;
                     }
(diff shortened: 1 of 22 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
```diff
@@ -117,7 +117,7 @@ public JID getValidJID(String username) {
             try {
                 if (username.contains("@")) {
                     JID jid = new JID(username);
-                    if (userManager.isRegisteredUser(jid)) {
+                    if (userManager.isRegisteredUser(jid, true)) {
```
Fishbowler, line 120: Would this be faster if this came after the check on L123? Do the local search before the remote one?

Fishbowler:
LGTM

There is an overload for the method that takes a string (with some vaguely duplicated code to the "false" path). It's probably useful in some cases, but:

- I found it confusing when looking for unconverted usages of the updated API call.
- Whilst both seem to get roughly equal usage, the String version is usually called with `someJID.getNode()` - is that subtly different from calling `.isRegisteredUser(someJID, false)` in a way I don't understand?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubSubServiceInfo.java
guusdk: I don't think so. The line on L123 only works if the provided value (confusingly named 'username') is _not_ a JID (or at least, doesn't have an `@` character). It wouldn't work with a JID.

The JID processing bit in `isRegisteredUser` does first check if the JID is local, preventing remote lookups if possible. I'm not seeing how to improve on the execution time significantly.

guusdk:
I think you're on to something. I initially thought that the overloaded `isRegisteredUser` that takes a username String would be just fine to have. By looking at its usages, I see an awful lot of instances where the node-part of a JID is used, assuming that the JID refers to a local entity. That's mostly not verified though. It will be a lot safer to use the JID-overloaded `isRegisteredUser`. I'll mark the String-based one as deprecated too, and update all usages to use the JID-based one.

PR #1734, opened 2020-10-16
OF-2092: Additional logging around pubsub persistence

guusdk:


Fishbowler (approved):
LGTM 👍

I've not run this, but all log lines bring in values with sensible string representations 🤷‍♂️

PR #1652, opened 2020-06-08
OF-2110: Refactoring of conditions for C2S

promiselaoliu:

PR #1735, opened 2020-10-16
Refactor ConnectionTypeTest to be parameterised

Fishbowler:
Test from #1652 had a typo, so a good opportunity.

### DIFF
--- a/xmppserver/src/test/java/org/jivesoftware/openfire/spi/ConnectionTypeTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/openfire/spi/ConnectionTypeTest.java
@@ -21,7 +21,7 @@ import org.junit.runners.Parameterized;
 
 import java.util.Arrays;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
 
 /**
  * Unit test that verify the implementation of {@link ConnectionType}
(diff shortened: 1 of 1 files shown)


Fishbowler:
hattip @promiselaoliu for spotting the typo 😉

xmppserver/src/test/java/org/jivesoftware/openfire/spi/ConnectionTypeTest.java
```diff
@@ -16,110 +16,55 @@
...
-        // Setup test fixture.
-        final ConnectionType input = ConnectionType.CONNECTION_MANAGER;
-
         // Execute system under test.
-        final boolean result = input.isClientOriented();
+        final boolean result = connType.isClientOriented();
 
         // Verify result.
-        assertFalse(result);
+        assertTrue(result == expected);
```
guusdk, line 68: Could you change this into `assertEquals(expected, result)` ?

Fishbowler: Good call. Got thrown by junit having no `assertEquals(boolean, boolean)` but the `(Object,Object)` variant works fine. When I change an expected result, the output looks much nicer.

PR #1736, opened 2020-10-16
Add GitHub CodeQL to Openfire

Fishbowler:
Uses the GitHub template with some tweaks.
Integrates some of the high-level security stuff you get from LGTM into the GitHub security dashboard for the project.

Whilst there's no new functionality in GitHub's version (yet) and LGTM offers more (like Warnings) the ease of use makes this useful to org members.


guusdk (approved):
This seems like a sensible addition. Being less verbose than LGTM might be a bonus, as I assume it makes it easier to figure out what things should have more priority. Also, I'm hoping that this is a good deal faster then LGTM?

PR #1737, opened 2020-10-16
Github Actions: Only run maven publish on IgniteRealtime, not on forks

Fishbowler:
It fails on forks, since it has no secrets, but would be nice not to fail.

e.g. https://github.com/akrherz/Openfire/actions/runs/310663956

Oops.


Fishbowler:
Can't test this outside master.
Relies on these docs being correct: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context (less that 100% hitrate on that so far...)

PR #1738, opened 2020-10-16
OF-1789: Explicitly close all handlers when stopping Jetty

guusdk:
Without this, restarting Jetty (sometimes) causes an error, indiating that we're trying to start something that's already started. This seems to happen only when the handler was being used when it was supposed to be stopped (eg: have an active BOSH connection).


akrherz (approved):
given the aggressive try/catch, I would hope this change does not have any side effects.

guusdk:
My concern is that something is not _supposed_ to be stopped when restarting (I'm kind of confused that we need to explicitly stop elements in a collection, even after we've stopped the collection itself) I've done some basic testing, which didn't show any issues.

PR #1739, opened 2020-10-16
Remove unused webdav servlet

Fishbowler:
It was added for "future use" in 2007, and introduces some security warnings, which, whilst benign, are a good call to clean things up.


Fishbowler:
This resolves 6 of the issues highlighted in the security dashboard by #1736

akrherz:
You have any concerns with this @jadestorm ? :)

jadestorm:
=)  I don't honestly even use Openfire anymore so... there's no real reason for me to have any concerns!  Glad to see Openfire is still trucking along though!

guusdk:
> I don't honestly even use Openfire anymore 
😱

PR #1740, opened 2020-10-19
4.6

lakadbagha:
- Fixed bug to revoke self role in conference.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>parent</artifactId>
         <groupId>org.igniterealtime.openfire</groupId>
-        <version>4.6.0</version>
+        <version>4.6.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQAdminHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQAdminHandler.java
@@ -98,6 +98,8 @@ public class IQAdminHandler {
         
         if (!itemsList.isEmpty()) {
             handleItemsElement(role, itemsList, reply);
+        } else if (!itemsList.isEmpty() && !itemsList.toString().contains(packet.getFrom().toBareJID())) {
+            handleItemsElement(role, itemsList, reply);
         }
         else {
             // An unknown and possibly incorrect element was included in the query
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQAdminHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQAdminHandler.java
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.jivesoftware.openfire.muc.spi;
 
 import java.util.ArrayList;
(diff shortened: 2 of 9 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQAdminHandler.java
```diff
@@ -98,6 +98,8 @@ public void handleIQ(IQ packet, MUCRole role) throws ForbiddenException, Conflic
         
         if (!itemsList.isEmpty()) {
             handleItemsElement(role, itemsList, reply);
+        } else if (!itemsList.isEmpty() && !itemsList.toString().contains(packet.getFrom().toBareJID())) {
```
guusdk, line 101: This will never be invoked. When the condition that is added here evaluates to `true`, then the condition in the `if` statement just above the one that you added will also be true. That will prevent the `else` statement from being invoked.

guusdk: The way in which this code checks if a particular bare JID is in a list of elements is very unsafe. Doing a `toString` on the collection might, depending on the implementation of the collection, not even give a result that reflects the String-based equivalent of each of the elements. This should be replaced by a more explicit check that iterates over each element in the collection, somehow.

distribution/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>parent</artifactId>
         <groupId>org.igniterealtime.openfire</groupId>
-        <version>4.6.0</version>
+        <version>4.6.1-SNAPSHOT</version>
```
guusdk, line 8: Changing version identifiers as used in the Maven infrastructure might be desirable, but has nothing to do with the changes suggested in this PR. Please take them out. By doing that, the PR will be a lot easier to review and merge, which helps a lot in us reviewing and merging it.

As an aside: these changes are already in the code: you'll find that similar changes have already been made on the current HEAD of the primary branch.

guusdk (changes requested):
Hello! Thanks for your contribution! I have found a couple of issues with it, that we'll need to fix, before this can be merged.

My biggest concern is that the changes that you provide do not make a functional difference. The code that you added cannot be executed, ever (which I've explained in more detail in a comment made elsewhere in this review).

I'm also not exactly sure what problem you're trying to address. It would be good to first exactly describe what the problem is that you're trying to fix, so that we can work towards a solution together.

Neustradamus:
@lakadbagha: Have you seen comments?

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQAdminHandler.java
lakadbagha: > @lakadbagha: Have you seen comments?

Hi am very sorry for busy. I get new mess in openfire suppose if i joined any conference and my network start dancing(fluctuate) then ID get error. And then ID not joining and getting room info ever again. I asked in community and nobody replied me there.

distribution/pom.xml
lakadbagha: @guusdk  hi first of all am sorry replying late. I found you are correct and it happened because when i check out before commit from 4.5 to 4.6 my codes conflicted. so i checked it again and now it was just one more condition need to added in first IF case and thats we want to run when somebody modify from openfire admin portal.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQAdminHandler.java
lakadbagha: @guusdk Can you please review and update the code.

if (!itemsList.isEmpty() && packet.getFrom() == null) {
          // Case when openfire admin portal is used to modify role.
            handleItemsElement(role, itemsList, reply);
        } else if (!itemsList.isEmpty() && !itemsList.toString().contains(packet.getFrom().toBareJID())) {
         // Case when client try to modify the role.
            handleItemsElement(role, itemsList, reply);
        } else {
            // An unknown and possibly incorrect element was included in the query
            // element so answer a BAD_REQUEST error
            reply.setChildElement(packet.getChildElement().createCopy());
            reply.setError(PacketError.Condition.bad_request);
        }

guusdk:
I'm closing this issue for inactivity, and a lack of a clear objective of the code change.

PR #1745, opened 2020-10-21
[OF-2112] Fixed NPE when routing an IQ without a "from" attribute

nsobadzhiev:
Stanzas without a from attribute are generally fine. So it's not ok to raise an exception when trying to route one in `IQRouter`.

In the PR, no exception will be thrown but also the stanza will NOT be routed if it's S2S and also doesn't have a "from".

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
@@ -330,8 +330,19 @@ public class IQRouter extends BasicModule {
         }
         try {
             // Check for registered components, services or remote servers
-            // It is generally permissible to route stanzas that have no 'from' attribute. However,
+
+            // OF-2112: It is generally permissible to route stanzas that have no 'from' attribute. However,
             // they are not allowed in s2s traffic.
+            if (packet.getFrom() == null && !XMPPServer.getInstance().isLocal(recipientJID)) {
+                // Stanzas that originate from clients _always_ have a 'from' attribute (as that attribute value is set/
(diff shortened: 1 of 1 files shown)


nsobadzhiev:
Hmm, idk why also the change in the javadoc in LocalClientSession is shown...
It's merged already.

guusdk:
> idk why also the change in the javadoc in LocalClientSession is shown.

That can be fixed by rebasing to the current HEAD

Before merging this, I'd like to find out what happens if a stanza without 'from' address but with a 'to' address that refers to a remote domain is being processed. We might need a more explicit error here, but I'm not sure.

nsobadzhiev:
How do you want to do this? I, personally, don't have a setup with remote domains and I'm probably generally lacking knowledge in this so I have to idea how to test that.

guusdk:
I can do that pretty easily, but it takes some time that I cannot immediately spend. I'll have a look later though.

guusdk:
I played with this for a while, and added some code. My reasoning is that, if a scenario unfolds where a stanza is being routed to another domain, but has no 'from', this nearly always is a bug in Openfire itself, and as such, should be clearly logged.

I've also made the rejection of the stanza more explicit (by returning immediately). I've also rebased this branch for good measure.

What do you think?

nsobadzhiev:
I really think it's a good idea to return immediately and explicitly. Otherwise, theoretically, some code after can wrongfully process that packet, which wouldn't be good.

The comment is nice, too. If a packet is dropped and someone is wondering why, they SHOULD see something in the logs.
A step even further would be to even somehow return an error stanza, but you wouldn't know to whom to send it to. So... IGTM!

guusdk:
Excellent. Thanks for the time you've been putting into this. Much appreciated.

PR #1746, opened 2020-10-29
update Jira links to new atlassian cloud instance

akrherz:
see https://discourse.igniterealtime.org/t/89041


akrherz:
One downside to this for consideration is that we "control" the FQDN `issues.igniterealtime.org` and have a working redirect in-place to the cloud instance.  So we'd have to code churn this again if we ever had the Jira cloud URL change or moved to some different Jira cloud/hosted instance in the future.  If this PR is turned down, that's fine.  We should at least update to https for the links, which I can PR if necessary.

My vote is to take this change just to keep things straight-forward.

guusdk (approved):

PR #1748, opened 2020-11-03
Fix that a user can not rejoin a muc after connecting to another node…

cpetzka:
… in the cluster with the same resource and nickname.

See https://discourse.igniterealtime.org/t/null-pointer-exception-joining-a-room/88995/6 for description an solution


guusdk (changes requested):
In the follow-up discussion that we had, it became apparent that these changes do at the very least not fully resolve the issue.

I'm marking this PR as 'request changes' to prevent it from being accidentally merged, until we've sorted out exactly what's needed here.

guusdk:
I'm unable to progress beyond merging this, and restarting analysis to find any remaining issues.

PR #1749, opened 2020-11-03
WIP Fix sending unnecessary broadcast cluster tasks to other nodes

cpetzka:
This can happen, if a cluster node goes down.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -1083,6 +1083,11 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
      * @param leaveRole the role of the occupant that is leaving.
      */
     CompletableFuture<Void> sendLeavePresenceToExistingOccupants(MUCRole leaveRole) {
+        // Do not send the presence for occupants on other nodes
+        if (!leaveRole.isLocal() || leaveRole.isRemoteFmuc()) {
+            return CompletableFuture.completedFuture(null);
+        }
+		
         // Send the presence of this new occupant to existing occupants
         Log.trace( "Send presence of leaving occupant '{}' to existing occupants of room '{}'.", leaveRole.getUserAddress(), leaveRole.getChatRoom().getJID() );
         try {


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -1082,7 +1082,11 @@ void addOccupantRole( @Nonnull MUCRole role )
      *
      * @param leaveRole the role of the occupant that is leaving.
      */
-    CompletableFuture<Void> sendLeavePresenceToExistingOccupants(MUCRole leaveRole) {
+    CompletableFuture<Void> sendLeavePresenceToExistingOccupants(MUCRole leaveRole) {	
+		// Do not send the presence for occupants on other nodes
```
nsobadzhiev, line 1086: Incorrect/inconsistent indentation...

cpetzka: fixed

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -1083,6 +1083,10 @@ void addOccupantRole( @Nonnull MUCRole role )
      * @param leaveRole the role of the occupant that is leaving.
      */
     CompletableFuture<Void> sendLeavePresenceToExistingOccupants(MUCRole leaveRole) {
+        // Do not send the presence for occupants on other nodes
+        if (!leaveRole.isLocal() || leaveRole.isRemoteFmuc())
```
GregDThomas, line 1087: This is picky style thing, I know, but can we have braces around the statement for this if() statement please? It may help prevent bugs in the future as this code changes.

GregDThomas:
Actually, I think there's a bigger issue with this. 

From my reading, without this change the code sends a "Leave" presence message to every occupant of the room on every node. This can be in one of two cases; 

a) If another node has left the cluster. In which case we only want to send to "local" users, or

b) org.jivesoftware.openfire.muc.spi.MultiUserChatServiceImpl#removeUser is invoked, e.g. because a Presence message of type ERROR is received. In which case, the notification needs to be sent to /every/ user, not just the ones on the local node where the message was received. 

So you may need to somehow indicate which of the two cases you're dealing with, and treat accordingly.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
cpetzka: changed

cpetzka:
@GregDThomas  and @nsobadzhiev  thanks for your review!

> Actually, I think there's a bigger issue with this.
> 
> From my reading, without this change the code sends a "Leave" presence message to every occupant of the room on every node. This can be in one of two cases;
> 
> a) If another node has left the cluster. In which case we only want to send to "local" users, or
> 
> b) org.jivesoftware.openfire.muc.spi.MultiUserChatServiceImpl#removeUser is invoked, e.g. because a Presence message of type ERROR is received. In which case, the notification needs to be sent to /every/ user, not just the ones on the local node where the message was received.
> 
> So you may need to somehow indicate which of the two cases you're dealing with, and treat accordingly.

To case b: If I understand it correctly, the presence should be received on the node where the client has the physical connection. There it would be a local muc role (isLocal()==true) and the change would be sent to everyone.

cpetzka:
@GregDThomas thanks for the review!

I see what you mean and will rework the pull request. This may take some time due to other work.

akrherz:
:wave: @cpetzka, do you still have interest in pushing this PR over the finish line?

cpetzka:
> 👋 @cpetzka, do you still have interest in pushing this PR over the finish line?

I don't think so, as the file itself no longer exists and the MUC processing as a whole has been reworked.

PR #1750, opened 2020-11-04
OF-2129: Remove Flash support from Openfire

guusdk:
Openfire’s code seems to support Flash clients. Flash is an ancient technology, that has no use anymore. Ignite’s Sparkweb and XIFF projects are long dead. In a quick test, I couldn’t even get the crossdomain.xml file to work.

Spark-specific code in Openfire:

- adds a bit of complexity (and thus, maintenance)
- raises security concerns
- specifically allows the null character to be parsed in XML, which according to the XMPP specification is not allowed.

Support for Flash should be dropped, which is what this commit does.


xmppserver/src/main/java/org/jivesoftware/util/cache/CacheMXBean.java
```diff
@@ -0,0 +1,4 @@
+package org.jivesoftware.util.cache;
+
+public interface CacheMXBean {
```
GregDThomas, line 3: Probably a bit of JavaDoc here explaining what this is for may help.

guusdk: That should not be in this commit at all, I think. I think I messed up.

guusdk: Should be gone now.

PR #1752, opened 2020-11-11
OF-2077: Improvements around cluster tasks that relate to pubsub.

guusdk:
This PR contains various commits that:
- improve the documentation to express the intent of the implementation
- mark code that should not be used as deprecated
- Fix implementations that unconditionally attempted to execute changes on pubsub entities that potentially are not loaded (OF-2077)
- rename some methods, also to better express intent (to avoid future variants of OF-2077).

Please refer to the comments on individual commits for more details.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/NodeSubscription.java
@@ -177,7 +177,7 @@ public class NodeSubscription {
      * owner JID or a full JID if the owner wants to receive the notification at a particular
      * resource.<p>
      *
-     * Moreover, since subscriber and owner are separated it should be theorically possible to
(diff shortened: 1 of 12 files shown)


GregDThomas:
This looks good - a whole bunch of improvements. My very slight niggle is the "ifLoaded()" methods return null. This sounds like a classic use case for an Optional. But that's probably just personal preference.

akrherz:
@guusdk which release of openfire do you want this PR to target?

guusdk:
I would not mind this going into 4.6.x

guusdk:
I've now replaced the null values with Optional

GregDThomas (approved):

Fishbowler:
I can reproduce this pretty easily on Master, but not on the branch.

Repro steps:

- Have a user on both nodes subscribe to a PubSub node

That’s it. That’s the test.

Have looked at Pubsub, PEP, single servers and clustered operations.

Fix LGTM 👍

PR #1753, opened 2020-11-18
OF-2146: Fix affiliation and role change broadcasts in MUC

guusdk:
This effectively undoes a workaround that was introduced when implementing FMUC. That workaround introces the issue described in OF-2146.

After many hours of testing, we've not been able to reproduce the original issue that the workaround intends to fix.


Fishbowler:
Added new tests to https://github.com/igniterealtime/Smack/pull/446 - these fail on master/4.6, but pass on this branch.

PR #1756, opened 2020-11-18
OF-2148: Change to wider type to prevent information loss

guusdk:
Prevent a possible issue where a wider type gets added to narrower type.

Ideally, this is replaced by proper types (eg: Duration, Instant), but the cost/benefit of applying that larger change does not pay off.


akrherz (approved):

akrherz:
cherry-picked to 4.6 branch

PR #1759, opened 2020-11-20
OF-2152: Apply more explicit address verification

guusdk:
This commit takes the explicit 'from' address verification that was put in place for Presence broadcast handling, and applies the same to Message broadcast handling.

The check has been moved up a bit, and has been applied at the constructor of the tasks that are executed asynchronously. These can be expected to generate more useful stack traces.


guusdk:
Note that this does not fix OF-2152: it only adds more diagnostics.

PR #1760, opened 2020-11-20
OF-2153: Gracefully reject SM resumption on wrong cluster node

guusdk:
Openfire currently does not allow XEP-0198 resumption on a cluster node different than the one that originally had the session.

This commit adds exception handling for this scenario.


akrherz (approved):

akrherz:
cherry picked onto 4.6 branch

PR #1761, opened 2020-11-27
New CI that performs a basic connectivity test

guusdk:


guusdk:
The new test fails, as it is exposing a bug in Openfire (OF-2156).

akrherz (approved):
I defer to Dan on this one.

runConnectivityIntegrationTests
```diff
@@ -0,0 +1,299 @@
...
+        final int iterations = 25;
+
+        // Execute system under test.
+        int result = 0;
+        for (int i=0; i<iterations; i++) {
+            final XMPPBOSHConnection con = new XMPPBOSHConnection(config);
+            con.connect();
+            try {
+                con.login("john", "secret");
+                Thread.sleep(200);
```
Fishbowler, line 174: This is okay, but if we're seeing it flake in the future (e.g. if the local Openfire isn't performant on the GH CI infrastructure), we could replace this magic number with a listener pattern instead.

Fishbowler (approved):
I've left one comment for "future us" but LGTM 👍

PR #1762, opened 2020-11-27
OF-2156: Fix for BOSH async issues

guusdk:


guusdk:
This PR includes the change from https://github.com/igniterealtime/Openfire/pull/1761 as I'm hoping that the changes provided in the later commit fix the issue exposed by the former.

Fishbowler:
There's some conflicts following the merge of #1761. There's no thread.sleep, and you've upped the connections to 50 consecutive connections. This is intentional? This PR contains the intended state of these tests?

Fishbowler:
nm. seen the commit history :)

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpConnection.java
```diff
@@ -113,10 +143,17 @@ public void deliverBody(String body, boolean async) throws HttpConnectionClosedE
             isClosed = true;
         }
 
-        if (body == null) {
-            body = getSession().createEmptyBody(false);
-        }
-        HttpBindServlet.respond(getSession(), this.context, body, async);
+        HttpBindServlet.respond(getSession(), this.context, body != null ? body : session.createEmptyBody(false), async);
```
Fishbowler, line 146: Why the mix of session and getSession() here? Are they equivalent here?

Fishbowler (approved):
Refactoring looks sound. I can't validate the deep side of the business logic, but I'm convinced that if the tests pass that the intended problem is solved.

PR #1763, opened 2020-12-01
Fix unavailable session (revert OF-2012)

Zyndoras:
Hi there,

in previous Openfire versions it was possible for a client connection to exist but to be "unavailable". 
In the admin console it looked like this: 

(see Presence: offline)

[OF-2012](https://igniterealtime.atlassian.net/browse/OF-2012) removed this feature by killing off the Client Route (Client Session) altogether when it sends an `unavailable` presence.

This makes it impossible for Openfire to process any packets by this client, including presences that would make it `available` again.
Besides the hassle to setup a new connection everytime this happens, it seems to me that many XMPP clients actually cannot handle this situation of an unusable connection properly, or do not notice that their connection is dead at all.

So if this effect was unintended I propose to revert [OF-2012](https://igniterealtime.atlassian.net/browse/OF-2012) (ede343d2) with this PR, to get the old behavior back.

Cheers

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/SessionManager.java
@@ -741,7 +741,7 @@ public class SessionManager extends BasicModule implements ClusterEventListener
         if (session.getAddress() != null && routingTable != null &&
                 session.getAddress().toBareJID().trim().length() != 0) {
             // Update route to unavailable session (anonymous or not)
-            routingTable.removeClientRoute(session.getAddress());
+            routingTable.addClientRoute(session.getAddress(), session);
         }
     }
 


Fishbowler:
@guusdk Didn't you do some work recently on reverting OF-2012?

Zyndoras:
After some digging it seems to me that [OF-2012](https://igniterealtime.atlassian.net/browse/OF-2012) was just implemented because of a poorly named function. 

`RoutingTable::addClientRoute` does not only add, but also update Client Sessions. So re-adding that Client Route to update it with the new availability status when it becomes unavailable should be the right thing to do.

`RoutingTable::removeClientRoute` on the other hand, while sounding more fitting for `sessionUnavailable` just removes the Client session, which is not what the availability status intends to do.

hamzaozturk:
It was reverted in 4.6.3.

PR #1764, opened 2020-12-03
OF-2157 cluster-aware sequencing

guusdk:


guusdk:
(lots of comment in the second commit message)

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
```diff
@@ -150,23 +155,29 @@ public SequenceManager(int seqType, int size) {
...
-    public synchronized long nextUniqueID() {
-        if (!(currentID < maxID)) {
-            // Get next block -- make 5 attempts at maximum.
-            getNextBlock(5);
+    public long nextUniqueID() {
+        final Lock lock = sequenceBlocks.getLock(type);
+        lock.lock();
+        try {
+            Data data = sequenceBlocks.get(type);
+            if (data == null || !(data.getCurrentID() < data.getMaxID())) {
```
GregDThomas, line 170: `!(data.getCurrentID() < data.getMaxID())`

Seems like an odd way of expressing this - is there a good reason for it being this way around?

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
```diff
@@ -217,35 +224,26 @@ private void getNextBlock(int count) {
...
-                this.maxID = newID;
+            // already changed the original id that we read. This code should be
+            // called under a lock, so this is not supposed to happen. Throw a
+            // very verbose error if it does.
+            if (pstmt.executeUpdate() == 1) {
+                final Data result = new Data(currentID, newID);
+                sequenceBlocks.put(type, result);
+                return result;
+            } else {
+                throw new IllegalStateException("Failed at attempt to obtain an ID, aborting...");
```
GregDThomas, line 235: As this doesn't set abortTransaction to true, it's not actually aborted? Is that desired?

FWIW, I'd default the rollback flag to true, and set it to false immediately before the the `return result;`, so it only needs to changed in one place.

GregDThomas: Hmm, I now see that the rollback isn't required because no update was performed. So perhaps it's simply the log message ('aborting') that is misleading?

GregDThomas:
Looks fine to me; a few possibly picky comments, feel free to do with them as you see fit!

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
Fishbowler: I like that it expresses the actual intent. The CurrentID _should_ be less than the MaxID and isn't.
Although by that logic, the line should begin with a check that the data isn't not null 😂

Fishbowler:
I've run this on a Docker cluster and fired messages from both sides in alternating and in no particular order, and I'm seeing tidy sequential IDs. That's how this was found. I need to spend more time on the other uses of SequenceManager to look for any unintended side effects.

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
guusdk: @matt-tucker deviced it like that in 2004. Who am I to argue with syntax that survived for 16 years? :)

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
guusdk: This code used to iterate a couple of times, before failing. I've removed that, as, under lock, that iteration should not be needed. With that, the flow became a bit... wonkey. I like the strategy that you're suggestion, as that's less ambiguous here.

guusdk:
I'm somewhat concerned that we're now performing database interaction while under a cluster-wide lock. The lock should be pretty specific to the use-case, but still. Can we improve on this?

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
GregDThomas: Actually, on reflection, I don't think a transaction is required at all. 

There's a single (ACID compliant) SQL statement.

If a row is updated, the returned value can be used. 
If no row is updated, there is nothing to roll back.
If a SQL exception occurs, there is nothing to roll back.

So perhaps simply remove all the unnecessary complexity?

Fishbowler:
Everything looks tidy, works as expected in a cluster, even with one node writing with a delay.

PR #1770, opened 2020-12-17
OF-2163: Prevent unexpected stanza modifications to leak information

guusdk:
Delivering a stanza to a MUC room occupant changes the addressing of the stanza (the room JID is replaced by the real JID). It is often undesirable that this change is applied to the original stanza (that can be processed further), as such changes can leak the privacy-sensitive real address of the recipient.

This commit creates defensive copies of the stanza that is being sent, to prevent leaking the real address of the recipient.


guusdk:
For four messages exchanged in one MUC room, between two people, this seems to result in sensible output:

Screenshot from 2020-12-17 14-37-34

Note that this was using the Monitoring plugin as being developed in https://github.com/igniterealtime/openfire-monitoring-plugin/pull/144

Fishbowler:
Tested with MUC, DMs and 1-2-1s. LGTM 👍

PR #1771, opened 2020-12-18
OF-370

mightymop:
add memberlist retrival when occupant is not in room

(but has to be an serveradmin, mucserviceadmin, muc owner, admin or member)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,36 @@ private void processDecliningInvitationMessage(
                                          info.elementTextTrim("reason"), packet.getFrom());
     }
 
+    /**
+    * checks for memberlist retrival query
+    * @return true if occupant is allowed to retrieve member list otherwise false
+    */
+    private boolean isAllowed(IQ packet)
```
guusdk, line 582: This is a confusing implementation, which would be very hard to maintain.

The method name is very generic - but is checks for a very specific thing. Reflecting that in the method name would be a good improvement.

The method checks if a particular entity has permission to do so. To check permission, it should not be needed to check what the content of the stanza is (if it is a muc#admin query). This probably does serve a purpose - but if a check like this is needed, then it should go in a dedicated method.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -587,16 +618,21 @@ private void process(
         @Nullable final MUCRole preExistingRole )
     {
         // Packets to a specific node/group/room
-        if ( preExistingRole == null )
+        boolean allowMemberListRetrival = false;
```
guusdk, line 621: Before your changes, the code that follows this if-statement assumes that there is a non-null `preExistingRole`. This commit modifies that code in various places to allow for null values. Although it probably works fine, it will also be very error-prone. I believe that it would be an improvement to take the case where `preExistingRole` is null (basically any use-case where a non-occupant is trying to interact with the room and let that be processed by a different method completely. That way, we need not make the existing code more complex.

guusdk (changes requested):

guusdk:
Thanks for your improvement! I left a couple of in-line comments. Apart from that, would you please squash your commits to get rid of the 'merges? Also, 'retrival' is spelled as 'retrieval'. You're doing it consistently the other way around.

mightymop:
Now it should be ok...

Neustradamus:
A big improvement!

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,71 @@ private void processDecliningInvitationMessage(
...
+    {
+        MUCRoom room = server.hasChatRoom(packet.getTo().getNode())?server.getChatRoom(packet.getTo().getNode()):null;
+        if (room!=null&&(room.getOwners().contains(packet.getFrom().asBareJID())||
+            room.getAdmins().contains(packet.getFrom().asBareJID())||
+            room.getMembers().contains(packet.getFrom().asBareJID()))||
+            server.getSysadmins().contains(packet.getFrom().asBareJID())||
+            XMPPServer.getInstance().getAdmins().contains(packet.getFrom().asBareJID()))
+        {
+            Element query = packet.getChildElement();
+            if (query!=null&&query.getName().equalsIgnoreCase("query"))
```
guusdk, line 592: Why is this check necessary? This seems to indicate that "permission" to do something is based on what the client sends. That does not sound safe to me, as we can't verify that what a client sends is truthful (it can easily be abused).

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,71 @@ private void processDecliningInvitationMessage(
...
+        }
+        catch ( ForbiddenException e )
+        {
+            Log.debug("Unable to process IQ stanza: sender don't have authorization to perform the request.", e);
+            sendErrorPacket(packet, PacketError.Condition.forbidden, "You don't have authorization to perform this request.");
+        }
+        catch ( ConflictException e )
+        {
+            Log.debug("Unable to process IQ stanza: processing this request would leave the room in an invalid state (eg: without owners).", e);
+            sendErrorPacket(packet, PacketError.Condition.conflict, "Processing this request would leave the room in an invalid state (eg: without owners).");
```
guusdk, line 624: This error message does not make sense in context of processing a request for a member list.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,71 @@ private void processDecliningInvitationMessage(
...
+        }
+        catch ( NotAllowedException e )
+        {
+            Log.debug("Unable to process IQ stanza: an owner or administrator cannot be banned from the room.", e);
+            sendErrorPacket(packet, PacketError.Condition.not_allowed, "An owner or administrator cannot be banned from the room.");
```
guusdk, line 629: This error message does not make sense in context of processing a request for a member list.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,71 @@ private void processDecliningInvitationMessage(
...
+        }
+        catch ( CannotBeInvitedException e )
+        {
+            Log.debug("Unable to process IQ stanza: user being invited as a result of being added to a members-only room still does not have permission.", e);
+            sendErrorPacket(packet, PacketError.Condition.not_acceptable, "User being invited as a result of being added to a members-only room still does not have permission.");
```
guusdk, line 634: This error message does not make sense in context of processing a request for a member list.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -589,12 +655,21 @@ private void process(
         // Packets to a specific node/group/room
         if ( preExistingRole == null )
         {
-            Log.debug("Ignoring stanza received from a non-occupant of '{}': {}", roomName, packet.toXML());
-            if ( packet.isRequest() )
+            //OF-370: check for memberlist retrieval
+            boolean allowMemberListRetrieval = checkForMemberListQuery(packet); 
+            if (allowMemberListRetrieval && packet.isRequest())
```
guusdk, line 660: This should probably be more specific than `packet.isRequest`. Instead, it should check if the request is a ListRetrieval request (which I think is now part of the implementation of `checkForMemberListQuery` - but I do not think it belongs there).

mightymop: divided into seperate methods now

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
mightymop: removed exception handler because it is not a invite case

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
mightymop: changed error message

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
mightymop: changed error message

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
mightymop: ok I've seperated the two checks.. it now checks for the listretrieval and then it will check if non-occupant is allowed to retrieve the list

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,76 @@ private void processDecliningInvitationMessage(
...
+     * Checks if packet sender (non-occupant) is allowed to retrieve a memberlist.
+     * Sender must have a member privileges at minimum to be allowed. 
+     * @param packet IQ
+     * @return true, if sender is allowed to retrieve or false if not
+     */
+    private boolean hasPermissionToRetrieveList(IQ packet)
+    {
+        JID from = packet.getFrom();
+        MUCRoom room = server.hasChatRoom(packet.getTo().getNode())?server.getChatRoom(packet.getTo().getNode()):null;
+        return  room!=null&&(room.getOwners().contains(from.asBareJID())||
```
guusdk, line 611: The XEP specifies that 'any occupant of a member-only room' should be allowed. I think it is correct to also allow it for server admins, but maybe it's better to check for occupants, rather than for owners, admins and members? I'm not sure how much work this will be (if it is lots, then don't bother).

guusdk, line 611: I believe that there is room for an additional feature. Although the XEP specifies that members SHOULD be able to retrieve the memberlist of a members-only room, it does not explicitly say what to do for rooms that are not members-only. Depending on the configuration of the room, all members are exposed through service discovery. I think that it is safe to say that when a sender is allowed to retrieve the members of a room through service discovery, then it should also be allowed to perform the memberlist query. The relevant code is in `org.jivesoftware.openfire.muc.spi.MultiUserChatServiceImpl#getItems(java.lang.String, java.lang.String, org.xmpp.packet.JID)`

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,76 @@ private void processDecliningInvitationMessage(
                                          info.elementTextTrim("reason"), packet.getFrom());
     }
 
+    /**
+     * This method will parse a packet and checks if it is a MEMBER! list retrieval packet or not. 
+     * @param packet IQ
+     * @return true, if it is or false if not
+     */
+    private boolean isListRetrieval(IQ packet)
```
guusdk, line 583: There might be a security problem here, as, when this method returns 'true', the packet is processed as if it is sent by an admin. That's dangerous. Can we somehow do that differently?

Also, this implementation does not check a couple of things:
- is this a retrieval _request_ (is it of type 'get')? This method will return 'true' even for a 'set' - would that allow people to apply changes, even if they're not supposed to?
- are there _other_ directives (apart from requesting the member list) in this stanza? Can this code be abused to ask for the member list _as well as something else_ to bypass this check?

guusdk (changes requested):
Thanks for your changes. I'm sorry to be pedantic, but I'm still seeing issues. Thank you for your perseverance!

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
mightymop: I am now checking for a request packet...
Furthermore the packet will be checked for not allowed tags... otherwise an exception will be thrown

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
```diff
@@ -574,6 +575,76 @@ private void processDecliningInvitationMessage(
...
+    /**
+     * Checks if packet sender (non-occupant) is allowed to retrieve a memberlist.
+     * Sender must have a member privileges at minimum to be allowed. 
+     * @param packet IQ
+     * @return true, if sender is allowed to retrieve or false if not
+     */
+    private boolean hasPermissionToRetrieveList(IQ packet)
+    {
+        JID from = packet.getFrom();
+        MUCRoom room = server.hasChatRoom(packet.getTo().getNode())?server.getChatRoom(packet.getTo().getNode()):null;
```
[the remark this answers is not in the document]
mightymop: added NPE check

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCUser.java
mightymop: But if the occupant did not join the room, but he wants the retrieve the list it is senceless to check for occupants whether he es allowed to do i think.
So only allowed users should be able to retrieve.

mightymop: I would check if it is a memberonly room then check like above and on the other hand everybody is allows to retrieve the list... that is more simple...

PR #1774, opened 2020-12-22
OF-2170: LDAP userDN cache should contain negative lookups

guusdk:
By caching lookup failures, subsequent lookups are prevented, which will increase performance.

@GregDThomas I would've used `Optional` if it was `Serializable`! ;)

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapManager.java
@@ -19,56 +19,23 @@ package org.jivesoftware.openfire.ldap;
 import org.jivesoftware.admin.LdapUserTester;
 import org.jivesoftware.openfire.group.GroupNotFoundException;
 import org.jivesoftware.openfire.user.UserNotFoundException;
-import org.jivesoftware.util.JiveGlobals;
-import org.jivesoftware.util.JiveInitialLdapContext;
-import org.jivesoftware.util.StringUtils;
-import org.jivesoftware.util.SystemProperty;
+import org.jivesoftware.util.*;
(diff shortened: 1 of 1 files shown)


GregDThomas:
Just leave this here: `org.jivesoftware.util.CacheableOptional`

guusdk:
I should have known better. :) I applied the change.

gjaekel:
What will be the cache time for this negative lookups?

guusdk:
There is no difference in duration between negative lookups and positive
ones, caching wise. The duration / maximum lifetime of an entry is
configurable, and can be looked up on the admin console. By default, it is
6 hours.

gjaekel:
May you consider to add a "negative" cache time in a second sprint? I think there's a conflict between the goals of performance for positive entries and error hunting for negatives.

guusdk:
I'm not sure if the benefit outweighs the complexity of adding/maintaining
that. Also: Using a different (lower) value for negative lookups undoes the
performance benefit that is the reason for having negative lookups cached
in the first place.

The consumer of the cache value will still report the same problems
associated with the value not being available. There shouldn't be a
functional effect. Generally speaking, the negative lookup is repeatable,
and will only be fixed after manual intervention. In such cases, an admin
can manually purge the case to get rid of the negative cached value.

gjaekel:
Thank you for discussion. I was thinking about the "helpdesk" question "Yes, first I forget to add the user in the LDAP. But then I add it, and it's *still* not working!"

It probably will be a complex code change, but what's about an (up to the limit) exponential rising cache time (e.g. starting with 1min and doubling), like it's typical for email queuing.

guusdk:
I do see some benefit in that, but I believe that the added complexity far
outweighs it. In my book, not worth it, to be honest.

PR #1783, opened 2020-12-26
OF-2180: Fixes that a local muc role is overridden by a remote muc role after …

cpetzka:
…cluster start up

It is a follow up to https://discourse.igniterealtime.org/t/null-pointer-exception-joining-a-room/88995/21?u=chp but not introduced by https://discourse.igniterealtime.org/t/null-pointer-exception-joining-a-room/88995/7?u=chp

If a cluster node starts, then it can happen that a client establishes a connection and joins the muc before the cluster connection is ready (e.g.: by a plugin, which needs very long time to start).

If the cluster connection is ready, then the senior member queries all the mucs including the online users of the mucs from this cluster node. The cluster node then queries the Mucs including the online users of the cluster from the senior node and enters them in its own maps. However, the cluster node receives its online users again as a RemoteMucRole and overwrites its own LocalMucRole. The consequence is that the user can still send messages, but no longer receive them.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MultiUserChatManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MultiUserChatManager.java
@@ -32,6 +32,7 @@ import org.jivesoftware.database.SequenceManager;
 import org.jivesoftware.openfire.XMPPServer;
 import org.jivesoftware.openfire.cluster.ClusterEventListener;
 import org.jivesoftware.openfire.cluster.ClusterManager;
+import org.jivesoftware.openfire.cluster.NodeID;
 import org.jivesoftware.openfire.container.BasicModule;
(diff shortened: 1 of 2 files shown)


guusdk:
I'd like to merge this - could you re-word the commit to include a reference to "OF-2180" in the commit message? That will allow our bugtracker to correspond changes with issues.

I can rebase this commit myself, but I'm not sure if I'm taking away the credits for the commit by doing so. I'll try that in an hour or so, unless you've beaten me to it. ;-)

cpetzka:
> I'd like to merge this - could you re-word the commit to include a reference to "OF-2180" in the commit message? That will allow our bugtracker to correspond changes with issues.
>
> I can rebase this commit myself, but I'm not sure if I'm taking away the credits for the commit by doing so. I'll try that in an hour or so, unless you've beaten me to it. ;-)

Thank you, i have changed the commit message and rebased the branch.

PR #1784, opened 2020-12-28
OF-2176 OF-763: Make Admin Console session timeout configurable

gjaekel:
Add a http session listener to the Admin Console plugin that implements a configurable session timeout.

The session timeout is taken from the property `adminConsole.sessionTimeout`, the default is 30min.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/admin/SessionListener.java
+++ b/xmppserver/src/main/java/org/jivesoftware/admin/SessionListener.java
@@ -12,7 +12,6 @@ import javax.servlet.annotation.WebListener;
  * The session timeout is taken from the property {@code adminConsole.sessionTimeout}, the default is 30min.
  *
  * @author Guido Jaekel, guido.jaekel@gmx.de
- * @author Guus der Kinderen, guus.der.kinderen@gmail.com
  */
 
 @WebListener
--- a/xmppserver/src/main/java/org/jivesoftware/admin/SessionListener.java
+++ b/xmppserver/src/main/java/org/jivesoftware/admin/SessionListener.java
@@ -1,0 +1,32 @@
+package org.jivesoftware.admin;
+
+import org.jivesoftware.util.JiveGlobals;
+
+import javax.servlet.http.HttpSessionEvent;
+import javax.servlet.http.HttpSessionListener;
+import javax.servlet.annotation.WebListener;
+
+/**
+ * A http session listener that implements a configurable session timeout.<p>
+ *
+ * The session timeout is taken from the property {@code adminConsole.sessionTimeout}, the default is 30min.
+ *
+ * @author Guido Jaekel, guido.jaekel@gmx.de
+ * @author Guus der Kinderen, guus.der.kinderen@gmail.com
+ */
+
+@WebListener
+public class SessionListener implements HttpSessionListener {
+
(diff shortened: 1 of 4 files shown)


guusdk:
This seems nice and concise. I'm assuming that you've tested it?

I like people signing their name to the implementation that they add. No need to include mine, though. I can't take credits for this one.

For bonus points, perhaps use `org.jivesoftware.util.SystemProperty` instead of `JiveGlobals`?

gjaekel:
> I like people signing their name to the implementation that they add. 

This is my first commit here. Please point me to an example where and how to sign.


> For bonus points, perhaps use org.jivesoftware.util.SystemProperty instead of JiveGlobals?

Something like

    public static final SystemProperty<Duration> STARTUP_RETRY_DELAY = SystemProperty.Builder.ofType(Duration.class)
        .setKey("xmpp.server.startup.retry.delay")
        .setDefaultValue(Duration.ofMillis(2000))
        .setMinValue(Duration.ofMillis(-1))
        .setChronoUnit(ChronoUnit.MILLIS)
        .setDynamic(Boolean.TRUE)
        .build();

?

guusdk:
> > This seems nice and concise. I'm assuming that you've tested it?
> 
> No, the core is taken from an example on the Internet. Maybe a testcase should be added.
> 

No requirement for a test-case, but doing a manual test (maybe with a low value of something like 2 minutes) would be good.

> > I like people signing their name to the implementation that they add.
> 
> This is my first commit here. Please point me to an example where and how to sign.
> 

You misunderstood. I was praising you for adding your name to your code. You do not need to add mine.

> > For bonus points, perhaps use org.jivesoftware.util.SystemProperty instead of JiveGlobals?
> 
> Something analogous to
> 
> ```
> public static final SystemProperty<Duration> STARTUP_RETRY_DELAY = SystemProperty.Builder.ofType(Duration.class)
>     .setKey("xmpp.server.startup.retry.delay")
>     .setDefaultValue(Duration.ofMillis(2000))
>     .setMinValue(Duration.ofMillis(-1))
>     .setChronoUnit(ChronoUnit.MILLIS)
>     .setDynamic(Boolean.TRUE)
>     .build();
> ```
> 

Exactly!

xmppserver/src/main/java/org/jivesoftware/admin/SessionListener.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.servlet.http.HttpSessionListener;
+import javax.servlet.annotation.WebListener;
+
+/**
+ * A http session listener that implements a configurable session timeout.<p>
+ *
+ * The session timeout is taken from the property {@code adminConsole.sessionTimeout}, the default is 30min.
+ *
+ * @author Guido Jaekel, guido.jaekel@gmx.de
+ * @author Guus der Kinderen, guus.der.kinderen@gmail.com
```
Fishbowler, line 15:
```suggestion
```

gjaekel:
> Exactly!

I understand your motivation, but this will need some experience with the spirit of this lib to use the appropriate keywords. I may guess some for sure, but also miss others.

guusdk:
> this will need some experience with the spirit of this lib to use the appropriate keywords. I may guess some for sure, but also miss others.

No worries, I can do that for you. I can push to this PR, I think. Let me figure that out.

I'll merge this in, generate a nightly to be able to test things.

PR #1787, opened 2021-01-05
OF-2191 fix: add type 'headline' to server messages

mightymop:
Until now, server messages do not have the type attribute. 
This could be problematic to some clients, that theese message are totaly ignored.


guusdk:
Thank you! I've squashed the commit into one. It will be part of release 4.6.1.

PR #1788, opened 2021-01-05
Add dependency checking maven profile

Fishbowler:
use `mvn verify -P deps` to see an HTML report outputted to to the target directory

### DIFF
--- a/.dependency-check-suppressions.xml
+++ b/.dependency-check-suppressions.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+   <suppress> <!-- Ignore Openfire for the search jar, else dependency-check picks up every Openfire CVE since Openfire v1.7.2-->
+      <notes><![CDATA[
+   file name: search.jar: search-1.7.2.jar
+   ]]>      </notes>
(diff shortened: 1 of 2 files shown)


.dependency-check-suppressions.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+   <suppress> <!-- Ignore Openfire for the search jar, else dependency-check picks up every Openfire CVE since Openfire v1.7.2-->
```
guusdk, line 3: There's no such thing as "Openfire v1.7.2" - I'm thinking you mean "the version of Openfire that the search plugin version v1.7.2 was build on"?

Fishbowler: Without this exclusion, Dependency Checker gets confused and returns all CVEs for Openfire versions > 1.7.2. Whilst there isn't a 1.7.2, it returns _everything_ since the name change.

guusdk: ah. That's not good. I'm wondering how it uses a version number of a plugin combined with the artifact name of the Openfire project, but that's probably beyond the scope of this PR.

Fishbowler: Yeah, I'd love to dig into DC code to see whether it's wrong or we're doing something wrong. But another problem for another time (or maybe not at all)

PR #1791, opened 2021-01-09
fix: handle result packets in IQVersionHandler

mightymop:
result packets coming through BOSH (may websocket too) were not processed by IQVersionHandler.
On TCP Connection theese results are being processed in org.jivesoftware.openfire.net.SocketRead.java - getIQ()


mightymop:
rebased PR opened...

PR #1792, opened 2021-01-09
OF-2241: handle result packets in IQVersionHandler

mightymop:
result packets coming through BOSH (maybe websocket too) were not processed by IQVersionHandler.
The result is that the software information from theese connections are not listed in session-details.jsp.
On TCP Connection theese results packets are being processed in org.jivesoftware.openfire.net.SocketRead.java - getIQ()
This PR fixes the problem, that only software infos from clients on tcp connection are displayed in the admin console on session details...


xmppserver/src/main/java/org/jivesoftware/openfire/handler/IQVersionHandler.java
```diff
@@ -76,6 +83,40 @@ public IQ handleIQ(IQ packet) throws PacketException {
...
+                            {
+                                localSession.setSoftwareVersionData(element.getName(), element.getStringValue());
+                            }
+                            else
+                            {
+                                /*
+                                  The result comes from a server 2 server connection, so we write the information
+                                  to the log, because we dont need it at this point.
+                                */
+                                Log.info(packet.getFrom()+" "+element.getName()+"="+element.getStringValue());
```
Fishbowler, line 108: Unsure of the value of logging it.
If it's useful, should this be at debug, and include some context? Could be pretty noisy.

mightymop: Sorry was a debug output... forgot to delete it

xmppserver/src/main/java/org/jivesoftware/openfire/handler/IQVersionHandler.java
```diff
@@ -76,6 +83,40 @@ public IQ handleIQ(IQ packet) throws PacketException {
...
+                            }
+                        } else {
+                            Log.warn("No software version data found in packet: " + packet.toXML());
+                        }
+                    }
+                }
+
+            } catch (Exception e) {
```
Fishbowler, line 116: What are you trying to defend against here? Could this be more explicit?

xmppserver/src/main/java/org/jivesoftware/openfire/handler/IQVersionHandler.java
mightymop: i will change it to debug output

xmppserver/src/main/java/org/jivesoftware/openfire/handler/IQVersionHandler.java
mightymop: oh long time ago:D i dont know exactly why .. but i think it was for development too... so i will remove it

Fishbowler:
Thanks. I've rebased, and taking a look now

Fishbowler:
The idea is that it'll populate this empty field, right?

pr1792

I've tested it cursorily using inVerse and haven't gotten anything to appear, but I'm not sure that it's got the capabilities to  necessarily trigger this, and doesn't have debug capabilities. Have you got a test client that will?

mightymop:
> 
> 
> The idea is that it'll populate this empty field, right?
> 
> pr1792
> 
> I've tested it cursorily using inVerse and haven't gotten anything to appear, but I'm not sure that it's got the capabilities to necessarily trigger this, and doesn't have debug capabilities. Have you got a test client that will?

you have to visit: session-details.jsp of a session...
session-details jsp

Fishbowler:
Took a little effort, but I got it working. Turns out it populates both!
That effort feels like an opportunity to get some better automated builds going on.

Fishbowler (approved):

xmppserver/src/main/java/org/jivesoftware/openfire/handler/IQVersionHandler.java
```diff
@@ -76,6 +83,34 @@ public IQ handleIQ(IQ packet) throws PacketException {
...
+                        {
+                            localSession.setSoftwareVersionData(element.getName(), element.getStringValue());
+                        }
+                        else
+                        {
+                            /*
+                              The result comes from a server 2 server connection, so we write the information
+                              only to the debug log, because we dont need it at this point.
+                            */
+                            Log.debug(packet.getFrom()+" "+element.getName()+"="+element.getStringValue());
```
akrherz, line 107: How expensive is this string concatenation? Would placeholders be better so to avoid doing this each time this code is hit, even when not in debug mode?

mightymop: I changed the logging to Log.debug("XEP-0092 Packet from={} {}={}",packet.getFrom(),element.getName(),element.getStringValue());

gjaekel:
> Would placeholders be better so to avoid doing this each time this code is hit, even when not in debug mode?

:+1: Such "unprotected" debug log statements might be rather expensive in "deep inner loops" or frequently used code paths. They don't only eat cpu cycles in a direct way, but -- on Java -- also produce senseless noise on the heap to be handled by the GC.

One may scan the whole codebase for such patterns to refactor.

Fishbowler (approved):

akrherz (approved):

PR #1796, opened 2021-01-16
OF-2195: Add Git SHA to Admin header

Fishbowler:
Add a subtle git sha to the footer so that we can observe the exact provenance of a given Admin interface

image

### DIFF
--- a/xmppserver/pom.xml
+++ b/xmppserver/pom.xml
@@ -57,6 +57,42 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>git-info</id>
+            <activation>
+                <file>
+                    <exists>${maven.multiModuleProjectDirectory}/.git</exists>
+                </file>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>pl.project13.maven</groupId>
+                        <artifactId>git-commit-id-plugin</artifactId>
+                        <version>4.0.3</version>
+                        <executions>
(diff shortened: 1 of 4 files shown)


guusdk:
Nitpicky: shouldn't this go with the version identifier, on the top-right?

Fishbowler:
Aw, man. I can't believe I didn't spot that. I actually looked for it.
Also, spotted that this breaks the Docker build. I'm working on another branch that's going to make improvements there, but will switch this to draft whilst I fix both of these things.

guusdk:
I like having an identification like this. My only concern: does it break the build if the source is not/no longer in a hot repository?

Fishbowler:
That's exactly what's happening in Docker, where it builds the source away
from the .git folder. Maybe this needs solving separate from Docker -
someone somewhere will hit this in a different context.
Good spot!

Fishbowler:
Build now appears with the version, and still succeeds when git isn't available, e.g. in Docker:

image

akrherz:
Thanks for doing this.  It is unclear to me what this change does for tagged/GA releases?  For a GA release, we would not want the git hash included, would we?

Fishbowler:
I think it's still useful. Better to know from a screenshot that it's definitely _that_ commit, and not someone else's build or someone's branched from the release or where they've accidentally got the POM in a wrong state. It's not foolproof, but it gives a good indication of providence.

PR #1801, opened 2021-01-30
OF-2190: Never lowercase the roomname when calling LocalMUCRoomManager#addRoom

Fishbowler:
Fixes https://igniterealtime.atlassian.net/browse/OF-2190

I've traced code paths and this appears to be the only one that requires changing.


Fishbowler:
Rebased

guusdk:
Thanks Dan!

PR #1803, opened 2021-02-05
denote Openfire 4.6.2 release

akrherz:


akrherz:
power cycling attempting to get github actions to awake from its slumber.

akrherz:
CI problem is that our workflow configs only specify master branch, will fix later today.

PR #1804, opened 2021-02-05
run CI on all branches

akrherz:
At Ignite Realtime, we care about all branches.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [master]
+    branches: "**"
   pull_request:
-    branches: [master]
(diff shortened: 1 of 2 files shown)


Fishbowler (commented):
I think it's just the pull request you care about here, isn't it? If someone wants to merge into a version branch? If it's for any push to any branch, that could be a lot of Actions activity on WIP.

Or perhaps it be selective, but maintain the list? All PRs, plus pushes to master, 4.5, 4.6, 4.7, etc.

akrherz:
I am thinking about forks and folks that push to their own branches for CI's opinion.  As was as ignite's branches.

Fishbowler:
True. This only runs on forks if you choose it, and we don't push features to the ignite repo. Good shout.

Fishbowler (approved):

PR #1806, opened 2021-02-12
OF-2200: Have alternating background colors when looking at group members

guusdk:


akrherz:
I have rebased this.

Fishbowler:
LGTM 👍 

Before:
before-of2200

After:
after-of2200

PR #1810, opened 2021-03-11
OF-2207: Ensure that the room exists on all nodes, before firing off …

guusdk:
…other room events to avoid race conditions

This commit turns the room creation event into a synchronous action. I'm hoping that the impact of this will be minimal, as creating a room should be a relatively uncommon occurrence.


gjaekel:
Hmm - room creation and destruction is a very common action while using OFmeet, isn't it?

guusdk:
Well, there are obviously use-cases where room creation happens more frequently than others, but in essence, a "room created" event should occur a _lot_ less frequent than events like "message sent in room" or "occupant joined room"

gjaekel:
I know what you mean, but also this _lot_ is a very small one, because an typical A/V meeting will also use -- if any -- just a very small number of text messages and also just a vew joins and leaves. :) :)

Let's stop here, because your goal is "to avoid race conditions" and the impact _should_ be low also in the OFmeet usecase!

akrherz (approved):

PR #1813, opened 2021-03-12
use python 3.9 for testing

akrherz:
Perhaps the flakiness we see in the aioxmpp tests are due to async bugs in python 3.6.  perhaps python 3.9 has less bugs like that :)


akrherz:
so puzzling, this continues to work in my fork, but not here.  I still think this is a good change to consider.

PR #1816, opened 2021-03-18
OF-2210: Restore SessionDestroyed event handling

guusdk:
This reverts the change for OF-2012, which erroneously assumed it was fixing a typo.

Marking a session as unavailable is different from removing a session. By removing the session in this method, it later can't be found, and event listeners are not triggered as a result.

This is where @GregDThomas owes me an "I told you so" - he challenged OF-2012 at the time. :)


GregDThomas:
:) Keep it in the back pocket for later.

akrherz (approved):

guusdk:
> :) Keep it in the back pocket for later.

Springing it on me while I least expect it, I bet? Well, I deserve this one. :)

PR #1817, opened 2021-03-21
Publish an official container image to Docker Hub

Fishbowler:
Adds to the existing CI build, and uses the existing build artifacts to create and publish a docker image to Docker Hub.

- Any tag pushed to the repo will be pushed as an equivalent tag to Docker Hub
- Anything pushed to master will be pushed to :latest on Docker Hub 

I've added some consideration for forks being able to push to Docker Hub. Anyone running the CI in their fork AND with the appropriate secrets set will attempt to use them to push an image.

To make it work:

- Create an account on Docker Hub
- Create a token for the user account (https://hub.docker.com/settings/security)
- Add a repository secret (https://github.com/igniterealtime/Openfire/settings/secrets/actions) for DOCKERHUB_USERNAME and DOCKERHUB_TOKEN - (do these want to live on the repo, in case we want to publish from a different repo in the future?)
- Add a repository secret called DOCKERHUB_OPENFIREIMAGE that contains the repo to push to (which might be assumed to be "dockerhub-username/openfire", but might not be, for example if it's a user pushing to an organisation) - this should look like "igniterealtime/openfire"

### DIFF
--- a/.github/dockerhub.env
+++ b/.github/dockerhub.env
@@ -1,2 +1,0 @@
-DOCKERHUB_USERNAME=fishbowler
-DOCKERHUB_OPENFIREIMAGE=fishbowler/openfire
--- a/.github/dockerhub.env
+++ b/.github/dockerhub.env
@@ -1,0 +1,2 @@
+DOCKERHUB_USERNAME=fishbowler
+DOCKERHUB_OPENFIREIMAGE=fishbowler/openfire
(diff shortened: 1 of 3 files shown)


GregDThomas:
This looks really useful. Is it worth adding the "to make it work" section to a README somewhere so it can be easily found without trawling through commit messages?

Fishbowler:
Good idea! 
I've also got a bigger problem - the Docker Hub images don't execute like the locally built ones do.
`start-stop-daemon: unable to start /usr/local/openfire/bin/openfire.sh (Permission denied)`

Fishbowler:
I wonder if there's a way to gently test an image

Fishbowler:
This looks ready for review 👍

akrherz (approved):

PR #1824, opened 2021-04-17
OF-2240: Add self-presence status to kick presence to "kickee"

Fishbowler:
Openfire has specific logic for detecting when self-presence needs to occur on a broadcast. Kick works by sending a specific presence to the kicked user, removing them from the MUC, then sending the broadcast - the logic is never invoked.

This change adds the self-presence to that kick.

I've used some yet-to-be-contributed additional tests for Smack to test that it both failed before the change and succeeded afterwards. See https://github.com/Fishbowler/Smack/commit/0e6963330804580a43f82b61f667fc40163189da

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -2854,7 +2854,10 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
 
                 // Send a defensive copy (to not leak a change to the 'to' address - this is possibly overprotective here,
                 // but we're erring on the side of caution) of the unavailable presence to the banned user.
-                kickedRole.send(kickPresence.createCopy());
+                Presence kickSelfPresence = kickPresence.createCopy();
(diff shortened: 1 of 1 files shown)


Fishbowler:
The kick self-presence test is into Smack, and will be in the next released version.

guusdk:
Does that code (under different circumstances perhaps) get executed with an element that already has a 'status' attribute. Is there a chance here that this ends up adding a second 'status' attribute to that element?

guusdk:
oh, this is very specific to the 'kick' scenario (I assumed it was in the more generic broadcast functionality). That makes my concern less likely to happen.

Fishbowler:
I found the issue whilst using Jitsi with Openfire, and saw the kicks failing there. A read of the code shows that whilst self-presence gets added in other situations, kick isn't one of them. The simple Smack test fails. Experientially, I've not found a situation where the kickee gets the self-presence, and I can only see the 110 being added [here](https://github.com/IgniteRealtime/Openfire/blob/9f024dbb235a2cc58fbdfdd6f79c7664d366382c/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java#L1558), on a code path not used by kick, but I wouldn't want to guarantee it could _never_ happen :D

PR #1825, opened 2021-04-18
Fix smack integration tests

Fishbowler:
Smack integration tests aren't building with latest gradle. This change:

* Forces the known-good Gradle version
* Re-enables a previously broken test
* Bumps the Smack version used for tests


guusdk:
Thanks!

PR #1826, opened 2021-04-19
use older ubuntu for aioxmpp tests (Epoll issue?)

akrherz:
Whilst pulling my hair out, I realized that Github Actions updated the `ubuntu-latest` about the time failures started to happen.  I then noticed python using the `epoll` reactor, which sometimes can be sensitive to the Linux distro being used.  Rolling back to `ubuntu-18.04` has worked twice for me, watch it fail here.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -54,7 +54,8 @@ jobs:
   aioxmpp:
 
     name: Execute aioxmpp-based CI tests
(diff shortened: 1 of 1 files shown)


guusdk:
@horazont This change seems to have fixed the issues that we were having with running aioxmpp in our CI. I'm not familiar enough with the technical details to be able to predict if this is going to be a problem anywhere other than in our CI environment, but I wanted you to be aware, just in case that doesn't turn out to be the case.

PR #1828, opened 2021-04-25
publish maven artifacts on any branch push

akrherz:


Fishbowler:
I know it's already the "standard pattern" here, but this means we probably don't want any maintainer pushing casual feature branches, as it'll overwrite the SNAPSHOT from master.

We could constrain it to branches that are master or that look like a version, if it's a real concern?

akrherz:
Thanks @Fishbowler, I have blissful ignorant on how to elegantly constrain branches other than listing them all out in the if statement.  Getting feature branch pushes here is rare, but it sometimes happens :)

gjaekel:
@Fishbowler I add a missing `{{ }}` in your code snipped above.

Fishbowler:
Thanks @gjaekel - the danger of coding straight into a comment box!

Fishbowler:
Still not quite right
https://github.com/igniterealtime/Openfire/pull/1828/checks?check_run_id=2610954090

> /home/runner/work/_temp/614a69f8-6b49-4c74-b0fe-ed21d3b43a1d.sh: line 1: [[refs/pull/1828/merge: No such file or directory
/home/runner/work/_temp/614a69f8-6b49-4c74-b0fe-ed21d3b43a1d.sh: line 1: refs/pull/1828/merge: No such file or directory

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -88,6 +88,20 @@ jobs:
...
 
+  check_branch:
+    runs-on: ubuntu-latest
+    outputs:
+      is_publishable_branch: ${{ steps.check-branch.outputs.is_publishable_branch }}
+    steps:
+      - name: check branch ${{ github.ref }} is either master or a version number
+        id: check-branch
+        run: |
+          if [[${{ github.ref }} == 'refs/heads/master' || ${{ github.ref }} =~ refs\/heads\/[0-9]+\.[0-9]+ ]]; then
```
Fishbowler, line 99: Double quotes around the `${{ github.ref }}`s?

gjaekel: If its BASH, then it need a space between `[[` and the expression.

gjaekel: And double quotes around will make it save against spaces in the path. But there may be probably other things that will break in this case :smile:

akrherz: Thanks, is this now resolved?

Fishbowler:
Merged. We won't know until after the merge whether it's definitely worked

akrherz:
> Merged. We won't know until after the merge whether it's definitely worked

Well, not until I just cherry-picked to the 4.6 branch :)

PR #1831, opened 2021-05-05
fix ClassCastExceptions

tenaciousRas:
fixes:

- https://discourse.igniterealtime.org/t/error-500-retrieving-rooms-info-with-rest-api-plugin/89851 and related :: cast enum to string by overriding Role#toString
- https://discourse.igniterealtime.org/t/hazelcast-pluginservlet-roomremovedevent-classcastexception/90088 - also happens when updating/saving a MUC created on local cluster node :: use collection serializer instead of stringlist -- comment/motivation about/for backwards compatibility w/4.6.0 removed

This resolves ClassCastExceptions being thrown in OF 4.6.x w/RestAPI plugin 1.5.0.
OF 4.6.4 should finally be compatible with RestAPI plugin. I suggest RestAPI plugin release v1.5.0 as STABLE.
As things currently stand, v1.4.0 is incompatible with OF v4.6.x.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -3283,7 +3283,7 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
         ExternalizableUtil.getInstance().writeSafeUTF(out, description);
         ExternalizableUtil.getInstance().writeBoolean(out, canOccupantsChangeSubject);
         ExternalizableUtil.getInstance().writeInt(out, maxUsers);
-        ExternalizableUtil.getInstance().writeStringList(out, rolesToBroadcastPresence.stream().map(Enum::name).collect(Collectors.toList())); // This uses stringlist for compatibility with Openfire 4.6.0. Can be replaced the next major release.
+        ExternalizableUtil.getInstance().writeSerializableCollection(out, rolesToBroadcastPresence);
         ExternalizableUtil.getInstance().writeBoolean(out, publicRoom);
         ExternalizableUtil.getInstance().writeBoolean(out, persistent);
         ExternalizableUtil.getInstance().writeBoolean(out, moderated);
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
```diff
@@ -3283,7 +3283,7 @@ public void writeExternal(ObjectOutput out) throws IOException {
         ExternalizableUtil.getInstance().writeSafeUTF(out, description);
         ExternalizableUtil.getInstance().writeBoolean(out, canOccupantsChangeSubject);
         ExternalizableUtil.getInstance().writeInt(out, maxUsers);
-        ExternalizableUtil.getInstance().writeStringList(out, rolesToBroadcastPresence.stream().map(Enum::name).collect(Collectors.toList())); // This uses stringlist for compatibility with Openfire 4.6.0. Can be replaced the next major release.
```
tenaciousRas, line 3286: I'm confused. The comment said we're using StringList (List<String>) for backwards compatibility with 4.6.0, but why does xmppserver.jar v4.6.3 need to be compatible with xmppserver.jar 4.6.0? I thought this JAR only needs to be compatible with clients of its API f.e. plugins.

mightymop:
There was a change in LocalMUCRoom class (OF-2168) 
The member
```
private List<String> rolesToBroadcastPresence
```
was changed to 
```
private List<MUCRole.Role> rolesToBroadcastPresence
```

I think this change was between 4.6.0 and 4.6.3 so newly compiled plugins which are using mucs and so on have to be compiled against 4.6.3 or 4.7.0-SNAPSHOT.

I had the same problem when compiling a custom plugin too. With depending on 4.7.0-SNAPSHOT i could resolve this.

tenaciousRas:
thanks @mightymop! We're testing this change to the RestAPI plugin - I added https://github.com/igniterealtime/openfire-restAPI-plugin/issues/56

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
guusdk: The reasoning here is that this would not cause issues in a cluster where one cluster node has this update, but another does not (yet - eg when updating nodes online).

guusdk:
I wonder if using the REST API plugin with the fix in https://github.com/igniterealtime/openfire-restAPI-plugin/pull/57 applied makes this PR superflicious.

guusdk:
I'm going to close this for now. I've been working with @tenaciousRas for a while now, and this hasn't come up again. We can re-open if needed.

PR #1833, opened 2021-05-10
OF-2245, OF-2246, OF-2247, OF-2248, OF-2249, OF-2250 and OF-2251: Various BOSH + backup packet delivery improvements

guusdk:
It is hard to reason over volatile booleans (or any volatile type, for that matter). With volatile booleans, a compare-and-set strictly speaking isn't atomic, but there's the 'happens-before' guarantee for volatile members, that might make things safe again. In all, I'm not sure if the volatile member tracking the 'isClosed' state of a HTTP session is sufficiently thread safe. Replacing it with AtomicBoolean will ensure it is, and will at the very least remove ambiguity.


gjaekel:
As long as the lines of code are not **very** frequently used -- as here -- safety is better than performance. In other cases, the code might be trimmed to behave in an opportunistic way, e.g. to "close double" in seldom cases and to catch the resulting exception which is caused by this race condition.

PR #1838, opened 2021-05-18
Increase sleep to 10 seconds awaiting socket to be open

akrherz:


Fishbowler (approved):
I've rerun these checks a few times and can't get them to fail. Seems like a very worthwhile change!

PR #1839, opened 2021-05-21
JMX support for thread pools

guusdk:
This exposes statistics of most of Openfire's thread pools via JMX (OF-2257). In the process of implementing this, I found a couple of thread pools that had hard-coded configuration. I made that configurable (OF-2258).


guusdk:
This could (also) go on the 4.6 branch.

PR #1841, opened 2021-05-25
OF-2254: Admin console to show if system property was set by user

guusdk:
Apart from distinguishing between a property value that is, or is not equal to the default value of the property, it is convenient to be able to tell if a property is equal to the default value (this helps to guard against property values for which the default is changed between Openfire releases). This commit adds a column that can be used to determine if the property value has been explicitly set (instead of it just being a default).

The commit also replaces the icon used to represent 'no' with a orange dash, to visually set it apart from the icon that is used by the 'delete' button.

The resulting admin console page looks like this. Note that the screenshot has two user-set property values (marked by a red box): one where the value that was set is equal to the default value, and one where the value is different.

image


guusdk:
Being able to tell if a property was explicitly set is a request that I got from the community. I am wondering to what extend the trade-off between "added value" and "added complexity in this page" is worth it. Is this trade-off acceptable? Can we express things differently?

GregDThomas:
a) I don't think the added complexity is an issue; this is an admin page, and arguably the most in-depth page we have (Perhaps a "Here be dragons" warning is required!)

b) However, you've chosen "set by user". Looking at your screen shot. the **_default_** "entry" for adminConsole.sessionTimeout is "null" - which is equivalent to 30 minutes. It looks to me like the value for adminConsole.sessionTimeout is set to 30 minutes. So whilst it's been set, the effective value as not actually changed. I am unsure as the the value of highlighting that. Would a "Changed from default" indicator be more appropriate (`!systemProperty.getValue().equals(systemProperty.getDefaultValue()`) ? (you'll have to use the `JiveGlobals.getProperty(key) != null` for the "legacy" properties as there is no clue as to what the default is).

c) No issue with the actual code changes.

guusdk:
> Would a "Changed from default" indicator be more appropriate (`!systemProperty.getValue().equals(systemProperty.getDefaultValue()`)

Isn't that what we have in the drop-down box that is above the 'default value' column (where you can select 'changed' and 'unchanged')?

What I'm adding in this commit intends to allow a user to see if a property value has been explicitly set, even if it is set to what is the default value for that property.

GregDThomas:
> What I'm adding in this commit intends to allow a user to see if a property value has been explicitly set, even if it is set to what is the default value for that property.

Ah, missed that. I struggle to see the exact need for that requirement, but I guess there is one!

guusdk:
> Ah, missed that. I struggle to see the exact need for that requirement, but I guess there is one!

It is a bit far-fetch: we ran in an issue after deploying a code change that modified the default value of a plugin.

akrherz:
I like this change, but wonder about the "Set by User" verbiage.  User sort of implies an openfire user and not necessarily an openfire admin changed the property.  Since an admin is reading the page and may think a user changed the property?

guusdk:
I am not at all comfortable with the wording. I was struggling a lot there. I'm very open to suggestions. They're best short, though, to not increase the width of the page to much.

GregDThomas:
Could it be done by overloading the existing "default" drop box, no extra width required?

- Unchanged (null)
- Unchanged (set)
- Changed
- Unknown

guusdk:
Leaving it open only to fuss about wording is somewhat counter-productive. Merging as-is for now. Happy to accept updates.

PR #1844, opened 2021-06-11
Various improvements for instances that have many MUC rooms

guusdk:
The commits in this PR should:
- increase performance when loading many MUC rooms from the database
- allow administrators to disable pre-loading all MUC rooms from the database at startup
- stop Openfire from ignoring rooms that are unloaded from memory
- improve the admin console

Please refer to the commit messages for more details.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
@@ -10,15 +10,25 @@ import org.jivesoftware.openfire.muc.MUCRoom;
 
 /**
  * this class supports the simple LocalMUCRoom management including remove,add and query.
+ *
+ * Note that this implementation provides a representation of rooms that are currently actively loaded in memory only.
+ * More rooms might exist in the database.
+ *
  * @author <a href="mailto:583424568@qq.com">wuchang</a>
  * 2016-1-14
  */
 public class LocalMUCRoomManager {
     private final Map<String, LocalMUCRoom> rooms = new ConcurrentHashMap<>();
-     
+
+    /**
+     * Returns the number of chat rooms that are currently actively loaded in memory.
+     *
+     * @return a chat room count.
+     */
     public int getNumberChatRooms(){
         return rooms.size();
     }
+
     public void addRoom(final String roomname, final LocalMUCRoom room){
         rooms.put(roomname, room);
         GroupEventDispatcher.addListener(room);
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -889,6 +889,8 @@ groupchat.service.properties.error_already_exists=A service with the specified n
 groupchat.service.properties.label_service_description=Group chat service description (optional):
 groupchat.service.properties.save=Save Properties
 groupchat.service.settings_affect=Changes you make here will affect the group chat service:
+groupchat.service.properties.label_service_preload=Pre-load MUC rooms in memory when initially starting the service
+groupchat.service.properties.label_service_preloaddays=Pre-load MUC rooms that have registered activity in the last ... days:
 groupchat.service.properties.label_service_cleanupdays=Remove empty MUC Rooms after ... days:
 groupchat.service.properties.label_service_muckeep=Disable MUC room cleanup for this service
 
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -889,10 +889,15 @@ groupchat.service.properties.error_already_exists=A service with the specified n
 groupchat.service.properties.label_service_description=Group chat service description (optional):
 groupchat.service.properties.save=Save Properties
 groupchat.service.settings_affect=Changes you make here will affect the group chat service:
+
+groupchat.service.properties.memory_management.legend=Memory Management
+groupchat.service.properties.memory_management_description=MUC rooms are persisted in the database, but can be loaded in \
+  memory to improve performance (when a room is being operated on that is not loaded in memory, it will be loaded in \
+  memory on demand). The settings below configure this behavior.
 groupchat.service.properties.label_service_preload=Pre-load MUC rooms in memory when initially starting the service
 groupchat.service.properties.label_service_preloaddays=Pre-load MUC rooms that have registered activity in the last ... days:
-groupchat.service.properties.label_service_cleanupdays=Remove empty MUC Rooms after ... days:
-groupchat.service.properties.label_service_muckeep=Disable MUC room cleanup for this service
+groupchat.service.properties.label_service_cleanupdays=Unload empty MUC rooms after ... days from memory:
+groupchat.service.properties.label_service_muckeep=Disable MUC room unloading for this service
 
 # Group Chat History Settings Page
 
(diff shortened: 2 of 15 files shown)


guusdk:
I'd like this to go in the 4.6 branch. To facilitate, I attempted to keep API changes to a minimum.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
```diff
@@ -10,15 +10,25 @@
 
 /**
  * this class supports the simple LocalMUCRoom management including remove,add and query.
+ *
+ * Note that this implementation provides a representation of rooms that are currently actively loaded in memory only.
+ * More rooms might exist in the database.
```
GregDThomas, line 15: Is it worth explaining why a room may be in the DB but not in memory?

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
```diff
@@ -10,15 +10,25 @@
...
+    /**
+     * Returns the number of chat rooms that are currently actively loaded in memory.
+     *
+     * @return a chat room count.
+     */
     public int getNumberChatRooms(){
         return rooms.size();
     }
+
     public void addRoom(final String roomname, final LocalMUCRoom room){
```
GregDThomas, line 32: Not suggesting you change this (backward compatibility), but can't help but think  that addRoom(final LocalMUCRoom room) would have been easier, with the name of room retrieved from the room itself :shrug:

GregDThomas:
Seems fine to me from what I can tell. One silly thing about a comment plus one observation only, neither of which are important.

akrherz:
> I attempted to keep API changes to a minimum.

As I reviewed this, I don't think I found any API breaks.  Is that actually the case?

GregDThomas:
> As I reviewed this, I don't think I found any API breaks. Is that actually the case?

I concur with that.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
guusdk: > > I attempted to keep API changes to a minimum.
> 
> As I reviewed this, I don't think I found any API breaks. Is that actually the case?

I think so. I added a couple of new methods, which, where possible, got a default implementation, so I think we're safe.

guusdk:
> > I attempted to keep API changes to a minimum.
> 
> As I reviewed this, I don't think I found any API breaks. Is that actually the case?

I think so. I added a couple of new methods, which, where possible, got a default implementation, so I think we're safe.

danieltet:
Thank you Guus for the heads up, we have tried these changes. We wanted to see if the rooms load faster, and therefore the server starts faster. 

Short answer: Yes these changes have improved our startup time from 15 minutes to 3 minutes - so very promising.

Our setup:
- Standalone dev server with prod copy of DB
- Server: r5.xlarge, with stock configuration, DB: RDS db.r5.large (aurora) - all in AWS
- JVM: 4 GB, OOTB, no special configs/changes, javac 1.8.0_241
- Rooms: 92713 - based on ROWS in ofMucRoom table (see side issue below for another observation here)
- Users: 270K total (not active)
- Users: 0 active, this was a dev server with no load

Normal/in the past, in all 4.x.x versions (since upgrade from 3.x) we have had a long delay in the startup of the server. The startup script would wait on this line for 15 to 30 minutes upon startup - not ideal as you can imagine.
org.jivesoftware.openfire.muc.spi.MUCPersistenceManager - Loading rooms for chat service conference

Even compiling the latest 4.7.0 from git it takes 16 min to start:
2021.06.15 04:59:03 DEBUG [main]: org.jivesoftware.openfire.muc.spi.MUCPersistenceManager - Loading rooms for chat service conference
...
2021.06.15 05:15:25 INFO  [main]: org.jivesoftware.openfire.XMPPServer - Openfire 4.7.0 Alpha [Jun 15, 2021 5:15:25 AM]
2021.06.15 05:15:25 DEBUG [pool-3-thread-1]: org.jivesoftware.openfire.container.PluginMonitor - Extracting plugin 'search'...
2021.06.15 05:15:25 DEBUG [pool-3-thread-1]: org.jivesoftware.openfire.container.PluginMonitor - Successfully extracted plugin 'search'.
2021.06.15 05:15:25 DEBUG [pool-3-thread-1]: org.jivesoftware.openfire.container.PluginManager - Loading plugin 'admin'...

With compiling of the latest 4.7.0 with these changes it takes 3 min to start:
2021.06.15 05:43:32 DEBUG [main]: org.jivesoftware.openfire.muc.spi.MUCPersistenceManager - Loading rooms for chat service conference
.....
2021.06.15 05:46:46 INFO  [main]: org.jivesoftware.openfire.XMPPServer - Openfire 4.7.0 Alpha [Jun 15, 2021 5:46:46 AM]
2021.06.15 05:46:46 DEBUG [pool-6-thread-1]: org.jivesoftware.openfire.container.PluginMonitor - Extracting plugin 'search'...
2021.06.15 05:46:46 DEBUG [pool-6-thread-1]: org.jivesoftware.openfire.container.PluginMonitor - Successfully extracted plugin 'search'.
2021.06.15 05:46:46 DEBUG [pool-6-thread-1]: org.jivesoftware.openfire.container.PluginManager - Loading plugin 'admin'...

In he first case we used these commands
git clone https://github.com/igniterealtime/Openfire.git of-4.7.0
cd of-4.7.0
mvn verify

To compile with PR chagnes, we used these commands:
git clone https://github.com/igniterealtime/Openfire.git of-4.7.0-1844
git fetch origin pull/1844/head:igniterealtime:master
git pull
git checkout igniterealtime/master
git diff 01286e7b90d457a3f34601cd6af982c7419176a7 e66882573ed83a1af9eb48317309f7cec522e705 --stat
and we saw (most of the files/changes listed in this PR)
cd of-4.7.0-1844
mvn verify

Side note 1:
- simple compile of 4.7.0 (and previous 4.x versions) ended up reporting only a fraction of all the rooms; in this case we got 40K, when we had 97K rows in ofMucRoom table. The fixes in 1844 fixed at least this file (muc-room-summary.jsp) so that we can now see all the rooms.

Hope this test helps. Thank you for these improvements. Let us know what else we can do to help, we would love to see these in the next release 4.6.x or 4.7.

DT.

guusdk:
Good to hear that you're seeing a nice improvement, Daniel. The fact that you're seeing 'more' rooms with the changes in this branch is caused by the fix for OF-2262 (without this fix, the admin panel would only show the rooms that were actively loaded in memory, not the ones that were only stored in the database).

The changes in this PR also introduce a new configuration option for the MUC service (you can set it in the admin console). This feature allows you to configure the amount of rooms that Openfire will load from the database into memory at startup. By reducing this number, you might be able to further speed up a server start. The drawback is that your server will be a bit more busy when users start interacting with rooms (as they need to be loaded on-demand at that point). Those instances will probably be smeared out over time better, which might make that less noticeable.

gjaekel:
Dear @danieltet,
> JVM: 4 GB, OOTB, no special configs/changes, javac 1.8.0_241
*  8u241 is an very old and outdated JVM version. Is this Oracle Java? May you consider to switch to AdoptOpenJDK (8u292)?
*  Even with 8u241, I strongly suggest to use the G1 garbage collector (`-XX:+UseG1GC`)

As a reference, I list the JVM option I use for the XMPP server at my installation. Of course, you have to scale the memory sizes for your usecase. And please also notice, that this service is nearly exclusive used for OpenMeeting (of Pàdé), just two of more than hundreds of user use text chat.
```
-XX:+UseG1GC -Xms64m -Xmx256m -XX:MaxMetaspaceSize=128M -XX:MaxDirectMemorySize=64M -XX:MaxGCPauseMillis=50 -XX:ConcGCThreads=5 -XX:+ParallelRefProcEnabled -XX:ParallelGCThreads=5 -XX:ActiveProcessorCount=20 -XX:+UseStringDeduplication
```

I wonder if you get additional speedup by tuning the JVM.

PR #1843, opened 2021-06-07
Check 'xmppSession' is not null

abe-ta:
Hi, I found XmppWebSocket#deliver has possibility of causing NullPointerException.
It is expensive to cause Exception. Downtime may occur when the connection with a large number of clients is disconnected.

I tested with a large number of clients, and LB health checks failed due to high CPU usage when many clients were disconnected unexpectedly.

Thanks.

PR #1846, opened 2021-06-17
OF-2265: Websocket delivery should not depend on XmppSession being present

guusdk:
Although normally, an XmppSession will be available, that won't be the case when, for example, sending an error about the XmppSession not being available.

Not doing packet accounting (and resetting an idle time) is preferable to not delivering the message due to a NullPointerException.


guusdk:
This is ment as an alternative to #1843. @abe-ta would you please review this?

abe-ta (approved):
LGTM

PR #1847, opened 2021-06-17
use older ubuntu for connectivity tests

akrherz:


guusdk:
Might as well reduce that sleep again.

PR #1848, opened 2021-06-17
Using unique values to prevent inter-test interference

guusdk:
By using different values for username and address, there's less of a chance of individual tests affecting the others (like what we've seen happen in Github Actions).


guusdk:
I'm hoping that this will prevent test failures like the one reported here: https://github.com/igniterealtime/Openfire/pull/1847/checks?check_run_id=2849974343#step:7:235

akrherz (approved):

PR #1851, opened 2021-06-18
Add warm-up connection to connectivity tests

guusdk:
For reasons beyond me, one connection (out of many) for the BOSH connectivity test seems to fail - but only on the infrastructure for Github Actions.

From the output, it appears to be the _first_ connection that fails. I suspect that this is triggered by a weird scheduling/timing/eventing thingy that is specific to that environment.

This commit adds an additional connection attempt, for which the outcome will be ignored. Hopefully by 'warming up' the system under test, we get more stable results.

### DIFF
--- a/runConnectivityIntegrationTests
+++ b/runConnectivityIntegrationTests
@@ -164,6 +164,18 @@ public class ConnectivityTest
                 .build();
         final int iterations = 50;
 
+        // For some reason, the first connection on a freshly started instance on CI seems to fail. This attempts to work
+        // around that problem by ignoring the first connection attempt.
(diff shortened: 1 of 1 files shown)


guusdk:
Earlier, we tried dropping the version of Ubuntu used for these tests, in order to work around the same problem. If that works, we might not need the change in this PR.

akrherz:
> Earlier, we tried dropping the version of Ubuntu used for these tests, in order to work around the same problem. If that works, we might not need the change in this PR.

Well, perhaps your change fixes the flapping we saw with newer ubuntu and someday, github actions will deprecate older ubuntu

PR #1852, opened 2021-06-18
Release 4.6.4

guusdk:


akrherz (changes requested):
Please remove the second SNAPSHOT commit as we currently build from bamboo against the branch and not tags.

akrherz (approved):

PR #1853, opened 2021-06-21
fix listpager on muc-room-summary.jsp when using multiple chatservices

mightymop:
Issue described in: https://discourse.igniterealtime.org/t/of-4-6-4-bug-in-adminconsole-when-paging-through-mucs/90362


guusdk:
@mightymop I've changed your fix to re-use more of the existing code, instead of adding new functionality. Can you please test if this works for you?

mightymop:
> 
> 
> @mightymop I've changed your fix to re-use more of the existing code, instead of adding new functionality. Can you please test if this works for you?

its working for me... thanks

PR #1854, opened 2021-06-21
fix numeration on muc-room-summary table

mightymop:
now starting with 1 on first line

Issue: the muc line number in table starts with 2, not with 1 on first line


guusdk:
Good catch! Thanks. I've added this as https://igniterealtime.atlassian.net/browse/OF-2277 in our issue tracker.

PR #1856, opened 2021-06-25
OF-2268: Prevent concurrency issue when starting JMX

guusdk:
This commit ensures that the JMX manager is not started more than once, preventing concurrency issues.


guusdk:
Candidate to be cherry-picked to the 4.6 branch, if we anticipate future released there.

PR #1858, opened 2021-07-01
Fix removing jars of a plugin after destroy.

annovanvliet:
A Plugin is should be 'hot' re-deployable, but this fails when jar files cannot be removed. An extra line is added to close the Plugin classloader to free the resources and jars used by the plugin.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/server/OutgoingSessionPromise.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/server/OutgoingSessionPromise.java
@@ -411,7 +411,12 @@ public class OutgoingSessionPromise {
             if (!packet.getTo().getDomain().equals(domainPair.getRemote())) {
                 throw new IllegalArgumentException("Cannot queue packet to intended recipient '" + packet.getTo() + "' in the outgoing session promise to domain " + domainPair + ". Remote domain does not match!");
             }
-            if (!packetQueue.offer(packet))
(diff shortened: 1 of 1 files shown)


guusdk:
Hi Anno, thanks for your contribution. Do you mind elaborating a bit more on how the issue that you're addressing with this PR is reproduced?

annovanvliet:
I added
https://discourse.igniterealtime.org/t/cannot-delete-or-refresh-or-update-plugin-because-of-jar-file-in-use/90414
to add some more info
.

Op do 1 jul. 2021 om 12:24 schreef Guus der Kinderen <
***@***.***>:

annovanvliet:
It is a proposal, it works when i add these line to the destroy of my
plugin, but i cannot see the consequence as the pluginclassloader is still
kept.

Op do 1 jul. 2021 om 12:31 schreef Anno van Vliet ***@***.***>:

GregDThomas:
I’d hazard a guess that you’re running on Windows with Java 9+? There’s a known bug with Jetty that prevents the correct resources being freed up. So the JVM keeps the JAR open so it can’t be deleted on Windows. There’s a work-around for Java 8 in the code base (look for a file in the org.apache package), but it doesn’t work with Java 9+

deleolajide:
I can confirm this is the current behaviour on my windows dev PC with java 15.0.2

guusdk:
I've been trying to rebase this PR, but I somehow failed. Apparently, that made Github close this PR. I don't quite understand why, to be honest.

In any case, I've re-applied the same change in a new branch/PR, here: https://github.com/igniterealtime/Openfire/pull/1938

PR #1863, opened 2021-07-17
OF-2273: Ensure that self-presence mods are made when joining anon room

guusdk:
When joining a room is configured to restrict presence broadcasts, the reflected 'self-presence' should still have all added annotations as it'd otherwise get.

This fixes a bug where people can't join such a room (as the client doesn't receive the proper self-presence stanza as a result of the join)


guusdk:
Candidate for 4.6 branch

PR #1867, opened 2021-07-22
fix wrong typo for overriden

willseeyou:


akrherz:
Thank you, would you mind fixing this in the other places too?

```
$ git grep overriden
documentation/pluggable-roster-support-guide.html:    version of this provider Openfire will use if not overriden. The DefaultRosterItemProvider uses the Openfire 
xmppserver/src/main/java/org/jivesoftware/admin/AdminConsole.java:            // See if global settings are overriden.
xmppserver/src/main/java/org/jivesoftware/openfire/ldap/LdapVCardProvider.java:     * fields that can be overriden are stored in the database.
xmppserver/src/main/java/org/jivesoftware/openfire/net/ClientStanzaHandler.java: * have their FROM attribute overriden to avoid spoofing.<p>
xmppserver/src/main/java/org/jivesoftware/openfire/net/ComponentStanzaHandler.java: * have their FROM attribute overriden to avoid spoofing.<p>
xmppserver/src/main/java/org/jivesoftware/openfire/privacy/PrivacyList.java:     * privacy lists can be overriden per session by setting an active privacy list.
xmppserver/src/main/java/org/jivesoftware/openfire/privacy/PrivacyList.java:     * privacy lists can be overriden per session by setting an active privacy list.
```

willseeyou:
@akrherz Have updated, please review again, thanks.

akrherz (approved):

PR #1869, opened 2021-07-23
fix wrong typo for connectoin,outoing,connecte and deliverying

willseeyou:
connectoin -> connection
outoing -> outgoing
connecte -> connect
deliverying -> delivering


guusdk:
Thank you!

PR #1872, opened 2021-08-24
Update LocalMUCRoom.java

dancosta93:
Added support to join as participant in moderated rooms even if user is not a member


dancosta93:
Not sure if this has been discussed in the past, but it was something I needed.

Is there any reason why the user must always start as a visitor if a room is moderated? IMO it makes sense that you can join as a participant and start chatting right away even if the room has moderation.

PR #1873, opened 2021-09-01
Added support to join as participant in moderated rooms even if user is not a member.

dancosta93:
Not sure if this has been discussed in the past, but it was something I needed.

Is there any reason why the user must always start as a visitor if a room is moderated? IMO it makes sense that you can join as a participant and start chatting right away even if the room has moderation.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoom.java
@@ -600,6 +600,10 @@ public class LocalMUCRoom implements MUCRoom, GroupEventListener {
             return MUCRole.Role.participant;
         }
         else {
+             // if true, default to participant even if the room is moderated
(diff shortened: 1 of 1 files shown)


akrherz:
I don't understand your usage case.  The point of moderated rooms is to prevent that from happening.

dancosta93:
Correct me if I'm wrong. 

But if the room is not moderated the moderator cannot kick/ban anyone (at least that's how it worked in my tests with openfire). And then when the room is moderated, every user joins as visitor which will be muted until the moderator allows.

What I'm trying to achieve is something in the middle, let the room be moderated (allowing all the functionalities for the moderator) but let users join already as participants so they can chat right away without moderator intervention.

akrherz:
I could be wrong, but I think that is what the `admin` room role is for.

dancosta93:
I get it, but is it really necessary that every user has to be manually accepted by the moderator? Figure a room that might have thousands of users joining, it's ok to let them start chatting right away but still have the ability to kick/ban users, therefore the room needs to be moderated.

My idea is not to overwrite any default behaviour, only to add a property that would allow this to happen.

dancosta93:
Will this be approved?

PR #1874, opened 2021-09-03
OF-1840: Change log file usage

guusdk:
This drops the separation of logged message over different log files by log level. Effectively, this removes all but the 'all.log' file, renaming that to Openfire log.

Additionally, this commit adds ANSI-based color to the log level as printed on each line. This provides a very helpful visual hint, drawing attention to the more important log lines.

This change introduces ANSI control characters to the log file. Native readers often can handle these (for example, 'less' has the -R option for this). It's good to remember that the log configuration can be easily modified by end users, by modifying the log4j2.xml configuration file with their own preferences (which can be used to undo this change).

Corresponding changes have been made to the admin console. As the default ANSI coloring as applied by Log4j2 works best on a dark background, the background of the displayed log is changed to be dark.


guusdk:
I am (very) open to suggestions to improve the UI.

guusdk:
On Windows, more work is possibly needed. See https://logging.apache.org/log4j/2.x/manual/layouts.html#enable-jansi

gjaekel:
> This change introduces ANSI control characters to the log file.

In common, the ANSI coloration of the log level is used for the console (stdout) stream of the logging but not for stored logfiles itself, which might be the source for technical processes like reporting. It would be more work to filter it out than to add it lateron if really required.
If you like the coloration, I suggest to add it on the fly by some simple search&replace of the Loglevel keywords at the GUI component. Or, maybe even better: Colorize the Background of the whole line.

gjaekel:
BTW: :+1: for the change. But for me, it comes to late ;) Because we rotate daily and keep and archive all logs, I changed the `log4j2.xml` at start of using _OpenFire_ at all. In those days, I also drop all other log channels except `all.log`.

guusdk:
My primary motivation is to have color in the file itself, that I mostly process with `less -SiR`. Including the same colors on the admin console is more of an afterthought for me, though.

I do see your point though. ANSI control characters in log files might introduce issues. Maybe we should use two copies (one with, and one without them) as a compromise?

akrherz:
Thanks for doing this Guus.  I am +0 on this change, but am doing none of the work, so alas.  Openfire's logging is always been confounding to me as there are so many messages that are more or less worthless to the admin.  The segregation of the log levels into different files at least helps separate some of the wheat from the chaff and prevent rapid turnover of files with important messages inside it.

GregDThomas:
FWIW, back in the day (!) when I used to look at Openfire logs I found the separate log files largely useless. Yes, the warnings are in `warn.log`, but often you needed to see the logs in info.log at the same time to determine exactly what was being warned about - or at the very least determine the context. So I relied entirely on `all.log`.

If it were me, I'd probably stick with the one file. I suspect (but have no evidence!) that most people consuming the log files by automated processes either already have their own logging config, or at the very least are capable of creating one. Perhaps suitable comments in the `log4j2.xml` showing how to disable the ANSI colouring?

guusdk:
I was under the impression that mostly everyone was using the log files in line with what Greg wrote. That certainly reflects my experience. A large motivator for this PR is to make that single log file a bit more useable, by using colors to quickly identify the more important messages in between the others.

The value of the separation by log level that Daryl describes surprises me a little. Although I don't recognize that myself, it's not to say that it therefore should be easily dismissed. Maybe admins of Openfire have a slightly different use-case for those log as compared to developers - or maybe there's just a personal preference that's perfectly acceptable to have?

There is little harm in having more log files - apart from them using a bit more disc space. If those additional tabs in the admin console are not as worthless as I assumed, why not keep them around? Maybe we can consider replacing them with a log-level-selector HTML input somehow, which might be a little less awkward to use.

akrherz:
> Maybe admins of Openfire have a slightly different use-case for those log as compared to developers

I think that is exactly my point.  The vast majority of Openfire users are admins and not developers.  For admins, they want actionable things to do.  "What is this log message telling me that I should do to fix it"  If it is just informational and there is nothing I can do to fix it, why would an admin want to see it?

For example
```
2021.09.05 21:16:34 org.jivesoftware.openfire.http.HttpBindServlet - Error sending packet to client.
org.jivesoftware.openfire.http.HttpConnectionClosedException: The http connection is no longer available to deliver content
        at org.jivesoftware.openfire.http.HttpConnection.deliverBody(HttpConnection.java:110) ~[xmppserver-4.6.4.jar:4.6.4]
        at org.jivesoftware.openfire.http.HttpSession.forwardRequest(HttpSession.java:570) ~[xmppserver-4.6.4.jar:4.6.4]
```

Error sending which packet to which client on which http connection?  Anyway, this is outside of what you are doing here with this PR.

>  log-level-selector HTML input somehow

That would be excellent and basically replicate previous functionality.

GregDThomas:
Nice!

guusdk:
I've reworked the original solution to remove a couple of oddities (mainly with regards to the new log-level filtering).

Also, an option to download the entire log file from the admin console was added (OF-2286).

My main concern is that the newly added ANSI control characters won't work on Windows. Would someone with access to a device that's running be so kind as to test the new functionality? If the log level (eg ERROR, WARN, INFO) on each log line as shown in the admin console is colored (like in the screenshot above), things should be fine.

PR #1875, opened 2021-09-08
- Added MUCDelegate getter/setter

dancosta93:
- Added MUCDelegate getter/setter to the interface instead of only in the implementation.


akrherz:
Thanks for this, but your commit is full of white space changes.  You want to separate those out to isolate your functional code changes.

PR #1876, opened 2021-09-23
feat: try to connect to parent xmpp domain 

mightymop:
If no dns entry was found for xmpp service subdomain, try to connect to xmpp domain host itself.

Example: 

If the dns entry for conference.xmppdomain, pubsub.xmppdomain, search.xmppdomain was not set the method will now try to connect to the (parent) xmppdomain, cause the service will mostly run on the same host as the subdomain service.


guusdk:
I'm thinking that this is behavior that's already in Openfire?

mightymop:
> I'm thinking that this is behavior that's already in Openfire?

Sadly not. For example:

xmpp domain 1 (mopsdom.de) has only dns entrys for the xmpp domain itself
xmpp domain 2 (mopbook) is my dev domain (debug)

i used gajim and connected with an account to mopbook to make some service discoverys on mopsdom.de and i ran into:

1. org.jivesoftware.openfire.net.SocketUtil.createSocketToXmppDomain which trys to get the host for mopsdom.de (thats working of cource)
2. when i wanted to scan conference.mopsdom.de it could not be resolved by dns (cause there is no entry) 
remoteHost.getHost(); returns "conference.mopsdom.de" to realHostname
so socket.connect( new InetSocketAddress... trys to open a connection to "conference.mopsdom.de" which of course will fail and then it jumps into catch block after timeout

i could reproduce this with other service subdomains (pubsub,...) too

so i inserted the fallback

mightymop:
rebased to current master...

PR #1886, opened 2021-10-14
OF-2296: Remove duplicate event listener registration

guusdk:
By registering MultiUserChatServiceImpl as a ClusterEventListener twice, it gets invoked more than once, which can cause problems.


evdherberg:
Good find!

PR #1891, opened 2021-10-19
OF-2293: Prevent duplicate MUC occupants from existing

guusdk:
This commit prevents an occupant to be added to a room that already has that same occupant. This has been observed in the wild (while testing new MUC/Clustering code).

The changes in this commit do not take away the cause of the duplicated addition, which are unknown at this time. The change does take away the direct event (preventing the duplicate from being added), and logs an error message that contains a stack trace. Hopefully that can be used in the future to find the cause of the issue.


guusdk:
Thanks! I will take your suggestions outside of the scope of this PR (as I think they're related, but not the same).

PR #1901, opened 2021-11-02
OF-2314 Kill s2s local connection if it already exists elsewhere in the cluster

evdherberg:
If multiple nodes have similar s2s connections before joining together in a cluster, the local s2s connections were allowed to keep existing on both nodes. That's actually not allowed, or at least not at this moment. This PR kills the s2s connection on the node that is considered to be 'joining' the cluster.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -1646,9 +1646,13 @@ public class RoutingTableImpl extends BasicModule implements RoutingTable, Clust
             if (route instanceof LocalOutgoingServerSession) {
                 // Terminating the connection should also trigger the OutgoingServerSessionListener#onConnectionClose in SessionManagerImpl.
                 // That will result in the s2s connection actually being removed from the LocalRoutingTable.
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
```diff
@@ -1621,7 +1621,36 @@ public void markedAsSeniorClusterMember() {
...
+                        } else {
+                            serversCache.put(address, server.getNodeID());
+                        }
+                    } finally {
+                        lock.unlock();
+                    }
+                })
+        );
+        for (DomainPair localServerRouteToRemove : localServerRoutesToRemove) {
+            final RoutableChannelHandler route = localRoutingTable.getRoute(localServerRouteToRemove);
```
guusdk, line 1645: Is it worth wrapping this in a `try/catch` to prevent closure of any one route to stop all remaining routes from being closed?

evdherberg: Yes.

evdherberg:
I rebased in the wrong order, leading to merge mess. New, clean PR is https://github.com/igniterealtime/Openfire/pull/1904. That's the one to merge.

PR #1903, opened 2021-11-02
Reflect aioxmpp project change: replace nose with pytest

guusdk:
In https://github.com/horazont/aioxmpp/commit/ccd00b7e377e65d884e12ce1f2167151612af9b8 'nose' was replaced by 'pytest' and 'pytest-cov'. As we're using the HEAD of their repository, we must have those packages available when running.

### DIFF
--- a/runAioxmppIntegrationTests
+++ b/runAioxmppIntegrationTests
@@ -89,10 +89,10 @@ function setUpAioxmppEnvironment {
     	git clone -b devel https://github.com/horazont/aioxmpp "${AIOXMPPDIR}"
     	pushd "${AIOXMPPDIR}"
 	fi
-    which pip3
-	pip3 install setuptools
-    pip3 install nose
-    pip3 install -e .
+	python3.9 -m pip install setuptools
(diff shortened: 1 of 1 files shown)


akrherz (approved):
things are green at least, but still unsure why more tests are being skipped than before

guusdk:
This change makes the build succeed again, but there is an unexpected inconsistency that I can't quite explain: In the last few successful builds prior to the migration to pytest, this was reported:

```
Ran 67 tests in 16.467s

OK (SKIP=4469)
```

After the migration to pytest and with the changes in this PR, we're getting:

```
======== 44 passed, 4468 skipped, 4 deselected, 874 warnings in 39.44s =========
```

Apart from us deselecting one more test than before (which maybe is caused by the `-k` option value not being specific enough), we also ran 23 (67-44) fewer tests. @horazont do you have any thought on what might be causing that?

guusdk:
I've merged this to at least get the builds go green again, with some aioxmpp coverage. If we can improve on that later, all the better.

PR #1909, opened 2021-11-05
OF-2322: Ensure outgoing queue is emptied before routing newer stanzas

guusdk:
The existing code is based around a route of last resort, the
OutgoingSessionPromise, which builds a helper - the PacketsProcessor - and
creates the new session. The PacketsProcessors operate from within a thread
pool, and thus have contended execution time.

As new packets arrives from onward transmission during the session setup, the
PacketsProcessor queues them for retransmission. However, the act of session
creation also creates the outgoing route, meaning that later packets will
bypass the queue, and be transmitted out of order.

This should give the impression of a randomly reordered set of stanzas, but
would in fact be two consecutive sequences of ordered stanzas which are
interleaved, hence if the two sequences are A-E, F-J, one might see:

A, B, F, G, C, H, I, D, J, E

The fix here is two ensure that all packets are queued until the queue itself
is empty.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/server/OutgoingSessionPromise.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/server/OutgoingSessionPromise.java
@@ -208,6 +208,19 @@ public class OutgoingSessionPromise implements RoutableChannelHandler {
         }
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


guusdk:
I believe that there's a good chance that this also fixes https://igniterealtime.atlassian.net/browse/OF-2321 as a side-effect.

dwd (approved):
Looks good to me (and nice and simple, too)

PR #1912, opened 2021-11-09
OF-2327: Remove 'dev mode'

guusdk:


GregDThomas:
Be aware that Ant is used for JSP compilation (I know!). This does means if a plugin depends on Openfire v.(without Ant), the plugin will have to explicitly add Ant as a compile dependency.

guusdk:
Woah, that'd be a dealbreaker to me. I thought that JSP compilation for plugins depended on a JSPC task that's defined as a Maven plugin execution - not doing it runtime?

GregDThomas:
I might be wrong, but at one time the JspC class had a dependency on Ant. A quick google turns up https://tomcat.apache.org/tomcat-8.0-doc/api/org/apache/jasper/JspC.html which shows

`public class JspC extends org.apache.tools.ant.Task`

More recent versions of Tomcat may be less ... surprising.

GregDThomas:
(but yes, would be a compile time dependency, not runtime - but would need to be explicitly added)

Fishbowler:
Changeset looks good. Running demoboot shows no problems. I've installed the Monitoring plugin to add additional pages, and it all looks good, and the logs look clean.

LGTM :+1:

Fishbowler:
The CodeQL alert isn't new. It's just complaining about you touching code near an old one.

guusdk:
Greg's concern is something we need to look at. Might be a reason to revert this. Let's try building a plugin that depends on an Openfire artifact that includes this change.

GregDThomas:
You probably just need to add Ant as an explicit dependency if it is an issue.

Fishbowler:
I think we don't use Apache Jasper and use the Mortbay one that's packaged with Jetty, that doesn't include any of the Tomcat bits.

> Let's try building a plugin that depends on an Openfire artifact that includes this change.

I've just rebuilt Hazelcast that depends on 4.7.0 and will have used my latest build. That seems to have all the right JSP in it.

GregDThomas:
Glad to hear. It was a long time since I hit this issue, so glad it’s no longer present.

PR #1913, opened 2021-11-09
Update dependencies

Fishbowler:
Fixes a bunch of CVEs.
Doesn't touch Mina.
Suppresses one false positive from JDOM2, which is included by Rome for RSS, but isn't used in a way that could expose the problem.


Fishbowler:
There's one more to resolve that I'll save for later - Jetty should go to 9.4.44 rather than 9.4.43, but the jasper JSP compiler won't compile our pages there :thinking:

guusdk:
Note that I'm trying to take out Ant completely in https://github.com/igniterealtime/Openfire/pull/1912/

Fishbowler:
We'll merge #1912 first, then deal with the rebase :+1:

Fishbowler:
^ Done this, awaiting CI

PR #1915, opened 2021-11-10
OF-2329: Fix XML parser behavior when detecting the name of the start tag

guusdk:
The name of the start tag is possibly terminated by any newline character (eg: a newline character), not only the space character.


Fishbowler (approved):

Fishbowler:
I've played with the original stanza, tabs, leading and trailing whitespace, newlines. All looking healthy :+1:

PR #1916, opened 2021-11-10
OF-2328: Implement optional Lock functionality for local locks

guusdk:
The clustered locks (based on Hazelcast) have some methods implemented that are not available in the non-clustered code.

The changes that have been applied to MUC's clustering implementation (OF-2220 for example) has introduced usage of these optional methods. This results in failing functionality when running Openfire in a non-clustered setup.

This commit adds an implementation for these optional methods.


Fishbowler:
Set the service to check every 2 minutes (rather than the default of 8) and created a non-persistent room from Spark, sent a message then idled. Logs show successful locks every 2 minutes. :+1:

Fishbowler (approved):

PR #1920, opened 2021-11-16
OF-2158 Update notifications should include cluster node reference

DJDANNY123:
Updates the updates notifications to include the hostname of the node which have plugins to be updated.

Tidies up an iterator in the same method

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/update/UpdateManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/update/UpdateManager.java
@@ -598,9 +598,9 @@ public class UpdateManager extends BasicModule {
 
         // Parse response and keep info as AvailablePlugin objects
         Element xmlResponse = SAXReaderUtil.readRootElement(response);
-        Iterator plugins = xmlResponse.elementIterator("plugin");
+        Iterator<Element> plugins = xmlResponse.elementIterator("plugin");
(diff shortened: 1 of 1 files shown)


Fishbowler:
Tested using https://github.com/surevine/openfire-docker-compose

image

Works on my machine :+1: 

Does the wording make sense in the context of a single server? i.e. Does "node" make sense outside of a cluster?

akrherz:
> Does "node" make sense outside of a cluster?

I see your point, but think is isn't too confusing.  A "node" is the place of installation, in my mind.

Fishbowler:
Took me a while, but got it tested on a standalone server. Tricky to get the UpdateManager to fire after the client connects. 
Future Us: Delete `update.lastCheck`, then stop/start the update service (Server Settings) then wait 30s.

akrherz:
> Future Us: Delete `update.lastCheck`, then stop/start the update service (Server Settings) then wait 30s.

I thought http://localhost:9090/available-plugins.jsp, doom scroll to the bottom and then click 'Update Now' ?

Fishbowler:
Aw man. That's sooo much easier :facepalm:

guusdk:
> I see your point, but think is isn't too confusing. A "node" is the place of installation, in my mind.

It's not to confusing, but why add even a little? A simple if-statement to check if clustering is enabled could be used to conditionally add the hostname.

PR #1922, opened 2021-11-16
OF-2338: fix npe on muc search if name is empty

mightymop:
fix a NPE when name = null in IQMUCSearchHandler

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/IQMUCSearchHandler.java
@@ -159,7 +159,7 @@ public class IQMUCSearchHandler
             }
         }
 
-        final FormField matchFF = df.getField("nameIsExactMatch");
+        final FormField matchFF = df.getField("name_is_exact_match");
         if (matchFF != null)
         {
             final String b = matchFF.getFirstValue();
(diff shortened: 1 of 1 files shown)


guusdk:
Ah, good catch, thanks.

I'd prefer if you fixed this by preventing the name to be added to the `names` collection when parsing the data form fields. That way, that collection will never hold `null` values.

I've raised https://igniterealtime.atlassian.net/browse/OF-2338 to track this issue. Can you please add a reference to that issue in your commit message? That helps the automated processes to correlate issue and commit.

Would you mind commenting on that issue with an example of the stack trace? I'm also interested in knowing what client has this behavior. How did you find the bug?

mightymop:
> Ah, good catch, thanks.
> 
> I'd prefer if you fixed this by preventing the name to be added to the `names` collection when parsing the data form fields. That way, that collection will never hold `null` values.
> 
> I've raised https://igniterealtime.atlassian.net/browse/OF-2338 to track this issue. Can you please add a reference to that issue in your commit message? That helps the automated processes to correlate issue and commit.
> 
> Would you mind commenting on that issue with an example of the stack trace?
> I'm also interested in knowing what client has this behavior. How did you find the bug?

I used Gajim and opened the service discovery dialog.
There i selected the muc service and because the service is publishing the jabber:iq:search feature the search button was enabled. A click on it opend the dynamic form to search for muc rooms. If i dont put any values into the search fields the issue occurs.

PR #1923, opened 2021-11-17
OF-2336: Return removed value when adding oversided new value

guusdk:
The javadoc contract defines that the removed value is to be returned.


guusdk:
@GregDThomas I'm not missing something obvious here, am I?

PR #1924, opened 2021-11-17
OF-264 Add group selection field in Create New User page

DJDANNY123:
Updates the create user page to include an option for assigning a group directly from that first page.

This implementation uses a text field expecting the exact name of a group, could potentially be improved with a dropdown list of available groups.


guusdk:
This look fine, and works as intended. @DJDANNY123 can you suppress the 'group' field when there are no groups in the system? That reduces complexity from (and thus questions from) unexpecting users.

PR #1926, opened 2021-11-18
OF-2239: Introduce new cache to be used by plugins

guusdk:
When plugins use the standard caches (in a cluster), ClassCastExceptions will occur after a plugin that has seen usage is reloaded/updated.

To prevent these exceptions, a new cache is introduced by this commit that stores data without a reference to the classes that define the data.


akrherz:
Thanks @guusdk.  I gather as an implication of this is that all our plugins will need updated to take advantage of this new code?

guusdk:
Plugins that use a cache, yes.

GregDThomas (approved):
LGTM

PR #1927, opened 2021-11-19
OF-2165 handle nickname clash

evdherberg:
When nodes come together to form a new cluster, room occupants with the same nickname cannot be allowed to remain in the room. This PR detects the situation, kicks all users with the offending nickname from the room, and informs all nodes in the cluster about the matter.


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
```diff
@@ -285,11 +300,67 @@ void restoreCacheContentAfterJoin(@Nullable final Set<OccupantManager.Occupant>
...
+
+                            // OF-2165
+                            // Check if the nickname of this occupant already existed for another user in the room.
+                            // If it did, we need to kick the users out. With sincere apologies.
+                            String nickBeingAddedToRoom = localOccupantRole.getNickname();
+                            boolean occupantWasKicked = false;
+                            try {
+                                final List<MUCRole> existingOccupantsWithSameNick = roomInCluster.getOccupantsByNickname(nickBeingAddedToRoom);
+                                final List<JID> otherUsersWithSameNick = existingOccupantsWithSameNick.stream().map(MUCRole::getUserAddress).filter(bareJid -> !bareJid.equals(localOccupantRole.getUserAddress())).collect(Collectors.toList());
+                                Log.debug("Current cluster room occupants: {}", roomInCluster.getOccupants());
```
guusdk, line 319: There's a bit of surplus debug logging here that should probably be removed.

xmppserver/src/main/java/org/jivesoftware/util/cache/ConsistencyChecks.java
```diff
@@ -720,8 +720,7 @@
             .collect(Collectors.toList());
         final List<String> allOccupantsJids = allOccupants
             .stream()
-            .map(OccupantManager.Occupant::getRealJID)
-            .map(JID::toFullJID)
+            .map(occupant -> occupant.getRealJID().toFullJID() + "/" + occupant.getNickname())
```
guusdk, line 723: This is weird to me. It combines a 'real jid' of an occupant, which is suggested to be a full JID, and adds a nickname to that.

So, in an example, giving a user's real JID `john@example.org/mobilephone` that is in a chatroom `myroom@conference.example.org` using nickname `Johnny` (this would make their "room JID" `myroom@conference.example.org/Johnny`) this value would be `john@example.org/mobilephone/Johnny?

If this is the intended end result (and to be used only to distinguish between entries), then this is OK. I would still suggest to use another way to concatenate `real JID` and `nickname` together, to make that more obvious. Something like `john@example.org/mobilephone with nickname: Johnny` for example.

My concern is that the Full JID in reality turns out to be a bare JID, which is then "completed" with the nickname. That would result in a full JID that mixes parts of the real JID and the room JID. That would be a JID that is not usable to route messages.

xmppserver/src/main/java/org/jivesoftware/util/cache/ConsistencyChecks.java
```diff
@@ -731,8 +730,7 @@
             .collect(Collectors.toList());
         final List<String> allMucRolesOccupantsJids = allMucRoles
             .stream()
-            .map(MUCRole::getUserAddress)
-            .map(JID::toFullJID)
+            .map(mucRole -> mucRole.getUserAddress().toFullJID() + "/" + mucRole.getNickname())
```
guusdk, line 733: Same concern as my earlier comment.

xmppserver/src/main/java/org/jivesoftware/util/cache/ConsistencyChecks.java
evdherberg: Valid concern. Intention was not to create a new JID, but to clarify why the same occupant JID could be in the list of occupants multiple times. I now improved the code to not suggest constructing some new JID.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
evdherberg: Missed that one. Removed now.

xmppserver/src/main/java/org/jivesoftware/util/cache/ConsistencyChecks.java
evdherberg: Same response.

guusdk:
Thanks!

PR #1928, opened 2021-11-22
OF-2199 Remove JRE Bundling with artifacts

akrherz:
Note: we need to address [INFRA-44](https://igniterealtime.atlassian.net/browse/INFRA-44), to get install4j updated on ignite's bamboo instance.


akrherz:
Ignite's bamboo instance should be ready to go for this change now.

guusdk:
No-one voiced a concern so far. Let's merge and see how far this gets us.

PR #1929, opened 2021-11-23
OF-2340: send stanzas that are 'responses' through PacketRouter, not RoutingTable

guusdk:
In various places in the code, Openfire generates a response to a stanza, to be delivered back to the sender of the original stanza. Typically, this involves errors, such as delivery errors.

Some of the existing code uses the RoutingTable to deliver these stanzas. This works, but bypasses a lot of checks and flags (as well as the interceptors). For "new" stanzas, it's more appropriate to use PacketRouter, as this will more closely resemble the normal stanza processing.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -551,6 +551,9 @@ public class MultiUserChatServiceImpl implements Component, MultiUserChatService
      */
     private void sendErrorPacket( Packet packet, PacketError.Condition error, String message )
     {
+        if (packet.getError() != null) {
+            Log.debug("Avoid generating an error in response to a stanza that itself has an error (to avoid the chance of entering an endless back-and-forth of exchanging errors). Suppress sending an {} error (with message '{}') in response to: {}", error, message, packet);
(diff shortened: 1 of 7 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -551,6 +551,9 @@ else if (this.iqHandlers != null) {
      */
     private void sendErrorPacket( Packet packet, PacketError.Condition error, String message )
     {
+        if (packet.getError() != null) {
+            Log.debug("Avoid generating an error in response to a stanza that itself has an error (to avoid the chance of entering an endless back-and-forth of exchanging errors). Suppress sending an {} error (with message '{}') in response to: {}", error, message, packet);
+        }
```
Fishbowler, line 556: Does this need a return statement to avoid sending the packet?

Fishbowler, line 556: Does this need a return statement to avoid sending the packet?

guusdk: Yes, thank you.

guusdk:
Rebased

PR #1931, opened 2021-11-29
Add embedded database feature to autosetup.

TorbenHaug:


TorbenHaug:
Hi,

I need the embedded database in the autosetup and I would be happy when it becomes part of the main project, so I do not need have an own fork for it.

Best regards 

Torben

TorbenHaug:
Found out, that it is already working, without these changes, but I do not know why.
And I didn't fond anything in the documentation

guusdk:
It "works", because the embedded database does not require any setup. If you configure Openfire to use the embedded database, by setting

```
  <connectionProvider>
    <className>org.jivesoftware.database.EmbeddedConnectionProvider</className>
  </connectionProvider>
```

Then there is nothing else that needs to be setup by an auto-setup process. Your change was effectively trying to do through "auto-setup" what is already done in the regular (non-auto-setup) configuration: set the connectionProvider to use the embedded variant.

GregDThomas:
Though if the documentation can be improved, I’m sure a PR on just that would be appreciated.

PR #1933, opened 2021-11-30
OF-2343: Consistently sending leave presence to occupant that's leaving

guusdk:


guusdk:
I'd love for someone to review not only the code change itself, but also the reasoning behind it.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -1196,18 +1196,17 @@ public void addOccupantRole(@Nonnull final MUCRole role)
             }
             item.addAttribute("role", "none");
 
-            // Check to see if the user's original role is one we should broadcast a leave packet for.
-            if(!canBroadcastPresence(leaveRole.getRole())){
+            // Check to see if the user's original role is one we should broadcast a leave packet for,
+            // or if the leaving role is using multi-session nick (in which case _only_ the leaving client should be informed).
+            if(!canBroadcastPresence(leaveRole.getRole()) || getOccupantsByNickname(leaveRole.getNickname()).size() > 1){
```
Fishbowler, line 1201: Is the first check necessary? The implementation of `broadcastPresence` has an identical check and logic within it.

Fishbowler:
Beyond the one query above, which is still healthy, LGTM :+1:

Fishbowler (approved):

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
guusdk: Interestingly, that duplication is exactly what started me looking into this. I think removing it would be good, but I first wanted to make sure that any change in the order in which checks were performed would not affect functionality. That's how I ended up with this PR. I'll likely follow-up with another PR to take care of the duplication with regards to `canBroadcastPresence`.

PR #1935, opened 2021-12-01
Changes to how component address matching is performed.

guusdk:


Fishbowler:
LGTM, although I haven't checked all the places this could affect. The few I have checked would've experienced this bug.

Fishbowler (approved):

PR #1944, opened 2021-12-14
Upgrade log4j to 2.16.0 for CVE-2021-4428

aldexis:
Upgrading log4j to 2.15.0 was unfortunately not enough per https://issues.apache.org/jira/browse/LOG4J2-3221

This should further address the CVE-2021-44228 "LogJam" vulnerability.


guusdk:
Unsure why this was closed. I generally think upgrading to 2.16.0 is a good idea. I've created a similar change in https://github.com/igniterealtime/Openfire/pull/1946

gjaekel:
This is a **very** good idea, indeed! ;)

PR #1949, opened 2021-12-16
OF-2354: Allow whitespace around comma-separated string values for list

guusdk:
When reading a comma-separated string to be parsed as a list of strings, whitespace surrounding the comma's should be ignored.

eg: "user@example.org, user@example.com" should be parsed as two values that contain no whitespace.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/SystemProperty.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/SystemProperty.java
@@ -119,7 +119,7 @@ public final class SystemProperty<T> {
         if (StringUtils.isEmpty(value)) {
             return Stream.empty();
         }
-        final List<String> strings = Arrays.asList(value.split(","));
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/util/SystemProperty.java
```diff
@@ -119,7 +119,7 @@
         if (StringUtils.isEmpty(value)) {
             return Stream.empty();
         }
-        final List<String> strings = Arrays.asList(value.split(","));
+        final List<String> strings = Arrays.asList(value.trim().split("[\\s,]+"));
```
akrherz, line 122: I am curious what happens with a string like so "1 , 2 ,  3" ?  Could that be added as a test?

guusdk: Force-pushed an update that includes this in a new unit test called `willCreateAListOfCommaWhitespaceSeparatedString2`. The test passes.

akrherz (approved):

PR #1950, opened 2021-12-16
Denote Openfire 4.6.6 Release

akrherz:


xmppserver/changelog.html
```diff
@@ -162,6 +162,13 @@ <h1>Openfire Changelog</h1>
     
     <div id="pageBody">
 
+<h2>4.6.6 -- <span style="font-weight: normal;">December 10, 2021</span></h2>
+<h3>Bug</h3>
+<ul>
+        <li><a href="https://igniterealtime.atlassian.net/browse/OF-2353">[OF-2353]</a> - Update Log4j dependency to 2.16.0 to address Log4Shell/LogJam.</li>
+</ul>
+
+
```
guusdk, line 171: Do we also want to include an entry for the release of 4.5.5 that'll happen today? The entry is a duplicate, with exception of the version number.

akrherz: I think being explicit is always better than implicit.

xmppserver/changelog.html
```diff
@@ -162,6 +162,13 @@ <h1>Openfire Changelog</h1>
     
     <div id="pageBody">
 
+<h2>4.6.6 -- <span style="font-weight: normal;">December 10, 2021</span></h2>
```
guusdk, line 165: Today is December 16th, not 10.

akrherz: yikes, fix ahoy

xmppserver/changelog.html
akrherz: yes, adding too.

PR #1951, opened 2021-12-16
Denote Openfire 4.5.5 release

guusdk:


akrherz (changes requested):
What about `plugins/openfire-plugin-assembly-descriptor/pom.xml` ?

guusdk:
That's why we don't let me push important buttons. Updated.

akrherz (approved):

PR #1958, opened 2021-12-19
Update MINA to 2.1.5

Neustradamus:
To fix https://nvd.nist.gov/vuln/detail/CVE-2021-41973.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,7 +120,7 @@
         <!-- Note; the following jetty.version should be identical to the jetty.version in plugins/pom.xml -->
         <jetty.version>9.4.43.v20210629</jetty.version>
         <standard-taglib.version>1.2.5</standard-taglib.version>
-        <mina.version>2.1.3</mina.version>
+        <mina.version>2.1.5</mina.version>
         <bouncycastle.version>1.68</bouncycastle.version>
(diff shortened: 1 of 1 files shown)


guusdk:
This CVE affects MINA's HTTP listener, which Openfire does not use. Openfire is therefor not affected by this problem.

In the past, we've had major issues caused by MINA upgrades. This makes me doubt if we should perform this upgrade.

Thoughts?

Neustradamus:
@guusdk: It will be better to update MINA, impacted CVE or not.

Neustradamus:
@guusdk: It will be better to update MINA, impacted CVE or not.
It is only an update in the same branch -> 2.1.x:
- https://mina.apache.org/mina-project/
- https://mina.apache.org/downloads-mina_2_1.html

guusdk:
"I will be better" is not a proper motivation. Please provide a reason for this update to be performed, that outweighs our experience that MINA upgrades can be very painful. We need to know if it's worth taking that risk.

Neustradamus:
@guusdk: Please merge it, it is easy to do it.

guusdk:
As I explained before: unless there is a good reason to do so, I do not think upgrading MINA is a good idea.

PR #1962, opened 2021-12-20
OF-2319

DJDANNY123:
Introduces system property xmpp.muc.allowpm.blockall to toggle whether all packets should be blocked, or just private messages.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
@@ -1402,23 +1402,31 @@ public class MUCRoom implements GroupEventListener, Externalizable, Result, Cach
      * conversation between room occupants or IQ packets when an occupant wants to send IQ packets
      * to other room occupants.
      *
+     * If the system property xmpp.muc.allowpm.blockall is set to true, this will block all private packets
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -1402,23 +1402,31 @@ public void sendPublicMessage(Message message, MUCRole senderRole) throws Forbid
...
-                throw new ForbiddenException();
-            default:
-            case visitor:
-                if (canSendPrivateMessage().equals( "participants" )) throw new ForbiddenException();
-            case participant:
-                if (canSendPrivateMessage().equals( "moderators" )) throw new ForbiddenException();
-            case moderator:
-                if (canSendPrivateMessage().equals( "none" )) throw new ForbiddenException();
+
+        if (packet instanceof Message || JiveProperties.getInstance().getBooleanProperty("xmpp.muc.allowpm.blockall", false)){
```
guusdk, line 1415: `JiveProperties` shouldn't be used anymore (we aught to deprecate that, I think). Instead, try using a `SystemProperty`. That's the new implementation for properties. The i18n line that you added will be picked up by that (and won't be in the API that you're now using).

guusdk (changes requested):

PR #1963, opened 2021-12-21
adds lgtm alert [java/zipslip] to ignore false positive

DJDANNY123:
for this line to be ignored from the dependency scan as a false positive


DJDANNY123:
I'm not interested in this alert
If you want to remove only this alert (and not other similar alerts), you can add the comment lgtm [java/zipslip] to the end of line 482. If you make changes to the suppression of alerts in a pull request, please note that theses changes will only be effective once that pull request is merged. See Alert suppression for more details.

guusdk:
This is indeed a false positive. Thanks!

PR #1964, opened 2021-12-22
Replaces the usages of synchronized(string) with synchronized(lock object) in the GroupManager class

DJDANNY123:
Replaces the usages of synchronized(string) with synchronized(lock object) in the GroupManager class.

No tidy method of replacing the strings that were previously used (e.g. GROUP_COUNT_KEY) as these are referred to from some static contexts and the values are needed so synchronizing on an instance of a "string lock" object isn't a great strategy.


guusdk (approved):
Looks good, thanks! Would you mind creating a JIRA issue for this change (doesn't need to be overly descriptive) and re-word the commit message to contain a reference to the issue number? We use that to generate changelogs.

PR #1965, opened 2021-12-22
JDBCAdminProvider replaces sync with synchronizing on an object lock

DJDANNY123:
Simple change to switch from locking on the string SQL statement to using a dedicated lock object in the class.


guusdk:
Thanks! Is this further improved upon by dropping that mutex object completely, and simply making the entire method synchronized?

Also, please create a JIRA issue and reword the commit message to include the JIRA identifier. That helps us track things in changelogs.

guusdk:
Thanks!

PR #1966, opened 2021-12-22
OF-2360 Move plugin / OF version retrieval to ClusterManager

evdherberg:
Logic for clusterwide fetching of plugin versions was previously located in a jsp. This PR moves that logic to the cluster manager, so it becomes available for reuse.

Furthermore a new method is introduced that detects version discrepancies of plugins between the local and other nodes. This will be useful for (at least) improving the monitoring plugin.

### DIFF
--- a/xmppserver/changelog.html
+++ b/xmppserver/changelog.html
@@ -162,6 +162,13 @@ hr {
     
     <div id="pageBody">
 
+<h2>(Next version) <span style="font-weight: normal;">(release date)</span></h2>
+
+<h2>Improvement</h2>
+<ul>
+    <li>[<a href="https://igniterealtime.atlassian.net/browse/OF-2360">OF-2360</a>] - Plugin version information accessible through cluster manager</li>
+</ul>
+
 <h2>4.7.0 Beta -- <span style="font-weight: normal;">December 3, 2021</span></h2>
 
 <h2>Bug</h2>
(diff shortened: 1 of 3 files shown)


xmppserver/changelog.html
```diff
@@ -162,6 +162,13 @@ <h1>Openfire Changelog</h1>
     
     <div id="pageBody">
 
+<h2>(Next version) <span style="font-weight: normal;">(release date)</span></h2>
+
+<h2>Improvement</h2>
+<ul>
+    <li>[<a href="https://igniterealtime.atlassian.net/browse/OF-2360">OF-2360</a>] - Plugin version information accessible through cluster manager</li>
+</ul>
+
```
guusdk, line 171: Unlike with plugins, we typically do _not_ update the changelog for Openfire when changes are made. Instead, we'll generate the entire changelog in one go when a release is made.

evdherberg: 👍

guusdk:
I tested only after merging. All admin consoles report the same plugin version on all nodes, even though I installed different versions.

evdherberg:
> I tested only after merging. All admin consoles report the same plugin version on all nodes, even though I installed different versions.

Fixed on the branch now. I guess that needs to go with a new PR?

guusdk:
Yes please.

evdherberg:
New PR at https://github.com/igniterealtime/Openfire/pull/1968

PR #1967, opened 2021-12-22
OF-2363 Assorted improvements to javascript files

DJDANNY123:
Updates assorted javascript files under webapp with sonarqube and LGTM alerts, some emphasis replacing unnecessary global variables with the correct assignment (e.g. let/const) to correctly scope.


guusdk:
Most of these are changes to third-party code - but then again, it's in third-party code that we last updated in ~2006 ...

PR #1969, opened 2021-12-23
OF-2360: Also show plugins not installed locally

guusdk:
Prior to this change, plugins that were not installed locally (only on other nodes) were not included in the report that includes plugin version differences.


guusdk:
This complements #1966 and #1968.

PR #1975, opened 2022-01-06
OF-2366: Drop commons-io

guusdk:
Replace the one single usage that we have with a equivalent bit of code.

### DIFF
--- a/xmppserver/pom.xml
+++ b/xmppserver/pom.xml
@@ -395,11 +395,6 @@
             <version>1.4</version>
         </dependency>
         <dependency>
-            <groupId>commons-io</groupId>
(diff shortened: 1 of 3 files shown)


akrherz (approved):

Fishbowler:
Have fed it a locally built JAR and a log file. It liked one and didn't like the other :+1:
LGTM

Fishbowler (approved):

Fishbowler:
Botheration. This still gets included by commons-fileupload.

guusdk:
That doesn't matter much. I primarily wanted to get rid of this as a direct dependency of Openfire, as we didn't really use it for much. There is not something inherently bad with this library.

PR #1978, opened 2022-01-13
OF-2368: Improve MUC ghost user detection

guusdk:
This commit combines approaches introduced by OF-910 and OF-2209 to detect a 'ghost user' in MUC.

Per OF-910, any Presence or Message stanza that contains an error is considered indicative of a ghost.

Per OF-2209, an IQ 'Ping' is used to determine ghosts.

Unlike the original solution for OF-2209, not all IQ error responses should be considered as indicative of a ghost. Firstly, the IQ error must be in response to an IQ Ping that was sent by the service itself. Furthermore, 'service-unavailable' and 'feature-not-implemented' can be considered as responses sent by a client (that does not support the IQ Ping protocol).

Finally, this commit adds a kick of users that have been pinged, but for which no response has been received at all.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -66,6 +66,7 @@ import org.jivesoftware.util.LocaleUtils;
 import org.jivesoftware.util.NotFoundException;
 import org.jivesoftware.util.TaskEngine;
 import org.jivesoftware.util.XMPPDateTimeFormat;
+import org.jivesoftware.util.cache.Cache;
 import org.jivesoftware.util.cache.CacheFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
```diff
@@ -389,28 +389,18 @@ public void processPacket(final Packet packet) {
...
+                removeChatUser(packet.getFrom());
+                return;
+            }
             // Check if the packet is a disco request or a packet with namespace iq:register
             if (packet instanceof IQ) {
                 if (process((IQ)packet)) {
                     Log.trace( "Done processing IQ stanza." );
                     return;
                 }
-            } else if (packet instanceof Message) {
```
akrherz, line 398: This removal sort of scares ignorant me.  Is the thought that this was previously unreachable code?

guusdk: No, it was reachable. It should now have been replaced with the invocation of `isDeliveryRelatedErrorResponse(packet)` a few lines higher.

akrherz: Thanks, but doesn't that code only handle stanzas of type `IQ` ?

guusdk: That's not the intention. About half-way through that method there's a comment that reads "The remainder of this implementation applies to Message and Presence stanzas."

akrherz:
Gonna run with this on ignite and see how it goes!

PR #1980, opened 2022-01-18
OF-2371: Increase queue capacity for OutgoingSessionPromise

guusdk:
The queue should be large enough to easily hold the stanzas sent back to a federated user that is joining a MUC room.


guusdk:
The default is the value used if nothing is in the database. Changing the default therefor doesn't require a database upgrade. If people have explicitly set the property, we probably shouldn't be changing it.

akrherz (approved):

PR #1981, opened 2022-01-18
OF-2372: Add support for 'forwarded for' HTTP headers to admin console

guusdk:
This optimizes detection of headers (notably the peers), for requests made against the admin console. Better registration of what IPs are accessing the console can't be bad...


guusdk:
This doesn't include a UI to change the new settings. Should we add that? If so: where?

gmournos (approved):
Fixes the problem.

PR #1986, opened 2022-01-25
Improvements to XMLProperties

guusdk:
The changes in this PR are motivated by observed thread contention in a high-volume environment. That's where the fixes in the first two commits stem from. I've added some minor fixes in subsequent commits.

I started off by using a ConcurrentMap to fix OF-2377, but that didn't quite cover all cases (in particular with protecting access to the 'document' field of this class). Instead, I've opted to go with a ReadWriteLock. This probably adds a bit of overhead, but should be more thread-safe.

One concession that I had to make for being able to use a ReadWriteLock, is to move all modifications that happened while a 'read' lock is acquired, outside of the lock (as ReentrantReadWriteLock guarantees that trying to obtain a 'write' lock while a 'read' lock is being held will fail). I've marked one operation that applies a change to an XML property without holding a 'write' lock. Triggering that code should be rare enough for me to believe that this won't be a practical problem.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
@@ -167,7 +167,7 @@ public class XMLProperties {
      * @param ignoreEmpty Ignore empty property values (return null)
      * @return the value of the specified property.
      */
-    public synchronized String getProperty(String name, boolean ignoreEmpty) {
+    public String getProperty(String name, boolean ignoreEmpty) {
         String value = null;
         boolean mustRewrite = false;
 
--- a/xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
@@ -328,8 +328,9 @@ public class XMLProperties {
      * @param name the name of the property to retrieve
      * @return all child property values for the given node name.
      */
+    @Deprecated
     public String[] getProperties(String name) {
-        return (String[]) getProperties(name, false).toArray();
+        return getProperties(name, false).toArray(new String[0]);
     }
 
     /**
--- a/xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
@@ -799,7 +799,7 @@ public class XMLProperties {
                         JiveGlobals.setPropertyEncrypted(name, true);
                     }
                     deleteProperty(name);
-                } else if (!databasePropertyValue.equals(xmlPropertyValue)) {
+                } else {
                     Log.warn("XML Property '" + name + "' differs from what is stored in the database.  Please make property changes in the database instead of the configuration file.");
                 }
             }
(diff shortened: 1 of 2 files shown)


GregDThomas:
Will look at this later - but am slightly curious as to the how it came about. My fading memory was that the XML file was pretty much only used to bootstrap the app until it could read the DB, so I am surprised that there was any contention at all. 

Note that I'm saying it's unnecessary, I'm sure it is, but don't know exactly why!

guusdk:
Usage of A SystemProperty that was not backed by an actual database entry which will attempt a lookup of an XMLProperty before eventually returning the default value.

GregDThomas:
> Usage of a SystemProperty that was not backed by an actual database entry which will attempt a lookup (every time) of an XMLProperty before eventually returning the default value.

That seems like an odd choice I made! 🤷

guusdk:
This probably isn't you. It might not be in SystemProperties directly, but a side-effect of a call that it does to an older API.

xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
```diff
@@ -182,43 +168,57 @@ public synchronized String getProperty(String name) {
      * @return the value of the specified property.
      */
     public synchronized String getProperty(String name, boolean ignoreEmpty) {
```
GregDThomas, line 170: Still `synchronized` ?

GregDThomas:
The code seems fine - no quibbles with the concession over the locks. I did find one `synchronized` method that needs the qualifier removing - a quick search may turn up others I've missed too.

The answer my next question is that this is probably an edge case that us used during an install or similar, but I am wondering if a better approach would be simply read in the whole XML file in to a Map of properties, and return values direct from the Map - a ConcurrentHashMap would mean no locking would be required. The Map & File can then be updated in any write statements.

GregDThomas (approved):

xmppserver/src/main/java/org/jivesoftware/util/XMLProperties.java
guusdk: hargh! I CTRL-F'ed for that! Good catch, thanks!

guusdk:
> The code seems fine - no quibbles with the concession over the locks. I did find one `synchronized` method that needs the qualifier removing - a quick search may turn up others I've missed too.
> 
> The answer my next question is that this is probably an edge case that us used during an install or similar, but I am wondering if a better approach would be simply read in the whole XML file in to a Map of properties, and return values direct from the Map - a ConcurrentHashMap would mean no locking would be required. The Map & File can then be updated in any write statements.

I think that this is good improvement, but I'd like to push that to a major release. I'm dreading some weird edge-case that's broken by this appoach.

Fishbowler:
This is fundamental stuff, and the tests pass which gives confidence. Have you got any real world testers that can help validate this? Given the database activity, would this be a candidate for the new DB tests you'll introduce with #1990, or is that overkill? Moreover: how can I help get this finished?

guusdk:
>

guusdk:
> Have you got any real world testers that can help validate this?

This now has been ported to a custom fork of Openfire, where it helped reducing the startup time for Openfire from ~1h15 minutes to ~30 minutes (this wasn't the only change to go in, and their use of Properties is _very_ peculiar - but still).

> Given the database activity, would this be a candidate for the new DB tests you'll introduce with #1990, or is that overkill?

I'm not going to claim that unit tests are bad to have (feel free to add some more :smile:), but the main point of this PR is to improve performance when there is _no_ database content. Unit tests aren't really suitable for performance-measuring things. It might be good to verify behavior more, but I do think that quickly gets a bit excessive for this PR.

> how can I help get this finished?

Hit the green button?

PR #1988, opened 2022-01-26
OF-2380 pubsub inmemory collection types

guusdk:
This pulls in the fix for OF-2379, which is a requirement for OF-2380.

The fix for OF-2380 introduces a non-compatible change to a cache definition used in clustering. We'd best not include this in a patch release, and instead do this only for a major release.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/InMemoryPubSubPersistenceProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/InMemoryPubSubPersistenceProvider.java
@@ -47,12 +47,12 @@ public class InMemoryPubSubPersistenceProvider implements PubSubPersistenceProvi
     /**
      * Cache that holds all nodes (mapped by service ID).
      */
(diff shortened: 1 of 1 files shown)


akrherz:
rebased and fixed conflict

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/InMemoryPubSubPersistenceProvider.java
```diff
@@ -349,7 +349,7 @@ public PublishedItem getPublishedItem( LeafNode node, PublishedItem.UniqueIdenti
         log.debug( "Getting published item {} for node {}", itemIdentifier.getItemId(), node.getUniqueIdentifier() );
 
         PublishedItem lastPublishedItem = null;
-        List<PublishedItem> publishedItems;
+        Set<PublishedItem> publishedItems;
```
akrherz, line 352: @guusdk Is this change still right?  This was the conflict I had to resolve.

PR #1990, opened 2022-01-28
OF-2382 / OF-2383 : Shared Group rework

guusdk:
`AbstractGroupProvider.getSharedGroupNames(JID)` returns _all_ groups, instead of the groups that are shared with the provided user.

This commit fixes that problem, and introduces a set of unit tests that verify the behavior.


guusdk:
Reverted to draft, as:
- more tests should be added to the new unit test. I believe that the 'shared group' functionality is covered, but unrelated other methods of the GroupProvider should be tested too.
- we need verification if the incorrect behavior of this method (corrected in this PR) wasn't depended on by other code in Openfire.

akrherz:
> * we need verification if the incorrect behavior of this method (corrected in this PR) wasn't depended on by other code in Openfire.

My suspicion was that the admin console used that, but am not seeing it.  Interesting find here, thanks

Fishbowler:
I've validated all of the new unit tests - that's some good coverage already

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
```diff
@@ -204,7 +235,10 @@ public void deleteGroup(String groupName) {
             // Remove the group entry.
             pstmt = con.prepareStatement(DELETE_GROUP);
             pstmt.setString(1, groupName);
-            pstmt.executeUpdate();
+            int groupsDeleted = pstmt.executeUpdate();
+            if (groupsDeleted == 0){
+                throw new GroupNotFoundException("Did not find group to delete");
+            }
```
Fishbowler, line 241: Is this better than an explicit check for the existence of the group? I realised I've used a mix of both. This way is cheaper, but it's less explicit.

guusdk: My primary concern would be to not change the API if we don't have to. This kind of thing is very likely to be implemented in third party integrations. We shouldn't break those for the sake of making things more consistent. That said, I wish we would've had more consistency in the first place. In this specific example, there also is an argument to be made for silently ignoring the fact that a group that is being deleted doesn't exist in the first place: the end result is going to be the same. That's again that we really _should_ be doing (or _not_ be doing) consistently, but alas...

xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupNameInvalidException.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (C) 2004-2008 Jive Software. All rights reserved.
```
guusdk, line 2: When adding a new file, lets update the Copyright notice to this year and "Ignite Realtime Foundation"

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.jivesoftware.openfire.group;
 
+import net.bytebuddy.implementation.bytecode.Throw;
```
guusdk, line 18: que?

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
```diff
@@ -369,7 +369,7 @@ public void addMember(String groupName, JID user, boolean administrator) throws
             con = DbConnectionManager.getConnection();
             pstmt = con.prepareStatement(ADD_USER);
             pstmt.setString(1, groupName);
-            pstmt.setString(2, server.isLocal(user) ? user.getNode() : user.toString());
+            pstmt.setString(2, server.isLocal(user) ? user.getNode() : user.asBareJID().toString());
```
guusdk, line 372: Although this makes sense to me, does this break existing behavior? Is there a chance that someone relies on this functionality?

xmppserver/src/main/java/org/jivesoftware/openfire/group/AbstractGroupProvider.java
```diff
@@ -95,8 +95,7 @@ public void addMember(String groupName, JID user, boolean administrator) throws
      * @throws UnsupportedOperationException if the provider is read only
      */
     @Override
-    public void updateMember(String groupName, JID user, boolean administrator)
-    {
+    public void updateMember(String groupName, JID user, boolean administrator) throws GroupNotFoundException {
```
guusdk, line 98: We're changing the 'contract' of this method, by having it throw GroupNotFoundException. We should add that to the interface, so that we force all implementations to adhere to that.

guusdk: ... which you did. Sorry! :)

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
```diff
@@ -1902,4 +1902,59 @@ public void testUpdateMemberOfNonExistentGroup() throws Exception {
...
+    public void testDeleteNonExistentMember() throws Exception {
+        final String GROUP_NAME = "Test Group A";
+        final JID needle = new JID("jane@" + Fixtures.XMPP_DOMAIN);
+        final DefaultGroupProvider provider = new DefaultGroupProvider();
+        provider.createGroup(GROUP_NAME);
+
+        try {
+            provider.deleteMember(GROUP_NAME, needle);
+        } catch (Throwable t){
+            throw new Exception("Exception deleting a non-existent group member. This should fail silently as a no-op.");
```
guusdk, line 1934: Nit: should we throw an Exception (which indicates that the implementation of the test has an error), or should this fail the test (indicating that the code under test has an error)?

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
```diff
@@ -1902,4 +1902,59 @@ public void testUpdateMemberOfNonExistentGroup() throws Exception {
...
+     */
+    public void testDeleteMemberFromNonExistentGroup() throws Exception {
+        final String GROUP_NAME = "Test Group A";
+        final JID needle = new JID("jane@" + Fixtures.XMPP_DOMAIN);
+        final DefaultGroupProvider provider = new DefaultGroupProvider();
+
+        try {
+            provider.deleteMember(GROUP_NAME, needle);
+        } catch (Throwable t){
+            throw new Exception("Exception deleting a group member from a non-existent group. This should fail silently as a no-op.");
```
guusdk, line 1953: same nit as above

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
Fishbowler: Oops. Fixed.

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
Fishbowler: Good point. That's lazy. Fixed.

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
Fishbowler: same fix as above :)

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
guusdk: Nah.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
guusdk: I believe that since this feedback, we've opted to add these changes to a non-patch release, and live with the consequences.

xmppserver/src/main/java/org/jivesoftware/openfire/group/AbstractGroupProvider.java
```diff
@@ -274,7 +266,7 @@ public boolean hasSharedGroups() {
                 if (DbConnectionManager.getDatabaseType() == DbConnectionManager.DatabaseType.oracle) {
                     pstmt = con.prepareStatement("SELECT 1 FROM dual " + HAS_SHARED_GROUPS_PARTIAL);
                 } else {
-                    pstmt = con.prepareStatement("SELECT 1 " + HAS_SHARED_GROUPS_PARTIAL);
+                    pstmt = con.prepareStatement("SELECT 1 FROM ofGroupProp " + HAS_SHARED_GROUPS_PARTIAL);
```
Fishbowler, line 269: I've updated this line, as it fails on HSQLDB, and so always returns true in the catch block

guusdk: With this, is it needed to differentiate between 'Oracle' and the other databases?

Reading up on this, it appears that HSQLDB should support `con.prepareStatement("VALUES('1') " + HAS_SHARED_GROUPS_PARTIAL`, but:
- I'm not sure if this actually works
- It probably needs to be specific to HSQLDB (adding another branch to that if-statement
- I'm not sure if this has any benefit over `"SELECT 1 FROM ofGroupProp"`

Fishbowler: > With this, is it needed to differentiate between 'Oracle' and the other databases?

Yeah, looks like the `dual` is the necessary method for Oracle.

> I'm not sure if this has any benefit over `"SELECT 1 FROM ofGroupProp"`

I'd prefer to leave it as-is. This makes sense to me, and is easy to parse.

guusdk:
Thank you for all of the hard work that went into this one!

PR #1997, opened 2022-02-07
OF-113: Limit lenght of group name/description to database restrictions

guusdk:
The database allows for up to 50 characters for a group name, and up to 255 for a description. The admin panel should not allow the end-user to provide longer values.


guusdk:
I fully acknowledge that I took the easy way out. :)

Fishbowler (approved):
LGTM 👍
No migration needed, since it's impossible to go longer.

PR #1998, opened 2022-02-09
OF-2393: Update slf4j to 1.7.36

Neustradamus:
https://www.slf4j.org/news.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
         <standard-taglib.version>1.2.5</standard-taglib.version>
         <mina.version>2.1.3</mina.version>
         <bouncycastle.version>1.70</bouncycastle.version>
-        <slf4j.version>1.7.32</slf4j.version>
+        <slf4j.version>1.7.36</slf4j.version>
         <log4j.version>2.17.1</log4j.version>
     </properties>
 


Fishbowler:
Something failed when you tried a higher version?

guusdk:
@Neustradamus why update to slf4j 1.7.33 when version 1.7.36 currently is the latest release?

Neustradamus:
1.7.36 fails, so we need to do 1.7.33, 1.7.34, 1.7.35 to see if the problem is here too.

guusdk:
What do you mean with "1.7.36 fails"? When I update slf4j to version 1.7.36 on my local computer rebuild and start Openfire, I'm not immediately seeing any problem.

Neustradamus:
The Openfire CI has failed when I have published for 1.7.36, have you looked why?

The log:
```
https://gist.github.com/Neustradamus/2bdba2c3974108d6e8f681379a62b181
```

guusdk:
On first glance, I would say that this build failed because of a problem on the server that was doing the build:

```
Error:  error reading /home/runner/.m2/repository/org/apache/logging/log4j/log4j-core/2.17.1/log4j-core-2.17.1.jar; zip END header not found
```

That sounds as if a file is corrupt. My advise would be to clean caches, and try again.

Neustradamus:
It is good, really strange the first Openfire CI test.

guusdk:
Thank you!

PR #1999, opened 2022-02-13
Bugfix/iq bind handler jid route miss

ferrantedev:
Sometimes user session JIDs are not correctly matched / routed, as explained in [Here](https://discourse.igniterealtime.org/t/unable-to-match-user-session-in-routingtable-correctly/91354/2)

You may validate the case by restoring the `String username = authToken.getUsername().toLowerCase()` method call in IQBindHandler.java

Added a couple of unit tests. Coverage is not 100%, there are 2 cases missing:

`String resource = packet.getChildElement().elementTextTrim("resource");
  if (resource == null || resource.length() == 0) {}
  else { //This case is missing, my knowledge of dom4j is limited

   }`

And also:

`AuthToken authToken = session.getAuthToken();
  if (authToken == null) { } // Even if I mock session.getAuthToken() as null, the sessionManager will throw a NPE before it reaches this point.`

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/handler/IQBindHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/handler/IQBindHandler.java
@@ -49,7 +49,7 @@ public class IQBindHandler extends IQHandler {
 
     private static final Logger Log = LoggerFactory.getLogger(IQBindHandler.class);
 
-    private IQHandlerInfo info;
+    private final IQHandlerInfo info;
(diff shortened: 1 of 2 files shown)


guusdk:
Thank you for taking the time to provide this change.

My primary concern is that by loosening checks on case sensitivity, we might end up opening a possibility for usernames to be spoofed. It might very well be the case that the changes in this PR do not increase the chance of that at all, but we will need to investigate that thoroughly before accepting the PR.

PR #2004, opened 2022-02-16
Release 4.7.1

guusdk:


akrherz:
Eh, we can't do the march within the PR dance like we do for plugins as I don't have bamboo configured to build against anything other than tags.  I'll update this PR accordingly

akrherz (approved):

PR #2005, opened 2022-02-17
OF-2395: Remove deprecated code

guusdk:
This commit removes code that was marked as being deprecated, if it was straightforward to remove / replace.

These changes are likely to break older code, but at some point, we'll need to bite the bullet, as the accumulated technical debt was getting pretty high (I've removed code that was marked as deprecated almost a decade ago).


guusdk:
This was one of those 'start pulling a thread' changes. I've somewhat arbitrary stopped pulling, which resulted in having done some work on moving int/long based durations to Duration instances, but not migrating corresponding `JiveGlobals.getIntProperty()` calls to a Duration-based `SystemProperty`. That too is a good thing (c)(r), but I had to stop myself somewhere, for this PR.

akrherz:
Thanks @guusdk for getting such a change in early on the 4.8 process!

akrherz:
rebased and resolved conflict

akrherz (approved):

PR #2006, opened 2022-02-24
OF-2398: Admin console: add visual indication of MUC room being locked

guusdk:
MUC rooms can be 'locked', which typically occurs when they are initially being created and configured. When a MUC room is 'locked', it cannot be entered by users.

The Openfire admin console should make it possible for an administrator to see if a room is locked.

This commit replaces the value in the 'users' column on the room summary page with a 'lock' icon, and displays a warning on the room settings page, when the room is locked.

image
image


akrherz (approved):

Fishbowler:
Does this actually want to be in 4.7.x stream?

guusdk:
@Fishbowler I'm at the very least not against putting it in the 4.7 branch. It is pretty much a cosmetic change, that is very unlikely to have side effects. Do you have any particular reasons for not adding it to 4.7?

Fishbowler:
No objection whatsoever 👍

PR #2007, opened 2022-02-27
OF-2399: When a property is migrated, always assume that no restart is required for it to take effect

GregDThomas:


GregDThomas:
Note; this is entirely untested. It looks like it should do the right thing, but I'm not entirely sure how to reproduce the issue.

guusdk:
Ah, that's clever. Thanks. I've tested it locally: it seems to work without side-effects.

Reproducing the issue is dead-simple. Start a new Openfire instance (I'm using 'demoboot') and look at the system properties. Without this fix, you'll see that the `xmpp.domain` property is requesting an Openfire restart.

PR #2009, opened 2022-03-03
OF-2403: Improve memory usage reporting on admin console.

guusdk:


guusdk:
There's a small chance that this breaks stuff (see first commit message). Let's not add this to a patch-release.

Fishbowler (approved):
Tested using a Temurin JDK and running a container based on ibmjava:11. LGTM :+1:

PR #2011, opened 2022-03-04
OF-2405: Enhance log info of loaded plugin by version

gjaekel:
Add version info to the "Successfully loaded plugin"-message

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginManager.java
@@ -691,7 +691,7 @@ public class PluginManager
                 AdminConsole.addModel( canonicalName, adminElement );
             }
             firePluginCreatedEvent( canonicalName, plugin );
-            Log.info( "Successfully loaded plugin '{}'.", canonicalName );
+            Log.info( "Successfully loaded plugin '{}-{}'.", canonicalName, metadata.getVersion() );
             return true;
         }
         catch ( Throwable e )


guusdk (approved):

guusdk:
Looks good. Thanks!

gjaekel:
May you please check if this change looks good in the Openfire logfile? Or is there such a logfile available in the CI buildlogs?

guusdk:
*gasp* you didn't test it? ;)

image

gjaekel:
No, I can't in an easy way. But a syntax error had have break the build an a semantic error just disturb the log output. But I take a look at the sources and typical plugin.xml's what the content of this string would be. 

Thank you.

guusdk:
Hah, I _did_ find a small issue: the admin console is loaded as a plugin, but has no version number. Upon start of the server, this is printed: 

```Successfully loaded plugin 'admin-null'.```

gjaekel:
Hm, then you discovered a miss in the corresponding plugin.xml :smile:

guusdk:
It doesn't have one. It's extraspecial! Let's just write a bit of code that ignores missing metadata or missing versions, to be safe.

gjaekel:
> It doesn't have one. 

https://github.com/igniterealtime/Openfire/blob/master/xmppserver/plugin.xml seems to be copied to `.../plugins/admin/`

> a bit of code

`.toString().replace("null","<unknown>")`?

guusdk:
I would just modify your original commit:

```java
if (metadata != null && metadata.getVersion != null) {
    Log.info( "Successfully loaded plugin '{}-{}'.", canonicalName, metadata.getVersion());
} else {
    Log.info( "Successfully loaded plugin '{}'.", canonicalName);
}

gjaekel:
That look much more professional.

PR #2018, opened 2022-03-05
OF-2411: Prevent deadlock in XmlProperties

guusdk:
OF-2377 introduced an updated implementation of the thread-safety module of XmlProperties (the old one wasn't fully thread-safe).

This new implementation causes a very specific problem at startup: When the initial attempt to migrate an XmlProperty is undertaken, TaskEngine is instantiated, which in turn tries to read/write properties (while under lock). This causes a deadlock.

To work around this issue, this commit replaces the usage of TaskEngine with a throw-away thread (that does not use any properties).

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/JiveGlobals.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/JiveGlobals.java
@@ -1289,15 +1289,14 @@ public class JiveGlobals {
                 try {
                     securityProperties = new XMLProperties(home + File.separator + JIVE_SECURITY_FILENAME);
                     setupPropertyEncryption();
-                    new Thread(() -> {
-                        // Migrate all secure XML properties into the database automatically
-                        for (String propertyName : securityProperties.getAllPropertyNames()) {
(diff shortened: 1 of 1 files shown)


guusdk:
This fixes the issue at hand, but I'm wondering if having the ReadWriteLock is a good idea in the first place.

Fishbowler (approved):

GregDThomas:
LGTM. However, I do wonder if doing it in a separate thread at all is wise. It means at startup some will have been migrated, and some not. Probably OK, but? Unless this is really slow (suspect it it’s only a couple of passwords? DB and LDAP at most?) perhaps just do it synchronously?

guusdk:
I wondered about that too. I'm not sure why that was made asynchronously, but that does open the door for things to be mid-transition while they're used. I've now raised https://igniterealtime.atlassian.net/browse/OF-2412 to change this, and added a commit with a fix for that issue to this PR.

Fishbowler:
I've hit this issue whilst loading XML config in https://github.com/surevine/openfire-docker-compose - something recently has affected timings perhaps?

This branch solves the issue.

Fishbowler:
OF-2377 wasn't ported to 4.7, so no cherry-pick required

PR #2026, opened 2022-03-07
OF-2414: Where appropriate, send a stream error before closing a stream.

guusdk:
When a stream is being closed, it's often helpful to signal why this happens, especially when the stream is closed due to an error.

This commit adds stream errors in various places where such context is desirable.

The API to close a connection has been modified to be able to include an optional stream error. This replaces an older structure where a stream error is first sent manually, followed by a call to the 'close' API. With this change, only one attempt to send data is made instead of two (one for the stream error, and another one for the end-stream tag).

As there's an API change, we'd better not include this change in a patch release.


Fishbowler:
Have you been able to test any number of these connection types already?

guusdk:
I have not.

akrherz (approved):

akrherz:
I think this commit should have referenced OF-2413

guusdk:
Sadly, this contained a bug. This is being addressed in https://github.com/igniterealtime/Openfire/pull/2054

PR #2027, opened 2022-03-11
OF-2415: Allow empty plugins directory at boot time in Docker entrypoint

Fishbowler:
Without this fix, mapping an empty folder prevents Openfire from booting

### DIFF
--- a/build/docker/entrypoint.sh
+++ b/build/docker/entrypoint.sh
@@ -43,10 +43,12 @@ initialize_log_dir() {
 }
 
 copy_provided_plugins() {
-  if [ -d "/opt/plugins" ]; then
-    echo "Copying user-provided plugins"
-    cp -nR /opt/plugins/* ${OPENFIRE_DIR}/plugins/
-  fi
+  for file in /opt/plugins/*.jar; do # Workaround for https://github.com/koalaman/shellcheck/wiki/SC2144 that doesn't involve redirecting terminal output
+    if [ -f "$file" ]; then
+      echo "Copying user-provided plugin: $file"
+      cp -n "$file" "${OPENFIRE_DIR}/plugins/"
+    fi
+  done
 }
 
 # allow arguments to be passed to openfire launch


build/docker/entrypoint.sh
```diff
@@ -43,7 +43,7 @@ initialize_log_dir() {
 }
 
 copy_provided_plugins() {
-  if [ -d "/opt/plugins" ]; then
+  if ls "/opt/plugins/*.jar" > /dev/null 2>&1; then
```
guusdk, line 46: I'm going full-power pedantic mode here: we're checking for _jar_ files, but we copy _everything_. That's probably not right.

pedantic++: I don't like that output redirection. If we don't need to check for jar files specifically (but can trigger the copy for _any_ content, we might be able to get away with something like `[ "$(ls -A /opt/plugins)" ]`

Fishbowler:
Rebased

Fishbowler:
Feedback applied

build/docker/entrypoint.sh
```diff
@@ -43,10 +43,13 @@ initialize_log_dir() {
...
 
 copy_provided_plugins() {
-  if [ -d "/opt/plugins" ]; then
-    echo "Copying user-provided plugins"
-    cp -nR /opt/plugins/* ${OPENFIRE_DIR}/plugins/
-  fi
+  for file in /opt/plugins/*.jar; do # Workaround for https://github.com/koalaman/shellcheck/wiki/SC2144 that doesn't involve redirecting terminal output
+    if [ -f $file ]; then
+      echo "Copying user-provided plugins"
+      cp -nR /opt/plugins/*.jar "${OPENFIRE_DIR}/plugins/"
```
akrherz, line 49: What is going on here with that '*.jar' when you are looping with $file variable?

Fishbowler: You're not wrong.
Let me give that another... bash :)

build/docker/entrypoint.sh
```diff
@@ -43,10 +43,13 @@ initialize_log_dir() {
...
 copy_provided_plugins() {
-  if [ -d "/opt/plugins" ]; then
-    echo "Copying user-provided plugins"
-    cp -nR /opt/plugins/* ${OPENFIRE_DIR}/plugins/
-  fi
+  for file in /opt/plugins/*.jar; do # Workaround for https://github.com/koalaman/shellcheck/wiki/SC2144 that doesn't involve redirecting terminal output
+    if [ -f "$file" ]; then
+      echo "Copying user-provided plugins"
+      cp -n "$file" "${OPENFIRE_DIR}/plugins/"
+      break
```
akrherz, line 50: Why is there a `break` here now?

Fishbowler: Oh, wow. That's rushing. Fixed.

Fishbowler:
Rebased.

Fishbowler:
Tested with 0, 1 and 2 plugins.

build/docker/entrypoint.sh
```diff
@@ -43,10 +43,12 @@ initialize_log_dir() {
 }
 
 copy_provided_plugins() {
-  if [ -d "/opt/plugins" ]; then
-    echo "Copying user-provided plugins"
-    cp -nR /opt/plugins/* ${OPENFIRE_DIR}/plugins/
-  fi
+  for file in /opt/plugins/*.jar; do # Workaround for https://github.com/koalaman/shellcheck/wiki/SC2144 that doesn't involve redirecting terminal output
+    if [ -f "$file" ]; then
+      echo "Copying user-provided plugins"
```
akrherz, line 48: Can we add the $file in the echo message?  Otherwise, that echo comes for each plugin within the loop?

akrherz (approved):

PR #2033, opened 2022-03-17
OF-2420 phase out lightbox

guusdk:
This replaces the outdated lightbox library (used to show modal popups) with the `<dialog>` HTML tag.


guusdk:
This was only used in the AD / LDAP configuration pages. I was not able to test anything beyond the first page, as for some reason, I can't quite get OpenLDAP up-and-running. I don't think that this has anything to do with the changes in this commit.

Fishbowler:
Something a little squiff with the testing popup being transparent, but as far as I can see, this isn't new, and is only visible when the browser is at a particular width.

image

Fishbowler:
The test on User Mapping ruins the styling

image

I think I've fixed this by removing some tags to ensure those pages are HTML fragments rather than full-blown pages.

Fishbowler:
Although the fix doesn't solve the z-ordering problem

image

guusdk:
This now works for me, with the exception of what appears a CSRF issue when testing an admin password. I don't think that this is new.

PR #2036, opened 2022-04-04
OF-2426: Refactor Group implementation

guusdk:
Reduce complexity by centralizing the responsiblity of dispatching events to event listeners and cache updates.

Avoid using the event listener mechanism that's also being provided to users of this API in the implementation itself.

As a byproduct of the latter, this fixes OF-2426: an event dispatched by the implementation no longer re-populates the cache unexpectedly.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupAdminRemoved.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupAdminRemoved.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004-2008 Jive Software. All rights reserved.
+ * Copyright (C) 2004-2008 Jive Software, 2022 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,7 +19,6 @@ import org.dom4j.Element;
 import org.jivesoftware.openfire.commands.AdHocCommand;
 import org.jivesoftware.openfire.commands.SessionData;
 import org.jivesoftware.openfire.component.InternalComponentManager;
-import org.jivesoftware.openfire.event.GroupEventDispatcher;
 import org.jivesoftware.openfire.group.Group;
 import org.jivesoftware.openfire.group.GroupManager;
 import org.jivesoftware.openfire.group.GroupNotFoundException;
@@ -28,7 +27,6 @@ import org.xmpp.forms.FormField;
 import org.xmpp.packet.JID;
 
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -71,36 +69,34 @@ public class GroupAdminRemoved extends AdHocCommand {
             return;
         }
 
-        // Creates event params.
-        Map<String, Object> params = null;
-
+        JID admin;
         try {
-
             // Get the admin
-            String admin = get(data, "admin", 0);
-
-            // Adds the admin
-            params = new HashMap<>();
-            params.put("admin", admin);
+            String adminValue = get(data, "admin", 0);
+            admin = new JID(adminValue);
         }
         catch (NullPointerException npe) {
             note.addAttribute("type", "error");
             note.setText("Admin required parameter.");
             return;
         }
+        catch (IllegalArgumentException e) {
+            note.addAttribute("type", "error");
+            note.setText("Admin value needs to be a JID.");
+            return;
+        }
 
-        // Sends the event
-        Group group;
+        final Group group;
         try {
-            group = GroupManager.getInstance().getGroup(groupname, true);
-
-            // Fire event.
-            GroupEventDispatcher.dispatchEvent(group, GroupEventDispatcher.EventType.admin_removed, params);
-
+            group = GroupManager.getInstance().getGroup(groupname);
         } catch (GroupNotFoundException e) {
             note.addAttribute("type", "error");
             note.setText("Group not found.");
+            return;
         }
+
+        // Perform post-processing (cache updates and event notifications).
+        GroupManager.getInstance().adminRemovedPostProcess(group, admin);
 
         // Answer that the operation was successful
         note.addAttribute("type", "info");
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupAdminRemoved.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupAdminRemoved.java
@@ -15,6 +15,7 @@
  */
 package org.jivesoftware.openfire.commands.event;
 
+import org.apache.commons.lang3.StringUtils;
 import org.dom4j.Element;
 import org.jivesoftware.openfire.commands.AdHocCommand;
 import org.jivesoftware.openfire.commands.SessionData;
@@ -26,9 +27,7 @@ import org.xmpp.forms.DataForm;
 import org.xmpp.forms.FormField;
 import org.xmpp.packet.JID;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Notifies the that a admin was removed from the group. It can be used by user providers to notify Openfire of the
@@ -58,40 +57,38 @@ public class GroupAdminRemoved extends AdHocCommand {
 
         Map<String, List<String>> data = sessionData.getData();
 
-        // Get the group name
-        String groupname;
-        try {
-            groupname = get(data, "groupName", 0);
-        }
-        catch (NullPointerException npe) {
-            note.addAttribute("type", "error");
-            note.setText("Group name required parameter.");
-            return;
+        // Input validation
+        final Set<String> inputValidationErrors = new HashSet<>();
+
+        Group group = null;
+        final String groupName = get(data, "groupName", 0);
+        if (StringUtils.isBlank(groupName)) {
+            inputValidationErrors.add("The parameter 'groupName' is required, but is missing.");
+        } else {
+            try {
+                group = GroupManager.getInstance().getGroup(groupName);
+            } catch (GroupNotFoundException e) {
+                inputValidationErrors.add("The group '" + groupName + "' does not exist.");
+            }
         }
 
         JID admin;
-        try {
-            // Get the admin
-            String adminValue = get(data, "admin", 0);
-            admin = new JID(adminValue);
-        }
-        catch (NullPointerException npe) {
-            note.addAttribute("type", "error");
-            note.setText("Admin required parameter.");
+        final String adminValue = get(data, "admin", 0);
+        if (StringUtils.isBlank(adminValue)) {
+            inputValidationErrors.add("The parameter 'admin' is required, but is missing.");
             return;
-        }
-        catch (IllegalArgumentException e) {
-            note.addAttribute("type", "error");
-            note.setText("Admin value needs to be a JID.");
-            return;
+        } else {
+            try {
+                admin = new JID(adminValue);
+            } catch (IllegalArgumentException e) {
+                inputValidationErrors.add("The value for parameter 'admin' should be a valid JID, but '" + adminValue + "' is not.");
+                return;
+            }
         }
 
-        final Group group;
-        try {
-            group = GroupManager.getInstance().getGroup(groupname);
-        } catch (GroupNotFoundException e) {
+        if (!inputValidationErrors.isEmpty()) {
             note.addAttribute("type", "error");
-            note.setText("Group not found.");
+            note.setText(StringUtils.join(inputValidationErrors, " "));
             return;
         }
 
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupMemberAdded.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupMemberAdded.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004-2008 Jive Software. All rights reserved.
+ * Copyright (C) 2004-2008 Jive Software, 2022 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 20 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupMemberAdded.java
```diff
@@ -71,37 +69,43 @@ public void execute(SessionData sessionData, Element command) {
...
-        Map<String, Object> params = null;
-        
-        try {
+        final String wasAdminValue = get(data, "wasAdmin", 0);
+        if (wasAdminValue == null) {
+            note.addAttribute("type", "error");
+            note.setText("WasAdmin required parameter.");
+            return;
+        }
+        final boolean wasAdmin = "1".equals(wasAdminValue) || Boolean.parseBoolean(wasAdminValue);
```
Fishbowler, line 78: What's populating the "wasAdmin"?

xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
```diff
@@ -789,6 +625,423 @@ private void evictCachedUserSharedGroups() {
...
+     */
+    public void createGroupPostProcess(@Nonnull final Group group)
+    {
+        // Adds default properties if they don't exist.
+        if (group.getSharedWith() == null) {
+            group.shareWithNobody();
+        }
+
+        // Update caches.
+        clearGroupNameCache();
```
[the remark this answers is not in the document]
guusdk: Possibly, but much of the code follows the "if anything changed, clear all caches." Although that leaves room for improvement, it's a whole lot easier to work with, when diagnosing issues. And this code has issues...

xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupMemberAdded.java
guusdk: A new data field in the form that is passed into this method by the `data` variable.

xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupAdminRemoved.java
```diff
@@ -71,37 +69,35 @@ public void execute(SessionData sessionData, Element command) {
...
-            group = GroupManager.getInstance().getGroup(groupname, true);
-
-            // Fire event.
-            GroupEventDispatcher.dispatchEvent(group, GroupEventDispatcher.EventType.admin_removed, params);
-
+            group = GroupManager.getInstance().getGroup(groupname);
         } catch (GroupNotFoundException e) {
             note.addAttribute("type", "error");
             note.setText("Group not found.");
+            return;
```
GregDThomas, line 95: "Ditto" - IYSWIM

xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupMemberRemoved.java
```diff
@@ -71,37 +69,35 @@ public void execute(SessionData sessionData, Element command) {
...
             // Get the member
-            String member = get(data, "member", 0);
-
-            // Adds the member
-            params = new HashMap<>();
-            params.put("member", member);
+            String memberValue = get(data, "member", 0);
+            member = new JID(memberValue);
         }
         catch (NullPointerException npe) {
```
GregDThomas, line 78: Catching a NPE again

xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupMemberRemoved.java
```diff
@@ -71,37 +69,35 @@ public void execute(SessionData sessionData, Element command) {
...
-            group = GroupManager.getInstance().getGroup(groupname, true);
-
-            // Fire event.
-            GroupEventDispatcher.dispatchEvent(group, GroupEventDispatcher.EventType.member_removed, params);
-
+            group = GroupManager.getInstance().getGroup(groupname);
         } catch (GroupNotFoundException e) {
             note.addAttribute("type", "error");
             note.setText("Group not found.");
+            return;
```
GregDThomas, line 95: duplicated code

xmppserver/src/main/java/org/jivesoftware/openfire/commands/event/GroupModified.java
```diff
@@ -82,56 +80,92 @@ public void execute(SessionData sessionData, Element command) {
...
         try {
             group = GroupManager.getInstance().getGroup(groupname, true);
-
-            // Fire event.
-            GroupEventDispatcher.dispatchEvent(group, GroupEventDispatcher.EventType.group_modified, params);
-
         } catch (GroupNotFoundException e) {
             note.addAttribute("type", "error");
             note.setText("Group not found.");
+            return;
```
GregDThomas, line 145: duplicated code

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
```diff
@@ -1,19 +1,27 @@
...
 import java.sql.SQLException;
-import java.util.AbstractSet;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Set;
+import java.util.*;
```
GregDThomas, line 21: Style thing, I know, but I don't like wildcard imports.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
```diff
@@ -32,7 +40,7 @@
 
 public class DefaultGroupPropertyMap<K,V> extends PersistableMap<K,V> {
 
-    private static final long serialVersionUID = 3128889631577167040L;
+    private static final long serialVersionUID = -7501088917416583815L;
```
GregDThomas, line 43: Could this change in serialVersionUID cause problems in mixed-version clusters, or during upgrade procedures?

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
```diff
@@ -94,17 +102,28 @@ public V put(K key, V value) {
...
         return result;
     }
 
     @Override
     public void clear() {
+        final Map<K,V> originalMap = new HashMap<>(this); // copy to be used by event handling.
         super.clear();
-        deleteAllProperties();
+
+        // Ignore all entries that are not strings.
```
GregDThomas, line 115: Might be worth clarifying /why/ you are ignoring them

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
```diff
@@ -191,7 +210,7 @@ public K next() {
         public void remove() {
             delegate.remove();
             if (current instanceof String) {
-                deleteProperty((String)current);
+                deleteProperty((String)current, null); // FIXME by not providing the original value, some event handlers may not be able to handle this.
```
GregDThomas, line 213: Does this FIXME need a JIRA reference?

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
```diff
@@ -242,8 +261,14 @@ public boolean remove(Object o) {
...
         public void clear() {
+            final Map<String,String> originalMap = new HashMap<>(); // copy to be used by event handling.
+            for (Entry<K,V> entry : delegate) {
+                if (entry.getKey() instanceof String && entry.getValue() instanceof String) {
+                    originalMap.put((String) entry.getKey(), (String) entry.getValue());
+                }
+            }
             delegate.clear();
-            deleteAllProperties();
+            deleteAllProperties(originalMap);
```
GregDThomas, line 271: This method doesn't do what the Javadoc says ("Removes all the elements in the set") - possibly "Removes all the elements in the set whose values are Strings" ?

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
```diff
@@ -389,15 +414,16 @@ public boolean hasNext() {
         public void remove() {
             delegate.remove();
             K key = current.getKey();
+            V value = current.getValue();
             if (key instanceof String) {
-                deleteProperty((String)key);
+                deleteProperty((String)key, value instanceof String ? (String)value : null);
```
GregDThomas, line 419: (there's clearly a lot of String protection going on here that I don't understand)

xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
```diff
@@ -789,6 +625,423 @@ private void evictCachedUserSharedGroups() {
...
+     */
+    public void memberAddedPostProcess(@Nonnull final Group group, @Nonnull final JID member, final boolean wasAdmin)
+    {
+        evictCachedUserForGroup(member);
+        final Group updatedGroup;
+        try {
+            // By forcing the lookup, the cache will automatically receive the required refresh.
+            updatedGroup = getGroup(group.getName(), true);
+        } catch (GroupNotFoundException e) {
+            Log.error("Group '{}' was not found after a member was added to it. This is indicative of a bug in Openfire. Please consider reporting it.", group.getName(), e);
```
GregDThomas, line 769: This might be also just a co-incidence, if someone deletes the group at about the same time someone is added to it.

Is it possible to single thread some of these actions? Adding a member requires a read lock, deleting the group a read-write lock? Suspect it's tricky across the cluster.

Probably more effort than it's worth.

xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
```diff
@@ -789,6 +625,423 @@ private void evictCachedUserSharedGroups() {
...
+        event.put("type", "propertyDeleted");
+        event.put("propertyKey", "*");
+        GroupEventDispatcher.dispatchEvent(updatedGroup, GroupEventDispatcher.EventType.group_modified, event);
+    }
+
+    // A generic method to process the effect of property changes to contact list sharing. Re-used by all property change post-processing methods.
+    private void propertyChangePostProcess(final Group group, final String key, final String originalValue)
+    {
+        switch (key) {
+            case "sharedRoster.showInRoster": {
```
GregDThomas, line 1010: Should these magic Strings be constants so that typo's don't creep in?

* "sharedRoster.showInRoster"
* "everybody"
* "sharedRoster.groupList"

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
```diff
@@ -92,11 +91,16 @@ public void tearDown() throws Exception {
         // Reset static fields after use (to not confuse other test classes).
         // TODO: this ideally goes in a static @AfterClass method, but that's not supported in JUnit 3.
         for (String fieldName : Arrays.asList("INSTANCE", "provider")) {
-            Field field = GroupManager.class.getDeclaredField(fieldName);
+            final Field field = GroupManager.class.getDeclaredField(fieldName);
```
GregDThomas, line 94: Yuck. Is there way to avoid this? Perhaps a package-protected method to set the listeners, to be used for test purposes?

GregDThomas (approved):
A good improvement, I think. I've commented on a few picky style issues, but nothing of any real substance I think.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
guusdk: I am not bothered by them in the slightest. :) I can count the number of times that these actually caused an issue for me on the fingers of one hand that belongs to the worst employee of a saw mill.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
guusdk: I'm not sure, but I don't think that the change is needed at all. The composition of the object did not change. I'll revert this.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
GregDThomas: That's fine. It has caused me problems in the past, which is why I try and avoid them.

(At the time, we had one person using IntelliJ and another with Eclipse. From memory Eclipse imported a wildcard which tripped up IntelliJ because of two different classes with the same name (or vice-versa? I forget)

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
guusdk: Ack. I've raised OF-2430 for this and modified the comment.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
guusdk: Actually, they _are_ all deleted. The map that's passed in is only used for event dispatching. I'm unsure if there is any sense in having this be a generic map - if we're not using Strings, this implementation seems to choke.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupPropertyMap.java
guusdk: This implementation apparently only works with Strings, while it accepts a generic type. Something to dig into another time...

xmppserver/src/main/java/org/jivesoftware/openfire/group/GroupManager.java
guusdk: I considered that, but given the complexity of the code, and the way how groups are nested, possibly retrieved from external sources (AD/LDAP), and used all over the place, and clustering, as you mentioned, my fear of introducing deadlocks won the argument.

xmppserver/src/test/java/org/jivesoftware/openfire/group/DefaultGroupProviderTest.java
guusdk: That's trading in one yuck for another yuck. The test code is a bit of a mess already (with regards to not using `@AfterClass` and the concessions that it needs to do for that). I'm opting to keep the yuckiness in one class.

guusdk:
Thanks Greg! I believe I've addressed most of your concerns, and choose to not act on a couple of them. Notably, the code duplication in the ad hoc commands. Although there is some room for improvement, I think it'd hurt readability more than that it brings benefit.

PR #2043, opened 2022-04-20
fix: prevent inserting affiliations and subscriptions to db if allready exists

mightymop:


guusdk:
Thanks for the PR! I do have a concern that this ignores a potential data corruption. Can you please explain how the problem can be triggered?

mightymop:
> Thanks for the PR! I do have a concern that this ignores a potential data corruption. Can you please explain how the problem can be triggered?

The issue only occurs on our production system but not on the testsystem… so hard to say what is causing it… the difference is the amount of users currently logged in but they are mostly not more then 400!?
I saw in the code (in one class) that you allready had a member variable to check if the object is allready in the db or not and i clonded the checks for the other classes too.
The PR fixes the issue on my system. Maybe it is a timing problem between flushing/clearing the cache and writing data to DB

PS:
i am using SQL Server

Fishbowler:
What happens to the state when a Node is loaded from the database on boot? From a plain reading, the `isSavedToDB` would be false?

PR #2044, opened 2022-04-25
Fix for multiple private messages being sent when user signed in from multiple locations

qoole:
Fixes: https://discourse.igniterealtime.org/t/muc-rooms-and-carbon-copies-sending-duplicate-messages-when-connected-with-multiple-sessions/91622


qoole:
Apologies - this must have been because I based it off the v4.7.1 tag - I'll redo it from main and re-request.
My first ever PR was never going to go smoothly!

PR #2047, opened 2022-06-02
allow Openfire to operate as a gateway for XMPP traffic

gmournos:
If Openfire does not validate that the messages it receives are addressed to its domain, then it will go ahead and forward messages to the target servers and it will effectively operate as a gateway for XMPP traffic.

This PR makes this validation configurable by setting system property "xmpp.server.validate.host" to "false" (default is "true"). The name of this property is similar to property "xmpp.client.validate.host", used by ClientStanzaHandler.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerSocketReader.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerSocketReader.java
@@ -23,6 +23,7 @@ import org.jivesoftware.openfire.auth.UnauthorizedException;
 import org.jivesoftware.openfire.event.ServerSessionEventDispatcher;
 import org.jivesoftware.openfire.interceptor.PacketRejectedException;
 import org.jivesoftware.openfire.session.LocalIncomingServerSession;
+import org.jivesoftware.util.JiveGlobals;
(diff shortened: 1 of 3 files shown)


guusdk:
Hi George, thanks for this! First off, I'm thoroughly impressed by someone figuring out that Openfire supports this kind of behavior. I was not aware that it could be used in this way.

My primary concern with this solution is that it foregoes all authentication/validation of remote domains. I can see how this might be acceptable in a completely closed down network, but the security implications out that could be significant. I'm worried that this can be leveraged to resort to open relay-like abuse, or scenarios where remote users can impersonate others, or more issues of that nature.

After experimentation, I believe that I have found a solution that does not require the code to be modified to skip all authentication/validation of streams. Instead, it takes a configurable list of domains that are allowed to communicate 'through' the gateway (interestingly enough, the change-set is even smaller than yours). The obvious downside of this approach is that all domains need to be pre-configured/pre-approved. Is this acceptable to your use-case?

guusdk:
I have raised https://igniterealtime.atlassian.net/browse/OF-2476 to track this feature request. I will follow up with a PR of my own that implements the feature as described in my previous message.

PR #2048, opened 2022-06-06
Added code of conduct

patelkh:
Adopting an explicit code of conduct per guidelines in Producing Open Source Software by Karl Fogel.


guusdk:
Excuse me, who are you?

I'm not against adopting a code of conduct for this project, but if that is to be done, it would be an effort that is driven by members of the IgniteRealtime community, not by a seemingly anonymous passerby that drops a pull request, without there having been any discussion had on the topic, or a rationale given.

PR #2049, opened 2022-06-08
OF-2444: Deadlock in BOSH implementation

guusdk:
The changes in this PR should lead to a fix for a BOSH related deadlock in Openfire 4.7.0 and 4.7.1, as described in [OF-2444](https://igniterealtime.atlassian.net/browse/OF-2444). When the PR was created, this wasn't yet achieved. I have created the PR as a draft for that reason.

My goal is to incrementally reduce complexity and, in particular, the usage of locks in the `HttpSession` class. In particular, I believe that the deadlock described in OF-2444 is prevented when the processing of inbound data (provided by a BOSH request) no longer occurs under guard of the lock represented by the `connectionQueue` field.

This PR builds up to a refactoring that would allow for that change to become possible.

The refactoring is sufficiently complex that I would welcome feedback on my approach while I'm making process. That is why I'm creating this PR, prior to it being ready. I am doing my best to make each commit in this PR to be a isolated change, which hopefully facilitates reviewing things.

As an aside, while trying to refactor the code, I am running into what I believe to be other issues. I am raising new JIRA issues as I go along. Fixes might be included in this PR, but until they are included, I have left TODO and FIXME comments that reference the JIRA issues.

So far, I think these I have identified these issues:
- https://igniterealtime.atlassian.net/browse/OF-2445
- https://igniterealtime.atlassian.net/browse/OF-2446
- https://igniterealtime.atlassian.net/browse/OF-2447
- https://igniterealtime.atlassian.net/browse/OF-2448
- https://igniterealtime.atlassian.net/browse/OF-2449
- https://igniterealtime.atlassian.net/browse/OF-2450
- https://igniterealtime.atlassian.net/browse/OF-2451

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
@@ -160,7 +160,13 @@ public class HttpSession extends LocalClientSession {
     // FIXME: OF-2445: Prevent elements to be scheduled for delivery in #pendingElements after the session is closed (as they would never be delivered).
     private final ConcurrentLinkedQueue<Deliverable> pendingElements = new ConcurrentLinkedQueue<>();
 
-    private final List<Delivered> sentElements = new ArrayList<>();
+    /**
+     * A list of data that has been delivered, for potential future retransmission.
+     *
+     * The size of this collection is limited. It will contain only the last few transmitted elements.
+     */
+    @GuardedBy("itself")
+    private final List<Delivered> sentElements = new ArrayList<>(); // FIXME: OF-2446 Use a more appropriate data type than ArrayList.
 
     private Instant lastPoll = Instant.EPOCH;
     private Duration inactivityTimeout;
@@ -661,13 +667,14 @@ public class HttpSession extends LocalClientSession {
      * @return previously delivered data when found.
      */
     @Nullable
-    @GuardedBy("connectionQueue")
     private Delivered retrieveDeliverable(final long rid) {
         Delivered result = null;
-        for (Delivered delivered : sentElements) {
-            if (delivered.getRequestID() == rid) {
-                result = delivered;
-                break;
+        synchronized (sentElements) {
+            for (Delivered delivered : sentElements) {
+                if (delivered.getRequestID() == rid) {
+                    result = delivered;
+                    break;
+                }
             }
         }
         return result;
@@ -967,11 +974,13 @@ public class HttpSession extends LocalClientSession {
 
         // Keep track of data that has been delivered, for potential future retransmission.
         final Delivered delivered = new Delivered(deliverable, connection.getRequestId());
-        while (sentElements.size() > maxRequests) {
-            sentElements.remove(0);
-        }
+        synchronized (sentElements) {
+            while (sentElements.size() > maxRequests) {
+                sentElements.remove(0);
+            }
 
-        sentElements.add(delivered);
+            sentElements.add(delivered);
+        }
     }
 
     private void closeSession(@Nullable final StreamError error) {
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
@@ -662,24 +662,22 @@ public class HttpSession extends LocalClientSession {
      * Attempts to find data that was previously sent back to the client, using a particular request ID. This is
      * expected to be used to process requests for retransmission.
      *
-     * The implementation only holds a limited amount of data. It will return null when no data that matches the ID can
-     * be found.
+     * The implementation only holds a limited amount of data. It will return an empty Optional when no data that
+     * matches the ID can be found.
      *
      * @param rid The request ID for which to find previously delivered data.
      * @return previously delivered data when found.
      */
-    @Nullable
-    private Delivered retrieveDeliverable(final long rid) {
-        Delivered result = null;
+    @Nonnull
+    private Optional<Delivered> retrieveDeliverable(final long rid) {
         synchronized (sentElements) {
             for (Delivered delivered : sentElements) {
                 if (delivered.getRequestID() == rid) {
-                    result = delivered;
-                    break;
+                    return Optional.of(delivered);
                 }
             }
         }
-        return result;
+        return Optional.empty();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
```diff
@@ -503,26 +511,24 @@ public void forwardRequest(HttpBindBody body, AsyncContext context)
...
-                        connection.close();
-                    }
+                    // Note that the old connection is removed here, but the new connection will not be added back before the mutex is released. This leaves
+                    // room for another thread to try and consume a connection before this connection has been restored. This should be safe, as the consumer
+                    // should not be allowed to consume a connection with a RID that, compared to the previously consumed RID, leaves a 'gap'. In that case,
+                    // the to-be-delivered data is expected to be queued, which will be picked up as soon as this connection (for which the RID 'fills the gap',
+                    // is being processed by org.jivesoftware.openfire.http.HttpSession.processConnection.
+                    iter.remove();
+
+                    assert !queuedConnection.isClosed();
```
GregDThomas, line 527: Is assert the right thing to do here? Perhaps a log statement with some tidy up?

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
```diff
@@ -687,41 +698,44 @@ private void processConnection(@Nonnull final HttpConnection connection, @Nonnul
...
         synchronized (connectionQueue) {
+            // Note that this queue will automatically order its entities.
             connectionQueue.add(connection);
 
-            // Order the connection queue, and determine if the sequence of consecutive request IDs can be updated.
-            connectionQueue.sort(connectionComparator);
+            final Iterator<HttpConnection> iter = connectionQueue.iterator();
+            while (iter.hasNext()) {
+                final HttpConnection queuedConnection = iter.next();
+                assert !queuedConnection.isClosed();
```
GregDThomas, line 712: Another assert?

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
```diff
@@ -740,39 +754,43 @@ private void processConnection(@Nonnull final HttpConnection connection, @Nonnul
...
-            while( openConnections.size() > hold) {
+            while (!connectionQueue.isEmpty() && connectionQueue.size() > hold) {
                 if (Log.isTraceEnabled()) {
-                    Log.trace("Stream {}: releasing oldest connection (rid {}), as the amount of open connections ({}) is higher than the requested amount to hold ({}).", streamid, rid, openConnections.size(), hold);
+                    Log.trace("Stream {}: releasing oldest connection (rid {}), as the amount of open connections ({}) is higher than the requested amount to hold ({}).", streamid, rid, connectionQueue.size(), hold);
+                }
+                final HttpConnection openConnection = connectionQueue.peek();
+                assert openConnection != null;
+                if (openConnection.getRequestId() > lastSequentialRequestID) {
+                    break; // There's a gap.
```
GregDThomas, line 782: break or continue? Or perhaps a better comment as to what a gap actually means / is interpreted

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
```diff
@@ -804,26 +822,19 @@ private enum OveractivityType {
...
      * @param connection the new connection.
      * @throws HttpBindException if the connection has violated a facet of the HTTP binding
      *         protocol.
      */
     private void checkOveractivity(HttpConnection connection) throws HttpBindException {
-        int pendingConnections = 0;
+        int pendingConnections;
         OveractivityType overactivity = OveractivityType.NONE;
 
         synchronized (connectionQueue) {
```
GregDThomas, line 836: Not sure this `synchronized` is necessary, given the scope of the containing block?

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
```diff
@@ -979,22 +991,22 @@ private void closeSession(@Nullable final StreamError error) {
             // open connections in this method.
             synchronized (connectionQueue) {
                 boolean isFirst = true;
-                for (HttpConnection toClose : connectionQueue) {
+                while (!connectionQueue.isEmpty()) {
+                    final HttpConnection toClose = connectionQueue.poll();
+                    assert !toClose.isClosed();
```
GregDThomas, line 996: assert?

GregDThomas (commented):
A couple of mostly minor comments (not sure about the use of assert; I'd rather see some logging/tidy up, but that's just me).

I've also just spotted that there is a ThreadLocal XMPP parser in this class (unrelated to this change). Wonder if that's a potential memory leak?

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
guusdk: Recently, I have started leaning towards failing fast(er) than what we used to do. The connection at this stage is really supposed to be closed. If it's not, that is a bug in code. My current way of thinking is that I'd rather assert this state (which has the benefit of documenting the intent of this code) than write additional code that kicks in only when the preceding code has a bug. That feels like putting on bandages, just in case an injury will occur.

Over time, it's hard to distinguish between the "just in case" code, and the code that is actually supposed to implement the desired functionality. I'm hoping that with using assertions, this will be prevented.

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
guusdk: Break. The rationale is provided in comments earlier in the same method, but if you missed that, than others will too. I have modified this comment to explicitly refer to the earlier comments.

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
guusdk: You might be right. As I'm working on this PR, my plan was to incrementally progress to a point where we can hopefully significantly reduce the usage of locks. Up until that time, I've not yet spent to much thought-cycles on evaluating the existing locks, as that's very much on my to-do list for this PR. Work in progress!

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
guusdk: Same rationale as above. The statement asserts the same condition: the queued connections should not be in a 'closed' state.

guusdk:
I've now added a commit that hopefully fully resolves the issue. More reviewing would be very welcome. :)

I have not included fixes for all issues found during the course of this implementation, as I didn't want to cloudy the waters more than that I'm already doing. I'll create separate PRs for those, I think.

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
```diff
@@ -909,41 +926,79 @@ private void deliver(@Nonnull final Deliverable stanza) {
...
-            if (!delivered) {
-                Log.debug("Unable to immediately deliver a stanza on session {}. It is being queued instead).", getStreamID());
-                pendingElements.addAll( deliverable );
+    /**
+     * Returns a connection that can be used to deliver data to the client, if such a connection is currently available.
+     *
+     * @return A connection that is ready to be responded to.
+     */
+    @Nullable
+    private HttpConnection getConnectionReadyForOutboundDelivery()
```
GregDThomas, line 981: Might be tempted to use an Optional here, to force handling of the 'not present' case

GregDThomas (approved):
Looks fine; one trivial comment but from what I can see looks like it's a good clean up of the existing code.

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
guusdk: I never learn to use those. Makes more sense than passing around null values. :)

I have added a commit to this PR that makes the change (found one more method where it applied).

PR #2055, opened 2022-06-23
OF-2455: Make websockets stand out more in admin console

guusdk:
Openfire supports both BOSH and websockets. The latter is ofthen the preferred choice (or should be), as it generally is a less complex implementation.

The Openfire admin console does not really advertise that websockets are supported. This makes some people miss out on the fact that it is usable, making them go with a less optimal solution.

This commit renames the 'http bind' admin console pages to 'web binding', and explicitly mentions websocket next to BOSH, as being an option. With the exception of new API methods to expose the websocket endpoints, this is a cosmetic change.

image

image


akrherz (approved):

Fishbowler:
Rebased.

Fishbowler (approved):

PR #2058, opened 2022-06-23
server can reply with message error stanza with error information

aakashkumarjee:
Currently, If we reject packet from any Message Interceptor, we can't set error information. As we have access of packet but in PacketRejectionException handling, we create a copy of reply. 
So added this feature, so that if packet contains any error info added by any Intercepted, it should be saved in reply packet too and sent to client.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/MessageRouter.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/MessageRouter.java
@@ -179,6 +179,10 @@ public class MessageRouter extends BasicModule {
                 reply.setTo(session.getAddress());
                 reply.setFrom(packet.getTo());
                 reply.setType(packet.getType());
+                if(packet.getError()!= null){
+                    PacketError packetError = packet.getError();
+                    reply.setError(new PacketError(packetError.getCondition(), packetError.getType(), packetError.getText()));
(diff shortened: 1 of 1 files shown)


guusdk (changes requested):
Thank you for your contribution! I like the feature that you have added, but I would like to see a change.

Instead of modifying the original packet to relay the error from the filter to the reply, please use the `PacketRejectedException`. It can be modified to include a `PacketError`, which, if present, can be added to the reply.

aakashkumarjee:
Hey, Thanks @guusdk  for the suggestion. 
Although, I don't see the reason, why can't we add an error to the original packet? As I can see in other cases like "`not_authorized`" etc are directly been added to the original Packet. 

And also, I haven't found an easy way to add `PacketError` in `PacketRejectionException` other than extending, modifying the original class, or adding errors in String format inside the Exception. 
And after this catch statement, our original packet object would get destroyed by GC as there's no use for its reference after this block, right? 
Or am I missing something here?

guusdk:
I do not like modifying the stanza that is being rejected to specify what error should be returned. That's an undesirable hack, in my opinion.

I have created #2310 to replace this PR. In that PR, I have made the modifications that I suggested (as of now, they're untested).

PR #2064, opened 2022-06-29
OF-2449: BOSH should handle invalid pause requests

guusdk:
If a BOSH client requests a pause that's longer than the maximum value, error out.


guusdk:
On second thought: this could break existing clients. Lets make it configurable.

akrherz (approved):

PR #2067, opened 2022-07-05
OF-2464: Admin console needs not define a locale

guusdk:


Fishbowler (approved):
This is a fallback for a fallback. I don't expect this is ever needed or used.

PR #2072, opened 2022-07-15
OF-2470: Protect against NPE from IQ stanza with no child elements

Fishbowler:
Claude and Dele isolatedf a case where a Jitsi client sends an IQ Result stanza with no child elements, causing an NPE in a new feature introduced in Openfire 4.7.2.
This defends against it. I can't see any other instances of this risk in this class.

Cherrypick for 4.7


guusdk (approved):
IQ result stanzas without child elements are perfectly acceptable. The code should indeed guard against that.

PR #2073, opened 2022-07-19
fixes OF-2472: Pubsub node summary pages should be sortable

guusdk:
This adds 'sorting' functionality to the admin console pages that show pubsub node listings (for both the pubsub service, as well as each PEP service).

The changes are largely implemented in the generic, pre-existing ListPager implementation, in the hope that they are reusable for other pages.

Resulting page:
image


xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
```diff
@@ -45,10 +45,21 @@
...
+    private static final String REQUEST_PARAMETER_KEY_SORT_ORDER = "sortOrder";
+    private static final String REQUEST_PARAMETER_KEY_SORT_COLUMN_NUMBER = "sortColumnNumber";
     private static final String PAGINATION_FORM_ID = "paginationForm";
     private static final int DEFAULT_PAGE_SIZE = 25;
     private static final int[] PAGE_SIZES = {25, 50, 100, 1000};
 
+    /**
+     * Descending sort (ie 3, 2, 1...).
+     */
+    public static final int DESCENDING = 0;
```
GregDThomas, line 57: Perhaps an enum? Or given it only takes two values, would an Boolean suffice?

`private final Boolean sortAscending =` True, False or null (null implying don't care)?

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubsubNodeResultFilter.java
```diff
@@ -0,0 +1,318 @@
...
+{
+    private static final Logger Log = LoggerFactory.getLogger(PubsubNodeResultFilter.class);
+
+    // ############################################################
+    // Search order criteria
+    // ############################################################
+    /**
+     * Descending sort (ie 3, 2, 1...).
+     */
+    public static final int DESCENDING = 0;
```
GregDThomas, line 48: Wonder if this should be set to `ListPager.DESCENDING` ?

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
```diff
@@ -83,6 +96,21 @@ public ListPager(final HttpServletRequest request, final HttpServletResponse res
...
+     *
+     * @param request              the request object for the page in question
+     * @param response             the response object for the page in question
+     * @param items                the complete list of items to display on the page
+     * @param sortColumnNumber     the index of the column (0-based) of which values are to be ordered
+     * @param sortDescending       true if the column is to be in descending order, false for ascending order.
+     * @param filter               the filter to apply to the complete list
+     * @param additionalFormFields 0 or more input field identifiers (<strong>NOT form field names</strong>) to include in requests for other pages
+     */
+    public ListPager(final HttpServletRequest request, final HttpServletResponse response, final List<T> items, final Predicate<T> filter, final int sortColumnNumber, final boolean sortDescending, final String... additionalFormFields) {
```
GregDThomas, line 113: I wonder if `sortDescending `should be an `Boolean `rather than a `boolean`, to allow some pages that support sorting to also specify "I don't care" in the same constructor?

GregDThomas: (I can see this being useful on, e.g., the page that shows users, but for servers backed by large corporate LDAP servers they may not want to sort the whole lot unless they have to)

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
```diff
@@ -333,6 +373,18 @@ public String getPageFunctions() {
             .append("\t}\n")
             .append("\n");
 
+        sb.append("\tfunction toggleColumnOrder(sortColumnNumber) {\n")
+            // Orders based on a particular column, reversing the order if that column was already ordered on.
+            .append(String.format("\t\tvar formObject = document.getElementById('%s');\n", PAGINATION_FORM_ID))
+            .append(String.format("\t\tvar previousSortColumnNumber = formObject.%s.value || '0';\n", "sortColumnNumber"))
```
GregDThomas, line 379: These next few lines seem to use `String.format` unnecessarily - passing in a string constant that may as well just have been in the format string? should `"sortColumnNumber"` have been a `private static final String` definition and used elsewhere? Possibly `REQUEST_PARAMETER_KEY_SORT_COLUMN_NUMBER `(and similarly for the next few lines)

xmppserver/src/main/webapp/pubsub-node-summary.jsp
```diff
@@ -23,6 +22,9 @@
 <% webManager.init(request, response, session, application, out ); %>
 
 <%  // Get parameters
+    int sortOrder = ParamUtils.getIntParameter(request, "sortOrder", webManager.getPageProperty("pubsub-node-summary", "console.sortOrder", PubsubNodeResultFilter.ASCENDING));
```
GregDThomas, line 25: (probably not as part of this MR, but...)

I wonder if there could/should be static methods/classes with ListPager that will fetch the current "stuff" - filters, page number,  sort order, etc. off a submitted form - otherwise everyone using it will end up fetching basically the same information off - just from different field names depending on the exact usage.

Could/should have been written with the original code, but hindsight ...

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubsubNodeResultFilter.java
GregDThomas: (or even a `Boolean`, in line with my comments on `ListPager`

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
GregDThomas: (and if you don't like "tri state" Booleans, perhaps an Enum to avoid having to catch the ~4 billion invalid values for an integer?)

GregDThomas (commented):
A couple of things - nothing major, though - a good enhancement.

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
guusdk: I agree that this isn't the most elegant of solutions. I used this, as this is also how it is used in `SessionResultFilter`. The numeric value is what is being passed as a query parameter in the JSPs, which is convenient.

I am liking the clarity that using words like 'DESCENDING' or 'ASCENDING' give, when applied to code, which is why I prefer this over a boolean value. We could use an enum with the same effect, while also capturing the query param value. Maybe that's the better approach.

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubsubNodeResultFilter.java
guusdk: I have been struggling with this a bit. On the other hand, it would be great to use the same field for all usages, but on the other hand, the two are also distinct, in the sense that the implementation of sorting cannot easily be generic: it is specific to the types of the particular element being processed.

GregDThomas: Perhaps pass in a `Comparable<T>` to the `ListPager` to provide the sorting?

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
guusdk: For those, the overloaded constructors (the ones that does not have the `final int sortColumnNumber` and `sortDescending` parameters) are available. The implementation currently does not need a null-value, as 'I do not want ordering' is expressed by using an invalid column number value. That one might instead be a `null`, perhaps. However, as users of that particular overloaded constructors _want_ ordering of a column, I don't think it matters to much.

GregDThomas: Expanding on my thinking;

The first time you hit the users page, you wouldn't want sorting - to avoid large numbers of users throwing things off.

However, installations with small numbers of users may which to sort by name, which they could do by clicking on a column.

Rather than the code backing this relying on calling one of two constructors depending on if the sorting was selected or not, a single constructor with a nullable param might help. But perhaps YAGNI :)

xmppserver/src/main/java/org/jivesoftware/openfire/pubsub/PubsubNodeResultFilter.java
guusdk: I liked this idea, but then ran into issues trying to implement it.

The `ListPager` doesn't really do any sorting. We could make it do that generically by providing a Comparator to replace the ugly integer-based arguments that we now have, but that introduces another issue. Those integer-based arguments are used to populate a hidden form, which is used to transfer the entire 'state' of sorting/filtering/pagination to the next request.

I've been looking at how to make this all a bit prettier, but after every change, something else keeps failing. I might be missing something obvious, but I'm getting a bit frustrated with fixing one thing, to find two new issues pop up. I'm thinking of merging this as-is, leaving improvements to future PRs.

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
guusdk: The actual sorting is not done in ListPager, it's done by the caller. Currently, that defaults to a non-null value, but I assume that the vast majority of people prefer to have some kind of order.

We can optimize this later, but we're not introducing a new overhead here: at least the pubsub node stuff already had sorting applied. We are now only expanding the possible columns to sort on.

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
guusdk: Usage of `String.format` with constants predates my involvement here :) and to be honest, I like how it improves readability of the text in which the value is injected. The escaped javascript is difficult enough to read in itself. Doing more string concatting would not improve readability, I think.

guusdk: I definitely intended those strings to be replaced by the constants that were already defined. I'll push that change.

xmppserver/src/main/webapp/pubsub-node-summary.jsp
guusdk: You had me at 'probably not as part of this MR' ;)

akrherz (approved):

PR #2074, opened 2022-07-20
OF-2474: IP-based access control for the admin console

guusdk:
This adds an 'allow' and 'block' list of IP(v4) addresses, that controls access to the admin console.

A third property is added that can be used to configure if these lists should also be applied when accessing pages that are otherwise 'excluded' from access control by the AuthCheckFilter.

An admin console page has been added to easily configure the new properties.

image

### DIFF
--- a/xmppserver/src/main/webapp/system-admin-console-access.jsp
+++ b/xmppserver/src/main/webapp/system-admin-console-access.jsp
@@ -1,0 +1,173 @@
+<%@ page contentType="text/html; charset=UTF-8" %>
+<%@ page errorPage="error.jsp" %>
+<%@ page import="org.jivesoftware.util.StringUtils" %>
+<%@ page import="org.jivesoftware.util.CookieUtils" %>
+<%@ page import="org.jivesoftware.util.ParamUtils" %>
+<%@ page import="java.util.*" %>
+<%@ page import="java.util.regex.Pattern" %>
+<%@ page import="org.jivesoftware.admin.AuthCheckFilter" %>
+
+<%@ taglib uri="admin" prefix="admin" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/fmt" prefix="fmt" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn" %>
+
+<jsp:useBean id="webManager" class="org.jivesoftware.util.WebManager" />
+<% webManager.init(request, response, session, application, out ); %>
+
+<%  final Map<String, String> errors = new HashMap<>();
+
+    // Get parameters
+    final String blockedIPs = request.getParameter("blockedIPs");
+    final String allowedIPs = request.getParameter("allowedIPs");
+    final boolean ignoreExcludes= ParamUtils.getBooleanParameter( request, "ignore-excludes" );
+
+    boolean save = ParamUtils.getParameter(request, "save") != null;
+    boolean success = ParamUtils.getBooleanParameter(request, "success");
+
+    // Handle a save
+    Cookie csrfCookie = CookieUtils.getCookie(request, "csrf");
+    String csrfParam = ParamUtils.getParameter(request, "csrf");
+
+    if (save) {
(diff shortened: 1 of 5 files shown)


Fishbowler:
Rebased.

xmppserver/src/main/webapp/system-admin-console-access.jsp
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        if (errors.isEmpty()) {
+            AuthCheckFilter.IP_ACCESS_BLOCKLIST.setValue(blockedSet);
+            AuthCheckFilter.IP_ACCESS_ALLOWLIST.setValue(allowedSet);
+            AuthCheckFilter.IP_ACCESS_IGNORE_EXCLUDES.setValue(ignoreExcludes);
+
+            // Log the event
+            webManager.logEvent("Updated Admin Console access configuration.", "Blocklist = " + String.join(", ", blockedSet) +"\nAllowlist = " + String.join(", ", allowedSet)+"\nignore excludes = "+ignoreExcludes);
```
Fishbowler, line 83: Newlines don't show in the Audit Viewer. Semicolons instead?

image

Fishbowler: Perhaps use explicit quotes to show that the allowlist is empty in the above example?

Fishbowler:
```suggestion
            webManager.logEvent("Updated Admin Console access configuration.", "Blocklist = {" + String.join(", ", blockedSet) +"}, Allowlist = {" + String.join(", ", allowedSet)+"}, Ignore excludes = {"+ignoreExcludes+"}");
```

How about this? Are braces too unweildy for this kind of logging?

image

Fishbowler:
When trying to set an invalid IP address in the Admin console, the value returns the previous value rather than allowing me to correct it. Is this the same behaviour in the rest of the Admin console?

Fishbowler:
When using a proxy that passes the X-Forwarded-For, and `adminConsole.forwarded.enabled` is set, the page correctly shows the client IP address.

When the proxy address is added to the blocklist, the proxy (and its clients) are not blocked. Intentional behaviour?

xmppserver/src/main/webapp/system-admin-console-access.jsp
guusdk: I think that the braces are a good idea. I have add those.

As for the newlines: I've seen them used in different places, as a kind of convention. The details don't show unless they're opened, so I think those newlines should simply be rendered properly. I've raised https://igniterealtime.atlassian.net/browse/OF-2503 for this.

guusdk:
> When trying to set an invalid IP address in the Admin console, the value returns the previous value rather than allowing me to correct it. Is this the same behaviour in the rest of the Admin console?

This is the same behavior in the rest of the Admin Console, but we should do better. I've now pushed a change that will populate the input fields with the unsaved / erroneous input values.

I've also modified the error messages to be more explicit in the fact that none of the settings have been saved. This should rule out any misconception in that other values (that were not erroneous) might have been saved.

guusdk:
> When the proxy address is added to the blocklist, the proxy (and its clients) are not blocked. Intentional behaviour?

I'm not sure if I interpret what you're asking correctly. When a proxy is in play (and `adminConsole.forwarded.enabled` is enabled), the intentional behavior (which as far as I'm concerned can be up for debate), is that the blocklist and allowlist apply to the IP addresses of the clients _as seen by the proxy_ (the client's "public" address).

Do we need to worry about a rogue proxy server being used?

Fishbowler:
That makes sense. I've confirmed that the help text shown on the page uses the correct client IP when `adminConsole.forwarded.enabled` is set to True. Is there an argument for making that property true by default for new installations?

Fishbowler (approved):

Fishbowler:
On the topic of a rogue proxy, that's an industry-wide problem, and not something we should try to solve, I think

guusdk:
> Is there an argument for making that property true by default for new installations?

Maybe - the same could be said for the very similar property that's used by BOSH. The downside is that it would, by default, allow clients to 'spoof' their IP by using a X-Forwarded-For header.

If we do decide to change the default, let's take that on under a different issue number.

> On the topic of a rogue proxy, that's an industry-wide problem, and not something we should try to solve, I think

Makes sense.

PR #2075, opened 2022-07-26
OF-2479: Allow clients that do websockets without the required XMPP framing

guusdk:
This basically is a hack to get Tsung to connect to Openfire over websockets.

The functionality introduced here is switched off by default. If enabled, it replaces `stream` element names with `open` and `close` where appropriate, to conform to 'XMPP framing' as described in https://datatracker.ietf.org/doc/html/rfc7395#section-3.3


guusdk:
The Tsung bug that this new feature in Openfire works around is tracked as https://github.com/processone/tsung/issues/164

GregDThomas:
Is there a follow-on ticket to remove this code when the Tsung bug is fixed? 

(you can tell I spend less time coding, now, right?)

guusdk:
I didn't create such an issue. I do not anticipate that this code will ever be taken out (at least not for the reason of Tsung having fixed that issue). I highly doubt that the Tsung bug will ever be fixed, after talking to a developer in that general vicinity. Even if it does, that fix needs a Tsung release (which has not happened in a while), and even then, people need to adopt the new release. The change in Openfire is fairly small, and very localized, and disabled by default. I'm happy to leave this in.

akrherz (approved):

PR #2076, opened 2022-07-27
OF-2473: Prevent websocket deadlock

guusdk:
Jetty's websocket implementation invokes `@OnWebSocketError` and `@OnWebSocketClose` annotated methods when certain erroneous conditions occur. These invocations happen synchronously.

As a result, a 'close' operation is called synchronously to the processing of the original request, which can happen under guard of a mutex. This leads to the 'close' implementation being called while a mutex is held that is intended to be held only while inbound data is being processed. This has lead to deadlocks.

An example scenario is when a websocket request arrives that asks to close the connection. Sometimes (presumably if the per immediately disconnects the connection, rather than waiting for an answer), returning a response fails ('EOF'). This then triggers error handling.

This commit makes the `@OnWebSocketError` and `@OnWebSocketClose` methods be asynchronous, to ensure that its processing is decoupled from the inbound data that might have tirggered their invocation.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
@@ -25,6 +25,7 @@ import org.eclipse.jetty.websocket.api.Session;
 import org.eclipse.jetty.websocket.api.annotations.*;
 import org.jivesoftware.openfire.*;
 import org.jivesoftware.openfire.entitycaps.EntityCapabilitiesManager;
+import org.jivesoftware.openfire.http.HttpBindManager;
 import org.jivesoftware.openfire.multiplex.UnknownStanzaException;
 import org.jivesoftware.openfire.net.SASLAuthentication;
 import org.jivesoftware.openfire.net.SASLAuthentication.Status;
@@ -96,7 +97,8 @@ public class XmppWebSocket {
     @OnWebSocketClose
     public void onClose(int statusCode, String reason)
     {
-        closeSession();
+        // Handle asynchronously, to prevent deadlocks. See OF-2473.
+        HttpBindManager.getInstance().getSessionManager().execute(this::closeSession);
     }
 
     @OnWebSocketMessage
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
```diff
@@ -125,14 +127,21 @@ public void onTextMethod(String stanza)
...
-        closeStream(new StreamError(StreamError.Condition.internal_server_error));
-        try {
-            if (wsSession != null) {
-                wsSession.disconnect();
+
+        // Handle asynchronously, to prevent deadlocks. See OF-2473.
+        HttpBindManager.getInstance().getSessionManager().execute(() -> {
+            try {
+                Log.debug("Attempting to close session on which an error occurred.");
+                closeStream(new StreamError(StreamError.Condition.internal_server_error));
```
GregDThomas, line 135: Does closeStream() do the right thing if it's already been closed - now things are no longer synchronous need to be more careful over the processing here

xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
```diff
@@ -96,7 +97,8 @@ public void onConnect(Session session)
     @OnWebSocketClose
     public void onClose(int statusCode, String reason)
     {
-        closeSession();
+        // Handle asynchronously, to prevent deadlocks. See OF-2473.
+        HttpBindManager.getInstance().getSessionManager().execute(this::closeSession);
```
GregDThomas, line 101: You'll need to wrap this in a try/catch to ensure any errors during processing are properly logged and not lost in the ether. Either in this call, or in closeSession itself.

xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
guusdk: I think you're right. By making things more asynchronous, more guards are probably in order. I've added some `synchronous` blocks.

Sideways-related: The code is the exact same snippet that was there before, now only wrapped in a Runnable. Even the old code had some apparent duplication going on (`closeStream` shuts down the session, while that's also done a few lines further down?).

xmppserver/src/main/java/org/jivesoftware/openfire/websocket/XmppWebSocket.java
guusdk: Oh, absolutely! Losing errors like that can be very confusing. I've modified the asynchronous call. Thanks!

GregDThomas (approved):
LGTM 👍

akrherz (approved):

PR #2078, opened 2022-07-28
OF-2471: Display GroupJID human readable name in Audit viewer

guusdk:
When modifying MUC permissions for a Group, display the Group name, not its base-32 identifier.

image

### DIFF
--- a/xmppserver/src/main/webapp/muc-room-affiliations.jsp
+++ b/xmppserver/src/main/webapp/muc-room-affiliations.jsp
@@ -84,53 +84,53 @@
             errors.put("userJID","userJID");
         }
 
-        if (errors.size() == 0) {
+        if (errors.isEmpty()) {
             try {
-                ArrayList<String> memberJIDs = new ArrayList<String>();
+                List<JID> memberJIDs = new ArrayList<>();
                 if (userJID != null) {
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/webapp/muc-room-affiliations.jsp
```diff
@@ -84,53 +84,53 @@
...
                         String domain = webManager.getXMPPServer().getServerInfo().getXMPPDomain();
+                        addMe = new JID(username, domain, null);
                         userJID = username + '@' + domain;
                     }
                     else {
                         String username = JID.escapeNode(userJID.substring(0, userJID.indexOf('@')));
-                        String rest = userJID.substring(userJID.indexOf('@'), userJID.length());
+                        String rest = userJID.substring(userJID.indexOf('@') + 1);
+                        addMe = new JID(username, rest.trim(), null);
                         userJID = username + rest.trim();
```
Fishbowler, line 103: Delete the userJID variable now? It doesn't appear to be used

guusdk: I initially had that, but it is used somewhere way down the page. I actually reinstated it to prevent an issue there.

PR #2091, opened 2022-08-02
OF-400: Log an error when saving a MUCRoom with a subject > 100 chars

Fishbowler:
In addition to #2086

Adds another clear logging point when the subject is too long. This might not be all of them, but this is called by the Openfire REST API Plugin.


Fishbowler:
This isn't the way. If we can't honour the client request, we should do so more noisily.

PR #2100, opened 2022-08-08
OF-2492 - mvnw fixes

Fishbowler:
Fixes OF-2492 by making mvnw executable
Whilst I was there, updated wrapper scripts and properties (via `mvn wrapper:wrapper`)


guusdk:
Is there value in updating our CI scripts to use the wrapper, rather than an externally provided version of Maven to run the builds?

Fishbowler:
That's probably a good test of our code being buildable away from CI :+1:

PR #2103, opened 2022-08-26
OF-2495: Websocket onError handler prevents earlier data to be processed

guusdk:
This commit ensures that the handling of the onError event is done in the same thread pool as that is using data. This should help ensure that those events are handled in roughly the same order. This in turn prevents data that has been supplied previously from going unprocessed, because the onError handler is executed 'early'.


guusdk:

PR #2105, opened 2022-09-01
OF-2499: No longer serialize MUC history with room

guusdk:
As suggested in a 'TODO' comment in code, this replaces the MUC history that is stored inline with each MUC room, with a cache lookup.

This change intends to result in a performance boost, as retrieving/storing MUC rooms from a cache (and serializing them between cluster nodes) no longer includes the history. The data associated with history is typically very large (~50kb for 25 messages), but changes a lot less often than the room itself.


guusdk:
This commit introduces an API change, which make servers running this code incompatible (in a cluster) with servers running older code. That should be a reason to not include it in a patch release, but in a major release instead. However: some Openfire domains suffer badly from this issue. Waiting for a major release isn't much of an option for them. With that in mind, I'd like to add this to the next patch release in the 4.7 branch.

guusdk:

PR #2106, opened 2022-09-08
OF-2502: Do not iterate over clustered cache when calculating stats

guusdk:
When collecting statistics, two MUC methods iterate over caches. That is inefficient in itself, but the cache can also be clustered. If that's the case, then a lot of data serialization, which is a waste of resources.

This commit introduces a new cache in which statistics can be stored. This way, the statistics need not be recalculated every time, saving resources at the expense of potential inaccuracy.


guusdk:

PR #2109, opened 2022-09-13
OF-2505: BOSH: add back-pressure to network IO when processing can't keep up

guusdk:
For BOSH and websocket connections, this commit introduces a new feature (disabled by default) that allows Openfire to be configured to apply 'back pressure' that could be useful in a scenario where the network IO thread pool is outpacing the processing thread pool.

By enabling the feature using `xmpp.httpbind.worker.backpressure-enabled` (and setting the queue to a limited value with `xmpp.httpbind.worker.queue-capacity`), the queuing of new tasks by the network IO thread pool blocks when the queue is at capacity.


akrherz:
@guusdk Did you get any feedback on your 4.7 branch build with this change included?

guusdk:
I did - although it seemed to work as designed, the net effect of the change wasn't what we were hoping for in the very specific load test that was being used. We seem to be running into race conditions / locking scenario.

There still might be merit in this change, even if we're not going to use it in the specific use-case that I had in mind.

guusdk:
I think this at the very least needs a modification. The enum `ThreadPoolExecutorRejectionPolicy` that is added in this PR is not used at all. I think that enum is an artifact of an earlier attempt to get this functionality implemented. It can be removed.

Fishbowler:
Converted to draft whilst it awaits changes.

PR #2110, opened 2022-09-14
OF-2475: Allow 'future' users

guusdk:
Openfire supports mechanisms (which are unlikely to be used by default) that allow User representations to be created on an ad-hoc basis: typically when the user successfully authenticates. This leaves some room for users to be interacted with, prior to Openfire realizing that the user actually exists.

This commit introduces a new feature that allows stanzas addressed to users that could be potential future users to be handled as if that user exists, but is offline.

The feature can be enabled using the new system property `usermanager.future-users.enable`


Fishbowler:
Do I read the implementation correctly, that this not necessarily need be implemented in an abstracted/federated authentication setup, but could be enabled in a purely local fashion? i.e. could we:
* enable it on the ignite server
* send messages to `test.person@igniterealtime.org` 
* manually create the user 
* see those messages delivered to the user when they log in for the first time?

Fishbowler:
Confirmed locally. That's really quite nifty!

Fishbowler:
Some interesting thoughts on how this interacts with other functionality that I've started considering.
What happens when the Offline Message Store is set to bounce? Is that even a use case?
MUC invitations are messages, so they should work like any other, right?

guusdk:
> Do I read the implementation correctly, that this not necessarily need be implemented in an abstracted/federated authentication setup, but could be enabled in a purely local fashion? i.e. could we:
> 
> * enable it on the ignite server
> * send messages to `test.person@igniterealtime.org`
> * manually create the user
> * see those messages delivered to the user when they log in for the first time?

I believe so, yes!

> Some interesting thoughts on how this interacts with other functionality that I've started considering.
> What happens when the Offline Message Store is set to bounce? Is that even a use case?
> MUC invitations are messages, so they should work like any other, right?

For all (most) intents & purposes, a 'future user' would be equal to a regular, offline user. Ideally, we don't see a difference between the two. I would not be surprised though if there are some. We might be able to improve on some things that we find, while others, we might deem reasonable enough to fail to not warrant further development.

guusdk:
Should this feature go on the admin console somewhere?

Fishbowler:
What happens in a large scale deployment with this enabled?
Does the Offline Message Cache size need to be Unlimited in this setup?

Fishbowler:
Disco to a non-existent user gives a success with an empty disco reponse
VCard request to a non-existent user gives a success with an empty vcard-tmp
Messages to a non-existent user are stored and delivered once the user comes online
Ditto contact requests
Ditto MUC invitations
If Monitoring is installed and configured to store 1-to-1 messages, messages to the non-existent user can be queried via MAM2 by the user, once created

Fishbowler:
Spent a while considering Admin Console - if you're gonna add it, I _think_ that `/reg-settings.jsp` is the right place

guusdk:
> Spent a while considering Admin Console - if you're gonna add it, I _think_ that `/reg-settings.jsp` is the right place

Agreed. I have now added a section to that page that should allow an admin to easily toggle support for this new feature.

Fishbowler (approved):
Having already looked at the rest, the Admin console changes look good too. Text is clear. Sets and unsets and reflects the system property state.

PR #2112, opened 2022-09-20
OF-2508: Nodeprep room name when creating through Ad-hoc command

guusdk:
Openfire provides an ad-hoc command that can be used to create a room. Openfire must ensure that room name values provided through that command are properly normalized (nodeprepped).

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/commands/admin/muc/CreateMUCRoom.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/commands/admin/muc/CreateMUCRoom.java
@@ -93,6 +93,7 @@ public class CreateMUCRoom extends AdHocCommand {
             note.setText("Room name must be specified.");
             return;
         }
+        roomname = JID.nodeprep(roomname);
         JID admin = admins.iterator().next();
 
         boolean isPersistent;


guusdk:
I'm contemplating if we should add a type-safe class, to ensure that all room names that are passed around internally are properly normalized.

GregDThomas:
Always yes.

guusdk:
It is nice how either mentioning your name or 'type-safety' summons you. ;)

The downside of that is that it's a major rewrite - and if we're going down that path, we might as well provide type-safe classes for each JID-part through Tinder, which would even be a larger undertaking.

GregDThomas:
> It is nice how either mentioning your name or 'type-safety' summons you. ;)

Just don't say my names five times in front of the mirror ;)

> The downside of that is that it's a major rewrite

You don't have to implement it all at once. For example, your strongly typed room name should do the "right thing" when called with `toString()`. So use `name.toString()` in the code that has yet to be migrated, perhaps with a @deprecated tag on the toString() (I know) to highlight that the usage needs updating.

guusdk:
If I understand you correctly, that would require me to pass an entire JID instance. We typically just want the node-part.

I'm tempted to put this in as-is (at least for the 4.7.x branch), perhaps scheduling a larger rewrite for later.

Fishbowler (approved):

PR #2113, opened 2022-09-21
OF-2476: Trunking/Gateway support

guusdk:
Openfire can be used to accept connections on behalf of remote domains, acting as a 'trunk' or 'gateway' for/to that domain. This feature can be useful when centralized orchestration of data (monitoring, filtering) is needed. This commit adds basic support and documentation for this feature.

To enable the functionality, all participating servers need to have specific configuration. This is explained in documentation that is added in this commit.


dwd (approved):
Nice docs! I think that if Openfire sees an inbound S2S for domain X, it'll look for a matching cert in its identity store, won't it? So some of the TLS caveats, at least for inbound, might benefit from expansion.

Similarly, I wonder if dnsOverride could be easily expanded to support an identity name, which might be the simplest option for allowing some form of TLS authentication?

documentation/trunking.html
```diff
@@ -0,0 +1,169 @@
...
+            <h2 id="setup">Setup</h2>
+            <p>
+                The essence in getting an Openfire instance to act as a Trunking entity in your network is to have all other XMPP domains connect to this server, when they intend to connect to another remote domain. The Trunking entity needs to be configured to accept
+                stanzas for the target domain. This is typically achieved through DNS manipulation. It will then process the data, and pass it on to the target domain.
+            </p>
+
+            <p>
+                The server that is sending the message will attempt to establish a server-to-server connection with the target server. The sending server perform DNS lookups based on the domain name of the target server. The results of this lookup must be manipulated
+                in such a way that the resulting IP address is that of the trunking server. Openfire can facilitate such DNS manipulation, as it allows DNS lookups to be skipped in favor of preconfigured hard-coded values. To configure such values, the
+                property <tt>dnsutil.dnsOverride</tt> can be used.
```
Fishbowler, line 78: Clarify that this applies when Openfire is used as the _sending_ server. 
I know most orgs that might have this use case are likely to have a homogenous estate, but worth calling it out.

documentation/trunking.html
```diff
@@ -0,0 +1,169 @@
...
+            </p>
+            <p>
+                The expected format of the value of the <tt>dnsutil.dnsOverride</tt> property is <code>{HOST,IP:PORT}</code>, where multiple entries are separated by a comma. Note that whitespace is not allowed, as shown in <a href="#example-dnsutil-propvalue">example 1</a>:
```
Fishbowler, line 81: Word reshuffling.

```suggestion
                The expected format of the value of the <tt>dnsutil.dnsOverride</tt> property is a comma-sparated list of <code>{HOST,IP:PORT}</code>. Note that whitespace is not allowed, as shown in <a href="#example-dnsutil-propvalue">example 1</a>:
```

documentation/trunking.html
```diff
@@ -0,0 +1,169 @@
...
+
+            <ol>
+                <li>
+                    <p>
+                        Openfire at <tt>trunk.example.com</tt> needs to be configured to allow trunking to both XMPP domains. To achieve this, it's property <tt>xmpp.gateway.domains</tt> is configured to use this value: <code>saturn.example.org,mercury.example.net</code>
+                    </p>
+                </li>
+                <li>
+                    <p>
+                        When one of its users of <tt>saturn.example.org</tt> sends a stanza to one of the users of <tt>mercury.example.net</tt>, <tt>saturn.example.org</tt> will attempt to establish a server-to-server connection to <tt>mercury.example.net</tt>. The objective here is to
```
Fishbowler, line 120:
```suggestion
                        When a user of <tt>saturn.example.org</tt> sends a stanza to one of the users of <tt>mercury.example.net</tt>, <tt>saturn.example.org</tt> will attempt to establish a server-to-server connection to <tt>mercury.example.net</tt>. The objective here is to
```

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketReader.java
```diff
@@ -461,8 +463,12 @@ private boolean isHostUnknown(String host) {
...
             return false;
         }
-        // Check if the host matches a subdomain of this host
-        return !routingTable.hasComponentRoute(new JID(host));
+        if (routingTable.hasComponentRoute(new JID(host))) {
+            // Check if the host matches a subdomain of this host
+            return false;
+        }
+
+        return !JiveGlobals.getListProperty("xmpp.gateway.domains", Collections.emptyList()).contains(host);
```
Fishbowler, line 471: Why JiveGlobals rather than a SystemProperty?

Fishbowler, line 471: How would this work when the stanza is for a service domain, e.g. conference or pubsub?

documentation/trunking.html
guusdk: I've added another sentence to call that out.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketReader.java
guusdk: Interesting thought. This is something that we should test for.

The problem might also pop up when resolving the IP address for the remote XMPP host that is serving that component.

I assume that most setups will not have DNS entries for those component domains (although they might have, when they do trickery to get Let's Encrypt certificates generated!)

This is what I think will mostly happen: When there are no DNS SRV or A entries for the component domains, then the lookup for XMPP-service-to-Server-hostname should fall back from using the XMPP domain of `conference.example.org` (no DNS SRV record) to the 'parent' XMPP domain of `example.org` (which does have a DNS SRV record), which then resolves to a hostname, and ultimately, an IP address.

When there are DNS entries for the component domains (A and/or SRV), then I'm not sure what happens.

Should we require administrators to add those to the `xmpp.gateway.domains` and/or `dns.override`, or can we, safely, facilitate things in code?

guusdk: Lazy. I will change this.

guusdk:
> Nice docs! I think that if Openfire sees an inbound S2S for domain X, it'll look for a matching cert in its identity store, won't it? So some of the TLS caveats, at least for inbound, might benefit from expansion.
> 
> Similarly, I wonder if dnsOverride could be easily expanded to support an identity name, which might be the simplest option for allowing some form of TLS authentication?

I'd like to be able to achieve this. However, for Openfire's EXTERNAL mechanism to become compatible, some work needs to happen (it currently strictly requires that the authzid equals the stream header). I have made a PR with what I think is the required preparation in #2118. I'd love your feedback on that @dwd.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketReader.java
dwd: Openfie used to, years ago, "walk the tree" to find DNS SRV records, but that has various risks when a parent domain injects records. If there's no conference.example.org DNS records, most servers won't connect to it. That's OK.

Fishbowler:
Having created a containerised test environment, and applied the correct config following the docs, I get the following error on the gateway server, preventing any server from talking "through" the gateway:

```
2022.10.10 21:36:55 WARN [Server SR - 1986725953]: org.jivesoftware.openfire.net.SocketReader - Closing session due to incorrect hostname in stream header. Host: xmpp1.localhost.example. Connection: org.jivesoftware.openfire.net.SocketConnection@33699048 socket: Socket[addr=/172.50.0.200,port=35132,localport=5269] session: null
2022.10.10 21:38:31 WARN [Server SR - 1205336629]: org.jivesoftware.openfire.net.SocketReader - Closing session due to incorrect hostname in stream header. Host: xmpp3.localhost.example. Connection: org.jivesoftware.openfire.net.SocketConnection@4095658f socket: Socket[addr=/172.50.0.100,port=34426,localport=5269] session: null
```

The setup is 3 servers, named (boringly) xmpp1, xmpp2, xmpp3, where 2 is the gatway through which 1 and 3 can communicate. Config is in https://github.com/Fishbowler/openfire-docker-compose/commit/5c4f22d74dc7044e69da71125fd4f3bb8985a6e4, and its parent commit (https://github.com/Fishbowler/openfire-docker-compose/commit/f2d0380333a962352479b0658fdd21abe351cc17) gives the fuller layout.

That error should only be raised if the isHostUnknown returns true, which should only happen if the domain isn't enabled for trunking. Will add some debug lines to work out what's going on....

documentation/trunking-guide.html
```diff
@@ -0,0 +1,169 @@
...
+
+            <div style="text-align: center;">
+                <figure id="example-dnsutil-propvalue" style="margin: 2em;">
+                    <code style="padding: 10px; border: 3px solid #e7eaee; border-radius: 5px">{saturn.example.org,198.51.100.4:5269},{jupiter.example.com,198.51.100.4:5269}</code>
+                    <figcaption style="margin-top: 1.5em;">Example 1: <tt>dnsutil.dnsOverride</tt> property value, mapping two hostnames to the same IP address and port.</figcaption>
+                </figure>
+            </div>
+
+            <p>
+                To enable the trunking functionality in Openfire, the property <tt>xmpp.gateway.domains</tt> needs to be populated with a (comma-separated) list of XMPP domain names of remote XMPP domains that the Openfire server will forward data to.
```
Fishbowler, line 92:
```suggestion
                To enable the trunking functionality in Openfire, the property <tt>xmpp.gateway.enabled</tt> needs to be set to <code>true</code>, and the property <tt>xmpp.gateway.domains</tt> needs to be populated with a (comma-separated) list of XMPP domain names of remote XMPP domains that the Openfire server will forward data to.
```

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
```diff
@@ -643,6 +643,10 @@ private boolean isHostUnknown(String recipient) {
         if (host_unknown && recipient.contains(domainPair.getLocal())) {
             host_unknown = !routingTable.hasComponentRoute(new JID(recipient));
         }
+
+        if (host_unknown) {
+            host_unknown = Trunking.isTrunkingEnabledFor(recipient);
+        }
```
Fishbowler, line 649: I think this logic got inverted when the switch to SystemProperties happened

```suggestion
        if (host_unknown) {
            host_unknown = !Trunking.isTrunkingEnabledFor(recipient);
        }
```

Fishbowler:
With the above changes applied locally, I'm down to boring TLS rejection errors, which just need configuring or disabling, but configuring seems like the better route :+1:

guusdk:
That negation bug was embarrassing. Thanks for weeding that one out.

The TLS exceptions are not unexpected (unless you would be using wildcard certificates or at least certificates that cover all domains). I do expect dialback to kick in (and succeed) after TLS EXTERNAL failed. Did that happen?

As an aside: I think that the default configuration of Openfire might disallow TLS EXTERNAL for s2s completely, as the 'mutual authentication' setting appears to be 'disabled' by default. That will prevent the remote certificates from being verified, which precludes EXTERNAL. I'm going to look into this more.

image

Fishbowler:
Rebased

Fishbowler:
Following the rebase, I've got a successful test in the containerised environment using /server-connectiontest.jsp from node1 to node3 and vice-versa, via node2 as the gateway.

Fishbowler:
I've tested messaging, and tested MUC across the gateway (by adding the conference subdomain to the list of gateway domains), and I've tested using the Content Filter plugin on the gateway server to prevent messages from crossing the gateway.

I've only tested using Openfire's own self-signed certs, but I think certificate trust is beyond the scope of this change.

Fishbowler (approved):

guusdk:
I have raised a new tracker to add SASL EXTERNAL support: https://igniterealtime.atlassian.net/browse/OF-2532

PR #2116, opened 2022-09-28
Problem with transferring 4 or more files

Nenormatica:
The openfire server has a dedicated ip address. All clients connect from the Internet. Everything works fine except file transfer. When 4 or more files are transmitted, 1 or 2 files arrive. 
And the rest are in the transfer status. 
And it happens that all the files just weigh in the transfer queue. 
I use File Transfer Proxy Settings. I specified the external ip in the settings. 
But it didn't work for me. What do I need to do so that files are transferred without problems? 

I'm really desperate.

[openfire (6).log](https://github.com/igniterealtime/Openfire/files/9666483/openfire.6.log)


akrherz:
Please create a [forums post](https://discourse.igniterealtime.org)

PR #2118, opened 2022-10-04
OF-2514 & OF-2513: Prepare for non-default authzid on S2S SASL EXTERNAL

guusdk:
This PR contains two related, yet distinct improvements that both intend to facilitate sending a different-than-default authzid when using the SASL EXTERNAL mechanism for S2S (in support of Trunking Support, #2113).

This fixes in this PR:
- Ensure that Openfire's SASL implementation can differentiate between an empty and a missing initial response (OF-2514)
- Allow the initial response used for SASL EXTERNAL to be empty (something else than the default authzid) (OF-2513).
 
With this merged, the intend is to add another change (perhaps as part of #2113) that allows to use a different authzid altogether.

Please refer to the commit messages for a detailed description of each change.


guusdk:
Force pushed a fix for a bug that was in the original commit.

Added unit tests, that depend on the changes in https://github.com/igniterealtime/Openfire/pull/2149 (this PR will need to be rebased).

guusdk:
Rebased.

guusdk:
Rebased. The fix for OF-2513 is largely redundant, as the same problem has already been addressed by the fix for OF-2639.

PR #2119, opened 2022-10-04
OF-2516: Show session TX/RX instead of prio on admin console

guusdk:
The admin console's "Session Summary" page lists all client sessions and shows some data.

With this commit, the 'priority' value (which is hardly used) is replaced by TX and RX values, that give an indication of activity (as it shows the amount of stanzas exchanged between client and server).

image


xmppserver/src/main/webapp/session-summary.jsp
```diff
@@ -303,7 +269,22 @@
...
+            <a href="session-summary.jsp" onclick='return toggleColumnOrder(${SessionResultFilter.SORT_NUM_CLIENT_PACKETS})'>
+                <fmt:message key="session.details.received-abbreviation" />
+                <c:if test="${listPager.sortColumnNumber == SessionResultFilter.SORT_NUM_CLIENT_PACKETS}">
+                    <img src="images/sort_${listPager.sortDescending ? "descending" : "ascending"}.gif" alt="The sort order in this column is ${listPager.sortDescending ? "descending" : "ascending"} (click to toggle).">
+                </c:if>
+            </a>
+        </th>
+        <th nowrap>
+            <a href="session-summary.jsp" onclick='return toggleColumnOrder(${SessionResultFilter.SORT_NUM_SERVER_PACKETS})'>
+                <fmt:message key="session.details.transmitted-abbreviation" />
```
Fishbowler, line 282: Should (can?) this have the unabbreviated text as a hover/tooltip?

Fishbowler: I've pushed a commit that does it, by adding a title attribute to the `th`, but I didn't have the right strings, so did a _tiny_ refactor to make everything tidy. Feel free to bin it or amend it as you like - this tweak is almost as big as the change that it tweaks :)

guusdk: Just tested it: works for me. I'm happy to have that go in.

Fishbowler (approved):

Fishbowler:
No change, just tidying commit msg

PR #2120, opened 2022-10-04
OF-2517: Fix 'packet sent' count for BOSH connections

guusdk:
Openfire keeps track of the amount of stanzas sent through a connection. For BOSH connections, the outgoing count was never updated. This is fixed in this commit.

This change also ensures that a 'last activity' field is updated more frequently. This could improve the stability of BOSH connections.


Fishbowler (approved):
LGTM, but I can't work out how this improves the frequency at which lastActivity is updated.

PR #2124, opened 2022-10-05
take install4j schema update verbatim

akrherz:
I just launched install4j and let it migrate the schema.


guusdk:
Cherry-picked to 4.7.

PR #2125, opened 2022-10-10
bump install4j maven plugin to latest release

akrherz:
The 1.1.2 version is 5 years newer than 1.1.1, so it has to be better, right?


Fishbowler:
I can't find any changelog. Feels like this is low risk, in any event.

Fishbowler (approved):

PR #2128, opened 2022-10-11
OF-2523: Do not re-use stream ID as the default resource ID

guusdk:
Let's decouple the stream ID and resource identifier that's used by default. This should help make both less predictable.


Fishbowler:
@dwd The resource is practically public domain (in many scenarios), but what's the risk in the stream being known?

guusdk:
After some [discussion in the XSF chatroom](https://logs.xmpp.org/xsf/2022-11-26?p=h#2022-11-26-8cffb0aaff3b5c77), I believe that it does not hurt to decouple the two.

PR #2129, opened 2022-10-11
OF-2524: Improving speed of loading MUC history from database

guusdk:
Two changes (see individual commit messages for details):
- add all messages to the history in one batch, to prevent repeated acquiring and releasing of a (cluster-wide) lock
- add database index to speed up query to retrieve messages from the database.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
@@ -25,6 +25,7 @@ import org.jivesoftware.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xmpp.packet.JID;
+import org.xmpp.packet.Message;
 
 import javax.annotation.Nonnull;
 import java.math.BigInteger;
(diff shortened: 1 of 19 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
```diff
@@ -796,23 +797,29 @@ public static void loadHistory(@Nonnull final MUCRoom room, final int maxNumber)
...
                     Date sentDate = new Date(Long.parseLong(rs.getString("logTime").trim()));
                     String subject = rs.getString("subject");
                     String body = rs.getString("body");
                     String stanza = rs.getString("stanza");
-                    room.getRoomHistory().addOldMessage(senderJID, nickname, sentDate, subject, body, stanza);
+                    oldMessages.add(room.getRoomHistory().parseHistoricMessage(senderJID, nickname, sentDate, subject, body, stanza));
+                }
+
+                if (!oldMessages.isEmpty()) {
+                    room.getRoomHistory().addOldMessages(oldMessages);
```
Fishbowler, line 812: Any memory exhaustion concerns here? 
For a short while here, there's double the committed memory.
You'll have seen some esoteric setups, as well as Ignite, which has significant history. You might judge this to be low risk.

guusdk: The data is limited to that of one room. I don't see how that would be that much of a problem.

Fishbowler: True. 100K msgs at 100b per msg would only be a temporary 9MB spike.

Fishbowler (approved):

guusdk:
cherry-picked to 4.7.

PR #2133, opened 2022-10-15
Update apache commons-text from 1.6 to 1.10.0

Fishbowler:
Fixes OF-2529

Updates Apache commons-text in Openfire to prevent it incorrectly appearing as vulnerable to CVE-2022-42889, even though it isn't, since we don't use the affected portions of the library.

### DIFF
--- a/xmppserver/pom.xml
+++ b/xmppserver/pom.xml
@@ -371,7 +371,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-text</artifactId>
-            <version>1.6</version>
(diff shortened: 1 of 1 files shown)


Fishbowler:
Escaping XML only.

pcullum (approved):

guusdk:
I had a quick look if we could remove this as a dependency, but escaping XML seems like a sufficient tricky beast that I'd rather not maintain a custom implementation.

Let's merge and cherry-pick to 4.7.x to squelch any alarms raised on those releases.

PR #2135, opened 2022-10-19
OF-2531: Remove unneeded JID-based lock in MUC

guusdk:
To fix https://igniterealtime.atlassian.net/browse/OF-1649 a user/JID-based lock was introduced. This lock made the creation of a user-and-joining-a-room atomic with respect to the cleanup action remove-all-users-that-are-not-in-any-room.

With the rewrite of MUC in https://igniterealtime.atlassian.net/browse/OF-2224 (and friends), the cleanup of users is no longer based on room membership, but on duration since last activity. This will prevent https://igniterealtime.atlassian.net/browse/OF-1649. The lock that was put in place now no longer serves a purpose, and should be removed.


guusdk:
I have already applied this change to the 4.7 branch, which helped me generate a distributable artifact that I needed for testing. If this PR is to be changed or rejected, the same needs to be applied to the commit already in the 4.7 branch.

GregDThomas (approved):
LGTM, as you say, it's no longer needed,

PR #2136, opened 2022-10-27
OF-2533: AdminConsole's IP access should allow ranges

guusdk:
This adds to the new IP-based access control for the admin console that was introduced in 4.8 via OF-2474 / https://github.com/igniterealtime/Openfire/pull/2074

Now, IP _ranges_ can also be used (eg, 192.168.2.0/24).

Also improved IPv6 support


Fishbowler (approved):
Approved after chatting through risks in open_chat

PR #2137, opened 2022-11-03
OF-2537: Advertise support for pubsub's 'multi-item' feature

guusdk:
XEP-0060 defines, amongst many others, the 'multi-items' service discovery feature.

This feature can be advertised by a pubsub service, when clients are allowed to configure a node in such a way that it can accept more-than-one item, by:

- setting `persist_items` to `true`
- setting `max_items` to a value higher than 1.

Openfire supports allows for both. It can/should advertise the feature `http://jabber.org/protocol/pubsub#multi-items` on its pubsub services.


guusdk:
This looks benign enough to be a candidate for 4.7.x

akrherz (approved):

PR #2145, opened 2022-11-10
Suppress LGTM alert for ZipSlip in PluginMonitor

guusdk:
ZipSlip is possible only when explicitly enabled by an administrator. No need to have a static analyzer alert for this particular case.


guusdk:
An additional observation: if a malicious attacker is able to upload a plugin (that might have a ZipSlip vulnerability), they are able to upload executable code. I wonder to what extend guarding against ZipSlip here actually improve security significantly.

PR #2147, opened 2022-11-10
OF-2543: When retrieving pubsub items, always include payload

guusdk:
The pubsub node configuration option `pubsub#deliver_payloads` is defined in XEP-0060 to control if payloads are included in pubsub notification events. Openfire should not use it to determine if payloads are to be included when answering item retrievals (those should _always_ have payloads, if they exist on the item).


guusdk:
We've tested this fix with Moveim (movim/movim#756) which seems to resolve an issue there.

guusdk:
Backported to 4.7.

PR #2156, opened 2022-12-03
OF-2556: Record namespace prefixes sent on root XML element

guusdk:
XMPP connections take the form of an XML document, starting with a 'stream' root element.

Openfire processes the direct child elements as if they were individual documents (for reasons beyond the scope of this issue). In that approach, any namespaces defined in the original root element are lost.

Generally, this isn't much of an issue, as a default namespace is assumed/used that is specific to the connection type. However, if the root element defined _other_ issues, that information is lost.

Defining additional namespaces is allowed in XML/XMPP, but isn't typically used. If it were, we would have hit problems long ago. There is one known application of this, which is Dialback (refactoring Openfire's S2S server to use MINA as part of OF-1112 is how this issue was found).

Openfire should be able to XML-parse child elements of the root 'stream' tags, if those child elements use a prefix that refers to a namespace defined on that root element.

This commit achieves that, by recording such declarations in the `Connection` representation over which they were sent. When such declarations are found, each stanza that is processed is wrapped in dummy root element, on which those definitions are placed, prioer to XML parsing. After parsing, the root element is discarded again.

### DIFF
--- a/xmppserver/src/main/java/org/dom4j/io/XMPPPacketReader.java
+++ b/xmppserver/src/main/java/org/dom4j/io/XMPPPacketReader.java
@@ -82,14 +82,15 @@ public class XMPPPacketReader {
 
     /**
      * Retrieves a collection of namespaces declared in the current element that the parser is on, that are not defined
-     * in {@link #IGNORED_NAMESPACE_ON_STANZA}
+     * in {@link #IGNORED_NAMESPACE_ON_STANZA}, and that are not the default namespace (the namespaces returned all have
+     * a defined prefix).
      *
      * @param xpp the parser
      * @return A collection of namespaces
-     * @throws XmlPullParserException
+     * @throws XmlPullParserException On any problem thrown by the XML parser.
      */
     @Nonnull
-    public static Set<Namespace> getNamespacesOnCurrentElement(XmlPullParser xpp) throws XmlPullParserException
+    public static Set<Namespace> getPrefixedNamespacesOnCurrentElement(XmlPullParser xpp) throws XmlPullParserException
     {
         final Set<Namespace> results = new HashSet<>();
         final int nsStart = xpp.getNamespaceCount(xpp.getDepth()-1);
--- a/xmppserver/src/main/java/org/dom4j/io/XMPPPacketReader.java
+++ b/xmppserver/src/main/java/org/dom4j/io/XMPPPacketReader.java
@@ -16,10 +16,13 @@ import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
 import org.xmlpull.v1.XmlPullParserFactory;
 
+import javax.annotation.Nonnull;
 import java.io.*;
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * <p><code>XMPPPacketReader</code> is a Reader of DOM4J documents that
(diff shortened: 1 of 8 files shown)


xmppserver/src/main/java/org/dom4j/io/XMPPPacketReader.java
```diff
@@ -77,6 +80,30 @@ public XMPPPacketReader(DocumentFactory factory) {
...
+    /**
+     * Retrieves a collection of namespaces declared in the current element that the parser is on, that are not defined
+     * in {@link #IGNORED_NAMESPACE_ON_STANZA}
+     *
+     * @param xpp the parser
+     * @return A collection of namespaces
+     * @throws XmlPullParserException
+     */
+    @Nonnull
+    public static Set<Namespace> getNamespacesOnCurrentElement(XmlPullParser xpp) throws XmlPullParserException
```
GregDThomas, line 92: I reckon this method would be a great place for a unit test. I can see what it should be doing, but a unit test would at least validate it does what I think it does.

xmppserver/src/main/java/org/jivesoftware/openfire/Connection.java
```diff
@@ -16,24 +16,38 @@
...
+
+    /**
+     * When a connection is used to transmit an XML data, the root element of that data can define XML namespaces other
+     * than the ones that are default (eg: 'jabber:client', 'jabber:server', etc). For an XML parser to be able to parse
+     * stanzas or other elements that are defined in that namespace (eg: are prefixed), these namespaces are recorded
+     * here.
+     *
+     * @see <a href="https://igniterealtime.atlassian.net/browse/OF-2556">Issue OF-2556</a>
+     */
+    Set<Namespace> additionalNamespaces = new HashSet<>();
```
GregDThomas, line 49: Given your current implementation, this could be a final field. If it's not supposed to be a final, why not simply re-initialise it in the setter method?

GregDThomas (approved):
Minor comments only (esp. some unit testing on that XPP utility method) but looks good overall

xmppserver/src/main/java/org/jivesoftware/openfire/Connection.java
```diff
@@ -368,6 +382,33 @@ public interface Connection extends Closeable {
...
+     * When a connection is used to transmit an XML data, the root element of that data can define XML namespaces other
+     * than the ones that are default (eg: 'jabber:client', 'jabber:server', etc). For an XML parser to be able to parse
+     * stanzas or other elements that are defined in that namespace (eg: are prefixed), these namespaces are recorded
+     * here.
+     *
+     * @return A collection that contains all non-default namespaces that the peer defined when last opening a new stream.
+     */
+    @Nonnull
+    default Set<Namespace> getAdditionalNamespaces() {
+        return additionalNamespaces;
```
GregDThomas, line 395: A possible loss of encapsulation here. Perhaps return an unmodifiable copy of the set?

xmppserver/src/main/java/org/dom4j/io/XMPPPacketReader.java
guusdk: Agreed! I've added some in a new commit. If anything, it showed that I could have better named the method (which I now did).

xmppserver/src/main/java/org/jivesoftware/openfire/Connection.java
guusdk: This will be invoked for _every_ stanza that is received over the connection, while the added benefit is low. I'm not sure if the added overhead is worth the safety that it adds.

xmppserver/src/main/java/org/jivesoftware/openfire/Connection.java
guusdk: IntelliJ tells me that the modifier 'final' is redundant on interface fields.

I'm not sure if this should be an interface field though. Maybe this needs to go in an abstract implementation? I am not used to using interfaces this way...

Fishbowler (approved):

guusdk:
Merging, as I have commits that depend on this. I'm happy to include further feedback in a new PR.

PR #2157, opened 2022-12-06
OF-2557: Show TLS config on each session/connection

guusdk:
The cipher suite used for encryption is currently exposed (and shown on the admin console) only for S2S sessions, but should be exposed for all session types.

Similarly, lets expose the TLS version that is being used (to complement the cipher suite). This will be useful to diagnose the TLS configuration for each session type.

Implementation-wise, the TLS characteristics should be exposed by Connections, not by Session objects.

Confusingly, Openfire refers to ‘securing’ a connection, which could refer to authentication or encryption. Rename to ‘encrypt’ where appropriate.

### DIFF
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -2057,6 +2057,8 @@ session.details.close=Closed
 session.details.connect=Connected
 session.details.streaming=Streaming
 session.details.authenticated=Authenticated
+session.details.tls_version=TLS Version
+session.details.cipher=Cipher Suite
 session.details.presence=Presence
 session.details.away=Away
 session.details.chat_available=Available to Chat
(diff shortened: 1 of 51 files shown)


guusdk:
This became somewhat of a large blob - apologies to reviewers.

akrherz (approved):
I am unsure of the semantics here though.  Is it possible to have an encrypted session over a non-TLS connection?  Or a non-encrypted session over a TLS connection? I obviously defer to Guus's opinion.

guusdk:
All encryption of XMPP connections is TLS based. Authentication can be based on several mechanisms, including ones that are TLS-based. I don't think it should typically possible to use TLS-based authentication when (TLS-based) encryption has not been established.

Fishbowler (approved):
LGTM. Most of the changes are s/secure/encrypted.
I don't know enough of the fine detail to know that the defaults are correct, but they look reasonable. I don't yet have the infrastructure to test this properly (and assume you've done the trivial tests that I could do without TLS).

guusdk:
Rebased to main.

PR #2165, opened 2023-01-11
OF-2611: Add unit tests for outbound server session

guusdk:
This PR contains two commits:
- the first refactors test code (in preparation of the second commit)
- the second commit adds unit tests that verify the behavior of establishing an outbound server-to-server connection. It does so by providing a dummy remote server implementation, against which a socket is being opened.

The second commit depends on way to much mocks for comfort. @GregDThomas do you have suggestions on how to improve this?

@Fishbowler: can you please help to validate and possibly suggest changes/additions the test scenarios that are implemented in the second commit?


xmppserver/src/test/java/org/jivesoftware/Fixtures.java
```diff
@@ -89,13 +98,15 @@ public static XMPPServer mockXMPPServer() {
...
         doReturn(mockIQRouter()).when(xmppServer).getIQRouter();
+        doReturn(mockConnectionManager()).when(xmppServer).getConnectionManager();
 
         return xmppServer;
     }
 
     public static XMPPServerInfo mockXMPPServerInfo() {
         final XMPPServerInfo xmppServerInfo = mock(XMPPServerInfo.class, withSettings().lenient());
         doReturn(XMPP_DOMAIN).when(xmppServerInfo).getXMPPDomain();
+        doReturn(XMPP_DOMAIN).when(xmppServerInfo).getHostname();
```
GregDThomas, line 109: Is it worth creating a new constant with a different value? I can see issues where the hostname and XMPP domain, not necessarily the same, could get confused with each other.

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalOutgoingServerSessionTest.java
```diff
@@ -0,0 +1,335 @@
...
+            final int port = remoteServerDummy.getPort();
+
+            // Execute system under test.
+            final LocalOutgoingServerSession result = LocalOutgoingServerSession.createOutgoingSession(domainPair, port);
+
+            // Verify results.
+            assertNotNull(result);
+            assertFalse(result.isClosed());
+            assertTrue(result.isEncrypted());
+            assertFalse(result.isUsingServerDialback());
```
GregDThomas, line 152: More a style thing, but more recently I've become a big fan of AssertJ - https://assertj.github.io/doc/

The above four assertions could then be written more fluently as
```java
assertThat(result).isNotNull();
assertThat(result.isClosed()).isFalse();
assertThat(result.isEncrypted).isTrue();
assertThat(result.isUsingDialback).isFalse();
```
Though the real benefit comes when using things like reflective comparisons on lists, etc.

GregDThomas (approved):
A couple of comments, personal preference really so nothing of note.

I see what you mean about the number of mocks. It seems to me that Openfire wasn't really originally developed with unit testing in mind (perhaps a sign of it's longevity), and trying to shoe-horn them in after the fact leads to more effort than would be ideally required.

But I can't see an alternative, and it's better than no unit tests :)

guusdk:
I have added a bunch of commits to this PR.

The first three commits are changes that are also in separate PRs (https://github.com/igniterealtime/Openfire/pull/2166 and https://github.com/igniterealtime/Openfire/pull/2167). When those PRs get merged, these commits can be taken out of this PR (assuming it will be rebased).

The next commit applies to more refactoring of the test infrastructure. This commits should probably be squashed/fixed up with the first commit in this branch (but I didn't want to mess up the commit/review history right now)

The last commit basically rewrites the unit tests, and expands on them. In my first draft, I made some incorrect assumptions. In this rewrite, I tried to describe (in javadoc) more concisely:
- what the test scenario is.
- what the expected results are.
- why those results are expected.

I then tried to reproduce these three in the implementation of the test.

Things to review:
- There is _a lot_ of copy/pasting going on, so I would love for someone to review if the unit test javadoc describes what actually is being tested (and if this conforms to the name of the test method)
- Stepping back a bit further, I would also love if someone could review the rationales that I put in each test. Is what I defined as the expected outcome for each scenario indeed the outcome that, according to XMPP specifications, is expected.

Furthermore, we can probably expand further on these tests. 

And, lastly - not all tests are succeeding. These might be bugs in Openfire that require fixing :). This should be accompanied with corresponding JIRA issues (and might even go in new PRs).

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalOutgoingServerSessionTest.java
guusdk: I myself have no strong preference for one style or the other. Currently, I'm struggling to get the changes in this PR to be correct. I fear that moving away from my style-based comfort zone (even for an improvement) might overload my poor brain. :)

xmppserver/src/test/java/org/jivesoftware/Fixtures.java
guusdk: Good idea. Done.

xmppserver/src/main/java/org/jivesoftware/openfire/session/ServerSession.java
```diff
@@ -1,10 +1,50 @@
...
+        SASL_EXTERNAL,
+        OTHER
+    }
+
+    /**
+     * Obtain method that was used to authenticate this session. Null when the session is not authenticated.
+     *
+     * @return the method used for authentication (possibly null).
+     */
+    AuthenticationMethod getAuthenticationMethod();
```
GregDThomas, line 31: Perhaps
`Optional<AuthenticationMethod> getAuthenticationMethod();`
to force the check for absence?

xmppserver/src/main/java/org/jivesoftware/openfire/session/RemoteSessionTask.java
```diff
@@ -181,7 +181,7 @@ public enum Operation {
          * Operations of outgoing server sessions
          */
         getOutgoingDomainPairs,
-        isUsingServerDialback,
+        getAuthenticationMethod,
```
GregDThomas, line 184: Is this going to cause compatibility issues with mixed version clusters? Does it matter?

xmppserver/src/main/java/org/jivesoftware/util/S2STestService.java
```diff
@@ -154,15 +171,14 @@ private void logSessionStatus() {
         final DomainPair pair = new DomainPair(XMPPServer.getInstance().getServerInfo().getXMPPDomain(), domain);
         OutgoingServerSession session = XMPPServer.getInstance().getSessionManager().getOutgoingServerSession(pair);
         if (session != null) {
-            int connectionStatus = session.getStatus();
-            switch(connectionStatus) {
-            case Session.STATUS_CONNECTED:
+            switch(session.getStatus()) {
```
GregDThomas, line 174: Is this switch statement necessary? Would not `log.info("Session is {}", session.getStatus())`; suffice?

xmppserver/src/main/webapp/server-session-details.jsp
```diff
@@ -318,12 +321,15 @@
...
-                                        <fmt:message key="server.session.details.dialback"/>
-                                    </c:when>
-                                    <c:otherwise>
-                                        <fmt:message key="server.session.details.tlsauth"/>
-                                    </c:otherwise>
+                                        <c:when test="${session.usingServerDialback}">
+                                            <fmt:message key="server.session.details.dialback"/>
+                                        </c:when>
+                                        <c:when test="${session.usingSaslExternal}">
+                                            <fmt:message key="server.session.details.tlsauth"/>
```
GregDThomas, line 328: The name of key doesn't seem to reflect it's usage here. It may be necessary to rename the key?

xmppserver/src/main/webapp/session-details.jsp
```diff
@@ -266,19 +266,19 @@
         </td>
         <td>
             <%
-                int status = currentSess.getStatus();
-                if (status == Session.STATUS_CLOSED) {
+                Session.Status status = currentSess.getStatus();
+                if (status == Session.Status.CLOSED) {
```
GregDThomas, line 270: Do you want to use the various `if(currentSess.isClosed/Authenticated/etc`) methods here?

xmppserver/src/main/webapp/session-row.jspf
```diff
@@ -58,21 +58,21 @@
...
     <td>
-        <%  int _status = sess.getStatus();
+        <%  Session.Status _status = sess.getStatus();
             boolean isDetached = sess instanceof LocalSession && ((LocalSession) sess).isDetached();
             if (isDetached) { %>
 
             <fmt:message key="session.details.sm-detached" />
         <%
-            } else if (_status == Session.STATUS_CLOSED) {
+            } else if (_status == Session.Status.CLOSED) {
```
GregDThomas, line 67: `sess.isClosed()` ?

xmppserver/src/main/java/org/jivesoftware/openfire/session/Session.java
```diff
@@ -47,9 +46,30 @@ public interface Session extends RoutableChannelHandler {
...
+        CLOSED,
+        CONNECTED,
+        AUTHENTICATED
+    }
+
+    /**
+     * @deprecated Use {@link Status#CLOSED} instead.
+     */
+    @Deprecated // Remove in or after Openfire 4.9.
+    Status STATUS_CLOSED = Status.CLOSED;
```
GregDThomas, line 59: You've marked this as deprecated, but by changing get/setStatus() you've already broken compatibility - is it worth keeping this at all?

GregDThomas (approved):
A few more comments. None of which identify any real issues, tbh.

GregDThomas (approved):
A few more comments. None of which identify any real issues, tbh.

xmppserver/src/main/java/org/jivesoftware/openfire/audit/spi/AuditorImpl.java
```diff
@@ -403,24 +403,28 @@ public AuditPacket(Packet packet, Session session) {
...
-                    // This is a workaround. Since we don't want to have an incorrect FROM attribute
-                    // value we need to clean up the FROM attribute. The FROM attribute will contain
-                    // an incorrect value since we are setting a fake JID until the user actually
-                    // authenticates with the server.
-                    packet.setFrom((String) null);
-                    break;
-                default:
-                    element.addAttribute("status", "unknown");
-                    break;
+            if (session == null) {
```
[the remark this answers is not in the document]
guusdk: I can see the code optimization, but I do not particularly like adding anything to the definition of the enum when it's not expected to be a re-usable thing - and I don't think that the status attribute of these specific values is re-used.

This attribute value is (as far as I can see) very specific to the Audit implementation (which itself is not even one of the primary features of Openfire). I don't think it's used elsewhere. Adding complexity in the central place to reduce complexity in only one of it's use-cases isn't a great trade-off.

I would prefer to simply use the `toString()` of the enum, but I don't like breaking compatibility in the Audit logs, as I suspect that people will have automated processing of these.

xmppserver/src/main/java/org/jivesoftware/openfire/handler/PresenceUpdateHandler.java
```diff
@@ -161,7 +149,7 @@ private void process(Presence presence, ClientSession session) throws Unauthoriz
             Presence.Type type = presence.getType();
             // Available
             if (type == null) {
-                if (session != null && session.getStatus() == Session.STATUS_CLOSED) {
+                if (session != null && session.getStatus() == Session.Status.CLOSED) {
```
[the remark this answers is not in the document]
guusdk: There already exists such a method, in `org.jivesoftware.openfire.session.Session#isClosed`:

    /**
     * Returns true if the connection/session is closed.
     *
     * @return true if the connection is closed.
     */
    default boolean isClosed() {
        return getStatus() == Status.CLOSED;
    };

The reason that I was not using it, is because some implementations override the definition (and not use the 'status' value to determine if the session is closed). I'm not sure if this is more, or less reason to start using the `isClosed()` method instead of comparing the `status` value directly. For now, I choose to not change behavior much. There probably is an argument to be made to do that, but perhaps that warrants it's own issue number?

xmppserver/src/main/java/org/jivesoftware/openfire/audit/spi/AuditorImpl.java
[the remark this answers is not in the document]
guusdk: I'm not sure. Three lines up, there is an explicit null-check for the session. As this bit of code seems to be designed to handle null sessions, I thought it best to continue doing that there.

xmppserver/src/main/java/org/jivesoftware/openfire/net/StanzaHandler.java
```diff
@@ -286,7 +286,7 @@ else if ("presence".equals(tag)) {
                 // an available presence show
                 packet.setShow(null);
             }
-            if (session.getStatus() == Session.STATUS_CLOSED && packet.isAvailable()) {
+            if (session.getStatus() == Session.Status.CLOSED && packet.isAvailable()) {
```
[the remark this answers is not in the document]
guusdk: I do not disagree, but I'm holding off doing that for reasons explained in https://github.com/igniterealtime/Openfire/pull/2165#discussion_r1071926857

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketReader.java
```diff
@@ -204,7 +204,7 @@ else if ("presence".equals(tag)) {
                 // an available presence show
                 packet.setShow(null);
             }
-            if (session.getStatus() == Session.STATUS_CLOSED && packet.isAvailable()) {
+            if (session.getStatus() == Session.Status.CLOSED && packet.isAvailable()) {
```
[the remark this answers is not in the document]
guusdk: I do not disagree, but I'm holding off doing that for reasons explained in https://github.com/igniterealtime/Openfire/pull/2165#discussion_r1071926857

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalServerSession.java
```diff
@@ -76,8 +95,32 @@ public Connection getConnection() {
...
+
+    /**
+     * Set the method that was used to authenticate this session. Setting a value will cause the status of this session
+     * to be updated to 'Authenticated'.
+     *
+     * @param authenticationMethod The new authentication method for this session
+     */
+    public void setAuthenticationMethod(@Nonnull final AuthenticationMethod authenticationMethod) {
+        this.authenticationMethod = authenticationMethod;
+        setStatus(Status.AUTHENTICATED);
```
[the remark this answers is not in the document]
guusdk: Exactly what are you questioning? If a session can be 'un-authenticated'?

Openfire currently does not support revalidating the authentication of a session during its lifetime (but perhaps should).

Until it does, this should be fine (and an improvement over the previous implementation, that marked a session as being Dialback-authenticated, even before it was set up at all).

xmppserver/src/main/java/org/jivesoftware/openfire/session/RemoteSessionTask.java
guusdk: I suspect it will cause compatibility issues in a cluster. I'm planning to get this in a more-than-patch release (eg, 4.8.0, not 4.7.5). That should lessen the expectancy of version compatibility in a cluster.

xmppserver/src/main/java/org/jivesoftware/openfire/session/ServerSession.java
guusdk: As I expected you to suggest this, initially implemented it as an Optional :).

I reverted that, as there was a silly amount of places where I had to 'get' the value without it doing a presence check.

xmppserver/src/main/java/org/jivesoftware/util/S2STestService.java
guusdk: Good catch. I'll add that.

xmppserver/src/main/webapp/server-session-details.jsp
guusdk: I don't see the problem? SASL EXTERNAL == TLS authentication (at least in the current implementation of Openfire).

xmppserver/src/main/webapp/session-details.jsp
guusdk: Maybe, but the reasoning in https://github.com/igniterealtime/Openfire/pull/2165#discussion_r1071926857 applies here too.

xmppserver/src/main/webapp/session-row.jspf
guusdk: Not disagreeing, but: https://github.com/igniterealtime/Openfire/pull/2165#discussion_r1071926857

xmppserver/src/main/java/org/jivesoftware/openfire/session/Session.java
guusdk: I had a plan for this (not sure if it holds up though): Those fields (constants really) initially were integers, but were refactored to enum values.

The `get/setStatus()` methods are changed, as you mentioned, but in the same way: they used integer values before, but now use the same enum values.

My reasoning for keeping the fields in (but marking them as being deprecated) is that I suspect that 99% of the usages will be combined with usage of the getter and setter. By changing the type of the fields _and_ the getter/setter to the same new types, compilation should not need to break (at least not when they're used together, like `if (session.getStatus() == Session.STATUS_CLOSED) { ... }` ).

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalServerSession.java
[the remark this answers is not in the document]
guusdk: The XMPP specifications in RFC 6120, section 5.3.5 do describe TLS renegotiation, but define it as OPTIONAL. The specification includes something of a warning, asking implementers to carefully weigh the costs and benefits of TLS renegotiation before supporting it. I'm not going to raise a feature request ticket for Openfire. :)

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
```diff
@@ -413,6 +413,11 @@ else if (ServerDialback.isEnabled() && features.element("dialback") != null) {
             if (connection != null) {
                 connection.forceClose();
             }
+
+            if (JiveGlobals.getBooleanProperty(ConnectionSettings.Server.STRICT_CERTIFICATE_VALIDATION, true)) {
+                log.warn( "Unable to create a new session: not trying dialback as a fallback, as server strictCertificateValidation is enabled" );
```
guusdk, line 418: It is better to avoid using the term 'dialback' here. 

When this message is logged, we do not know if Dialback is even enabled. In future versions of Openfire, Dialback support might be removed, or a third authentication mechanism (that we're not naming here) might be added. The message as logged now is based on assumptions, that might not be true in all cases when it is logged. This can add up to a lot of confusion in practice.

Instead, replace this with what's factually being done in the if-statement that is logging the message. Something like:
> Aborting attempt to create outgoing session as TLS handshake failed, and strictCertificateValidation is enabled.

xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionConfiguration.java
```diff
@@ -54,7 +55,7 @@ public class ConnectionConfiguration
      * @param compressionPolicy the compression policy
      */
     // TODO input validation
-    public ConnectionConfiguration( ConnectionType type, boolean enabled, int maxThreadPoolSize, int maxBufferSize, Connection.ClientAuth clientAuth, InetAddress bindAddress, int port, Connection.TLSPolicy tlsPolicy, CertificateStoreConfiguration identityStoreConfiguration, CertificateStoreConfiguration trustStoreConfiguration, boolean acceptSelfSignedCertificates, boolean verifyCertificateValidity, Set<String> encryptionProtocols, Set<String> encryptionCipherSuites, Connection.CompressionPolicy compressionPolicy )
+    public ConnectionConfiguration( ConnectionType type, boolean enabled, int maxThreadPoolSize, int maxBufferSize, Connection.ClientAuth clientAuth, InetAddress bindAddress, int port, Connection.TLSPolicy tlsPolicy, CertificateStoreConfiguration identityStoreConfiguration, CertificateStoreConfiguration trustStoreConfiguration, boolean acceptSelfSignedCertificates, boolean verifyCertificateValidity, Set<String> encryptionProtocols, Set<String> encryptionCipherSuites, Connection.CompressionPolicy compressionPolicy, boolean strictCertificateValidation )
```
guusdk, line 58: Please describe the new parameter in javadoc.

xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java
```diff
@@ -1026,4 +1027,31 @@ public String toString()
                 '}';
     }
 
+    public void setStrictCertificateValidation(boolean strictCertificateValidation) {
```
guusdk, line 1030: Please add Javadoc and \*twitch\* move it above the `toString()` method.

xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java
```diff
@@ -1026,4 +1027,31 @@ public String toString()
...
+        if ( oldValue == strictCertificateValidation ) {
+            Log.debug( "Ignoring strict certificate validation configuration change request (to '{}'): listener already in this state.", strictCertificateValidation );
+            return;
+        }
+
+        Log.debug( "Changing strict certificate validation configuration from '{}' to '{}'.", oldValue, strictCertificateValidation );
+        restart();
+    }
+
+    private boolean getStrictCertificateValidation() {
```
guusdk, line 1045: Please add Javadoc and \*twitch\* move it above the `toString()` method.

xmppserver/src/test/java/org/jivesoftware/Fixtures.java
```diff
@@ -217,4 +250,58 @@ public static IQ iqFrom(final String stanza) {
...
+        "yfZLwSrR3UNcNYk9W01/xjQVgH2JsF1B1nbn7eJt0mzr0arHp7VjtjdDJIfkZEEh\n" +
+        "5ZpOERffIINoEptoKMCfjbcp2+PLZ1TjL6MxtrVs+TQmX4i9wL03uNgItbFqYeYP\n" +
+        "RVeb9OaNj4NRMZB49D/jIaqmQWZi6u2ooOQv6AlyzeKInWm+aDmxCAX4IZAod18/\n" +
+        "1hI2qIN8Xj9GUT7wldL368Dq1fUI5g==\n" +
+        "-----END CERTIFICATE-----\n";
+
+    /**
+     * Private key for {@link #invalidX509Certificate} for testing purposes
+     */
+    public static final String invalidX509CertificatePrivateKey
```
guusdk, line 278: Maybe rename to `privateKeyForExpiredCert` to reduce the likelyhood that someone (like me) assumes that this private key is somehow invalid (which it isn't, judging by the javadoc).

guusdk:
I have created a separate Pull Request to bump junit from 4 to 5: https://github.com/igniterealtime/Openfire/pull/2204

I'm planning to remove that commit from this PR.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
AlexGidman: fixed

xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionConfiguration.java
AlexGidman: done

xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionConfiguration.java
```diff
@@ -201,4 +203,8 @@ public boolean isEnabled()
     {
         return enabled;
     }
+
```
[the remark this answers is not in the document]
AlexGidman: done

xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java
AlexGidman: done

xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java
AlexGidman: done

xmppserver/src/test/java/org/jivesoftware/Fixtures.java
```diff
@@ -217,4 +250,58 @@ public static IQ iqFrom(final String stanza) {
         }
     }
 
+    /**
+     * Self-Signed invalid X509 certificate for testing purposes (expired 17th June 2023)
+     */
+    public static final String invalidX509Certificate
```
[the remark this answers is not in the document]
AlexGidman: done

xmppserver/src/test/java/org/jivesoftware/Fixtures.java
AlexGidman: done

guusdk:
I have created another separate Pull Request for a Dialback fix: https://github.com/igniterealtime/Openfire/pull/2205

Again,[ the corresponding commit](7b257c7d1af187c92457a9576e7e3060a843e9e0) should probably be removed from this PR when it's rebased to something that includes the above, to keep things concise.

xmppserver/src/test/java/org/jivesoftware/util/CertificateManagerTest.java
```diff
@@ -70,12 +74,12 @@
     @BeforeAll
     public static void initialize() throws Exception
     {
-        keyPairGenerator = KeyPairGenerator.getInstance( "RSA" );
-        keyPairGenerator.initialize( 512 );
+        keyPairGenerator = KeyPairGenerator.getInstance( KEY_ALGORITHM );
+        keyPairGenerator.initialize( KEY_SIZE );
```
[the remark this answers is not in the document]
guusdk: I'm assuming that small key sizes make for faster operation. As this code is used in unit testing, I'm preferring speed over security.

guusdk:
This PR grew to quite the monster. We could slim it down a little bit by taking out the changes for OF-2622. There's already a pull request for that in https://github.com/igniterealtime/Openfire/pull/2211. When that gets merged, this PR can be rebased, and the corresponding commit can be removed.

Another commit that can be taken out of this PR is OF-2626. For that to be set apart in a new PR, the change in OF-2622 appears to be needed.

distribution/src/conf/openfire-demoboot.xml
```diff
@@ -50,5 +50,8 @@
               </roster>
           </user2>
       </users>
+      <authprovider>
+          <mode>custom</mode>
+      </authprovider>
```
Fishbowler, line 55: Will this have an effect on other things that use demoboot?

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
```diff
@@ -309,6 +315,11 @@ public boolean authenticateDomain(OutgoingServerSocketReader socketReader, Strin
 
         log.debug( "Authenticating domain ..." );
 
+        if (!ServerDialback.isEnabled() && !ServerDialback.isEnabledForSelfSigned()) {
+            log.info("Failed to authenticate domain: Dialback has been disabled by configuration.");
```
Fishbowler, line 319: Very nitpicky:

```suggestion
            log.info("Failed to authenticate domain via Dialback: Dialback has been disabled by configuration.");
```

Presumably this isn't the only way that domain authentication might occur. I think it might always be last, but is that an implementation detail, or a requirement of the standards?

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
```diff
@@ -473,35 +491,31 @@ protected void dialbackError(String from, String to, PacketError err) {
...
         log.debug( "Validating domain...");
-        if (connection.getTlsPolicy() == Connection.TLSPolicy.required &&
-                !connection.isEncrypted()) {
-            connection.deliverRawText(new StreamError(StreamError.Condition.policy_violation).toXML());
-            // Close the underlying connection
-            connection.close();
-            return false;
+
+        if (!ServerDialback.isEnabled() && !ServerDialback.isEnabledForSelfSigned()) {
+            throw new StreamErrorException(new StreamError(StreamError.Condition.policy_violation, "Dialback has been disabled by configuration."));
```
Fishbowler, line 507: Should the message be set on the StreamErrorException as well as the StreamError?
Or when this gets logged (e.g. on on L428) will the logger reach into StreamError to get its internal message?

(Of course, we could update the StreamErrorException constructor to set the Exception message to the Error's message if it is provided and the exception's message isn't, but it's getting close to fallable logic inside error handling, which is dangerous)

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
Fishbowler: Ignore. Logger is already going to know the class :facepalm:

xmppserver/src/test/java/org/jivesoftware/openfire/session/AbstractRemoteServerDummy.java
```diff
@@ -0,0 +1,222 @@
+package org.jivesoftware.openfire.session;
```
Fishbowler, line 1: No copyright header

xmppserver/src/test/java/org/jivesoftware/openfire/session/ExpectedOutcome.java
```diff
@@ -0,0 +1,298 @@
+/*
+ * Copyright (C) 2023 Guus der Kinderen. All rights reserved.
```
Fishbowler, line 2: Is this the copyright header you wanted for this file?

xmppserver/src/test/java/org/jivesoftware/openfire/session/ExpectedOutcome.java
```diff
@@ -0,0 +1,298 @@
...
+     *
+     * @param initiatingServer Configuration of the local server
+     * @param receivingServer Configuration of the remote server
+     * @return the expected outcome
+     */
+    public static ExpectedOutcome generateExpectedOutcome(final ServerSettings initiatingServer, final ServerSettings receivingServer) {
+        final ExpectedOutcome expectedOutcome = new ExpectedOutcome();
+
+        switch (initiatingServer.encryptionPolicy) {
+            case disabled: // <-- Initiating server's encryption policy.
```
Fishbowler, line 99: Do we still have the sources for all of these decisions?
I'd love to annotate where we got them, e.g. RFC sections.
Future Us will be grateful.

Fishbowler: There's also a lot of TODOs and conversations in the comments. I'd be keen to get those revalidated and squashed as much as we can.

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
```diff
@@ -0,0 +1,339 @@
...
+        final SessionManager sessionManager = new SessionManager(); // This is the system under test. We do not want to use a mock for this test!
+        sessionManager.initialize(xmppServer);
+        doReturn(sessionManager).when(xmppServer).getSessionManager();
+
+        final ConnectionManager connectionManager = Fixtures.mockConnectionManager();
+        final ConnectionListener connectionListener = Fixtures.mockConnectionListener();
+        doAnswer(new ConnectionConfigurationAnswer(identityStoreConfig, trustStoreConfig)).when(connectionListener).generateConnectionConfiguration();
+        doReturn(Set.of(connectionListener)).when(connectionManager).getListeners(any(ConnectionType.class));
+        doReturn(connectionListener).when(connectionManager).getListener(any(ConnectionType.class), anyBoolean());
+        doReturn(connectionManager).when(xmppServer).getConnectionManager();
```
Fishbowler, line 114: These mock behaviours already have equivalents in the Fixtures. Are they needed?

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
```diff
@@ -0,0 +1,339 @@
...
+    {
+        tmpIdentityStoreFile.delete();
+        tmpTrustStoreFile.delete();
+        identityStore = null;
+        trustStore = null;
+        DNSUtil.setDnsOverride(null);
+
+        if (remoteInitiatingServerDummy != null) {
+            //remoteInitiatingServerDummy.close();
+            remoteInitiatingServerDummy = null;
```
Fishbowler, line 162: Why not close()?
Should there be a comment here to let future maintainers know not to close()?

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalOutgoingServerSessionTest.java
```diff
@@ -0,0 +1,319 @@
...
+
+        // Mock the connection configuration.
+        doReturn(certificateStoreManager).when(xmppServer).getCertificateStoreManager();
+
+        final ConnectionManager connectionManager = Fixtures.mockConnectionManager();
+        final ConnectionListener connectionListener = Fixtures.mockConnectionListener();
+        doAnswer(new ConnectionConfigurationAnswer(identityStoreConfig, trustStoreConfig)).when(connectionListener).generateConnectionConfiguration();
+        doReturn(Set.of(connectionListener)).when(connectionManager).getListeners(any(ConnectionType.class));
+        doReturn(connectionListener).when(connectionManager).getListener(any(ConnectionType.class), anyBoolean());
+        doReturn(connectionManager).when(xmppServer).getConnectionManager();
```
Fishbowler, line 104: As above - do we need these explicit mock behaviours?

xmppserver/src/test/java/org/jivesoftware/openfire/session/RemoteInitiatingServerDummy.java
```diff
@@ -0,0 +1,555 @@
...
+            final SSLSocket sslSocket = (SSLSocket) ((SSLSocketFactory) SSLSocketFactory.getDefault()).createSocket(socket, null, socket.getPort(), true);
+            sslSocket.setSoTimeout((int) SO_TIMEOUT.toMillis());
+            sslSocket.addHandshakeCompletedListener(event -> { if (doLog) System.out.println("SSL handshake completed: " + event); });
+            sslSocket.startHandshake();
+
+            // Just indicate that we would like to authenticate the client but if client
+            // certificates are self-signed or have no certificate chain then we are still
+            // good
+            //sslSocket.setWantClientAuth(true); // DO WE NEED TO BRING THIS INTO OUR LocalOutgoingServerSessionTest MATRIX?
+            phaser.register();
```
Fishbowler, line 491: Is this a question that needs resolving?

xmppserver/src/test/java/org/jivesoftware/openfire/session/ServerSettings.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (C) 2023 Guus der Kinderen. All rights reserved.
```
Fishbowler, line 2: Intended copyright header?

Fishbowler (approved):
I've left a bunch of comments, mostly questions, but happy to move this through if folk decide there are no high value changes to be made.

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
guusdk: I think it's a good idea to provide as much context as reasonabily possible. That will help diagnosing problems. I have added some changes to the constructor. I believe they're fairly safe, but givin your last remark, it might be wise to review that carefully.

L428 is inside a method that doesn't seem to be used (I'm contemplating to remove it completely, but I wonder if maybe it _should_ be used somewhere, which is why I've not done that yet). In that method, the error isn't being transmitted to the client, nor is the connection closed. I've added that, too.

xmppserver/src/main/java/org/jivesoftware/util/StreamErrorException.java
```diff
@@ -57,6 +57,7 @@ public StreamErrorException(StreamError.Condition condition, String message, Thr
 
     public StreamErrorException(StreamError streamError)
     {
+        super(streamError != null && !streamError.getText().isEmpty() ? streamError.getText() : null);
```
Fishbowler, line 60: Can getText() ever return null?
If it can...

```suggestion
        super(streamError != null && streamError.getText() != null && !streamError.getText().isEmpty() ? streamError.getText() : null);
```

GregDThomas: Dunno why I'm on this thread, but 
```
 super(streamError != null && streamError.getText() != null && !streamError.getText().isEmpty() ? streamError.getText() : null);
```
Could be expressed a lot neater - to my mind anyway - as 
```
super( ofNullable(streamError).map(StreamError::getText).filter(StringUtils::isNotBlank).orElse(null) );
```

xmppserver/src/main/java/org/jivesoftware/openfire/server/ServerDialback.java
Fishbowler: Ace! I've added a comment on the constructor change separately, further down the PR :)

xmppserver/src/test/java/org/jivesoftware/openfire/session/ExpectedOutcome.java
guusdk: Many of the TODOs relate to one of two things:
- Possible usage of `anon` cypher suites, that would explicitly allow TLS for encryption but not authentication. Although theoretically possible, we've never worked with this or explicitly supported it. I don't know of this to be used 'in the wild'. I've left these TODOs in the code.
- Should we allow TLS encryption (not authentication) when an invalid certificate was used? @AlexGidman introduced a `strictCertificateValidation` setting, that can be used to explicitly configure Openfire either way. In the `ExpectedOutcome` I'm now supporting that setting, removing many of the TODOs. Note that the actual test execution always use `true` value for this setting. It is hard to configure the setting to a different value for each endpoint of the conversation.

I've also added some references to documentation where the calculated decisions are made of.

xmppserver/src/main/java/org/jivesoftware/util/StreamErrorException.java
guusdk: @Fishbowler text can be null. I'm not sure if it ever will be null, but it is allowed to be null. To avoid future breakage, I thought it best to code defensively here.

@GregDThomas You're on this thread because I've asked you to review this PR back when it was a single commit. ;)

I'm not a fan of the alternative annotation - it's more complex in my mind. That's probably a lot of personal preference going into this.

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
guusdk: Yes. The connection listener that gets mocked returns trust material that is specific to each test.

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
guusdk: There is no `close()` on that instance. There's a `disconnect()` that is called elsewhere, but probably should be called here for consistency. I've pushed a commit with that change.

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalOutgoingServerSessionTest.java
guusdk: Yes, they're installing test-specific trust material. Tests fail badly without them.

xmppserver/src/test/java/org/jivesoftware/openfire/session/RemoteInitiatingServerDummy.java
guusdk: I'm happy to let this one linger for now. :)

distribution/src/conf/openfire-demoboot.xml
guusdk: I don't think so, but it's a nasty hack. The best approach would be fixing OF-2592.

xmppserver/src/main/java/org/jivesoftware/util/StreamErrorException.java
Fishbowler: @guusdk If `getText()` can be null, do we need to add more defensive code, since `getText().isEmpty()` could NPE?

distribution/src/conf/openfire-demoboot.xml
guusdk: I've now created https://github.com/igniterealtime/Openfire/pull/2222 to address OF-2592. If that gets merged before this PR, we could drop this hack.

Fishbowler: Merged. Reverted the hack & rebased.

PR #2166, opened 2023-01-12
OF-2563: Replace Session status constants with enum values

guusdk:


xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpSession.java
```diff
@@ -1475,9 +1475,6 @@ public String toString()
             "address=" + getAddress() +
             ", streamID=" + getStreamID() +
             ", status=" + getStatus() +
-            (getStatus() == STATUS_AUTHENTICATED ? " (authenticated)" : "" ) +
-            (getStatus() == STATUS_CONNECTED ? " (connected)" : "" ) +
-            (getStatus() == STATUS_CLOSED ? " (closed)" : "" ) +
```
akrherz, line 1480: I'm curious about this removal?

guusdk: Previously, `getStatus()` would return a numeric value. To allow humans to make sense of that, that was followed by a text explaining what the numeric value means, by adding a string between parentheses after the numeric value.

With the change in this commit, getStatus will return an enum value, which shows as a human readable text.

akrherz (approved):

guusdk:
@GregDThomas provided feedback on these changes in https://github.com/igniterealtime/Openfire/pull/2165 (in which I also included these changes).

PR #2168, opened 2023-01-25
OF-2565: Should drop stream when client sends stanzas prior to resource binding

guusdk:
RFC 6120, section 7.1 dictates:
> If, before completing the resource binding step, the client attempts to send an XML
> stanza to an entity other than the server itself or the client's account, the server MUST NOT process the
> stanza and MUST close the stream with a `<not-authorized>` stream error.

Openfire, prior to this commit, would reply to pretty much all stanzas that are sent prior to resource binding with a stanza (not stream) error, and would not terminate the stream when the stanza is addressed to an entity other than the server itself or the client's account.

With this commit, stanzas that match the RFC definition will now cause the stream to be closed with a stream error. The prior behavior will remain in place for all other stanzas.

Arguably, there is room for further optimization:
- There appears to be duplication of processing in IQ-, Message- and PresenceRouter. However, they are invoked by other instances than those of SessionPacketRouter (which may in itself be something that's desirable to change)
- XmppWebSocket appears to duplicate part of the processing of data that is exchanged during session setup. Comments have been added for future improvement.
- The Openfire API makes it hard to distinghuish between 'client authenticated & completed resource binding' and 'client authenticated, but did not complete resource binding yet'. There are no explicit handling for the latter scenario, which again could be improved upon in the future.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/IQRouter.java
@@ -104,6 +104,10 @@ public class IQRouter extends BasicModule {
                         "jabber:iq:register".equals(childElement.getNamespaceURI()) ||
                         "urn:ietf:params:xml:ns:xmpp-bind".equals(childElement.getNamespaceURI())))) {
                 handle(packet);
+            } else if (SessionPacketRouter.isInvalidStanzaSentPriorToResourceBinding(packet, session)) {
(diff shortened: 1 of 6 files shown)


guusdk:
These changes are subject of a new integration test that is added in https://github.com/igniterealtime/smack-sint-server-extensions/pull/1

akrherz (approved):

PR #2169, opened 2023-02-07
Refactor WebSocket implementation

guusdk:
The PRs in this commit refactor the WebSocket implementation (and a bit of the other implementation, to facilitate this effort). The goal is to bring the way in which we process data to/from websockets more similar to how we're doing that with regular TCP sockets.

This should:
- make maintenance a bit easier (by doing the same stuff in more places);
- fix OF-2566 to restore Stream Management 'resume' functionality for websockets.

Please refer to commit messages of the individual commits for more details.


gjaekel:
Dear Guus,
inside the Pàdé plugin there's Reverse Websocket Proxy, too. It is used to forward the internet traffic to the Jitsi Ideo Bridge. It's function isto ease the "termination" of TLS in the one hand (instead of providing the certificate store to the JVM used for the JVB) and to avoid to need a dedicated port from the "outside" point of view for the connection handled by this. Because this port would be an additional requirement to use and e.g. must be open on all the "incoming" and "outgoing" firewalls between the OpenFire instance and the meeting participants.

Since a couple of month I observed problems with this Proxy: (Wrong?) Backend connections between the Pàdé plugin and the JVB are dropped for unknown reason. In addition, the corresponding frontend connection isn't closed. Therefore, the user's browser client don't get aware of this and don't recover. It still is able to send into void and didn't receive any message anymore.

The connection is used to interchange and spread the bandwith estimations and video resulutions between all participants and the bridge. By loosing this messages, the video streams will drop down to the lowest resolution or will be even switched off "to save bandwidth".

I'll do a look on this PR; maybe we can't take something here to enhance the reverse websocket proxy's code, too.

guusdk:
After testing, it appears that the changes in this PR also prevent OF-2473 (a deadlock). I have added a commit that removes the workarounds that were put in place to fix OF-2473.

akrherz (approved):

PR #2170, opened 2023-02-08
OF-2568: Improve Stream Management roll-over detection

guusdk:
When a client Ack's with a value _lower_ than the previous Ack'ed value, this should be interpreted as it Acking a stanza count that rolls over (2^32)-1. This very likely should cause the stream to be closed.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
@@ -174,6 +174,13 @@ public class StreamManager {
                 break;
             case "resume":
                 long h = new Long(element.attributeValue("h"));
+                if (h < 0) {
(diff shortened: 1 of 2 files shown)


gjaekel:
Dear Guus,
what to think about that the value must be *serious lower* than the previous, e.g half the integer range? This might offer the chance to distinguish this usecase from a short-term "inversion of time" caused by some delayed packages.

guusdk:
I'm not sure what you mean? What exactly are you suggesting to change?

XMPP dictates that stanzas are sent in order, so if you've seen a particular stanza, you can (must) assume that you've seen everything before it, too. Also, the `h` value in stream management is not identifying stanzas - it's only a stanza _count_. That really cannot go down at all, unless it's wrapped.

gjaekel:
Then it's all right already! I'm not familiar with XMPP. I just compare to IP, where you can not rely on that packages will the receiver in the "natural" order. But this was a wrong thought: XMPP is at an level above, therefore one CAN rely on that the IP level has done this.

akrherz (approved):

PR #2174, opened 2023-02-27
OF-2573 Add Name to Version in Session Summary

Fishbowler:
Tidies up the implementation from OF-2192, adding the name of the software being used, as well as the version.

Openfire appears to allow for other properties in the map created from the pairs received in the IQ, but Name and Version are the only props required by XEP-0092.

### DIFF
--- a/xmppserver/src/main/webapp/session-row.jspf
+++ b/xmppserver/src/main/webapp/session-row.jspf
@@ -43,10 +43,21 @@
     </td>
     <td style="width: 15%; white-space: nowrap">
         <% if (sess.getSoftwareVersion() != null) {
-            String version = sess.getSoftwareVersion().get("version");
-            if (version != null) { %>
-        <%= StringUtils.escapeHTMLTags(version) %>
-        <% } } %>
+            final String softwareName = sess.getSoftwareVersion().get("name");
+            final String softwareVersion = sess.getSoftwareVersion().get("version");
+
+            String softwareString = "";
+            if(softwareName != null && !softwareName.isBlank()){
+                softwareString += softwareName + " - ";
+            }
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/webapp/session-row.jspf
```diff
@@ -43,10 +43,21 @@
...
+            final String softwareName = sess.getSoftwareVersion().get("name");
+            final String softwareVersion = sess.getSoftwareVersion().get("version");
+
+            String softwareString = "";
+            if(softwareName != null && !softwareName.isBlank()){
+                softwareString += softwareName + " - ";
+            }
+            if(softwareVersion != null && !softwareVersion.isBlank()) {
+                softwareString += softwareVersion;
+            };
```
guusdk, line 55: If `name` has a value, but `version` has not, this seems to result in a string that ends with a dash. Maybe do this instead?

```suggestion
            String softwareString = "";
            if(softwareName != null && !softwareName.isBlank()){
                softwareString += softwareName;
            }
            if(softwareVersion != null && !softwareVersion.isBlank()) {
                if (!softwareString.isBlank()) {
                    softwareString += " - ";
                }
                softwareString += softwareVersion;
            };
```

guusdk (commented):
This can probably be improved to avoid the chance of having a dangling ` - `.

xmppserver/src/main/webapp/session-row.jspf
Fishbowler: Yep, you're absolutely right. I spent so long chasing down an NPE that occurs after the session is established but before the version info is fully available, that I completely missed this.

guusdk (approved):

PR #2184, opened 2023-03-06
CI: run on ubuntu-latest

akrherz:
We had hard coded to run on `ubuntu-18.04` due to perceived issues with python+epoll on later ubuntu releases.  Well, this runner is going away at GHA by the end of the month, so lets see what `ubuntu-latest` yields us now.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -51,8 +51,7 @@ jobs:
   aioxmpp:
 
     name: Execute aioxmpp-based CI tests
-    # ubuntu-latest may have issues with Epoll python events
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


guusdk:
You've added the 4.7.5 milestone to this. Note that the Main branch, against which this PR is raised, has already dropped support for Java 8 (which also affects the Github CI action definitions). That's not the case for the 4.7 branch.

akrherz:
Since we still run CI on the 4.7 branch, we need to update to `ubuntu-latest` there as well, since `ubuntu-18.04` is used there.

guusdk:
Yup, I'm not trying to argue differently. I just wanted to point out that the test that is _this_ PR might not be representative of what's going to happen on the 4.7 branch.

akrherz:
OK, that's fair.

PR #2187, opened 2023-05-08
OF-2594: When locating Openfire Home, consider 'tmp' file

guusdk:
It has been observed that under certain (unclean) circumstances, the shutdown of Openfire can leave the `conf` directory without an `openfire.xml` but with an `openfire.xml.tmp` file.

Openfire already has code that allows it to recover an XML configuration file form a `tmp` file.

In this commit, the code that attempts to locate the 'home directory' of the Openfire installation is modified to look not just for `openfire.xml` but also `openfire.xml.tmp`. This allows Openfire to restart after it was shut down uncleanly.


Fishbowler (approved):
LGTM

akrherz (approved):

PR #2188, opened 2023-05-16
OF-2600: Upgrade Jetty 9.4.43.v20210629 to 10.0.16

viv:
Remaining Issues:
- JSP compilation failure caused by: https://github.com/apache/tomcat/commit/224fb6c06528180213b6af28b28dee44029565d7#diff-ef02fd9c3f90ca4838d3cdaa051489556eca75d537fc396022e0ed456c24d895R329

- `WebSocketClientConnectionHandler.onConnect` assumes that `session.getRemoteAddress()` supplies an `InetSocketAddress` which might not always be true. What's the best approach in our context?

Fixed in this commit:
- Websocket server Maven artifact renamed from `websocket-server` to `websocket-jetty-server`
- `WebSocketServlet` renamed to `JettyWebSocketServlet`
- `WebSocketServletFactory` renamed to `JettyWebSocketServletFactory`
- `GzipHandler` is no longer part of `ServletContextHandler` https://github.com/eclipse/jetty.project/issues/4765
- Renamed `setWebInfClassesDirs` to `setWebInfClassesResources` https://github.com/eclipse/jetty.project/issues/4506
- Split `SslContextFactory` into Client and Server https://github.com/eclipse/jetty.project/issues/3464

See: https://www.eclipse.org/jetty/documentation/jetty-10/programming-guide/index.html#pg-migration-94-to-10


guusdk:
@GregDThomas didn't you do the last Jetty upgrade? Do you remember any hidden dragons?

GregDThomas:
My recollection is that it was fairly straightforward - I think one API changed so had to tweak the loading of plugins in some way, but nothing major. But that was a minor version update, looks like this major version update is more substantive :(

guusdk:
With @viv's latest tests:
- Admin console seems to work
- Admin console pages provided by plugins that were compiled using an older version still works.
- Connecting an XMPP client using BOSH works

GregDThomas:
Just remembered that the file org.eclipse.jetty.util.WebAppLoaderFix can probably be removed now - IIRC it only worked on older versions of JRE/Jetty, so it's probably long redundant now.

Originally created to fix https://igniterealtime.atlassian.net/browse/OF-1522 in https://github.com/igniterealtime/Openfire/pull/1228
(with comment "Note; we may need to revisit this come Java 9.")

viv:
Rebased onto latest from the main branch and updated to latest Jetty 10 version.

PR #2189, opened 2023-05-18
OF-2606: Prevent surplus DB connection errors during setup

guusdk:
When an invalid database connection string is provided during setup, the PluginMonitor tries to use it to read properties until it is corrected.

This leads to database connection errors being logged every time that the monitor runs (once every 30 seconds). It's mostly benign, but annoying.

With this commit, the PluginMonitor does not try to read from the database until setup has been completed (apart from the admin plugin, no other plugins are loaded during setup anyway).


Fishbowler (approved):
Tested. With a duff postgres connection string, it did the correct 11 attempts and stopped.

PR #2193, opened 2023-06-08
Revert "Limit aioxmpp's python to 3.11.2 "

guusdk:
This reverts commit 63e2e12d32c34ba11d855e73411fc74fce0f54f3.

https://github.com/horazont/aioxmpp/issues/391 suggests that with the new Python 3.11.4, the problem that occured in 3.11.3 has been resolved. If that's the case, we no longer have to pin to 3.11.2.


guusdk:
Sadly, the `setup-python@v4` action still uses Python 3.11.3. Maybe give it a couple of days and try again.

PR #2207, opened 2023-06-26
OF-2610 - Update truststore and script its update

Fishbowler:
Script pulls the latest set of certificates from Mozilla's authoritative list (as opposed to the old way, that used an installed version on Debian). 

Also includes an output from having run the script to update to the latest version. This could be added to CI, or we could set up a cron Action to generate and commit it.

Whilst it's a good set, I don't like the if/elseif/elseif/elseif chaos of trying to find a "good" name to import it as (the old way used filenames on disk). I can't find a better way. 

Ideally, you'd just trust the commonName, but they don't even all have one!


guusdk:
For future reference, can you please mention the JIRA issue in at least one commit message?

akrherz (approved):

PR #2208, opened 2023-06-27
OF-2615: Avoid direct usage of ConnectionManagerImpl

guusdk:
Openfire’s `ConnectionManager` module has one implementation: `ConnectionManagerImpl`. The two are often used interchangably.

Where possible, Openfire should use the interface, rather than the implemention. This can be easily achieved by defining often-used methods in the interface definition. If these methods are that important, that should be reason enough to define them in the interface contract.

The benefit of this approach is that mocking a `ConnectionManager` becomes easier.


guusdk:
I do not expect this commit to introduce any functional changes.

guusdk:
Asking for forgiveness rather than permission (I need this for a rebase).

PR #2211, opened 2023-07-04
OF-2622: Do not accept inbound Server Dialback when disabled

guusdk:
If the Server Dialback feature is disabled, Openfire should not allow peers to authenticate with that authentication mechanism.

Additionally, Openfire should not define the corresponding XML namespace when the feature is disabled, as other servers might use that to determine support.


Fishbowler:
Fixed in #2165

PR #2212, opened 2023-07-04
OF-2620: Use Openfire assets on Plugin pages

Fishbowler:
… rather than force them to duplicate

There's only one of these that I can find. And it's a 1 character change.

### DIFF
--- a/xmppserver/src/main/webapp/WEB-INF/classes/META-INF/tags/admin/infoBox.tagx
+++ b/xmppserver/src/main/webapp/WEB-INF/classes/META-INF/tags/admin/infoBox.tagx
@@ -7,7 +7,7 @@
         <table>
             <tbody>
                 <tr>
-                    <td class="jive-icon"><img src="images/${type}-16x16.gif" alt=""/></td>
+                    <td class="jive-icon"><img src="/images/${type}-16x16.gif" alt=""/></td>
                     <td class="jive-icon-label">
                         <jsp:doBody />
                     </td>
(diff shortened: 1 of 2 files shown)


guusdk:
It used to be possible to have the admin console be exposed on a URL that is not the context root of the domain (eg: `https://example.org/internal/` using URL rewriting magic). Although I'm not sure if that's even possible anymore, the change in this PR would most definately break it. Is there another way that we can achieve this, without turning the URL into something that's absolute rather than relative?

Fishbowler:
I haven't found a way through code or config to expose it on another URL, but I assume you could expose it via a reverse proxy.
A quick search of the codebase shows that there aren't any links that would be broken by that, but there are a few images. This would make 8, but one of the existing ones is the logo on the main page.

There's some server-side hackery to say "from the root, even if that's a directory", but that wouldn't help a reverse proxy.

I could put a safe number of instances of  `../` before the image reference to ensure it's getting to the root, but that's _more_ likely to break a reverse proxy.

I think the right solution is to rely on a reverse proxy being correctly configured for URL rewriting.

Fishbowler:
This now:
* shows info images in an infobox on plugin pages
* removes some console errors on plugin pages about not finding CSS

guusdk:
Ohh, what's this?

image

Fishbowler:
Rebased.

Fishbowler:
Fixed the path, but can't be tested until we fix OF-2668

guusdk:
Given that someone else just posted the exact same fix, I'm confident enough that this does more good than harm. I'm merging as-is.

PR #2213, opened 2023-07-06
OF-2623: Migrate LoginLimitManager's properties to SystemProperties

guusdk:
Introduces type-safety and makes properties be self-descriptive.

A side-effect of this change is that some of the properties no longer require a restart for changes to be applied.


Fishbowler (approved):
Quick test shows that I can still very quickly get locked out of my local instance :)

PR #2214, opened 2023-07-10
OF-2630 - Get parameter uses incorrect method

iarwen:
This param "encrypt" is in request form body, not in request attribute.


Fishbowler (approved):
That's an excellent spot! Thanks!

PR #2218, opened 2023-07-17
OF-2627: Correctly URL-escape group names in admin console

guusdk:
This prevents issues with working with group names that have URL-escaped characters in their name (like the `+` character).


Fishbowler (approved):
Checked with group names:

test+test
test/test
test test
test"test
test()test
test@test
test😳test

PR #2220, opened 2023-07-18
OF-2559: Introduce Netty for S2S & C2S

viv:
Initial attempt at introducing Netty for S2S and C2S connections. Still to do are:

- Implement `NettyConnectioHandler.reconfigure()`
- Properly handle the idle state
- Consider impact on stats counter

Needs extensive testing but should now support TLS 1.3.


pom.xml
```diff
@@ -121,6 +121,7 @@
         <jetty.version>9.4.43.v20210629</jetty.version>
         <standard-taglib.version>1.2.5</standard-taglib.version>
         <mina.version>2.2.1</mina.version>
```
guusdk, line 123: Let's drop the `mina.version` property.

xmppserver/src/main/java/org/jivesoftware/admin/PluginFilter.java
```diff
@@ -27,9 +26,10 @@
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * A servlet filter that plugin classes can use to dynamically register and un-register filter logic.
+ * A servlet filter that plugin classes can use to dynamically register and un-register filter logic.<p>
```
guusdk, line 32: Uck - can you revert the HTML-only changes in Javadoc please?

xmppserver/src/main/java/org/jivesoftware/openfire/spi/MINAConnectionAcceptor.java
```diff
@@ -1,15 +1,9 @@
 package org.jivesoftware.openfire.spi;
```
guusdk, line 1: Given that this class appears to be MINA-specific, shouldn't it be dropped completely?

PR #2222, opened 2023-07-20
OF-2592: autosetup: Do not force default DB when authprovider is default

guusdk:
When autosetup is running, existing code forced the database connection provider to be the default, when the authprovider was the default.

I'm not seeing why this is needed.

The result of this is that after running auto-setup (with a default auth provider config), all starts of Openfire except for the first one are likely to fail, when the original setup used something else than the default database provider (eg: the embedded provider).


Fishbowler (approved):
Any checks I'd do on this have already been done by CI :)

PR #2224, opened 2023-07-27
Documentation: add Load Balancing Guide

guusdk:
Rendered preview: https://htmlpreview.github.io/?https://github.com/igniterealtime/Openfire/blob/6431308e58fda3f3f6f643faa3bafc3cf3efb83f/documentation/load-balancing.html (this link points to a very specifc commit and is likely to outdate pretty fast).

### DIFF
--- a/documentation/index.html
+++ b/documentation/index.html
@@ -61,6 +61,9 @@
             <dt><a href="separating-admin-users-guide.html">Separating Administrative Users Guide</a></dt>
             <dd>A guide to setting up Openfire to work with different user stores for admins and non-administrative users.</dd>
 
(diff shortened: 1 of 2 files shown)


akrherz:
Thank you.  The only thing I can think about adding is a discussion of how bosh/httpbind is typically not exposed as 7443, but with some haproxy/nginx proxy listening on 443 and sending the traffic openfire's way.  I am unsure if you want to get into that discussion?

guusdk:
@akrherz That is a good idea, thank you. I've force-pushed a change to include a small new section on that (I've also updated the preview URL in my first comment). Can you think of more?

akrherz (approved):

PR #2225, opened 2023-07-30
CI/CD: Add Mellium-based integration test to Github workflow

guusdk:
This adds a very basic Mellium-based test to the Github CI workflow.

Mellium defines 'proper' integration tests, but its framework for IT is a bit different than what we're used to. Were we typically start a server, and tell a framework to start testing against it, the Mellium framework has hooks where it starts clean servers on demand. Our existing scripts do no facilitate that. To get a ball rolling, this commit introduces a Mellium 'test' that uses an examplebot that's included with Mellium, and make it connect against a running Openfire instance.


Fishbowler:
Worth a rebase and try again against vLatest?

PR #2226, opened 2023-08-01
OF-1913 Various S2S Interop Issues 

AlexGidman:
https://igniterealtime.atlassian.net/browse/OF-1913

- Fixes to and from attributes in stream initialisation stanzas
- Creates a new session and a new stream ID following TLS negotiation for inbound S2S.
- Detect new stream ID's on outbound S2S connections and move to new session

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
@@ -42,6 +42,12 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+/**
+ * Stanza handler for responding to incoming stanzas when the server is acting as the client in an S2S scenario.
+ *
+ * @author Alex Gidman
+ * @author Matthew Vivian
+ */
 public class RespondingServerStanzaHandler extends StanzaHandler {
 
     private static final Logger LOG = LoggerFactory.getLogger(RespondingServerStanzaHandler.class);
@@ -84,6 +90,12 @@ public class RespondingServerStanzaHandler extends StanzaHandler {
         return doc.element("dialback") != null;
     }
 
+    /**
+     * Checks if XML string is start of stream; if it is, extract namespaces for parsing followign stanzas, and create
+     * a new session if a new stream ID is detected (e.g. following TLS negotiation).
+     * @param xml The XML stanza to verify
+     * @return true if start of stream, else false
+     */
     @Override
     protected boolean isStartOfStream(String xml) {
         final boolean isStartOfStream = super.isStartOfStream(xml);
@@ -104,7 +116,10 @@ public class RespondingServerStanzaHandler extends StanzaHandler {
                 // following TLS negotiation
                 String newStreamId = rootElement.attribute("id").getValue();
                 if (sessionCreated && newStreamId != null) {
-                    session = createLocalOutgoingServerSession(newStreamId, connection);
+                    ServerSession.AuthenticationMethod existingAuthMethod = session instanceof LocalOutgoingServerSession
+                        ? ((LocalOutgoingServerSession) session).getAuthenticationMethod()
+                        : null;
+                    transferConnectionToNewSession(newStreamId, existingAuthMethod);
                 }
             } catch (DocumentException e) {
                 LOG.error("Failed extract additional namespaces", e);
@@ -112,6 +127,21 @@ public class RespondingServerStanzaHandler extends StanzaHandler {
         }
 
         return isStartOfStream;
+    }
+
+    /**
+     * Transfer an existing connection to a new session when a new streamID is detected. Also transfers existing
+     * authentication method if previously set.
+     * @param newStreamId new stream ID for the new session
+     * @param existingAuthMethod authentication method used previously (possibly null)
+     */
+    private void transferConnectionToNewSession(String newStreamId, ServerSession.AuthenticationMethod existingAuthMethod) {
+        session = createLocalOutgoingServerSession(newStreamId, connection);
+        if (isSessionAuthenticated() && session instanceof LocalOutgoingServerSession) {
+            ((LocalOutgoingServerSession) session).setAuthenticationMethod(existingAuthMethod);
+        } else {
+            LOG.debug("Expected session to be a LocalOutgoingServerSession but it isn't, unable to setAuthenticationMethod().");
+        }
     }
 
     @Override
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
@@ -92,12 +92,20 @@ public class RespondingServerStanzaHandler extends StanzaHandler {
             // Pull namespaces off of the stream:stream stanza and add them to the additional
             List<Namespace> receivedNamespaces = null;
             try {
-                receivedNamespaces = DocumentHelper.parseText(xml + "</stream:stream>").getRootElement().declaredNamespaces();
+                Element rootElement = DocumentHelper.parseText(xml + "</stream:stream>").getRootElement();
+                receivedNamespaces = rootElement.declaredNamespaces();
                 Set<Namespace> additionalNamespaces = receivedNamespaces
                     .stream()
                     .filter(RespondingServerStanzaHandler::isRelevantNamespace)
                     .collect(Collectors.toSet());
                 connection.setAdditionalNamespaces(additionalNamespaces);
+
+                // Create a new session with a new ID if a new stream has started on an existing connection
+                // following TLS negotiation
+                String newStreamId = rootElement.attribute("id").getValue();
+                if (sessionCreated && newStreamId != null) {
+                    session = createLocalOutgoingServerSession(newStreamId, connection);
+                }
             } catch (DocumentException e) {
                 LOG.error("Failed extract additional namespaces", e);
             }
@@ -313,7 +321,18 @@ public class RespondingServerStanzaHandler extends StanzaHandler {
     @Override
     void createSession(String serverName, XmlPullParser xpp, Connection connection) throws XmlPullParserException {
         String currentStreamId = xpp.getAttributeValue("", "id");
-        session = new LocalOutgoingServerSession(domainPair.getLocal(), connection, BasicStreamIDFactory.createStreamID(currentStreamId));
+        session = createLocalOutgoingServerSession(currentStreamId,  connection);
+    }
+
+    /**
+     * Creates a LocalOutgoingServerSession
+     *
+     * @param streamId id taken from the responding servers latest stream tag
+     * @param connection the connection between the servers
+     * @return a LocalOutgoingServerSession
+     */
+    private LocalOutgoingServerSession createLocalOutgoingServerSession(String streamId, Connection connection) {
+        return new LocalOutgoingServerSession(domainPair.getLocal(), connection, BasicStreamIDFactory.createStreamID(streamId));
     }
 
     public void setSessionAuthenticated(boolean authenticated) {
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
@@ -90,21 +90,33 @@ public class RespondingServerStanzaHandler extends StanzaHandler {
         return doc.element("dialback") != null;
     }
 
+
     /**
-     * Checks if XML string is start of stream; if it is, extract namespaces for parsing followign stanzas, and create
-     * a new session if a new stream ID is detected (e.g. following TLS negotiation).
-     * @param xml The XML stanza to verify
-     * @return true if start of stream, else false
+     * Transfer an existing connection to a new session when a new streamID is detected. Also transfers existing
+     * authentication method if previously set.
+     * @param newStreamId new stream ID for the new session
+     * @param existingAuthMethod authentication method used previously (possibly null)
      */
+    private void transferConnectionToNewSession(String newStreamId, ServerSession.AuthenticationMethod existingAuthMethod) {
+        session = createLocalOutgoingServerSession(newStreamId, connection);
+        connection.reinit(session);
+        if (isSessionAuthenticated() && session instanceof LocalOutgoingServerSession) {
+            ((LocalOutgoingServerSession) session).setAuthenticationMethod(existingAuthMethod);
+        } else {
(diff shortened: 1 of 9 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/net/StanzaHandler.java
```diff
@@ -98,6 +97,11 @@ public abstract class StanzaHandler {
      */
     protected PacketRouter router;
 
+    /**
+     * Domain of the local server and remote server or client.
+     */
+    private DomainPair domainPair;
```
guusdk, line 103: Isn't a DomainPair instance per definition server-to-server specific? I'm not sure if it should be used by clients, or, by extension, be used in this abstract class.

xmppserver/src/main/java/org/jivesoftware/openfire/net/StanzaHandler.java
```diff
@@ -484,7 +490,25 @@ protected boolean negotiateTLS() {
...
+        // Get remoteDomain from encrypted stream tag
+        for (int eventType = xpp.getEventType(); eventType != XmlPullParser.START_TAG;) {
+            eventType = xpp.next();
+        }
+        this.domainPair = new DomainPair(XMPPServer.getInstance().getServerInfo().getXMPPDomain(), xpp.getAttributeValue("", "from"));
+
+        // Discard session data obtained from unencrypted initiating stream
+        // in accordance with RFC 6120 § 5.4.3.3. See https://datatracker.ietf.org/doc/html/rfc6120#section-5.4.3.3
+        SessionManager sessionManager = SessionManager.getInstance();
+        sessionManager.unregisterIncomingServerSession(session.getStreamID());
```
guusdk, line 507: This is server-to-server specific. It should not be defined in an abstract class, as that shares this functionality with not onlty the Server-to-Server subclass, but also the Client-to-Server, ExternalComponent and Multiplexer subclasses.

guusdk (changes requested):
This PR fixes server-to-server issues, but applies changes to implementation that is shared by all stanza handlers (including others than the server-to-server one). That's undesirable from a separation of concerns perspective, but it also appears to introduce a problem: the aioxmpp integration test fails on these changes. Aioxmpp is a client library. Its execution _should_ be unaffected by changes to server-to-server functionality.

xmppserver/src/main/java/org/jivesoftware/openfire/net/StanzaHandler.java
AlexGidman: Moved to ServerStanzaHandler

xmppserver/src/main/java/org/jivesoftware/openfire/net/StanzaHandler.java
AlexGidman: Moved to ServerStanzaHandler

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
```diff
@@ -84,6 +90,12 @@ private static boolean isDialbackOffered(Element doc) {
         return doc.element("dialback") != null;
     }
 
+    /**
```
guusdk, line 93: Is there another solution to implement this functionality, other than overriding this method?

The original method definition was very simple: check if a provided value defines the start of a stream. With the changes in this override, there is a _lot_ of additional, almost implicit functionality being added.

This typically leads to weirdly interdependant code, that's hard to build further upon and hard to maintain. Can we somehow move the functionality that's now added _in_ this method, to something that's invoked based on the outcome of the method invocation?

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
```diff
@@ -84,6 +90,12 @@ private static boolean isDialbackOffered(Element doc) {
...
+     * Checks if XML string is start of stream; if it is, extract namespaces for parsing followign stanzas, and create
```
guusdk, line 94: Typo in 'followign'.

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
```diff
@@ -92,12 +104,23 @@ protected boolean isStartOfStream(String xml) {
...
-                receivedNamespaces = DocumentHelper.parseText(xml + "</stream:stream>").getRootElement().declaredNamespaces();
+                Element rootElement = DocumentHelper.parseText(xml + "</stream:stream>").getRootElement();
+                receivedNamespaces = rootElement.declaredNamespaces();
                 Set<Namespace> additionalNamespaces = receivedNamespaces
                     .stream()
                     .filter(RespondingServerStanzaHandler::isRelevantNamespace)
                     .collect(Collectors.toSet());
                 connection.setAdditionalNamespaces(additionalNamespaces);
+
+                // Create a new session with a new ID if a new stream has started on an existing connection
```
guusdk, line 115: Maybe add a comment to explain why the new session is created. Future people maintaining this code will need to understand that.

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
```diff
@@ -92,12 +104,23 @@ protected boolean isStartOfStream(String xml) {
...
+                // following TLS negotiation
+                String newStreamId = rootElement.attribute("id").getValue();
+                if (sessionCreated && newStreamId != null) {
+                    ServerSession.AuthenticationMethod existingAuthMethod = session instanceof LocalOutgoingServerSession
+                        ? ((LocalOutgoingServerSession) session).getAuthenticationMethod()
+                        : null;
+                    transferConnectionToNewSession(newStreamId, existingAuthMethod);
```
guusdk, line 122: What happens to the _old_ session? Does that linger somewhere? Is that a potential source of duplication, memory leaks, or even (delayed) errors (after something times out / gets cleaned up and triggers data)?

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerSocketReader.java
```diff
@@ -47,7 +47,9 @@
  * by changing the property <b>xmpp.server.processing.max.threads</b>.
  *
  * @author Gaston Dombiak
+ * @deprecated see {@link ServerStanzaHandler}
  */
+@Deprecated
```
guusdk, line 52: This entire class is probably eligble to be removed under OF-2635 right?

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerStanzaHandler.java
```diff
@@ -132,6 +150,48 @@ void startTLS() throws Exception {
         //needed ? Connection.ClientAuth.needed : Connection.ClientAuth.wanted
         connection.startTLS(false, false);
     }
+
+    @Override
+    protected String getStreamHeader() {
+        StringBuilder sb = new StringBuilder(200);
```
guusdk, line 156: Nit: There's no need to initialize the capacity of the StringBuilder. The theoretical improvement in speed that this gives is negated by the fact that changes in the variables (and/or future code modifications that neglect to update this value) won't make this value ever 'correct'. The biggest impact of this is that it might confuse someone in the future.

Maybe even replace this entire ting with simple String-concatination? An IDE should help you with that.

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerStanzaHandler.java
```diff
@@ -132,6 +150,48 @@ void startTLS() throws Exception {
...
+        sb.append(" id=\"").append(session.getStreamID()).append("\"");
+        sb.append(" xml:lang=\"").append(session.getLanguage().toLanguageTag()).append("\"");
+        sb.append(" version=\"").append(Session.MAJOR_VERSION).append('.').append(Session.MINOR_VERSION);
+        sb.append("\">");
+        return sb.toString();
+}
+
+    @Override
+    protected void tlsNegotiated(XmlPullParser xpp) throws XmlPullParserException, IOException {
+        // Discard domain values obtained from unencrypted initiating stream
```
guusdk, line 174: This code is server-to-server specific (with usage of DomainPair and ServerSession-usage), but the specification in the RFC that you're referncing hold true for all TLS negotations. Should something similar be built for the other StanzaHandlers?

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
```diff
@@ -110,11 +110,11 @@ public class LocalIncomingServerSession extends LocalServerSession implements In
      */
     public static LocalIncomingServerSession createSession(String serverName, XMPPPacketReader reader,
             SocketConnection connection, boolean directTLS) throws XmlPullParserException, IOException {
-        return createSession(serverName, reader.getXPPParser(), connection, directTLS);
+        return createSession(serverName, reader.getXPPParser(), connection, directTLS, false);
     }
 
     public static LocalIncomingServerSession createSession(String serverName, XmlPullParser xpp,
-                                                           Connection connection, boolean directTLS) throws XmlPullParserException, IOException {
+                                                           Connection connection, boolean directTLS, boolean doNotSendXMPPStream) throws XmlPullParserException, IOException {
```
guusdk, line 117: This doNotSendXMPPStream argument is added, because it's needed for 'recreating' the session object after TLS has completed, to get rid of any stale data in the old session, right? (The open `stream` element has already been sent at that stage).

Instead of trying to juggle with new Stream instances to work our way towards the goal of having a clean session object, would it be more straightforward to instead create a new method on the Session class/interface, that, if invoked, resets that state. It could be named very explicitly `resetState`. With that, there would be no need to try and force a new Session instance. That would avoid the need for error-prone nuances, like this additional argument only-to-be-used-in-a-very-specific-case, and the porting over of the authenticationMethod that was used in the new `transferConnectionToNewSession` method. That in turn would lead to code that's easier to understand and maintain.

There is a case to be made that with a new Session object, there simply is _no way_ that state from an earlier session survives, but I'd argue that having a very explicit 'reset' method, and what I expect the improvement in readability of the code, are advantageous enough to warrent the change.

guusdk (commented):
Thanks Alex. I've left a bunch of comments, that are leading up to a suggestion to rethink the approach. Maybe start with my comment on `LocalIncomingServerSession` (which was the last comment that I typed).

Note that my suggestion is very much up for debate. I'm happy to jump in a call.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
AlexGidman: I think a resetState method is harder to implement in this case given the way sessions are registered and unregistered. I think the problem lies in that the current method for starting a session stream is inside of LocalIncomingServerSession (whereas, we use  dedicated stanza handler logic or the NettySessionInitialiser for outbound connections). LocalIncomingServerSession should not be sending XMPP stanzas - this is a hangover from the old-old implementation pre-MINA.

I don't particularly like passing in a boolean to change behaviour of the function, but you could argue it already has a load of side effects already - when createSession is called, the session should be created and returned. The calling object should then choose whether to send the opening stream stanza or not.

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerStanzaHandler.java
AlexGidman: Possibly, but the ticket scope is for S2S issues.

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerStanzaHandler.java
AlexGidman: fixed

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
AlexGidman: Good spot - connection.reinit needed I think so that no references to the session are persisted

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerSocketReader.java
AlexGidman: Yes. Feel free to remove!

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
AlexGidman: Agree - I've moved this code to the initiate stream method.

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
AlexGidman: Removed

xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
AlexGidman: Added

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerStanzaHandler.java
guusdk: Fair. I've raised https://igniterealtime.atlassian.net/browse/OF-2645 to track this as a future improvement.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
guusdk: I think you're right. The sending of data and the session creation should probably be two separate things. This probably holds true for all types of Sessions. This probably warrants its own refactoring process (which likely goes hand-in-hand with OF-2645).

For now, I refactored just the local inbound server-to-server handling, which gets rid of that new argument (the double negation in there really bugged me :) ). Hopefully, this can be a starting point for a larger refactoring (which will go in a different PR).

guusdk (approved):

guusdk:
There appears to be a chicken and egg issue between this PR and https://github.com/igniterealtime/Openfire/pull/2232

I'll merge the latter (even though it's currently failing), and rebase this on the that. From there, we can hopefully fix issues in both PRs.

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
guusdk: My changes caused failures. I've reverted them for now. The commit is still in, for future reference.

PR #2227, opened 2023-08-01
OF-2559: Do not assume that all connections have established TLS

guusdk:
This intends to guard against null pointer exceptions.

With the new Netty code, I'm seeing null pointer exceptions on the admin console (see screenshot).

Although this PR might prevent the NPE, there migth be a deeper issue that is not addressed by this PR. At the time that this screenshot was taken, Openfire did not have any active server-to-server sessions (as verified by the `netstat` command).

image


AlexGidman:
Sensible checks, no reason this can't go in

AlexGidman (approved):

PR #2228, opened 2023-08-03
OF-2625 - accept self-signed certificates for both encryption and authentication

AlexGidman:
Accept self-signed now implicitly accepts any certificate for authentication.

Encryption requires that the certificate is valid (not expired), and we only authenticate using a self-signed cert after encryption. This does rely on the handshake verifying the cert for us.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
@@ -261,8 +261,10 @@ public class SASLAuthentication {
             if (trusted != null && session.getDefaultIdentity() != null) {
                 haveTrustedCertificate = verifyCertificate(trusted, session.getDefaultIdentity());
             }
-            if (haveTrustedCertificate) {
+            boolean acceptSelfSignedCerts = JiveGlobals.getBooleanProperty(ConnectionSettings.Server.TLS_ACCEPT_SELFSIGNED_CERTS);
+            if (haveTrustedCertificate || acceptSelfSignedCerts) {
                 // Offer SASL EXTERNAL only if TLS has already been negotiated and the peer has a trusted cert.
+                // OF-2625 - peer certificate is implicitly trusted if accept self-signed certificates enabled
                 final Element mechanism = result.addElement("mechanism");
                 mechanism.setText("EXTERNAL");
             }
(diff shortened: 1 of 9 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
```diff
@@ -261,8 +261,10 @@ public static Element getSASLMechanismsElement( LocalIncomingServerSession sessi
             if (trusted != null && session.getDefaultIdentity() != null) {
                 haveTrustedCertificate = verifyCertificate(trusted, session.getDefaultIdentity());
             }
-            if (haveTrustedCertificate) {
+            boolean acceptSelfSignedCerts = JiveGlobals.getBooleanProperty(ConnectionSettings.Server.TLS_ACCEPT_SELFSIGNED_CERTS);
```
guusdk, line 264: If I read this right, then the `EXTERNAL` mechanism will always be offered when the 'accept self signed' option is enabled.

It would be good to ensure that the certificate is indeed self-signed, before offering that mechanism. I think there's existing code that can be used for that, in `CertificateManager.isSelfSignedCertificate()`

xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
```diff
@@ -387,18 +389,15 @@ public static Status handle(LocalSession session, Element doc)
...
-                        {
-                            if ( verifyCertificates( incomingServerSession.getConnection().getPeerCertificates(), saslServer.getAuthorizationID(), true ) )
-                            {
-                                ( (LocalIncomingServerSession) session ).setAuthenticationMethod(ServerSession.AuthenticationMethod.SASL_EXTERNAL);
-                            }
-                            else
-                            {
-                                throw new SaslFailureException( Failure.NOT_AUTHORIZED, "Server-to-Server certificate verification failed." );
-                            }
+                        final boolean shouldVerifyCertificates = JiveGlobals.getBooleanProperty( ConnectionSettings.Server.TLS_CERTIFICATE_VERIFY, true );
```
guusdk, line 392: Does this guard against usage of a (non-self-signed) certificate that's signed by an entity that we do _not_ trust, when we allow for self-signed certificates? Maybe it's good to explicitly check if the certificate is self-signed?

xmppserver/src/main/java/org/jivesoftware/openfire/nio/NettyOutboundConnectionHandler.java
```diff
@@ -113,7 +113,11 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                 // RFC 7590 Section 3.4 writes: "In particular for XMPP server-to-server interactions, it can be reasonable
                 // for XMPP server implementations to accept encrypted but unauthenticated connections when Server Dialback
                 // keys [XEP-0220] are used." In short: if Dialback is allowed, unauthenticated TLS is better than no TLS.
-                if (SASLAuthentication.verifyCertificates(connection.getPeerCertificates(), domainPair.getRemote(), true)) {
+
```
guusdk, line 116: In line with earlier comments in this PR: I'm worried that with this code, the code will accept _any_ certificate when the 'accept-self-signed' option is enabled. We should only trust properly signed ones (as we already do without this option), and self-signed ones.

guusdk (changes requested):
Apart from the review feedback, would you mind splitting off the renaming of the method name in a separate commit? I suspect that somewhere in the future, someone is being hurt by this (as this is somewhat public API). It'd be good to have that as a separate change, instead of being embedded with other changes.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
AlexGidman: The client / initiating server certificate is not sent over and so cannot be evaluated.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
AlexGidman: As above - we do not have the peers certificate for verification.

xmppserver/src/main/java/org/jivesoftware/openfire/nio/NettyOutboundConnectionHandler.java
AlexGidman: If the remote server certificate is self signed we can add it to the local-server (acting as client) trust store for encryption, but this is not sufficient to prompt the local-server (acting as client) to send its certificate over without trying to manipulate the TLS handshake. 

True acceptance of self-signed certs can be achieved if the peers self-signed cert is added to the  trust store prior to initiating a connection (results in true encrypted + SASL S2S). 

We can arguably say that if we are encrypted and are accepting self-signed certs, then the certificate has been validated as part of the TLS handshake and so can skip authentication (effectively it looks like SASL occurs but without verifying certificates again), but this would be less secure than the current implementation.

AlexGidman:
From investigation, it looks like this behaviour may be less secure than the current implementation, and should perhaps be abandoned as not something we should implement.

PR #2229, opened 2023-08-03
OF-2559: EncryptionFactory improvements for Netty

guusdk:
Please review the individual commit messages. The first one fixes a bug in the new Netty implementation, the second one is more of a cleanup.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/EncryptionArtifactFactory.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/EncryptionArtifactFactory.java
@@ -367,6 +367,7 @@ public class EncryptionArtifactFactory
                 break;
         }
 
+        builder.trustManager(getTrustManagers()[0]); // The existing implementation never returns more than one trust manager.
         builder.protocols(configuration.getEncryptionProtocols());
         builder.ciphers(configuration.getEncryptionCipherSuites());
         builder.startTls(!directTLS);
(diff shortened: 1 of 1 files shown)


guusdk:
I did not check if there are methods that were only used by MINA and could potentially be removed.

xmppserver/src/main/java/org/jivesoftware/openfire/spi/EncryptionArtifactFactory.java
```diff
@@ -367,6 +367,7 @@ public SslContext createServerModeSslContext(boolean directTLS) throws Unrecover
                 break;
         }
 
+        builder.trustManager(getTrustManagers()[0]); // The existing implementation never returns more than one trust manager.
```
viv, line 385: Should we also set the keyManager here?
`builder.keyManager(getKeyManagerFactory());`

viv (approved):

AlexGidman (approved):

AlexGidman:
S2S docker tests still working as before with TLSv1.3

xmppserver/src/main/java/org/jivesoftware/openfire/spi/EncryptionArtifactFactory.java
guusdk: I don't think so, as the `keyManagerFactory` is already passed in de factory method of the `SslContextBuilder` isntance, a few lines higher.

PR #2230, opened 2023-08-03
OF-2559: When Inbound S2S session fails, return stream error

guusdk:
This might have been a problem that existed in the code for a long time, but only popped up with the migration from MINA to Netty: that's when ServerStanzaHandler started to being used in anger.

ServerStanzaHandler is invoked to create a session. The code more or less assumes that this always happens.

When a session is not created, the code should inform the peer that something went wrong. Without that, the peer is left hanging.

In this commit, any attempt to create a session that does not lead to a session being created results in a stream error to be returned to the peer, after which the connection is forcefully closed.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
@@ -125,6 +125,20 @@ public class LocalIncomingServerSession extends LocalServerSession implements In
             toDomain = serverName;
         }
 
+        boolean hasCertificates = false;
+        try {
+            hasCertificates = !connection.getConfiguration().getIdentityStore().getAllCertificates().isEmpty();
+        }
+        catch (Exception e) {
+            Log.error("Unable to find any content in the identity store. This connection won't be able to support TLS.", e);
+        }
+
+        if (!hasCertificates && connection.getConfiguration().getTlsPolicy() == Connection.TLSPolicy.required) {
+            Log.error("Server session rejected. TLS is required but no certificates " +
+                "were created.");
+            return null;
+        }
+
         // Retrieve list of namespaces declared in current element (OF-2556)
         connection.setAdditionalNamespaces(XMPPPacketReader.getPrefixedNamespacesOnCurrentElement(xpp));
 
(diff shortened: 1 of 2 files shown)


guusdk:
The change in this PR is the bare minimum needed to make unit tests happy again. I believe that there is room for further improvement here - maybe by refactoring the parent StanzaHandler implementation to return a Session when `createSession` is invoked, or make it throw a more appropriate (Stream)Exception - something like that.

I'm open to expanding this PR with such an improvement. If we choose not to do so now, we should probably log a JIRA issue for this improvement.

xmppserver/src/main/java/org/jivesoftware/openfire/net/ServerStanzaHandler.java
```diff
@@ -24,6 +24,7 @@
 import org.jivesoftware.openfire.session.ConnectionSettings;
 import org.jivesoftware.openfire.session.LocalIncomingServerSession;
 import org.jivesoftware.util.JiveGlobals;
+import org.jivesoftware.util.StreamErrorException;
```
viv, line 27: Unused import?

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
```diff
@@ -125,6 +125,20 @@ public static LocalIncomingServerSession createSession(String serverName, XmlPul
             toDomain = serverName;
         }
 
+        boolean hasCertificates = false;
+        try {
+            hasCertificates = !connection.getConfiguration().getIdentityStore().getAllCertificates().isEmpty();
+        }
+        catch (Exception e) {
+            Log.error("Unable to load find any content in the identity store. This connection won't be able to support TLS.", e);
```
viv, line 133: nit: `load` or `find` but not both (I see that was worded this way prior to your change)

viv (approved):

viv:
> I'm open to expanding this PR with such an improvement. If we choose not to do so now, we should probably log a JIRA issue for this improvement.

I'm keen to get this tested further with the rest of the in-flux Netty work so maybe log and come back to it?

AlexGidman (approved):

viv:
Just spotted the failing tests in CI:

```
Error:    LocalIncomingServerSessionTest.incomingTest:282 expected: <true> but was: <false>
Error:    LocalIncomingServerSessionTest.incomingTest:289 expected: <true> but was: <false>
Error:    LocalOutgoingServerSessionTest.outgoingTest:267 expected: <true> but was: <false>
```

viv:
That middle test failure in my previous comment seems to be failing on other PRs but not the first and third.

Just spotted your comment on: https://github.com/igniterealtime/Openfire/pull/2232#issuecomment-1664530279

Sounds like they're passing when combined.

guusdk:
I force-pushed the changes suggested by @viv, and raised a new ticket (https://igniterealtime.atlassian.net/browse/OF-2634) for desired refactoring of StanzaHandler.

It's indeed likely that tests will perform better once combined with the other changes, as noted in #2232. I'll merge this PR now. Let's see how thing go after we merged all of the related PRs.

PR #2231, opened 2023-08-03
OF-2633: When S2S TLS is required, announce that

guusdk:
When Openfire is configured to require TLS for S2S, the StartTLS feature that’s advertised should be marked as being “required”. It currently is only when Dialback is not available (making TLS needed for authentication purposes).


viv (approved):
Makes sense

AlexGidman (approved):
"Looks Good" - yes sensible addition.

PR #2232, opened 2023-08-03
OF-2559: Use Netty for LocalIncomingServerSessionTest

guusdk:
Prior to this commit, the unit test for LocalIncomingServerSession used the old, pre-Netty code. With this commit, the Netty code is being used instead.

Insteead of the old blocking code, the test now uses a (Netty-backed) ConnectionListener directly. To identify the LocalIncomingServderSession instance that is to be created, a streamID value is used, which is collected by the dummy server peer implementation for that purpose.

### DIFF
--- a/xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
@@ -27,6 +27,7 @@ import org.jivesoftware.openfire.spi.ConnectionManagerImpl;
 import org.jivesoftware.openfire.spi.ConnectionType;
 import org.jivesoftware.util.JiveGlobals;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -37,6 +38,8 @@ import org.mockito.junit.jupiter.MockitoExtension;
 import java.io.File;
 import java.io.IOException;
 import java.net.ServerSocket;
+import java.security.Key;
+import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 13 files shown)


guusdk:
I'm slighly concerned by the removal of the now-unused `ConnectionConfigurationAnswer` method in the test fixture. Shouldn't that be needed?

For me locally, all these tests are green when this PR is combined with the changes from #2229, #2230 and #2231.

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
```diff
@@ -40,6 +38,8 @@
 import org.mockito.stubbing.Answer;
```
AlexGidman, line 38: Unused imports:

`import org.mockito.stubbing.Answer;`
`import org.mockito.invocation.InvocationOnMock;`
`import org.jivesoftware.openfire.spi.ConnectionConfiguration;`

AlexGidman (commented):
Changes make sense: tests not passing without other changes mentioned so should probably wait until they have also been merged.

Can we delete SocketAcceptThread, BlockingAcceptingMode, SocketAcceptingMode, and ServerSocketReader now? It looks like these are all no longer used.

Some elements of the former implementation (SocketConnection etc) are still used for ServerDialback and other features of Openfire (perhaps these can be ticketed in OF Jira?)

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
```diff
@@ -106,43 +106,9 @@ public void setUpClass() throws Exception {
         sessionManager.initialize(xmppServer);
         doReturn(sessionManager).when(xmppServer).getSessionManager();
 
-        final ConnectionManager connectionManager = Fixtures.mockConnectionManager();
-        final ConnectionListener connectionListener = Fixtures.mockConnectionListener();
-        doAnswer(new ConnectionConfigurationAnswer(identityStoreConfig, trustStoreConfig)).when(connectionListener).generateConnectionConfiguration();
```
viv, line 111: I looked into your concern about removing `ConnectionConfigurationAnswer`. What I've seen is that `ConnectionConfigurationAnswer` was used to mock the call to `generateConnectionConfiguration()` on the `ConnectionListener`. 

With your new commit you are supplying the configuration previously defined in the `ConnectionConfigurationAnswer` directly to `ConnectionListener`, when `generateConnectionConfiguration()` is called it uses the config you passed in:

```
connectionListener = new ConnectionListener(ConnectionType.SOCKET_S2S,
                ConnectionSettings.Server.OLD_SSLPORT,
                ConnectionManagerImpl.DEFAULT_SERVER_SSL_PORT,
                ConnectionSettings.Server.ENABLE_OLD_SSLPORT,
                "xmpp.server.processing.threads",
                null,
                JiveGlobals.getProperty(ConnectionSettings.Server.TLS_POLICY),
                ConnectionSettings.Server.AUTH_PER_CLIENTCERT_POLICY,
                null,
                identityStore.getConfiguration(),
                trustStore.getConfiguration(),
                ConnectionSettings.Server.COMPRESSION_SETTINGS);
```

Not mocking the `ConnectionListener` makes sense to me in the context of an incoming server test.

viv (approved):

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
guusdk: Yeah, you're probably right. I was mostly worried about 'loosing' the ability to configure TLS-specifics - but I guess we can do that to the properties (as is done with most other configuration in this same test).

guusdk:
I very much want the old networking code to be removed, but didn't want to do it as part of this PR. As per @AlexGidman's suggestion, I have raised a new JIRA ticket for the removal effort: https://igniterealtime.atlassian.net/browse/OF-2635

The review suggestions were applied, and this PR has been rebased on the latest main.

guusdk:
My optimizations seem to have improved the LocalIncomingServerSessionTest (unexpectedly), but have also made the LocalOutgoingServerSessionTest fail its 13'th test - but only if it is run as part of multiple tests. I'm guessing that I've somehow made tests interdependant (which is bad).

guusdk:
Ah, I suspect that the new tests were failing off and on again from a race condition, not because session state was carried over. The race condition is fixed in commit 8f2f4f2

guusdk:
This PR is running into issues that are likely addressed in another PR: #2226. That PR is running into issues fixed by this PR.

To resolve this chicken/egg problem, I'm going to merge this PR, and rebase the other PR.

PR #2237, opened 2023-08-20
OF-1573: Add support for XEP-0352: Client State Indication

guusdk:
With this commit:
- the 'active' and 'inactive' commands of CSI are recognized by Openfire;
- the feature is advertised to clients;
- the current client state is displayed on the session details in the admin console;
- sessions that are resumed through Stream Management are automatically activated.

No actual optimizations have been implemented as of yet.

The implementation is loosely based on the implementation of Stream Management.

This new feature can be disabled by setting the property `xmpp.client.csi.enabled` to `false`


guusdk:
It is somewhat silly to merge this as-is: although the framework of a feature is now implemented (it can be toggled on and off), it doesn't do anything yet.

Ideally, some kind of optimization is added, before this is merged. Note that many easy-to-think-of optimizations break the XMPP specification (which is quite strict with regards to stanza delivery order).

A feasible optimization is a delay queue (of a certain capacity, with a given maximum delay) that would be used to store stanzas for later delivery, until an 'important' stanza comes along and pushes everything through.

guusdk:
In my second commit, I've added an optimization for inactive clients: unimportant stanzas are being delayed when a client is inactive.

Very basic testing succeed, but what remains to be tested is how this behaves when an inactive client gets disconnected and reattached through stream management. The intention is that the delayed stanzas are sent after session resumption, and the client is marked as being 'active' (the latter is being dictated by the XEP).

PR #2243, opened 2023-08-31
OF-2647: Remove deprecated code

guusdk:
This removes all code that was marked as being deprecated, and to be removed in or after release 4.8.0

Most of the removal revolves around making references to files more type-safe, by moving from String to Path.


guusdk:
I have dismissed the four CodeQL check failures. They all relate to uncontrolled (depending on user-provided value) data being used. However, a check exists (it was added as part of OF-1886) that should prevent abuse.

PR #2255, opened 2023-09-05
OF-2652 - Fix max buffer size error

AlexGidman:
The current code does not set the 'maxBufferSizeExceeded' flag to true until the parser.read function is called. This PR changes the boolean logic such that the NettyXMPPDecoder will release the byte buffer and return if the max buffer size is exceeded, preventing the error from being thrown

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/nio/NettyXMPPDecoder.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/nio/NettyXMPPDecoder.java
@@ -35,12 +35,10 @@ public class NettyXMPPDecoder extends ByteToMessageDecoder {
         // Get the XML parser from the channel
         XMLLightweightParser parser = ctx.channel().attr(NettyConnectionHandler.XML_PARSER).get();
 
-        // Check that the buffer is not bigger than 1 Megabyte. For security reasons
+        // Check that the stanza constructed by the parser is not bigger than 1 Megabyte. For security reasons
         // we will abort parsing when 1 Mega of queued chars was found.
         if (parser.isMaxBufferSizeExceeded()) {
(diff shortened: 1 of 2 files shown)


guusdk:
I do not quite understand this change. Is this resolving an issue where the old code incorrectly identified a state in which to much data was received, or does this no longer throw an exception when to much data is received?

When a remote entity is sending to much data, it is desirable to very verbosely kill the connection.

AlexGidman:
> I do not quite understand this change. Is this resolving an issue where the old code incorrectly identified a state in which to much data was received, or does this no longer throw an exception when to much data is received?
> 
> When a remote entity is sending to much data, it is desirable to very verbosely kill the connection.

Agreed - exception is now still thrown by the parser, this is then caught by the decoder and closes the connection. The maxBufferSizeExceeded check in the NettyXMPPDecoder is there in case any stray messages make there way to the decoder, and it will be discarded (though this is not likely to happen as the connection close should be a blocking process).

guusdk:
This sound s like an improvement to the detection of parsing stanzas that are oversized. I'm not sure if that is addressing the problem at hand though.

The primary concern around OF-2652 is that Openfire, with the new Netty network stack, might be _falsely detecting_ a never ending stanza. Hitting that maximum capacity should basically never happen with a buffer-size of 1MB (unless malicious users are active). Since the introduction of Netty, we're seeing it be reported regularly. That suggests a bug, somewhere.

guusdk:
To clarify, this commit does not prevent any issue that would lead to a buffer overflow, but if we have errors decoding the XML we now explicitly close the connection; what was happening we think is the error gets passed up the pipeline and doesn't get handled, the connection may remain open and that causes further errors (IllegalReferenceCount errors etc)

This PR explicitly catches and handles errors thrown in the handler in a way that we would expect (log it, close the connection and clear up any resources).

PR #2256, opened 2023-09-05
OF-2658 Netty dynamic channelhandler mods

guusdk:
This adds API to dynamically add and remove Netty ChannelHandler instances, which allows custom code to inject functionality in the Netty pipelines.

An example use case would be a logger, like the XML Debugger plugin, that dumps traffic that's sent and received through Netty.


guusdk:
I'm not confident that this is the cleanest implementation. Time constraints, and the desire to use this to revive the XmlDebugger plugin, to use it to track down issues, has made me go for the pragmatic approach at times. If you can suggest improvements, please do.

viv (approved):

Fishbowler (approved):
I don't pretend to know Netty innards, but patterns & logic look sensible

PR #2259, opened 2023-09-06
OF-2540: Update slf4j and log4j2

guusdk:
This updates the slf4j logging dependency from 1.7.36 to 2.0.9.

As no compatible log4j2 binding is available for the version of log4j2 that was used (2.17.1), this dependency has also been updated to 2.20.0.


Fishbowler (approved):

guusdk:
YUNOMERGE?

PR #2260, opened 2023-09-06
OF-1378: Renamed much of 'SSL' to 'TLS' (including 'legacy mode' to 'Direct TLS')

guusdk:
This is intended to be mostly a cosmetic change, but as labels are also renamed, this could introduce unexpected, undesired functional changes.


i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -2620,9 +2620,6 @@ ssl.settings.client.custom.mutualauth.bosh=Mutual authentication (BOSH connectio
 ssl.settings.client.plaintext.boxtitle=Plain-text (with STARTTLS) connections
 ssl.settings.client.plaintext.info=Openfire can accept plain-text connections, which, depending on the policy that is configured here, can be upgraded to encrypted connections (using the STARTTLS protocol).
 ssl.settings.client.plaintext.label_enable=Enabled
-ssl.settings.client.legacymode.boxtitle=Encrypted (legacy-mode) connections
-ssl.settings.client.legacymode.info=Connections of this type are established using encryption immediately (as opposed to using STARTTLS). This type of connectivity is commonly referred to as the "legacy" method of establishing encrypted communications.
-ssl.settings.client.legacymode.label_enable=Enabled
```
Fishbowler, line 2625: These got removed but not replaced which gives missing strings on connection-settings-socket-c2s.jsp

Fishbowler (changes requested):
One flippant, one actual fix needed.

guusdk:
Thanks! I believe I've now addressed both.

PR #2263, opened 2023-09-18
Various Netty improvements

guusdk:
This is a collection of smaller issues with the new Netty outgoing server-to-server implementation. Please refer to individual commit messages (and their JIRA issues) for details.

PR #2264, opened 2023-09-18
Prevent deadlock in Netty Outbound S2S

guusdk:
This, combined with fixes in https://github.com/igniterealtime/Openfire/pull/2263, seems to help getting S2S outbound connections established.

I'm not sure why though, and if by doing this, something important is neglected.

I got the idea of doing this, as I observed that when an outgoing session was being created to another Openfire server, a thread seemed to linger for a _long_ time waiting on obtaining a lock. I suspect a deadlock-like scenario.

I feel that more investigation is needed here, but I wanted to share this finding.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/RespondingServerStanzaHandler.java
@@ -100,11 +100,11 @@ public class RespondingServerStanzaHandler extends StanzaHandler {
     private void transferConnectionToNewSession(String newStreamId, ServerSession.AuthenticationMethod existingAuthMethod) {
         session = createLocalOutgoingServerSession(newStreamId, connection);
 
-        // Clear routing table cache of previous sessions for domain pair and re-add it.
-        RoutingTable routingTable = XMPPServer.getInstance().getRoutingTable();
-        routingTable.addServerRoute(domainPair, (LocalOutgoingServerSession) session); // This _replaces_ the pre-existing session in the routing table.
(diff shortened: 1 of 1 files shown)


viv (commented):
This makes sense as far as I can tell. Here's my understanding after chatting it through with @AlexG-Surevine: 

The `authenticateDomain()` method in `LocalOutgoingServerSession` takes care of registering the session after encryption and authentication (assuming that's what's configured) have been completed. 
https://github.com/guusdk/Openfire/blob/207ab49ef21fc493dd5077c1c9b9a558adddad1d/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java#L207

The `transferConnectionToNewSession` method is run in the context of **setting up** a new encrypted and authenticated session (again assuming that's what's configured). 

So during this setup of we don't want to clear routing table cache of previous sessions for domain pair and re-add it as there should be nothing there.

viv:
I've noticed that even with these changes I am seeing seemingly left over parts of S2S connections. I've seen outbound and inbound connections as well as none remain.

viv:
> I've noticed that even with these changes I am seeing seemingly left over parts of S2S connections. I've seen outbound and inbound connections as well as none remain.

Having said that, I've only since seen lingering outbound S2S connections, which is making me doubt whether I actually saw a lingering inbound connection (I'm fairly sure I did as it was unexpected).

I can see from the log that attempts are being made to clean up the lingering outbound connection but the connection remains:

`LocalRoutingTable - ServerCleanupTask is closing an outgoing server session ...`

guusdk:
Rebased to current HEAD of main branch.

PR #2271, opened 2023-09-21
OF-2560: Make RSS load on Admin Console async

guusdk:
This prevents the first page of the admin console to take up to TIMEOUT (2 minutes?) to load, when the server on which Openfire is installed cannot reach our RSS feed.


Fishbowler (approved):
LGTM
Read, but not tested.

PR #2273, opened 2023-09-22
fix: escape asterisk in remote debug command args

viv:
When debugging on an Amazon Linux EC2 instance we've found that the remote debug command failed as the asterisk character needed to be escaped.

@Fishbowler I see from your original commit message that you were working in Docker, I assume it already worked there so I'm not sure if this change breaks it for your Docker scenario.


Fishbowler (approved):
Works on my machine :+1:

PR #2274, opened 2023-09-22
OF-2677: Use UTF-8 character sets when Netty decodes strings

guusdk:
XMPP mandates that UTF-8 must be supported. Netty's default StringDecoder uses the character set that's the default of the server. Force UTF-8 instead.


Fishbowler (approved):
Looks rational

PR #2276, opened 2023-09-22
OF-2675: Make SNI configurable (defaults to off)

Fishbowler:
SNI checks hostnames. Admin is usually addressed via tunnels and via localhost, which fails the SNI check.

This makes SNI configurable (for TLS and for hostname verification via the same setting) and defaults it to false.

This puts the same configuration to Admin and to HTTPBind - we might want to extend this later to separate it.

image


guusdk:
Instead of having no SecureRequestCustomizer when SNI is undesired, does it work if we have one with it turned off?

Fishbowler:
I think that's what I've done - create an instance and turn off the SNI hostname validation if SNI is disabled at the factory level.

guusdk:
... you did.

On my first reading, I somehow ... imagined? ... that there was an if condition there.

PR #2285, opened 2023-10-10
OF-2685: Build Openfire once and reuse for all database tests

Fishbowler:


Fishbowler:
It needs to be installed to be used as a dependency of the Database Updater.
We could try to cache just the required things, but there are deps of deps to be considered.
The alternative is making the updater part of the parent/child pom relationship, but I don't think it's "first class enough".

PR #2287, opened 2023-10-12
OF-2660: Fix DirectTLS for S2S 

AlexGidman:
- DirectTLS now waits for TLS to be negotiated before attempting to start a stream


guusdk:
I'm merging these prior to review from @Fishbowler for us to be able to dogfood this easily.

PR #2289, opened 2023-10-13
fix: OF-2650 - improve strictCertificateValidation failure logging

AlexGidman:
Exception no longer thrown, logs now appropriate level for message and stack trace, and mention remote domain.


Fishbowler (approved):
LGTM
Simple logging change, and unlikely that domainPair would ever be null here.

PR #2299, opened 2023-10-21
OF-2693: Make configurable XML declaration generation

guusdk:
This commit replaces boilerplate code that generates a 'stream'-based XML document (and strips the closing tag) with a new utility method.

In this new method, new configuration is made available:
- `xmpp.xml.xml-declaration.suppress` Controls if an XML declaration is generated before a 'stream' open tag is sent.
- `xmpp.xml.xml-declaration.suppress-newline` Controls if a newline is added after a generated XML declaration and the 'stream' open tag.


Fishbowler (approved):

Fishbowler:
If dom4j changes their default, what happens to users who haven't overridden the default?

Which I guess is the same as: 

How do SystemProperties behave when we alter defaults?

guusdk:
When the default changes without a user having explicitly set a value, they will start using that default. That holds true for both scenarios.

PR #2302, opened 2023-10-23
test: OF-2665 - add unit test for Cache configuration to ensure that …

AlexGidman:
…unlimited and non-expiring caches are configured as such


guusdk (approved):
Nice and concise, I like it. Would you mind adding a copyright header (apache2) and a javadoc comment on the class level?

AlexGidman:
> Nice and concise, I like it. Would you mind adding a copyright header (apache2) and a javadoc comment on the class level?

Done

PR #2303, opened 2023-10-25
OF-2682: concurrent modification exception connection namespaces

guusdk:
The first commit fixes issue OF-2682, by replacing a static field with a non-static field in a new abstract class.

As we now have an abstract base class, the subsequent commits are used to replace some duplicate code in implementations, by moving it to that new base class. The migration of duplicated code is far from complete, but I didn't want to pile on to much for this commit.

Please refer to individual commit messages for a bit more detail on each change.


AlexGidman (approved):
Sensible changes - let's get this up on ignite for testing!

PR #2308, opened 2023-10-26
Fixed behaviour.js

bmatasar:
It should iterate over the `list` of rules instead of the non-iterable Behaviour object.
In several places `for..in` was used to iterate over the values of an array instead of `for..of`.


guusdk:
Thanks for your contribution! What kind of functional problems did this introduce?

bmatasar:
Errors in the console. If the css's are not needed, then there's no point in importing them. If they are, let's do it properly.

PR #2310, opened 2023-10-26
OF-2699: Define optional PacketError on PacketRejectedException

guusdk:
This change allows a PacketError to be added to the exception that is used to reject (incoming) stanzas in PacketInterceptors. This new field is used in the error that's sent back to the sender, allowing for more detailed error reporting.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/interceptor/PacketRejectedException.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/interceptor/PacketRejectedException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004-2008 Jive Software. All rights reserved.
+ * Copyright (C) 2004-2008 Jive Software, 2023 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/interceptor/PacketRejectedException.java
```diff
@@ -94,11 +105,41 @@ public String getRejectionMessage() {
...
+     * Adds an error that will be sent to the sender of the inbound stanza that got rejected.
+     *
+     * When the rejected stanza was a message, this error is added to a message that is sent (possibly also including
+     * the message body as specified by {@link #setRejectionMessage(String)}). When the rejected stanza was an IQ or
+     * Presence stanza, then the error is used in the error response. For these stanzas, this is a different stanza than
+     * the optional Message stanza that is sent when {@link #setRejectionMessage(String)}) is (also) used.
+     *
+     * @param rejectionError An optional error.
+     */
+    public void setRejectionError(PacketError rejectionError)
```
akrherz, line 141: Where is `setRejectionError` called?

guusdk: It is not used by Openfire itself. This is the new API that is the goal of this change, to be used by third parties / in plugins.

akrherz (approved):

PR #2313, opened 2023-10-31
New unit tests for Group

guusdk:
This new unit test verifies behavior of caching and event listeners in relation to operations on a Group.


guusdk:
I wrote this unit test to prove a bug in Openfire. Turns out that that bug didn't exist. Still, having the unit test is valuable.

akrherz (approved):

PR #2314, opened 2023-10-31
OF-2705: Improve routing stanzas to matching full JID

guusdk:
RFC 6121 section 8.5.3 defines how a server should process a stanza that is addressed to a full JID. When a matching resource is found, the specification does not distinghuis between an “available” resource and a “connected” resource. Instead, it refers to an “available resource or connected resource” (this is different for processing of stanzas to bare JIDs, in section 8.5.2).

Openfire’s implementation of routing stanzas addressed to a full JID of a resource for which a route exist should not evaluate the ‘availability’ of the route, as this contradicts the RFC, which doesn’t differentiate between ‘available’ and ‘connected’.


Fishbowler (approved):

Fishbowler:
Checked against RFC - looks legit 👍

PR #2317, opened 2023-11-01
Reduce verbosity in logs due to unloading inactive MUC rooms

guusdk:


xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/LocalMUCRoomManager.java
```diff
@@ -261,7 +261,7 @@ Duration unloadInactiveRooms(@Nonnull final Date cleanUpDate)
             try {
                 final MUCRoom room = get(roomName);
                 if (room != null && room.getEmptyDate() != null && room.getEmptyDate().before(cleanUpDate)) {
-                    Log.info("Unloading chat room (due to inactivity):" + roomName + "|" + room.getClass().getName());
+                    Log.debug("Unloading chat room (due to inactivity): {}", roomName);
```
akrherz, line 264: Did `room.getClass().getName()` provide anything of value before?

guusdk: Yes, before, we had two implementations: a 'local' and a 'remote' room (used in clustering). Somehwere in 4.7.x that got replaced by one implementation to rule them all. That's why I think we can safely drop it from the debug message now: it'll always be the same.

PR #2318, opened 2023-11-01
New Unit test that verify the implementation of ScramSha1SaslServer

guusdk:


guusdk:
As of now, the new unit test fails. I have failed to find out how/why. I've created a PR in the hope that a review by someone else helps to identify the issue.

guusdk:
Ah, found it. I was provisioning the AuthFactory with incorrectly formatted data.

PR #2319, opened 2023-11-01
Admin console Restructuring of 'session' pages

guusdk:
image


Fishbowler:
Nitpick: "Click to" on the button label is redundant, since it's a button.

guusdk:
You're not wrong, but I went with this, as there already was a translation for that particular label available.

Fishbowler (approved):

PR #2330, opened 2023-11-11
OF-2668 / OF-2723: Allow plugins to be compiled against Openfire 4.8

guusdk:
In this commit, a second artifact is produced from the xmppserver module. Apart from the prexisting 'jar' artifact, there now is a 'taglib' artifact.

This taglib artifact is a jar file (that needs to go into our maven repository), that holds the Admin Console taglibs, in a spec-compliant directory structure.

The plugin's parent pom-file is modified to be able to make use of this new artifact.

With this, compilation of a plugin that defins the plugin parent pom as its parent succeeds (although an unhealty amount of errors is logged). The resulting admin console pages appear functional, at least in Openfire 4.8.

### DIFF
--- a/plugins/pom.xml
+++ b/plugins/pom.xml
@@ -368,6 +368,13 @@
             <version>${openfire.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.igniterealtime.openfire</groupId>
+            <artifactId>xmppserver</artifactId>
+            <version>${openfire.version}</version>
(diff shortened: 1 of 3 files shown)


Fishbowler (approved):
Change looks sane

Testing a broken plugin:
* Fails to build when referencing 4.8.0 snapshot due to web compilation
* Builds after `mvn install`ing this branch
* Web assets appear correct when installing the plugin

✅

PR #2333, opened 2023-11-14
OF-2692: Explictly check if S2S Receiving Entity sets StreamID

guusdk:
RFC6120 dictates that during S2S, the Receiving Entity _must_ set a stream ID to the stream header. This commit fails explicitly when that's the case. Prior to this, a NullPointerException was thrown.

It is likely that this commit fixes a symptom rather than the cause of an issue. It appears that the stream-without-ID is sent by Prosody, that's attempting to send a stream error. Openfire might fail to close a socket, after giving up setting up a connection (or something of that nature).


akrherz (approved):
will definitely want to test this out on ignite and see what is found in the wild!

guusdk:
I suspect that the root cause for the issue is tackled in https://github.com/igniterealtime/Openfire/pull/2334

PR #2337, opened 2023-11-15
OF-2730: Change the way channels are closed

guusdk:
As per Alex' suggestion: When closing the channel we switched to using `ctx.channel().close();` rather than calling `close` directly on the context: `ctx.close()` starts to flow through the `ChannelPipeline` from the point of the `ChannelHandlerContext` while `ctx.channel().close()` will start from the tail of the `ChannelPipeline` all the time.


Fishbowler (approved):
The implications of this are beyond me, but the change is trivial that if testing points to this as a problem, we can revert it.

PR #2340, opened 2023-11-15
Revert "OF-317 Subscribe with Response Unsubscribed Causes Roster Push to Responding Client"

guusdk:
Fixes OF-1183 and OF-1224

This reverts commit a9c43db148902f3511f73a1235a252e5a49f4a56.

OF-317 seems to incorrectly state that the observed behavior is a bug. When a subscription request was processed (and rejected), it seems acceptable to have a roster item in both rosters (as long as they're of state 'None').

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/roster/Roster.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/roster/Roster.java
@@ -404,7 +404,7 @@ public class Roster implements Cacheable, Externalizable {
         // broadcast roster update
         // Do not push items with a state of "None + Pending In"
         if (item.getSubStatus() != RosterItem.SUB_NONE ||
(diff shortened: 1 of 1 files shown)


Fishbowler (approved):

akrherz:
I expressed my ignorant doubts about this change in open_chat.  If the subject of the subscription request denies the request, it should not be sent this roster item entry in the future.  This code accomplishes this, I believe.

guusdk:
The reverted commit prevents the roster push to the subject (with a 'None' subscription state)  but it also prevents other changes to be pushed (OF-1183 / OF-1224).

I believe it is reasonable to argue about the desirability of having a 'None' roster item, but the items that were suppressed by the same commit are definitely incorrectly suppressed - there's little argument there.

Reverting that commit replaces a definite bug with a maybe-to-be-improved-upon scenario.

PR #2344, opened 2023-11-15
OF-1382: Admin console should not autocomplete username/password in all forms

guusdk:
Auto-complete is helpful when logging into the admin console, but it is not helpful when creating new users or other forms where usernames and/or passwords are to be provided to create something new (like a MUC room).

### DIFF
--- a/xmppserver/src/main/webapp/group-edit.jsp
+++ b/xmppserver/src/main/webapp/group-edit.jsp
@@ -575,7 +575,7 @@
                         <fmt:message key="group.edit.add_user" />
                     </td>
                     <td class="c1" style="text-align: left; white-space: nowrap">
-                         <input type="text" size="45" name="username"/>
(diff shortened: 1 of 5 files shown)


Fishbowler:
This hasn't had the intended effect

image

guusdk:
It does seem to improve things for me, using Chrome without an external password manager. I no longer get auto-populated fields here (and in the 'password' input for MUC rooms). When I select the input fields, I still get hints from Chrome's stored passwords, but I think that's OK.

Is there any way that you can think of that we can further improve on this? Does your password manager give hints?

Fishbowler:
I'm just using the Chrome in-built password manager. I don't ever seem to get autofills locally, but just seen that I do on Ignite. Perhaps this will fix that?

Either way, it's a good change. Some googling says there's no great way to disable all password manager hints. There's this hack for LastPass, and that hack for Firefox, etc.

Fishbowler (approved):

PR #2346, opened 2023-11-16
OF-1407: Admin console should have a visual indicator when S2S test is running

guusdk:
The S2S test is a potentially long-running process. The admin console should show some kind of indicator that something is happening.

This is my poor attempt at adding a spinner. There will be much better approaches, but this is what my limited experience gets you within the amount of time I'd like to spend on something like this. Feel free to improve on this! :)


Fishbowler (approved):
That's some sweet colour matching too :heart:

PR #2347, opened 2023-11-16
OF-2734: Fix exception on PubSub Node config admin console page

guusdk:


Fishbowler (approved):
Works on my machine :+1:

PR #2348, opened 2023-11-16
OF-880: IQHandlers must comply with RFC 6121 Section 8.5.1 by default

guusdk:
This builds on a previous attempt to add support for IQ handling with regards to RFC 6121 Section 8.5.1 (No Such User), which was added in 15e6c5619d6ade7c40de95a1a7a158b1a9358ff0 and got partially reverted in 637cc81733092275c2726b43ebe603c0cbd438fe

The original fix got reverted, because pre-existing code (in IQHandlers of plugins) depends on non-compliant behavior.

With this new commit, I'm re-introducing the fix, but am adding a mechanism for individual plugins to opt out.


guusdk:
The original discussion around this was had in https://github.com/igniterealtime/Openfire/pull/38

@deleolajide is this approach acceptable to you? It will require the rayo plugin (and any other plugin that depends on non-spec compliant behavior) to override the new method in the IQHandler.

deleolajide:
> The original discussion around this was had in #38
> 
> @deleolajide is this approach acceptable to you? It will require the rayo plugin (and any other plugin that depends on non-spec compliant behavior) to override the new method in the IQHandler.

Yes it is. Rayo XEP has been updated to conform with RFC 6121. Any exceptions or edge cases can be implemented with the override method.

deleolajide (approved):
performNoSuchUserCheck is excellent solution for edge cases

PR #2354, opened 2023-11-17
OF-2738: Include SNI when setting up outbound server-to-server connec…

guusdk:
…tion

When setting up an outbound server-to-server connection, Openfire should send along an indication to what domain name it is attempting to connect. This will help the recipient server to process the connection for the correct vhost when that server is hosting multipe XMPP domains on the same server (ejabberd allows for this, for example).


guusdk:
I've tested this on nett05 against a server known to have multiple vhosts (that failed earlier), and a bunch of other servers that I expect do not, of different make and models (ignite, isode, xmpp.org). All seems fine.

Fishbowler (approved):

PR #2361, opened 2023-11-30
OF-2752: Sessions should be resumable only if closed due to 'unclean' disconnect

guusdk:
Prior to this commit, a session was (stream management) resumable whenever it got closed.

When a session is closed 'cleanly' (eg, with an exchange of `</stream>`), then a session should never be resumed.

This commit attempts to reverse the strategy. It is implied that a session is never resumable, unless it is closed unexpectedly (for example, as caused by a network disconnect).


akrherz:
Is this targeting 4.8.0 ?

guusdk:
I'm not sure yet, depends on how much review/testing we can get in. This feels like a bigger change.

Fishbowler (approved):

PR #2362, opened 2023-12-01
OF-2752: Refactor solution

guusdk:
In the previous commit for this issue, it was assumed that the state of a connection was always known 'good' or 'terminated'. In reality, there are lenghty periods in which Openfire cannot be sure of a connection is terminated, or simply unresponsive.

To avoid clients being disconnected by Openfire without their knowledge, Openfire should always try to send an end-of-stream indication. Without this, clients might assume for indefinite amounts of time that they're connected, while they are not.


Fishbowler (approved):
Read, not tested, but looks good to me. It all plugs together.

akrherz (approved):

PR #2363, opened 2023-12-01
OF-2753: Stream Management / Clustering - terminate on other cluster nodes

guusdk:
When a resumption attempt occurs on a cluster node that does not hold the detached session, terminate that session if it lives elsewhere in the cluster.

The rationale being that the client is unlikely to attempt resumption on the other cluster nodes. Lingering state might affect cluster functionality.

A new system property `stream.management.location.terminate-others.enabled` is introduced that controls this behavior.


akrherz (approved):

Fishbowler (approved):
Looks reasonable. One minor typo.

PR #2367, opened 2023-12-12
OF-2716: Fix Licences

Fishbowler:
This PR should be rebased in, not squashed.

Updates copyright headers on files.

* Adds headers to files that were missing them
* If they had headers, adds Ignite Realtime Foundation where missing
* If Ignite was there, updates the years
* Fixes some typos
* Fixes one file that had the wrong license header altogether

### DIFF
--- a/checkCopyrightHeader
+++ b/checkCopyrightHeader
@@ -172,6 +172,19 @@ fixMissingLicenseHeader() {
     echo "=> Fixed missing license header in $FILE"
 }
 
+fixMissingC() {
+    local FILE=$1
+    sed -i -E "1,5 s/Copyright/Copyright (C)/" "$FILE"
+    echo "=> Fixed missing (C) in $FILE"
+}
+
+fixMissingAllRightsReserved() {
+    local FILE=$1
+    #/192.168.1.2/s/$/ myalias/
+    sed -i -E "/Copyright/s/$/. All rights reserved/" "$FILE"
+    echo "=> Fixed missing 'All rights reserved' in $FILE"
+}
+
 processFiles(){
     # Only git tracked files
     git ls-tree -r main --name-only | grep '\.java$\|\.jsp$' | sort -u | while read line
@@ -188,6 +201,14 @@ processFiles(){
         MISSING_LICENSE_HEADER=0
         MISSING_IGNITE_MENTION=0
         INCORRECT_IGNITE_MENTION=0
+        MISSING_C=0
+        MISSING_RIGHTS_RESERVED=0
+
+        head -5 "$line" | grep -E "Copyright [0-9]{4}" > /dev/null
+        if [ $? -eq 0 ]; then
+            MISSING_C=1
+            echo "Missing (C) in Copyright line: $line"
+        fi
 
         # Check for license
         grep "Licensed under the Apache License" "$line" > /dev/null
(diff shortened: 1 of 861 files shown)


Fishbowler:
Flipping to draft, as Guus pointed out that I've missed all of the JSPs

xmppserver/src/test/java/org/jivesoftware/openfire/streammanagement/StreamManagerTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2023 Ignite Realtime Foundation. All rights reserved.
```
Fishbowler, line 2: I don't understand this one @guusdk.
You added this test file in 2023, but marked it as 2015-2023.
Reckon this was a copy/paste error, or was there some intent?

checkCopyrightHeader
```diff
@@ -0,0 +1,271 @@
...
+            JIVE_ATTRIBUTION=$(getCopyrightAttributionForJive)
+
+        # I'm not sure I'm happy fixing copyright headers for other folk.
+        # Jive already did this themselves in https://github.com/igniterealtime/Openfire/commit/ff0fa13dc3f67ca269da7bceef161e4a49663259
+        # So I'm going to leave this commented out for now.
+        #    grep "Copyright .* Jive Software" "$line" > /dev/null
+        #    if [ $? -eq 1 ]; then
+        #        MISSING_JIVE_MENTION=1
+        #        echo "Copyright header not found in: $line - needs to read: $JIVE_ATTRIBUTION"
+        #    fi
```
Fishbowler, line 214: This is the right decision, right?

xmppserver/src/test/java/org/jivesoftware/openfire/streammanagement/StreamManagerTest.java
guusdk: There unlikely was intent. It is likely a copy/paste thing.

checkCopyrightHeader
guusdk: I'm not sure what the decision is in the first place.

xmppserver/src/main/java/org/jivesoftware/openfire/auth/AuthProviderMapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 IgniteRealtime.org
+ * Copyright (C) 2016 IgniteRealtime.org, 2017-2018 Ignite Realtime Foundation. All rights reserved
```
guusdk, line 2: This happens in various places. I'm not against merging 'IgniteRealtime.org' (which in itself isn't a legal entity anyway - but it is 'us' so I'm happy to apply a change) with 'Ignite Realtime Foundation'.

Fishbowler: I merged the ones that are 2017 onwards, but kept 2016. I used the Ignite Realtime Foundation announcement blog post as the deciding date, which is October 2016, so there will have been commits that assign 'IgniteRealtime.org' from before the Foundation was announced. Playing it safe.

checkCopyrightHeader
Fishbowler: When I was fixing stuff up, I wrote fixes for Jive Software's copyright header too, then commented it out. I decided they took care of that to the extent that they wanted to way back when and we shouldn't touch it, regardless of what git history says.

guusdk:
Is there value in retaining the `checkCopyrightHeader` script? Are we realistically ever going to do this massive exercise again? By adding it to the repository, we're adding more complexity, more technical debt and more things for new developers to perceive as things being difficult-to-grasp.

If there is value in retaining it, I'd prefer it to be places somewhere away from the root directory - but my preference would be to not include it at all.

Fishbowler:
We could put the script somewhere else. We could also put it in a different repo, or maybe a gist. I think we should keep it.

I can imagine running it every few months and catching a few files that we've missed.

I'd like a better approach than that though... We could wrap the script in a Github Action and test that the git diff is zero?

guusdk:
Moving the script to a gist sounds fine to me.

I _really_ dislike additional hurdles for merging commits, and creating a Github Action comes dangerously close to that. I am confident that anything like this will reduce the amount of contributions that we get from third parties.

In the end, I prefer to have a PR from someone that doesn't have an updated copyright header than not having a PR because people give up (or don't even start) because of checks for trivial conventions. This is my concern with anything that uses checkstyle too, for example.

Fishbowler:
How do you feel about an action committing fixes automatically?

guusdk:
> How do you feel about an action committing fixes automatically?

Might be error-prone. I'm not against it, but I'm wondering if it's worth the effort.

PR #2369, opened 2023-12-15
OF-2756 Modify the method for determining JRE version

jqgsninimo:
Previously, the method assumed that the format of the JRE version is: `major.minor.build`. 
However, for `openjdk 21`, the value of `System.getProperty("java.version")` is `21`. 

This caused the previous method of determination to fail, resulting in the following error message:
> Error: Can not proceed with Openfire Setup.
Your current installation fails to meet minimum server requirements - please see the checklist below: ✗ At least JRE 11
✓ Servlet 2.3 Support
✓ JSP 1.2 Support
✓ Openfire Classes
✓ Openfire Home found (/usr/local/openfire)

Therefore, the method was modified to extract the significant digits using the regular expression `^\d+(\.\d+)?` for version comparison. 

The modified method for determination was tested in the following environments:
| **JRE Version** | 21 | 17.0.1 | 11.0.4 | 10.0.1 | 9.0.1 | 1.8.0_66 | 
|:---:|:---|:---|:---|:---|:---|:---|
| **matcher.group(0)** | 21 | 17.0 | 11.0 | 10.0 | 9.0 | 1.8 | 
| **jreVersionCompatible** | true | true | true | false | false | false |

---
In the previous pull request `OF-2756`, instead of using the outer variable 'jreVersionCompatible' as intended, a new variable was mistakenly defined within the try block. This has been corrected.

### DIFF
--- a/xmppserver/src/main/webapp/setup/index.jsp
+++ b/xmppserver/src/main/webapp/setup/index.jsp
@@ -5,7 +5,6 @@
 <%@ page import="java.util.HashMap"%>
 <%@ page import="java.util.Locale"%>
 <%@ page import="java.util.Map"%>
-<%@ page import="java.util.regex.*"%>
 <%@ page import="org.jivesoftware.util.*" %>
 
 <%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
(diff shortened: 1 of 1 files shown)


guusdk:
Thanks for your contribution! I can see no problem with your solution.

Can it be further improved by simplifying it? I do not think we need decimal points any more, so we do not need to use Double. There might be an opportunity to remove the regular expression and use a simple substring for all digits up to the first `.`. This can then be evaluated to be larger than the integer `11`.

jqgsninimo:
Code has been modified based on @guusdk's suggestion.
The version string is now split using a dot `.` as a delimiter, and the first part is parsed as an integer for comparison. 
During testing, if the version does not contain a dot `.`, the entire string is treated as the first part for parsing.

guusdk:
Thank you!

PR #2375, opened 2023-12-22
OF-2761: Prevent NPE when MUC service gets unexpected IQ result

guusdk:
When an IQ Service gets send an IQ result, it generates a null response, which it tries to route. That causes NullPointerExceptions to be logged.

With this commit, such null responses are silently dropped.


akrherz (approved):

Fishbowler (approved):
LGTM. Matches the fallthrough elif at the end.

PR #2386, opened 2024-01-18
OF-2774: Fix reset of 'idle' state of Netty connections

guusdk:
Whenever data is received, the 'idle' state of a Netty connection is to be reset. Prior to this commit, this occured conditionally on a particular handler being set (which was set only for s2s connections). With this change, the idle flag is always reset upon inbound data.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/nio/NettySessionInitializer.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/nio/NettySessionInitializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2023-2024 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


guusdk:
I have not tested this change. @akrherz can you?

akrherz:
Thanks @guusdk.  My test was not successful as it appears the XMPP ping logic still dumps the client connection when the client does not respond to the ping.
```xml
<stream:error xmlns:stream="http://etherx.jabber.org/streams">
<connection-timeout xmlns="urn:ietf:params:xml:ns:xmpp-streams"/>
<text xmlns="urn:ietf:params:xml:ns:xmpp-streams">Connection has been idle and did not respond to a keep-alive check.</text>
</stream:error>
</stream:stream>
```

guusdk:
I've updated the fix, and have tested it locally. This one should do better.

akrherz:
Thanks @guusdk , I have tested the following

:heavy_check_mark: client sends a whitespace every minute, stays connected and receives no ping requests from server

:heavy_check_mark: client sends no whitespace ping, gets ping request from server, does not respond, gets booted.

LGTM!

guusdk:
Excellent, thanks!

guusdk:
Cherry-picked to 4.8 branch.

PR #2393, opened 2024-01-20
OF-2778: Fix carbon copies of groupchat messages

guusdk:
Messages of type groupchat should not be carbon-copied. The previous implementation only prevented carbon copies for messages of type groupchat when they had a MUC child element - which typically does not happen in the wild.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/forward/Forwarded.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/forward/Forwarded.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2016-2024 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Fishbowler (approved):

xmppserver/src/main/java/org/jivesoftware/openfire/forward/Forwarded.java
```diff
@@ -111,41 +149,12 @@ public static boolean isEligibleForCarbonsDelivery(final Message stanza)
...
-            // TODO implement me.
-
-            // A private <message/> from a MUC participant (received from a full JID) to a local user SHOULD NOT be carbon-copied (these messages are already replicated by the MUC service to all joined client instances).
-            if (stanza.getFrom().getResource() != null) {
-                return false;
-            }
+        // A <message/> of type "groupchat" SHOULD NOT be carbon-copied.
+        if (stanza.getType() == Message.Type.groupchat) {
+            return false;
         }
```
Fishbowler, line 158: Nitpicking, but... This looks like one of the cheaper tests. Should it be higher up the list?

guusdk: Meh. Unless we have an indication that this would lead to real-world performance increase, I'd leave it as-is. The performance of this kind of thing often gets unexpectedly affected by JVM optimization anyway.

PR #2397, opened 2024-01-21
OF-2781 & OF-2782: Bugfixes for SerializingCache

guusdk:
The newly created cache gets wrapped (so that its implementation can be changed on the fly, when clustering starts or stops). The previous implementation caused a ClassCastException that is herein prevented.


guusdk:
@GregDThomas would you mind giving this a review, with regards to the implementation of generics here. I'm a bit lost. This code compiles, and passes the unit tests that I've added - but what _is_ the `T` type at this point?

guusdk:
I have now added a fix for a related issue (OF-2782) to this PR.

GregDThomas (approved):
LGTM, surprised we this hasn't been seen before!

Fishbowler (approved):

PR #2398, opened 2024-01-21
OF-2768: Use `null` instead of 'now' when `createTimestamp` or `modifyTimestamp` are not provided

jsattler:
Openfire users have an optional ‘creationTimestamp’ and ‘modifyTimestamp’ property. These are normally obtained from the Openfire database.

When users are provisioned through another `UserProvider` such as `LdapUserProvider`, then these values are often defaulted to ‘now’. That’s not appropriate, as it’s a wrong value. Additionally, these values are optional. If a `UserProvider` does not provide a value, `null` should be used instead.


guusdk (approved):
Thanks for your contribution @jsattler! I've tested it with and without `null` values. Both admin pages (user properties and user summary) show up with the expected content (thanks for going the extra mile and providing a nice i18n label). This looks good to me.

iarwen:
Hi @guusdk @jsattler, after I upgrade to openfire 4.8.1, and on page /user-summary.jsp, I occur below exception, I use hazelcast to build cluster, also use ldapuserprovider on my system.

image

PR #2406, opened 2024-01-26
OF-2788 & OF-2651: Netty threadpool usage

guusdk:
Fixes for two related issues. As with MINA, Netty _should_ use a distinct thread pool for processing data of a particular connection type (OF-2788). The threads that are used should be aptly named (OF-2651)


viv:
I think this was an oversight during the switch to Netty. This change looks good.

PR #2411, opened 2024-02-10
OF-2758: Remove admin flag when deleting user

guusdk:
When a user is being deleted, that user should no longer be marked as an admin.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/admin/AdminManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/admin/AdminManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2005-2008 Jive Software, 2017-2019 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2005-2008 Jive Software, 2017-2024 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


akrherz:
I wonder about needing to protect against removing the only admin account?

Fishbowler:
> I wonder about needing to protect against removing the only admin account?

Agree that's a risk, but that's a different problem from having the adminJids system property drift.

I had a quick look at deleting the last admin, which naturally requires you to be logged in as that admin. There's some specific detection for it:

> Warning! You are about to delete your own user account. Are you sure you want to do this? Doing so will log you out of the system immediately.

And when you confirm the deletion:

```
Exception:
	at org.jivesoftware.openfire.admin.decorators.main_jsp._jspService(main_jsp.java:228)
	at org.apache.jasper.runtime.HttpJspBase.service(HttpJspBase.java:70)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:590)
	at org.eclipse.jetty.servlet.ServletHolder$NotAsync.service(ServletHolder.java:1419)
```

I'll raise a new bug once I've played with it some more.

Fishbowler:
It's not about the last user, but about any admin deleting themselves. Even though there's an error raised, the `admin.authorizedJIDs` is updated appropriately :)

The same error occurs on main too. Created https://igniterealtime.atlassian.net/browse/OF-2796

Fishbowler (approved):

Fishbowler:
Have approved, but I haven't been anywhere the non-Default providers. I can't see a risk, given the abstractions via the provider in AdminManager, but thought I'd flag in case anyone else can think of one.

PR #2420, opened 2024-02-16
OF-2795 / OF-2166: Remove MUC affiliation of a user that is being deleted.

guusdk:
This prevents a new user by the same name to claim the affiliation.

I've not yet found the time to test these changes. I'd appreciate someone stepping up to help out with that.

I am somewhat worried about the performance of this solution in a big cluster full of chat rooms. Suggestions to improve on this are welcome.


Fishbowler:
LGTM. Definitely functions better compared with main.

You can set a permission for a local user before they exist. In demoboot, I can make 'test' an Admin and 'potato' a Member, then create the 'test' user and assume the permissions, but now, when 'test' is deleted the permission goes with it.

^ Not a bug, just an observation on the behaviour of Openfire now.

PR #2424, opened 2024-02-20
OF-2799: MUC occupant management for federated users in a cluster

guusdk:
When clustering, every Openfire cluster node tries to maintain a minimal set of data describing rooms and occupants. The primary reason to do this is to be able to reconstruct data in case of a cluster node unexpectedly dropping out of the cluster. This data is (mostly) managed by the OccupantManager class.

Prior to this commit, the OccupantManager associated every occupant with one or more cluster nodes (by cluster node ID). This commit changes that, in the following way:
- Users of the local/non-federated XMPP domain are associated with exactly one cluster node: the one that they're 'physically' connected to (eg: where their TCP connection terminates).
- Users of non-local/federated XMPP domains are no longer associated with any cluster node in particular.

The prior association of federated users with specific cluster nodes makes little sense. It _was_ based on the node where their server-to-server connection terminated. That, however, is a poor choice. Server-to-server connections may be teared down due to inactivity (without this leading to occupants being removed from a MUC). It is perfectly acceptable for the s2s connection to be re-established to another cluster node. It's also possible to have multiple s2s connections, potentially to different nodes in the local cluster. It is reasonable to assume that _any_ cluster node is able to (re)establish s2s with the domain of a federated user. This removes more reasons for keeping a (set of) nodeID(s) for a federated occupant.

It is assumed that drifts in data consistency are reduced/removed by no longer trying to maintain a occupant-to-clusternode binding for federated users.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MultiUserChatManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MultiUserChatManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2005-2008 Jive Software, 2017-2021 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2005-2008 Jive Software, 2017-2024 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,8 +19,6 @@ import com.google.common.collect.Multimap;
 import org.jivesoftware.database.DbConnectionManager;
 import org.jivesoftware.database.SequenceManager;
 import org.jivesoftware.openfire.XMPPServer;
-import org.jivesoftware.openfire.cluster.ClusterEventListener;
-import org.jivesoftware.openfire.cluster.ClusterManager;
 import org.jivesoftware.openfire.container.BasicModule;
 import org.jivesoftware.openfire.event.UserEventDispatcher;
 import org.jivesoftware.openfire.event.UserEventListener;
@@ -48,11 +46,7 @@ import org.xmpp.packet.JID;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.sql.Connection;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Types;
+import java.sql.*;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
@@ -990,8 +984,9 @@ public class MultiUserChatManager extends BasicModule implements MUCServicePrope
             .map(mucService -> ConsistencyChecks.generateReportForMucRooms(
                 mucService.getLocalMUCRoomManager().getROOM_CACHE(),
                 mucService.getLocalMUCRoomManager().getLocalRooms(),
-                mucService.getOccupantManager().getOccupantsByNode(),
-                mucService.getOccupantManager().getNodesByOccupant(),
+                mucService.getOccupantManager().getLocalOccupantsByNode(),
+                mucService.getOccupantManager().getNodeByLocalOccupant(),
+                mucService.getOccupantManager().getFederatedOccupants(),
                 mucService.getServiceName()
             )).collect(Collectors.toList());
     }
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/cluster/SyncLocalOccupantsAndSendJoinPresenceTask.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/cluster/SyncLocalOccupantsAndSendJoinPresenceTask.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2021-2024 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,7 +36,7 @@ import java.util.Set;
 /**
  * Task that is used by a cluster node to inform other cluster nodes of its local occupants. This is intended to be used
  * by a node that is joining a cluster (to send its local occupants to all other nodes), and by nodes in an existing
- * cluster that detect a new node joining (to send their local occupants to the joining node.
+ * cluster that detect a new node joining (to send their local occupants to the joining node).
  *
  * @author Guus der Kinderen
  */
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MultiUserChatServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004-2008 Jive Software, 2016-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2004-2008 Jive Software, 2016-2024 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -2779,9 +2779,10 @@ public class MultiUserChatServiceImpl implements Component, MultiUserChatService
     @Override
     public int getNumberRoomOccupants() {
         int total = 0;
-        for (final Set<OccupantManager.Occupant> nodeSet : occupantManager.getOccupantsByNode().values()) {
+        for (final Set<OccupantManager.Occupant> nodeSet : occupantManager.getLocalOccupantsByNode().values()) {
             total += nodeSet.size();
         }
+        total += occupantManager.getFederatedOccupants().size();
         return total;
     }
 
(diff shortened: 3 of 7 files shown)


Fishbowler:
The consistency check on /system-clustering-data-consistency-check.jsp appears to fail as soon as a remote participant joins a MUC

Fishbowler:
Another potential oddity.

I've got a cluster of 3 nodes as `xmpp.localhost.example`, plus another standalone server `otherxmpp.localhost.example`.

The following users joined muc1@conference.xmpp.localhost.example:
* user1@xmpp.localhost.example
* user2@xmpp.localhost.example
* user3@otherxmpp.localhost.example

user3 went idle, and sent an Away presence to the room, which was accepted

```
<presence xmlns="jabber:client" xml:lang="en-GB" to="user3@otherxmpp.localhost.example/Spark 3.0.2" from="muc1@conference.xmpp.localhost.example/user3" id="HBWV5-76" type="unavailable">
  <priority>0</priority>
  <show>away</show>
  <c xmlns="http://jabber.org/protocol/caps" hash="sha-1" node="https://igniterealtime.org/projects/smack" ver="WToosYwPOtUOiTdwT+SKfswnBvw="/>
  <x xmlns="http://jabber.org/protocol/muc#user">
    <item affiliation="none" jid="user3@otherxmpp.localhost.example/Spark 3.0.2" role="none"/>
    <status code="110"/>
  </x>
</presence>
```

user3 returned and sent an online presence:

```
<presence xmlns="jabber:client" to="muc1@conference.xmpp.localhost.example" id="HBWV5-78">
  <status>Online</status>
  <priority>1</priority>
  <c xmlns="http://jabber.org/protocol/caps" hash="sha-1" node="https://igniterealtime.org/projects/smack" ver="WToosYwPOtUOiTdwT+SKfswnBvw="/>
</presence>

```

but this was rejected:

```
<presence xmlns="jabber:client" xml:lang="en-GB" to="user3@otherxmpp.localhost.example/Spark 3.0.2" from="muc1@conference.xmpp.localhost.example" id="HBWV5-78" type="error">
  <status>Online</status>
  <priority>1</priority>
  <c xmlns="http://jabber.org/protocol/caps" hash="sha-1" node="https://igniterealtime.org/projects/smack" ver="WToosYwPOtUOiTdwT+SKfswnBvw="/>
  <error type="modify">
    <jid-malformed xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/>
    <text xmlns="urn:ietf:params:xml:ns:xmpp-stanzas">A nickname (resource-part) is required in order to join a room.</text>
  </error>
</presence>
```

guusdk:
I've now addressed the problem with the consistency check. There was no apparent consistency issue: the check itself simply needed to be updated to the new code.

guusdk:
Does the latter occur only with the changes in this commit, or is this a pre-existing condition. While trying to dig into this, I got sidetracked with an issue that I've recorded as https://igniterealtime.atlassian.net/browse/OF-2803

guusdk:
Working on OF-2803 gave me an idea to test the change in this PR. I have performed a test with the `xmpp.server.idle` value set to a 1 minute time-out. This means that all s2s connections are closed after a minute (when there's no data being exchanged).

Using Smack, I've created four clients, each connecting to one of the cluster nodes, and the last one to the federated node. They all join the same MUC room on the cluster.

All clients are silent, with the exception of the client that is connected to the federated room. It sends a message to the MUC every 80 seconds.

In this setup:
- the s2s connection gets teared down after one minute. 
- Twenty seconds later (when the message is sent), a new s2s connection is established. 
- This again gets teared down after 60 seconds
- Twenty seconds later, a new message is sent and a new s2s connection is established.
- gosub 1

I've observed:
- connections being re-established to different cluster nodes
- all MUC occupants remain in the room
- no data inconsistencies are reported

To me, this indicates that with regards to data consistency, the changes in this PR are at the very least not making things worse. I'll repeat this test with the code prior to my changes, to see if it reproduces the problem that this PR intends to fix.

guusdk:
I do believe that the changes in this PR have positive effect.

The test on the _old_ code was equally successful, during the runtime of the test. It was only after I stopped the test clients, that a data inconsistency was reported.

I have repeated that (stopping the test, checking for data inconsistencies) on servers using the code from this PR, and I did _not_ get data inconsistency.

There's a possibility that the timing of the stopping of the test makes some kind of difference - this isn't absolute proof - but it tracks with the intended change.

guusdk:
As @Fishbowler asked: this was the code I used to create the test clients:

```java
import org.jivesoftware.smack.ConnectionConfiguration;
import org.jivesoftware.smack.SmackConfiguration;
import org.jivesoftware.smack.packet.Presence;
import org.jivesoftware.smack.packet.PresenceBuilder;
import org.jivesoftware.smack.tcp.XMPPTCPConnection;
import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
import org.jivesoftware.smackx.caps.EntityCapsManager;
import org.jivesoftware.smackx.muc.MultiUserChat;
import org.jivesoftware.smackx.muc.MultiUserChatManager;
import org.jxmpp.jid.EntityBareJid;
import org.jxmpp.jid.impl.JidCreate;

import java.time.Duration;
import java.util.Set;

public class TestClusteredAndFederatedClients
{
    public static void main(String[] args) throws Exception
    {
        SmackConfiguration.DEBUG = true;
        System.setProperty("smack.debuggerClass","org.jivesoftware.smack.debugger.ConsoleDebugger");

        final XMPPTCPConnectionConfiguration user1 = XMPPTCPConnectionConfiguration.builder()
            .setUsernameAndPassword("user1", "password")
            .setXmppDomain("xmpp.localhost.example")
            .setResource("local-user-1")
            .setHost("localhost")
            .setPort(5221)
            .addEnabledSaslMechanism("PLAIN")
            .setSecurityMode(ConnectionConfiguration.SecurityMode.disabled)
            .build();

        final XMPPTCPConnectionConfiguration user2 = XMPPTCPConnectionConfiguration.builder()
            .setUsernameAndPassword("user2", "password")
            .setXmppDomain("xmpp.localhost.example")
            .setResource("local-user-2")
            .setHost("localhost")
            .setPort(5222)
            .addEnabledSaslMechanism("PLAIN")
            .setSecurityMode(ConnectionConfiguration.SecurityMode.disabled)
            .build();

        final XMPPTCPConnectionConfiguration user3 = XMPPTCPConnectionConfiguration.builder()
            .setUsernameAndPassword("user3", "password")
            .setXmppDomain("xmpp.localhost.example")
            .setResource("local-user-3")
            .setHost("localhost")
            .setPort(5223)
            .addEnabledSaslMechanism("PLAIN")
            .setSecurityMode(ConnectionConfiguration.SecurityMode.disabled)
            .build();

        final XMPPTCPConnectionConfiguration federatedUser3 = XMPPTCPConnectionConfiguration.builder()
            .setUsernameAndPassword("user3", "password")
            .setXmppDomain("otherxmpp.localhost.example")
            .setResource("federated-user-3")
            .setHost("localhost")
            .setPort(5229)
            .addEnabledSaslMechanism("PLAIN")
            .setSecurityMode(ConnectionConfiguration.SecurityMode.disabled)
            .build();

        final XMPPTCPConnection connection1 = new XMPPTCPConnection(user1);
        final XMPPTCPConnection connection2 = new XMPPTCPConnection(user2);
        final XMPPTCPConnection connection3 = new XMPPTCPConnection(user3);
        final XMPPTCPConnection connectionFederated3 = new XMPPTCPConnection(federatedUser3);

        final Set<XMPPTCPConnection> connections = Set.of(connection1, connection2, connection3, connectionFederated3);
        try
        {
            connections.forEach(connection -> connection.setUseStreamManagement(false)); // To reduce output in the debug log.
            connections.forEach(connection -> EntityCapsManager.getInstanceFor(connection).disableEntityCaps());  // To reduce output in the debug log.

            for (XMPPTCPConnection connection : connections) {
                connection.connect();
            }
            for (XMPPTCPConnection connection : connections) {
                connection.login();
            }

            Thread.sleep(Duration.ofSeconds(2).toMillis());

            // Make them all join the same MUC room
            final EntityBareJid roomAddress = JidCreate.entityBareFrom("muc1@conference.xmpp.localhost.example");
            for (XMPPTCPConnection connection : connections) {
                MultiUserChatManager.getInstanceFor(connection)
                    .getMultiUserChat(roomAddress)
                    .join(connection.getUser().getResourcepart()); // Cheating: using the real-JID's resource part as a nickname for the MUC.
            }

            final MultiUserChat multiUserChat = MultiUserChatManager.getInstanceFor(connectionFederated3).getMultiUserChat(roomAddress);
            Presence.Mode nextMode = Presence.Mode.away;
            for (int i=0; i<10; i++) {
                connectionFederated3.sendStanza(
                    PresenceBuilder.buildPresence()
                        .ofType(Presence.Type.available)
                        .setMode(nextMode)
                        .build()
                );
                nextMode = (nextMode == Presence.Mode.available ? Presence.Mode.away : Presence.Mode.available);
                if (i%2==0) {multiUserChat.sendMessage("test message " + (i+1));}
                Thread.sleep(Duration.ofSeconds(40).toMillis());
            }
        }
        finally
        {
            for (XMPPTCPConnection connection : connections) {
                try {
                    connection.disconnect();
                } catch (Throwable t) {}
            }
        }
    }
}
```

guusdk:
My management summary of the above:
- @Fishbowler's first reported issue (reported data inconsistency): fixed
- @Fishbowler's second reported issue: we need to check if this is introduced by this change, or pre-existing behavior (if so, a new issue is ideally created for this
- The changes in this PR appear to fix at least one issue that is reproducible in servers that do _not_ have this PR.

Fishbowler:
The latter problem is a bunch of different problems.

The problem with the second rejection is a Spark bug. It should be sending the nickname when it tries to go online.

I think there might still be a problem with the OccupantUpdatedTask - the representation of nick changes appears inconsistently in Spark for the client attached to the Senior node. On main, the nick changes disappear users from the memberlist for all other users. That's fixed here, but Spark doesn't recognise that they're renames. I haven't dug further into that - I'd want to check the behavior on a non-clustered setup before working out if this another Spark bug.

TL;DR - this seems like an improvement, and I think we should merge it and dogfood it and press forwards (predicated on folk being available to respond?)

Fishbowler (approved):

PR #2431, opened 2024-02-27
OF-2777: Allow groups to be shared with other groups

guusdk:
When a group is to be shared with other groups, this is stored in the database as a group property of name `sharedRoster.showInRoster` having a value of `onlyGroup`. The group property value for `sharedRoster.groupList` controls to what groups this group is shared. It either is a comma-separated list (denoting group names), or it's empty. An empty value signals that the group is shared only with members of the same group.

This commit corrects how these values are managed in the database.

This change largely fixes OF-2777, although one issue remains (or is possibly introduced by this change): When _disabling_ roster sharing for a group that was previously shared with other groups, then contacts (roster items) are NOT removed from the rosters of people that no longer have the shared group.


guusdk:
I've force-pushed changes to my original commit, correcting the fix that was in it. The current state of the code both passes the unit tests (that failed in my original fix) and pass manual testing.

akrherz (approved):

PR #2433, opened 2024-02-29
OF-2808: Fix race condition in connection close / stream management

guusdk:
The Netty connection implementation that replaces the MINA connection asynchronously closes a connection. This causes a race condition in Stream Management: when resuming a stream on a new connection, the close-listeners that are intended to operate on the old connection are sometimes erroneously operating on the new connection (which causes issues).

This commit makes the close method synchronous again, by introducing a latch. A new issue (OF-2811) was raised to address the issue without introducing blocking behavior.


guusdk:
This fix was tested by me and Daniel, the author of Conversations, in the wild.

akrherz (approved):

PR #2434, opened 2024-02-29
OF-2809: Guard against negative max idle time for websockets

guusdk:
When the max idle time is configured to be negative, then do not schedule the check to see if the connection is still active.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/websocket/WebSocketClientConnectionHandler.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/websocket/WebSocketClientConnectionHandler.java
@@ -113,8 +113,8 @@ public class WebSocketClientConnectionHandler
         }
 
         final Duration maxIdleTime = getMaxIdleTime();
+        xmppSessionIdleTask = new XmppSessionIdleTask();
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/websocket/WebSocketClientConnectionHandler.java
```diff
@@ -281,7 +286,7 @@ private final class xmppSessionIdleTask extends TimerTask {
         @Override
         public void run()
         {
-            if (!isWebSocketOpen()) {
+            if (!isWebSocketOpen() || getMaxIdleTime().isNegative() || getMaxIdleTime().isZero()) {
```
akrherz, line 289: Is this correct logic? seems like these should all be !checks

guusdk: No,  I believe this is correct. This tries to _not_ run the check if the socket is closed, or if the max-idle time is non-positive (and thus disabled).

Fishbowler (approved):

PR #2446, opened 2024-02-29
OF-2813: Stash the mvn repo after build, and re-use in later jobs

Fishbowler:
* Use mvn install to get Openfire into the mvn repo
* Stash the repo at the end of the Java 11 build
* Restore the repo before building the Database Updater

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -58,12 +58,12 @@ jobs:
         with:
           name: Coverage Report for 'xmppserver' module
           path: xmppserver/target/site/jacoco/
-      - name: Temporarily save mvn repo for later jobs
+      - name: Temporarily stash openfire artifacts from the mvn repo for later jobs
         if: ${{ matrix.distribution == 'zulu' && matrix.java == 11 }}
         uses: actions/upload-artifact@v4
         with:
           name: mvn-repo
-          path: ~/.m2/repository
+          path: ~/.m2/repository/org/igniterealtime/openfire/
           retention-days: 1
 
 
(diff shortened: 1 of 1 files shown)


Fishbowler:
Would it be useful to restore the repo before the `mvn deploy` step, or would that only be sped up by having the target directory?

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -58,6 +58,14 @@ jobs:
         with:
           name: Coverage Report for 'xmppserver' module
           path: xmppserver/target/site/jacoco/
+      - name: Temporarily save mvn repo for later jobs
+        if: ${{ matrix.distribution == 'zulu' && matrix.java == 11 }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: mvn-repo
+          path: ~/.m2/repository
```
guusdk, line 66: This potentially is a _lot_ of artifacts. Maybe scope it to ignite/jive?

Fishbowler: It's about 150MB, but only held for a day, and is pretty quick since it's internal to GHA

guusdk: I've noticed that 'pretty quick' is on a very variable scale. I suspect that subsequent runners are sometimes close together, but sometimes not. If we _can_ optimize this without to much trouble, it might be worth it.

Another thing that I'm not thinking of: this probably replaces the m2 repo on the runner where this is executed. If we're using caching there to retain its content between runs, that might have become pointless, if we're going to change the entire repo on every run anyway.

Fishbowler: > this probably replaces the m2 repo on the runner where this is executed

Yes, but it takes the output of job 1, and restores it to job 2. Whilst it could persist longer term, it would get overwritten by another job 1 before ever getting to another job 2, so the difference in the clone wouldn't ever make a difference.

Fishbowler: > maven install openfire into an ephemeral folder

That would be rather swish, I'll look into it!

Fishbowler: `mvn dependency:copy-dependencies` gives 129 items, weighing in at just over 50MB. There won't be a sensible prune of the m2 repo that will include everything we need, but exclude the other bits. We could just copy over the igniterealtime bit of the repo, but there might be new downloads off the internet. We'd save most of the copying, but perhaps lose more time in downloads. I'll give it a go and see how it looks.

Fishbowler: Looks tidy. Saves 7-12 seconds overall across all steps in the workflow.

akrherz (approved):

PR #2450, opened 2024-03-09
Always run database tests if we're gonna publish

Fishbowler:
* Have the database test jobs depend on the check_branch job that determines if this run will result in a publish
* Have the database tests run if either the relevant files have changed or the run will result in a publish (previously it was just the first check)
* Have the publish job await the database tests as well as the other tests

The publish now depends on jobs that might not run, but that seems fine, since the database tests will always run if the publish job would run (and at other times too)

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -230,9 +230,9 @@ jobs:
 
   sqlserver:
     name: Test SQL Server Upgrades
-    needs: [build, should-do-database-upgrade-tests]
+    needs: [build, should-do-database-upgrade-tests, check_branch]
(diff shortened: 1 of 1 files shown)


akrherz:
> The publish now depends on jobs that might not run, but that seems fine

This sounds like trouble to me, as the publish job would not run.  What's an example scenario that it could happen?

Fishbowler:
> > The publish now depends on jobs that might not run, but that seems fine
> 
> This sounds like trouble to me, as the publish job would not run. What's an example scenario that it could happen?

Bad wording. Whilst the jobs are optional, now that the database test jobs will always run if it's a run that would publish, the publish _actually_ depends on jobs that will always run.

akrherz (approved):

PR #2452, opened 2024-03-15
OF-2816: i18n Hebrew update

guusdk:
More translations from Schimon. Translation is now 54.64% complete.


Fishbowler (approved):
There's really nothing I could possibly object to here :)

Fishbowler:
Not merged yet - are there more translations incoming?

guusdk:
I just wanted to build to finish before pressing the green button, but forgot about it.

Hopefully, more translations will be forthcoming, but that shouldn't stop us from merging this.

PR #2455, opened 2024-03-27
OF-2818: Netty should not use 5MB buffers

guusdk:
This corrects a mistake where a configuration of having a _maximum_ buffer size was turned into having a _fixed_ buffer size.

Having 5MB buffers is bad, as with a limited amount of messages, a _lot_ of memory is being consumed.

This commit removes the hard-coded buffer size configuration, while retaining the maximum message length directive.

I've tested that, using a websocket client connection:
- a message of 128k characters can be sent using the default configuration;
- a message of 128k characters can not be sent when Openfire is reconfigured to use 4k for the `xmpp.parser.buffer.size` property


guusdk:
Note that I've _not_ tested that this actually gets rid of the off-heap memory usage. I'm not sure how to do that reliably. It probably requires some kind of load testing framework to be active?

Given the analysis that lead up to this change, I'm confident that the change touches on the appropriate areas of the code, and thus is likely to have the intended effect of getting rid of the 5MB byte buffers.

guusdk:
This solution has been deployed to a high-volume instance of Openfire. In a heap dump taken afterwards, the offending buffers are no longer present.

Fishbowler (approved):

PR #2458, opened 2024-04-05
OF-2814: Ensure that multi-byte characters in the XMPP data are parsed properly

guusdk:
Prioer to this change, parsing errors occurred when multi-byte characters where provided to the parser byte-by-byte (or in byte arrays where a multi-byte character was split over more than one byte array).

This commit re-introduces the fix for this problem that was created under OF-458. It uses a CharsetDecoder that will only read 'enough' bytes from an array, leaving bytes that form an incomplete character in the byte array (to be processed in the next iteration, when hopefully more data was added).

The original fix is refactored a little to account for the different API that is provided by Netty (as compared to the old MINA based code), and to facilitate unit testing (of which a couple are added).


guusdk:
I've not tested this with an actual XMPP client yet.

PR #2459, opened 2024-05-08
OF-2824: Addressing data consistency in caches of routing table 

guusdk:
The distinct commits in this PR each intend to address a separate issue, where the first three commits are preparing the code for easier application of the fourth commit.

The over-arching goal of all commits is to make the implementation of RoutingTable less error prone (in context of client sessions/routes). it does so by:
- adding strictness with regards to allowed `null` and bare vs. full JID usage
- simplifying the code (doing away with an optimization that I suspect plays a role in the cause of observed data inconsistency)
- replacing per-cache lock usage to usage of one lock, to guard the manipulation of three related caches. This intends to remove a potential synchronization issue.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -288,7 +288,7 @@ public class RoutingTableImpl extends BasicModule implements RoutingTable, Clust
             }
 
             Log.trace("Adding client full JID {} to users sessions cache under key {}", route, route.toBareJID());
-            // Acquires the same lock, which should not be an issue as the lock implementation (both Openfire's as Hazelcast's) is reentrant.
+            // Acquires the same lock, which should not be an issue as the lock implementation (both Openfire's and Hazelcast's) is reentrant.
             CacheUtil.addValueToMultiValuedCache(usersSessionsCache, route.toBareJID(), route.toFullJID(), HashSet::new);
         }
         finally {
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2005-2008 Jive Software, 2016-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2005-2008 Jive Software, 2016-2024 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -259,6 +259,9 @@ public class RoutingTableImpl extends BasicModule implements RoutingTable, Clust
 
     @Override
     public boolean addClientRoute(JID route, LocalClientSession destination) {
+        if (route.getResource() == null) {
+            throw new IllegalArgumentException("Route is not a full JID: " + route);
+        }
         boolean added;
         boolean available = destination.getPresence().isAvailable();
         Log.debug("Adding client route {}", route);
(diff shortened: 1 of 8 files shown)


akrherz (approved):
Don't see anything obviously wrong with this change.

Fishbowler (approved):

Fishbowler:
I've reviewed this commit by commit, and it seems sensible.
On locks, do those locks that we no longer acquire (Users, AnonymousUsers) get locked and edited anywhere else? i.e. Could this be trading race conditions?

Fishbowler:
Those smack tests though 😬

xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
```diff
@@ -263,51 +271,28 @@ public void addClientRoute(JID route, LocalClientSession destination) {
...
+            if (destination.getAuthToken().isAnonymous()) {
+                Log.trace("Adding client route {} to anonymous users cache under key {}", newClientRoute, route);
+                anonymousUsersCache.put(route.toFullJID(), newClientRoute);
+            } else {
+                Log.trace("Adding client route {} to users cache under key {}", newClientRoute, route);
+                usersCache.put(route.toFullJID(), newClientRoute);
             }
+
+            Log.trace("Adding client full JID {} to users sessions cache under key {}", route, route.toBareJID());
+            // Acquires the same lock, which should not be an issue as the lock implementation (both Openfire's as Hazelcast's) is reentrant.
```
GregDThomas, line 291: Trivial typo - both Openfire's **_as_** Hazelcast's -> both Openfire's **_and_** Hazelcast's

xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
```diff
@@ -921,17 +912,29 @@ public Collection<String> getComponentsDomains() {
 
     @Override
     public boolean hasClientRoute(JID jid) {
-        return usersCache.containsKey(jid.toString()) || isAnonymousRoute(jid);
+        final Lock lock = usersSessionsCache.getLock(jid.toBareJID());
+        lock.lock();
+        try {
+            return usersCache.containsKey(jid.toFullJID()) || isAnonymousRoute(jid);
```
GregDThomas, line 918: I forget how these locks work (it's been a while) but I assume they are re-entrant?

This method 
* invokes `usersSessionsCache.getLock(jid.toBareJID());`
* then calls `isAnonymousRoute(jid)`
  * and this method also invokes `usersSessionsCache.getLock(jid.toBareJID());`

I'd expect any tests to throw up issues, but that bit worries me. It looks wrong on inspection. Perhaps you need a `private boolean isAnonymousRouteUnlocked(JID jid)` that doesn't take out a lock, that this method (and the public one) can take out knowing that it's already locked?

GregDThomas (commented):
One typo, one thing that worries me but I may be off base there.

xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
guusdk: Yes, the locks are re-entrant. I've checked that for both Openfire's own caches, as well as Hazelcast's. Your other comment fixed a typo in a comment that explains that for another case that uses pretty much the same approach. :)

If your concern is optics on inspection, then I'd rather fix that by adding additional code commenting. I think that leads to easier-to-maintain code.

guusdk:
> do those locks that we no longer acquire (Users, AnonymousUsers) get locked and edited anywhere else?

No, I've checked (but cannot programatically ensure) that all interaction with those caches either use UsersSession's lock, or no lock at all (in cases were the cache interaction is not specific to one entry - I'm not sure how to improve on that).

guusdk:
> Those smack tests though 😬

I had introduced fail-fast behavior. Guess what: things started to fail rather fast. I was wrong to assume that the arguments used would always be of the expected type (a full JID rather than a bare JID).

There might be room to improve things, as the invocation with bare JIDs might be a symptom of an underlying issue, but that's potentially a larger refactoring - while it wouldn't fix the same problem in plugins or third-party code.

Given that it appears to be wide-spread, I've now modified to code to, instead of failing immediately, simply prevent an operation that would have a predictable null-result when a bare JID is provided.

Fishbowler:
LGTM 👍

PR #2461, opened 2024-05-17
OF-2827: JDBCAuthProvider should not lowercase/trim provided username

guusdk:
By lowercasing the input, a username with a capital letter can't be used.

The closely related JDBCUserProvider does not lowercase/trim values. It is probably best that both providers treat usernames in the same way.

I'm wondering if we even need a configuration option to make this behavior configurable. Isn't it just plain 'wrong' to have inconsistent processing of the `username` values - that's never going to work, is it?

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/auth/JDBCAuthProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/auth/JDBCAuthProvider.java
@@ -212,7 +212,6 @@ public class JDBCAuthProvider implements AuthProvider, PropertyEventListener {
         if (username == null || password == null) {
             throw new UnauthorizedException();
         }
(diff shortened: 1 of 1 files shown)


Fishbowler:
Seems reasonable to maintain the fidelity of the username. I can see an argument to maintain the trim, but I don't know whether `alice ` is a valid username.

Fishbowler (approved):

guusdk:
> I can see an argument to maintain the trim, but I don't know whether alice  is a valid username.

It likely isn't, but shouldn't that then be applied to where the user provides the input (eg: the login form)?

Fishbowler:
Exactly that. Let's trust the user.

guusdk:
Hehehe, let's not - but lets apply input validation near the input form maybe?

PR #2486, opened 2024-07-03
OF-2825: Moved implementation of Service Discovery security considerations to individual providers

guusdk:
Instead of applying the security considerations of section 8, XEP-0030 globally, they now are applied to the individual discovery data providers. This prevents them to be applied to providers where they do not apply (eg: MUC, that doesn't apply presence subscription based authorization at all).


akrherz:
failures addressed by #2487

PR #2487, opened 2024-07-03
OF-2841: Make CAPS advertised on Stream user specific

guusdk:
Registered users can get different features from the server than non-registered users. This should be expressed in the advertised CAPS.


akrherz:
failures addressed by #2486

PR #2490, opened 2024-07-05
OF-2843 & OF-2844: MUC ban improvements

guusdk:
An admin cannot ban an owner from a MUC room. XEP-0045 section 9.1 specifies:

> [A] user cannot be banned by an admin with a lower affiliation. Therefore, if an admin attempts to ban an owner, the service MUST deny the request and return a `<not-allowed/>` error to the sender

Openfire currently does send an error, but uses the `<conflict/>` condition. Instead, it must use the `<not-allowed/>` condition as specified by the XEP.

XEP-0045 Section 9.1 defines:

> If an admin or owner attempts to ban himself, the service MUST deny the request and return a `<conflict/>` error to the sender. (Note: This is different from the recommended service behavior on kicking oneself.)

Openfire currently allows admins to ban themselves. That should not be possible, as defined by the XEP.


akrherz:
What's the impact of this change on the `muc-room-affiliations.jsp` console page?  Too lazy to test at the moment.

guusdk:
The Admin Console applies changes by generating IQ stanzas that are always executed as if they were sent by someone with the 'owner' affiliation - but without actually using an existing user. That suggests that these changes _should not_ affect that functionality.

As a quick test, I have used the admin console to:

1. create a test room (which makes 'admin' an owner)
2. add a new owner ('john') to the room
3. remove the original owner (leaving just 'john' as the owner).

That all went without issue. Trying to remove 'john' as an owner then throws the 'there always must be at least one owner' error, as expected.

akrherz (approved):

PR #2497, opened 2024-07-13
OF-2848: Ensure `<destroy/>` element in presence unavailable when MUC is destroyed

guusdk:
Prior to this change, this element was present only when one of the optional attributes was provided.


guusdk:
Related clarification of the corresponding XEP: https://github.com/xsf/xeps/pull/1365

akrherz (approved):

PR #2498, opened 2024-07-13
Update aioxmpp testing

akrherz:
This PR:

1. Updates the aioxmpp git repo to its present home
2. Enables all tests that pass locally for me on Openfire 4.8.3

A before and after

```diff
- 46 passed, 4575 skipped, 4 deselected, 133 warnings in 7.01s
+ 4599 passed, 23 skipped, 3 deselected, 188 warnings in 31.61s
```

I have created Jira issues for the three tests currently being skipped and documented those within the CI config.  Watch how many of these tests now fail here :)

PS.  I have brought `aioxmpp` into [conda-forge](https://conda-forge.org/).  So our CI could be updated to make more speedy usage of it, but the difference is likely negligible

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -116,7 +116,8 @@ jobs:
           # OF-2849 test_publish_and_purge
           # OF-2850 test_publish_multiple_and_get_by_id
           # OF-2851 test_convert_field_datetime_default_locale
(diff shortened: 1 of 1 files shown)


akrherz:
I guess only having one failure here is a win

> FAILED tests/muc/test_e2e.py::TestMuc::test_set_topic - AssertionError: None != 'Wytches Brew!' : {<aioxmpp.structs.LanguageTag.fromstr('en')>: None}

[OF-2852](https://igniterealtime.atlassian.net/browse/OF-2852)

guusdk:
Thanks! This seems like an awesome improvement to Openfire's coverage.

On the conda-forge addition: am I wrong to see that not actually being used in this PR?

guusdk (approved):

akrherz:
I didn't convert our CI from using pypi/pip to conda-forge as I didn't think the speedup would have been meaningful, maybe a second or two faster.

PR #2505, opened 2024-08-16
OF-2858: Fix error condition for MUC PM

guusdk:
When sending a private message to a user that's not an occupant of the room, XEP-0045 section 7.5 defines that a different error condition is to be used than the one that Openfire has been using. This commit corrects the error condition.


akrherz (approved):

Fishbowler (approved):
Tests pass. The only other case in §7.5 is a occupant not in the room, but that's also covered by new SINT Extension tests. LGTM.

PR #2506, opened 2024-08-16
OF-2860: The non-breaking changes leading up to renaming the class MUCRole

guusdk:
See individual commit messages for more details.

The commits in this PR are intended to be suitable to be backported and included in a patch release.


akrherz (changes requested):
So we'll need a s/4.9.0/4.10.0/ and s/4.8.4/4.9.0/ done to this PR.

akrherz (approved):
Reviewed to believe that there are no API breaks with this.

PR #2509, opened 2024-08-22
OF-2864: Deprecate org.jivesoftware.util.Base64

guusdk:
Marked `org.jivesoftware.util.Base64` as deprecated, replaced all usage with `java.util.Base64`.

This replaces customer code with code provided by Java itself, reducing the maintenance load.


guusdk:
After a quick discussion with @akrherz we agreed to no longer mark this for removal in 4.9, but for 4.10 instead.

akrherz (approved):

akrherz:
current belief that `MultiUserChatOccupantIntegrationTest.mucLeaveTest` test failure is just flapping, rerunning

PR #2515, opened 2024-08-28
OF-2867: Openfire setup should not use removed code

guusdk:
By removing deprecated code, the Openfire setup failed. Specifically, a validation that checks for some runtime configuration failed, because it no longer could identify the 'home' directory. As this implementation used Java reflection, it went below the radar until now.


guusdk:
Without this change, Openfire setup cannot be started in 4.9.0 and later.

akrherz (approved):

PR #2523, opened 2024-09-17
OF-2873 Improve debian package

stokito:
I added a link to the Transifex to make it easier to find. It would be great to add it also on the website.

In the README I also spitted long lines to separate lines to make it easier to read without any work wrapping.

The debian lintian reports a lot of warnings for the deb package itself and I tried to fix some of them.

The biggest change is in the last commit d530fe9bbea1996b281149c7e5c582d0216e9b8a
The linitan reported: `W: openfire: recursive-privilege-change "chmod -R" [postinst:39]`:

```sh
mkdir -p /var/log/openfire
mkdir -p /var/lib/openfire/embedded-db
chown -R openfire:openfire /usr/share/openfire
chown -R openfire:openfire /var/lib/openfire
chown -R openfire:openfire /var/log/openfire
chown -R openfire:openfire /etc/openfire
chmod -R o-rwx /usr/share/openfire
chmod -R o-rwx /var/lib/openfire
chmod -R o-rwx /var/log/openfire
chmod -R o-rwx /etc/openfire
```

Here we have a few problems:
1. The dirs can be created by the `openfire.dirs` or even better by the `make install` and then deb package will contain them too.
2. The `/usr/share/openfire` should contain things like images and assets so it should be read only for anyone and there should be no any need to chown it to the `openfire`. I removed the chown.
3. The `/var/log/openfire` actually may be removed and logs by default should be printed to stdout or syslog where they can be collected by the systemd-journald and then available from the `journalctl -u openfire -f` command. This is not our business how to persist logs. The Prosody allows both: syslog and the append to a log file. This needs to be discussed separately.
4. The `/etc/openfire` folder is also shouldn't be writable but I see that it may be used for some backup (?). Maybe it would be better to save backups to the `/var/lib/openfire` folder instead.


guusdk:
On a cursory look, I can't spot major issues with this, but I'm not familiar with Debian packaging at all.

akrherz (approved):

PR #2531, opened 2024-09-19
OF-2875 CI: Add build deb job

akrherz:
This adds a step that builds the deb artifact.  It is not published anywhere at the moment, but is good to check that it works without having to mess with bamboo

PR #2532, opened 2024-09-19
Debian build

stokito:
The debian/rules was lack of "catch all" rule so `make clean` was failed because there is no clean target in the Makefile. I fixed that.

Basically the `rules` file is intended to change something in the build of the original package. But here we have an own native package so the rules should be empty with the only catch all target.

Later (maybe) I'l send another PR that will remove the `override_dh_install` target and use instead Makefile or debian/install or even maybe that it would be possible with Maven assembly plugin. Will check.

Other commits reduce warnings and makes a build log more clear

CC @akrherz


akrherz:
Thanks @stokito , could you kindly cherry pick #2531 here, so to see things in action?

akrherz:
>  tar -xf distribution/target/dist/distribution-artifact.tar
will extract into your cwd and not into that dist folder.  `-C` may be of usage

akrherz:
@stokito my suggestion for easier debugging is to add this step to the workflow prior to the step that fails
```yaml
      - name: Setup upterm session
        uses: lhotari/action-upterm@v1
        with:
          limit-access-to-actor: true 
```
watch the build logs and when it gets to that step, the SSH command you need to run will be shown and you can ssh in and manually debug what is happening within CI at that step.

akrherz:
I'll take a hack at this PR now and see how far I can get, thanks for all this effort so far!

akrherz:
sweet, now for the epic dread of having this fail in bamboo for some strange reason

PR #2534, opened 2024-09-21
Debian build fix

stokito:
I broke the deb build with invalid openfire.install file.
Also the changelog was broken and empty in the resulted deb package.
Another problem that we do need to change onwer for the /usr/shared/openfire that I removed chown.
Other changes are to reduce lintian warnings.

Sorry that missed the errors, I had to purge the openfire package fully from a system and tested on the clean setup and only then noticed the problems.

CC @akrherz


xmppserver/src/main/webapp/style/framework/fonts/fontawesome-webfont3e6e.svg
akrherz, line 1: Hmm, why is github's UI indicating that this file changed???

akrherz (approved):

build/debian/build_debs.sh
```diff
@@ -7,7 +7,8 @@ pushd .
 if [ $1 ]; then
     OPENFIRE_FULLVERSION=$1
 else
-    OPENFIRE_FULLVERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+    # extract version from pom.xml
+    OPENFIRE_FULLVERSION=$(grep -oP "<version>(.*)</version>" -m 1 pom.xml | cut -d ">" -f 2 | cut -d "-" -f 1)
```
akrherz, line 11: darn, this doesn't work for non-SNAPSHOT versions...

PR #2536, opened 2024-09-21
Admin panel: Add new user: groups lookup

stokito:
I tried to slightly improve the Admin Console to improve first impression.

When creating a user we may set a group name. I added an autocompletion for it.

Also on the user create form the browser tries to autocomplete username and other fields of the admin itself which is not needed here. I added the `autocomplete="off"` for inputs to fix this but Firefox may ignore it 👎 

There is a task [Admin console design improvements](https://igniterealtime.atlassian.net/browse/OF-1823) and I committed with its ticket number.


akrherz:
Thanks, this PR should not be backported to 4.9 branch, so I've added 324785d to an isolated PR #2537

i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -3245,6 +3245,7 @@ plugin.admin.reload_failure=Unable to reload the plugin. Refer to the log files
 plugin.admin.monitortask_running=The plugin manager is currently running. It may take a short time for the list of installed plugins to be complete.
 plugin.admin.name=Plugins
 plugin.admin.description=Description
+plugin.admin.documentation=Documentation
```
guusdk, line 2807: Did you intend to change the title from `README` to `Documentation`?

guusdk (changes requested):
Thanks for this! Would you mind applying similar changes to the room configuration screen? It's been driving me nuts that it autopopulates the password field from my browser's password manager.

I believe that the layout changes to the plugin pages can be improved a lot by a couple of small changes.

Can you move the header that reads "Description" a bit to the left, so that it aligns on the left side with the icon?

image

In the 'version' column, the version identifier (in the first line) gets centered, which is visually appealing, as the release date on the second line typically has more characters / is wider. There are some plugins where the version identifier has more characters than the release date. In the screenshot below, an example is shown of both. When the release date is shorter than the version identifier, things look a bit funky now. I would suggest centering the release date in such case.

image

I don't like the icon on the first line of the version column. It makes for a messy look. Can you please move it to the right of the version, vertically centered (so that it is in line with all other icons in the row)?

image

One final improvement that I think would be good to have is an aid for users to help them click the right 'install' button. On wider screens it's often a bit hard to make sure that the `+` button that's all the way on the right hand side corresponds with the name of the plugin that is all the way on the left. I think this can be improved in a couple of ways. I would choose one of these:
- Make the background color of each row alternate (white/gray/white/gray...)
- Make the background of a row over which the cursor is located change. I _think_ we had this in the past, where such a row would get a yellow-ish background color.

guusdk:
> Thanks, this PR should not be backported to 4.9 branch, so I've added [324785d](https://github.com/igniterealtime/Openfire/commit/324785db9e755460ab2cbc2caf4739f7e080691c) to an isolated PR #2537

That probably means that the commit should be removed from this PR?

PR #2539, opened 2024-09-22
OF-2877 Reproducible builds

stokito:
Pin the build timestamp and disable Built-By, Created-By and Build-Jdk-Spec fields from generated manifest. The Maven JAR plugin is updated to latest version that has the new option addDefaultEntries

https://igniterealtime.atlassian.net/browse/OF-2877

### DIFF
--- a/i18n/pom.xml
+++ b/i18n/pom.xml
@@ -9,6 +9,10 @@
     <artifactId>i18n</artifactId>
     <name>Internationalization files for Openfire</name>
     <description>These files are shared among the starter and xmppserver modules</description>
+    <properties>
+        <!-- Use static build timestamp for reproducible builds -->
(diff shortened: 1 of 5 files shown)


guusdk (commented):
Can/should we modify the static dates that are introduced in this PR with one that is more tied to the state of the code, like the date of the last commit, or something?

Some of these changes are likely going to confuse future developers ("This date shouldn't be static! Lets make it a variable!"). Can you please add a reference to OF-2877 in the comments that you've already added. That will help people find more of the background of this change.

Fishbowler (approved):
Whilst I don't quite understand the end goal beyond what's stated on the Reproducible Builds website, I similarly have no objections to moving in that direction, assuming we're not making unreasonable compromises for other users' needs along the way.

guusdk:
Thank you!

PR #2540, opened 2024-09-22
plugins parent pom.xml: move maven compiler plugin configuration to <pluginManagement>

stokito:
This will allow to inherit the configuration to plugins.
E.g. target Java version


guusdk (approved):
That looks sensible, thanks!

PR #2541, opened 2024-09-22
OF-2892 Available Plugins page: redesign to make Documentation easier to see

stokito:
The Plugins and Available plugins pages have two icons for documentation (README) and changelog.

The icons are very small and placed together that makes it difficult to understand them.

The icons are only 16px and the only way to understand is to put a cursor over them to see a title. This wouldn't be possible to to from a mobile at all (minor issue).

As a solution we can switch to a table where each cell can have a few lines.

I moved the changelog icon into the Version column as a an additional row. The re

Similarly the Documentation link moved into the Description column, and the Description merged into the Plugin Name.

available plugins

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/update/AvailablePlugin.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/update/AvailablePlugin.java
@@ -37,8 +37,6 @@ import org.slf4j.LoggerFactory;
 public class AvailablePlugin extends PluginMetadata
 {
     private static final Logger Log = LoggerFactory.getLogger( AvailablePlugin.class );
-    private static final DateFormat RELEASE_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
-    private static final DateFormat RELEASE_DATE_DISPLAY_FORMAT = DateFormat.getDateInstance(DateFormat.MEDIUM);
 
     /**
      * URL from where the latest version of the plugin can be downloaded.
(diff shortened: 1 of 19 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/update/AvailablePlugin.java
```diff
@@ -141,15 +139,7 @@ public static AvailablePlugin getInstance( Element plugin )
...
 
-        String releaseDate = null;
-        final String releaseDateString = plugin.attributeValue("releaseDate");
-        if( releaseDateString!= null) {
-            try {
-                releaseDate = RELEASE_DATE_DISPLAY_FORMAT.format(RELEASE_DATE_FORMAT.parse(releaseDateString));
-            } catch (final ParseException e) {
-                Log.warn("Unexpected exception parsing release date: " + releaseDateString, e);
-            }
-        }
```
Fishbowler, line 152: I think this exists because there are some older plugins out there with non-ISO release dates

Fishbowler:
I've not tested this (yet) but this looks to be an awesome change!
Those icons have bothered me massively. I'd love to get this into vNext.

Fishbowler:
I've run this on top of main. Dates render, docs and changelog links work.

There's one rendering bug when installing a plugin.

Before:
image

After:
image

The columns all narrow to display the :heavy_check_mark: for a few seconds, then returns to the before state.

Fishbowler:
Pushed the rebase.

guusdk:
Thanks! I have added a small change to this: I always have trouble tracking the lines of each plugin. To have a bit of a better visual indication, I made the background of a plugin-line change color slightly (from light-grey to white) when the mouse cursor hovers over it.

guusdk:
I don't think that that the failure of the Docker image build is related to changes in this PR. I'm happy to ignore it.

PR #2546, opened 2024-10-04
OF-2786: Happy Eyeballs (for outbound S2S connections)

guusdk:
The commits in this PR implement (much of the) Happy Eyeballs algorithm, as defined in the protoXEP in https://xmpp.org/extensions/inbox/xep-happy-eyeballs.html

The commits in this PR isolate distinct changes, which preferably are not to be squashed together.

This commit also explicitly sets `-Djava.net.preferIPv6Addresses=system` which causes the Java virtual machine to follow the system's preference with regards to IPv6 vs IPv4 usage (as opposed to the default, which is 'IPv4', at least in Java 11. Is there a more appropriate way of going about this?

Some aspects that are excluded from this implementation:
- Separate resolution of A and AAAA records (the Java API does that in one go - it seems silly to bypass that only to adhere to the specifications)
- Much of the 'sorting' as defined in [RFC 8305](http://tools.ietf.org/html/rfc8305) [[3](https://xmpp.org/extensions/inbox/xep-happy-eyeballs.html#nt-idm45857903655040)], Section 4. The implementation in this PR is driven by the SRV priority and weight attributes. Other selection mechanisms may be subject of future enhancement.

I've tested the implementation against various endpoints provided by badxmpp.org (eg `srv20.badxmpp.org` and `drop.badxmpp.org` which give encouraging results.

There likely is room to clean up the implementation through refactoring. One of the reasons that I'm asking for a review of this is to get feedback on how to do that best.

### DIFF
--- a/distribution/src/bin/openfire.sh
+++ b/distribution/src/bin/openfire.sh
@@ -139,7 +139,7 @@ esac
 done
 
 
-JAVACMD="${JAVACMD} -Dlog4j.configurationFile=${OPENFIRE_LIB}/log4j2.xml -Dlog4j2.formatMsgNoLookups=true -Djdk.tls.ephemeralDHKeySize=matched -Djsse.SSLEngine.acceptLargeFragments=true"
+JAVACMD="${JAVACMD} -Dlog4j.configurationFile=${OPENFIRE_LIB}/log4j2.xml -Dlog4j2.formatMsgNoLookups=true -Djdk.tls.ephemeralDHKeySize=matched -Djsse.SSLEngine.acceptLargeFragments=true -Djava.net.preferIPv6Addresses=system"
 
 if [ -z "$LOCALCLASSPATH" ] ; then
 	LOCALCLASSPATH=$OPENFIRE_LIB/startup.jar
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
@@ -156,7 +156,19 @@ public class DNSUtil {
 
         // Use domain and default port as fallback, if that's not already in the list.
         if (results.stream().flatMap(Set::stream).noneMatch(h -> h.getHost().equals(domain) && h.getPort() == defaultPort && !h.isDirectTLS())) {
-            results.add(Set.of(new WeightedHostAddress(domain, defaultPort, false, 0, 0)));
+            results.add(Set.of(new WeightedHostAddress(domain, defaultPort, false, Integer.MAX_VALUE, 0)));
+        }
+        if (logger.isDebugEnabled()) {
+            logger.debug("Answering lookup for domain '{}' from DNS responses. Returning:", domain);
+            for (final Set<WeightedHostAddress> srvLookup : results) {
+                for (final WeightedHostAddress hostAddress : srvLookup) {
+                    if (hostAddress.getPriority() != Integer.MAX_VALUE) {
+                        logger.debug(" - {} (based on a DNS lookup)", hostAddress);
+                    } else {
+                        logger.debug(" - {} (a fallback, based on the XMPP domain and default port)", hostAddress);
+                    }
+                }
+            }
         }
         return results;
     }
@@ -224,7 +236,7 @@ public class DNSUtil {
      * @return An ordered list of results (possibly empty, never null).
      */
     public static List<WeightedHostAddress> srvLookup(@Nonnull final String service, @Nonnull final String proto, @Nonnull final String name) {
-        logger.trace("DNS SRV Lookup for service '{}', protocol '{}' and name '{}'", service, proto, name);
+        logger.debug("DNS SRV Lookup for service '{}', protocol '{}' and name '{}'", service, proto, name);
 
         final String lookup = constructLookup(service, proto, name);
 
@@ -233,12 +245,12 @@ public class DNSUtil {
         if (optional != null) {
             // Return a cached result.
             if (optional.isAbsent()) {
-                logger.warn("DNS SRV lookup previously failed for '{}' (negative cache result)", lookup);
+                logger.debug("DNS SRV lookup previously failed for '{}' (negative cache result)", lookup);
                 result = new WeightedHostAddress[0];
             } else {
                 result = optional.get();
                 if ( result.length == 0 ) {
-                    logger.debug("No SRV record found for '{}' (cached result)", lookup);
+                    logger.info("No SRV record found for '{}' (cached result)", lookup);
                 } else {
                     logger.trace("{} SRV record(s) found for '{}' (cached result)", result.length, lookup);
                 }
@@ -249,7 +261,7 @@ public class DNSUtil {
                 final Attributes dnsLookup = context.getAttributes(lookup, new String[]{"SRV"});
                 final Attribute srvRecords = dnsLookup.get("SRV");
                 if (srvRecords == null) {
-                    logger.debug("No SRV record found for '{}'", lookup);
+                    logger.trace("No SRV record found for '{}'", lookup);
                     result = new WeightedHostAddress[0];
                 } else {
                     result = new WeightedHostAddress[srvRecords.size()];
@@ -257,15 +269,18 @@ public class DNSUtil {
                     for (int i = 0; i < srvRecords.size(); i++) {
                         result[i] = new WeightedHostAddress(((String) srvRecords.get(i)).split(" "), directTLS);
                     }
-                    logger.trace("{} SRV record(s) found for '{}'", result.length, lookup);
+                    logger.trace("{} SRV record(s) found for '{}':", result.length, lookup);
+                    for (WeightedHostAddress weightedHostAddress : result) {
+                        logger.trace(" - {}", weightedHostAddress);
+                    }
                 }
                 LOOKUP_CACHE.put(lookup, CacheableOptional.of(result));
             } catch (NameNotFoundException e) {
-                logger.debug("No SRV record found for '{}'", lookup, e);
+                logger.trace("No SRV record found for '{}'", lookup);
                 LOOKUP_CACHE.put(lookup, CacheableOptional.of(new WeightedHostAddress[0])); // Empty result (different from negative result!)
                 result = new WeightedHostAddress[0];
             } catch (NamingException e) {
-                logger.info("DNS SRV lookup was unsuccessful for '{}': {}", lookup, e.getMessage());
+                logger.info("DNS SRV lookup was unsuccessful for '{}': {}", lookup, e);
                 LOOKUP_CACHE.put(lookup, CacheableOptional.of(null)); // Negative result cache (different from empty result!)
                 result = new WeightedHostAddress[0];
             }
@@ -513,6 +528,11 @@ public class DNSUtil {
          */
         public int getWeight() {
             return weight;
+        }
+
+        @Override
+        public String toString() {
+            return "prio: " + priority + ", weight: " + weight + ", port: " + getPort() + ", host: " + getHost()  + ", isDirectTLS: " + isDirectTLS();
         }
     }
 }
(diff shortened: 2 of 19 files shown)


distribution/src/bin/openfire.sh
```diff
@@ -138,7 +138,7 @@ esac
 done
 
 
-JAVACMD="${JAVACMD} -Dlog4j.configurationFile=${OPENFIRE_LIB}/log4j2.xml -Dlog4j2.formatMsgNoLookups=true -Djdk.tls.ephemeralDHKeySize=matched -Djsse.SSLEngine.acceptLargeFragments=true"
+JAVACMD="${JAVACMD} -Dlog4j.configurationFile=${OPENFIRE_LIB}/log4j2.xml -Dlog4j2.formatMsgNoLookups=true -Djdk.tls.ephemeralDHKeySize=matched -Djsse.SSLEngine.acceptLargeFragments=true -Djava.net.preferIPv6Addresses=system"
```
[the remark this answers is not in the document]
guusdk: I believe that it was added somewhere between Java 8 (doesn't have it) and 11 (does have it): https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/doc-files/net-properties.html

xmppserver/src/main/java/org/jivesoftware/openfire/net/HappyEyeballsResolver.java
```diff
@@ -0,0 +1,324 @@
...
+        final Iterator<IndexedInetAddress> iterator = resolvedHosts.iterator();
+        while (iterator.hasNext()) {
+            final IndexedInetAddress resolvedAddress = iterator.next();
+            if (resolvedAddress.getIndex() == preferredNextIndex && (resolvedAddress.isIPv6() != preferredNextFamilyIsIpv4)) {
+                iterator.remove();
+
+                final int remainingCount = resultCountByIndex.merge(resolvedAddress.getIndex(), -1, Integer::sum);
+                assert remainingCount >= 0 : "After completing a test, the amount of remaining tasks cannot be negative (but was for index '"+resolvedAddress.getIndex()+"': " + remainingCount + ").";
+                if (remainingCount == 0) {
+                    do {
```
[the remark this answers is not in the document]
guusdk: I don't see how this is potentially endless? The check for that value in the map should always terminate, and in by far the most cases, pretty quickly.

xmppserver/src/main/java/org/jivesoftware/openfire/net/HappyEyeballsResolver.java
```diff
@@ -0,0 +1,324 @@
...
+                        preferredNextIndex++;
+                    }
+                    // If this value was set and is now 0, then all of its addresses have already been used up. Move up the next index!
+                    while (resultCountByIndex.containsKey(preferredNextIndex) && resultCountByIndex.get(preferredNextIndex) == 0);
```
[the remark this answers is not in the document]
guusdk: I prefer not to. I like how this is both descriptive and simple to understand.

Without additional checks (which wouldn't make the code more efficient or easier to understand anyway) that would introduce the risk of NullPointerExceptions being thrown (as a result of autounboxing a non-existing value).

Access to the collection is guarded by a mutex, so I cannot think of any concurrency issues that are in play.

If there is any efficiency penalty to this (which I do not think would be even measurable anyway), I expect the JVM to optimize things.

xmppserver/src/main/java/org/jivesoftware/openfire/net/HappyEyeballsResolver.java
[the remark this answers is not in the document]
Fishbowler: I agree with Guus. Whilst it's non obvious on a quick skim, the while is bound on an index, which can't be endless.

guusdk:
I've added a commit that refactors bits of the code, to remove some partial duplication, and have a more concise representation of data structures used when DNS-resolving XMPP domains into socket addresses.

guusdk:
Rebased to the latest HEAD of the Main branch.

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
```diff
@@ -90,28 +91,39 @@ public class DNSUtil {
...
+     * speaking, all values from the first set of the outer list should be
+     * processed, before elements from the next set of that list are to be processed.
+     *
+     * The Sets that are returned as elements of the list have a predictable
+     * iteration order. This order is based on a randomization based on the
+     * SRV 'weight' value of each host.
+     *
      * @param domain the domain.
      * @param defaultPort default port to return if the DNS look up fails.
      * @return a list of  HostAddresses, which encompasses the hostname and port
```
GregDThomas, line 111: Trivial, but this bit of Javadoc is no longer correct.

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
```diff
@@ -90,28 +91,39 @@ public class DNSUtil {
...
      *         that the XMPP server can be reached at for the specified domain.
      * @see <a href="https://tools.ietf.org/html/rfc6120#section-3.2">XMPP CORE</a>
      * @see <a href="https://xmpp.org/extensions/xep-0368.html">XEP-0368</a>
      */
-    public static List<HostAddress> resolveXMPPDomain(String domain, int defaultPort) {
+    public static List<Set<WeightedHostAddress>> resolveXMPPDomain(String domain, int defaultPort) {
```
GregDThomas: Given the only place (in this commit - feel free to ignore if this turns out to be wrong in subsequent commits) that this is used, the return value is immediately transformed in to an ordered list, perhaps this method should return that ordered list?

xmppserver/src/main/java/org/jivesoftware/openfire/net/HappyEyeballsResolver.java
```diff
@@ -0,0 +1,324 @@
...
+                    return IndexedInetAddress.from(index, InetAddress.getAllByName(hostAddress.getHost()), hostAddress.getPort(), hostAddress.isDirectTLS());
+                } catch (UnknownHostException e) {
+                    throw new RuntimeException(e);
+                }
+            };
+            solve(solve, index);
+        }
+    }
+
+    // Exposed for unit testing.
```
GregDThomas, line 97: I forget if Guava is in your class path, if it is, `@VisibleForTesting` is handy.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketUtil.java
```diff
@@ -60,59 +129,132 @@ public static Map.Entry<Socket, Boolean> createSocketToXmppDomain( String xmppDo
...
-            {
-                Log.debug( "An exception occurred while trying to create a socket connection to XMPP domain '{}' using remote host {}:{}", xmppDomain, realHostname, realPort, e );
-                Log.warn( "Unable to create a socket connection to XMPP domain '{}' using remote host: {}:{}. Cause: {} (a full stacktrace is logged on debug level)", xmppDomain, realHostname, realPort, e.getMessage() );
-                try
-                {
-                    if ( socket != null )
-                    {
-                        socket.close();
-                        socket = null;
+                        if (!JiveGlobals.getBooleanProperty(ConnectionSettings.Server.ENABLE_OLD_SSLPORT, true) && resolvedAddress.isDirectTLS()) {
```
GregDThomas, line 154: Couple of old-style settings here might be worth considering migrating while you're in the area - this line plus 5 lines later.

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
```diff
@@ -366,128 +372,66 @@ public static boolean isNameCoveredByPattern( String name, String pattern )
...
+     */
+    @Deprecated(since = "4.10.0", forRemoval = true) // Remove in or after Openfire 4.11.0
     public static List<Set<WeightedHostAddress>> prioritize(WeightedHostAddress[] records) {
         return prioritize(Arrays.asList(records));
     }
 
+    /**
+     * @deprecated Replaced by {@link SrvRecord#prioritize(Collection)}
+     */
+    @Deprecated(since = "4.10.0", forRemoval = true) // Remove in or after Openfire 4.11.0
```
GregDThomas, line 416: You changed the return value of this method in this set of commits already. Existing code is already not backward compatible - perhaps just remove it?

GregDThomas (commented):
Sorry for the delay, been offline for a week or so.

I've assumed that the protocol handling itself is correct, not being familiar with it. But a few trivial comments on things I've spotted - feel free to ignore any or all.

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
guusdk: How is this not correct? The list returns Sets that are instances of LinkedHashSet, which retain insertion order - or at least, that's the idea.

The purpose of this is to have a collection that, once randomized at initialization, has a predictable iteration order when it's being used.

I was debating if it would be more appropriate to use a List or Set here. List would be more indicative of a strict iteration order, while Set usage hints at the uniqueness of each element that it contains.

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
guusdk: I think that's not correct: the return value is used in SocketUtil to process all elements of the first set, before moving to the elements of the next set. 

I think one of my earlier commits didn't modify the SocketUtil usage of this code yet (apart from changes to make the compiler happy). This comment is highlighted by GitHub to be 'outdated', so I'm guessing this comment is based on a since-modified bit of code.

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
GregDThomas: JavaDoc says you are returning a list of HostAddresses. 
In practice, you're returning a list of Sets. Each Set contains a set of HostAddresses.
Iterating over the response requires different behaviour.

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
GregDThomas: Yes, sounds like it :)

xmppserver/src/main/java/org/jivesoftware/openfire/net/SocketUtil.java
guusdk: I tried that, but it quickly spiraled out of control. Those same settings are used elsewhere, and when you start migrating things in `ConnectionSettings` you really aught to do all of the settings in there. In the end, I reverted and created https://igniterealtime.atlassian.net/browse/OF-2888 (that I hope you will self-assign ;) ) to tackle this in a separate effort.

xmppserver/src/main/java/org/jivesoftware/openfire/net/HappyEyeballsResolver.java
guusdk: oooh, that's handy, thanks!

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
guusdk: If possible, it would be good to retain that (to prevent issues with plugins or third parties), and keep it in as a deprecated method.

I believe I've resolved this by restoring the original method signature now.

GregDThomas: Sorry, got confused between the commits :)

GregDThomas (approved):

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
guusdk: No, you correctly flagged it. I changed the code earlier today. Thanks!

xmppserver/src/main/java/org/jivesoftware/openfire/net/DNSUtil.java
guusdk: Oooh, I was looking at what GitHub renders as the green bit :) Sure, lets fix that.

guusdk:
Rebased.

PR #2557, opened 2024-10-16
Passthrough unrecognized openfire.sh arguments

guusdk:
openfire.sh can be invoked with a number of arguments (eg `-demoboot`, `-debug`). Prior to this commit, unrecognized arguments were ignored.

With this commit, unrecognized argumetns are passed through to the java command that starts Openfire. This allows anyone to add arguments, like system properties, by doing something like `./bin/openfire.sh -Dfoo=bar`


akrherz (approved):
local testing seems to confirm this works as intended.

PR #2560, opened 2024-10-17
OF-2826: Setup should check access to config files

guusdk:
When configuration files do not exist, then errors pop up during setup. This commit adds explicit file checks to ensure that a more descriptive error is shown prior to the setup being started.

image

### DIFF
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -1837,6 +1837,8 @@ setup.env.check.class=Classes
 setup.env.check.jive=Openfire Home found
 setup.env.check.not_home=Home not found. Define system property "openfireHome" or create and add the openfire_init.xml file to the classpath
 setup.env.check.doc=Please read the installation documentation and try setting up your environment again. After making changes, restart your appserver and load this page again.
+setup.env.check.config_found=Successfully loaded configuration file <tt>{0}</tt>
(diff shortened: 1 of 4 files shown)


i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -1837,6 +1837,8 @@ setup.env.check.class=Classes
 setup.env.check.jive=Openfire Home found
 setup.env.check.not_home=Home not found. Define system property "openfireHome" or create and add the openfire_init.xml file to the classpath
 setup.env.check.doc=Please read the installation documentation and try setting up your environment again. After making changes, restart your appserver and load this page again.
+setup.env.check.config_found=Successfully loaded configuration file <tt>{0}</tt>
+setup.env.check.config_not_loaded=Unable to loaded configuration file <tt>{0}</tt>. Check if the file exist and if Openfire has read and write access to it.
```
Fishbowler, line 1841:
```suggestion
setup.env.check.config_not_loaded=Unable to load configuration file <tt>{0}</tt>. Check if the file exists, and if Openfire has read and write access to it.
```

Fishbowler (changes requested):
One tweak, but otherwise a great QOL improvement

guusdk:
This JIRA issue that this relates to is OF-2868 not OF-2826.

PR #2563, opened 2024-10-19
OF-2526 Support SystemD for Debian

stokito:
https://igniterealtime.atlassian.net/browse/OF-2526

I added the openfire.service to run the Openfire directly from SystemD.
The init.d script remains and even installed but it looks like it's not used if the unit's service file exists.
The SystemD itself will call the openfire.sh script. So I added the script to be installed into `/usr/share/openfire/bin/openfire.sh`.

The existing file `/etc/default/openfire` is also supported by the SystemD unit.

There are additional commits to improve:
1. Added UFW firewall rules to make it easier to open ports. The Openfire has a lot of ports and it's difficult to collect the full list of them. The 8010 port is mentioned as Direct File Transfers and I'm not sure if this is a server thing of c2c connections. I added it anyway.
2. Added openfire.slice file to easier limit resources by editing it with `sudo systemctl edit openfire.slice`. The JRE by default uses only 25% of total RAM. So many servers may simply not use the whole RAM that they have. We may change this with Java's `-Xmx` but I think it would be more correct to use the whole RAM with the `-XX:MaxRAMPercentage=100` but limit with the SystemD slice settings if needed. This will be simpler for people who aren't familiar with Java options to reuse their existing knowledge.

### DIFF
--- a/build/debian/openfire.install
+++ b/build/debian/openfire.install
@@ -1,4 +1,4 @@
-bin/openfire.sh /usr/share/openfire/bin
+bin/openfire.sh usr/share/openfire/bin
 lib/*.jar usr/share/openfire/lib
 lib/log4j2.xml etc/openfire
 resources/database usr/share/openfire/resources


Fishbowler:
Rebased

distribution/src/dist/etc/ufw/applications.d/openfire
```diff
@@ -0,0 +1,39 @@
+[Openfire C2S]
+title=Openfire XMPP Server - Client to Server (C2S) connections
+description=Allow for XMPP clients to connect to the server and also use File Transfers
+ports=5222|5223|8010|7777
```
Fishbowler, line 4: Given that the ports are configurable, wouldn't these need to be set at runtime?

Fishbowler:
If I wanted to test this, how would I go about it?

build/debian/openfire.install
```diff
@@ -1,7 +1,9 @@
+bin/openfire.sh /usr/share/openfire/bin
 lib/*.jar usr/share/openfire/lib
 lib/log4j2.xml etc/openfire
 resources/database usr/share/openfire/resources
 conf/openfire.xml etc/openfire
 conf/security.xml etc/openfire
 resources/security etc/openfire
 plugins var/lib/openfire
+dist/* /
```
akrherz, line 9: What's happening with this line? and why is `/usr` above being mixed with `usr`

akrherz:
It would be very nice if this work could somehow be shared with the RPM based installer `distribution/src/bin/extra/redhat` only has the SysVInit file presently.

build/debian/openfire.install
akrherz: I still don't understand why `/usr` is being used vs `usr` ?

distribution/src/dist/etc/ufw/applications.d/openfire
```diff
@@ -0,0 +1,39 @@
...
+
+[Openfire Proxy]
+title=Openfire XMPP Server - TURN Proxy and STUN server
+description=This allows users to use calls and file transfer between each others when they are behind NAT.
+ports=3478|5349
+
+[Openfire Connection Manager]
+title=Openfire XMPP Server - Connection Manager
+description=Openfire Connection Manager improves the scalability of server by handling a portion of the client connections
+ports=5262|5263
```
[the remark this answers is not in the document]
guusdk: The Connection Managers are software components that connect to Openfire over TCP. Their purpose is to offload basic IO handling to other hosts. It is rarely used these days, but it does see use. I'm also in talks with someone who is contemplating reviving it. For now, lets keep these ports.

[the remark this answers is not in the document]
guusdk: The same rationale applies to External Components, I think.

I'm not sure of the implications of adding those ports to this file. We shouldn't make it _impossible_ for people to use these ports, but I'm OK with the ports being usable only when some tweaking is done by an admin.

PR #2564, opened 2024-10-20
OF-2893: Warn administrator when a wildcard pattern is loaded

Fishbowler:
Plugins like the REST API Plugin require wildcard routes (because they do HTTP calls that include username or JID in the path). Some of these require the new(-ish) adminConsole.access.allow-wildcards-in-excludes property to be set so that they can either be open, or so that they can apply their own authentication outside that of the normal Admin Console access.

Whilst there have been attempts to update documentation, not everyone reads it. This even trips up the folk most familiar with it (meaning: me, almost every time).

This adds a warning in the Plugin Admin page that displays when a loaded plugin has a wildcard exception declared, but the System Property is not set.

https://igniterealtime.atlassian.net/browse/OF-2893


guusdk (approved):

guusdk:
That's really useful, thanks!

PR #2566, opened 2024-10-26
OF-2786: Add Networking Guide to documentation

guusdk:
Describes how to configure DNS and IPv6.

The website needs an update after this gets released (to update the Openfire documentation page to point to the new guide introduced by this PR).


guusdk:
Should this be split up into two guides?

documentation/network-configuration-guide.html
```diff
@@ -0,0 +1,403 @@
...
+            </p>
+            <fieldset>
+                <legend>Example DNS SRV records for a group chat service</legend>
+                <pre><code>
+_xmpp-server._tcp.conference.example.org. 86400 IN SRV 0 5 5269 xmpp.example.org.
+_xmpps-server._tcp.conference.example.org. 86400 IN SRV 0 5 5270 xmpp.example.org.
+                </code></pre>
+            </fieldset>
+            <p>
+                Note <code>xmpp-client</code> and <code>xmpps-client</code> records are not needed for these server-sided services.
```
Fishbowler, line 211: Tiny bit of clarity
```suggestion
                Note <code>xmpp-client</code> and <code>xmpps-client</code> records are not needed for these server-sided services - the user's server connects to remote services on their behalf.
```

documentation/network-configuration-guide.html
```diff
@@ -0,0 +1,403 @@
...
+
+        <section id="ipv6-java">
+
+            <h3>Java Networking Configuration</h3>
+
+            <p>
+                Openfire is an application that runs in a Java virtual machine. Openfire's IPv6 support is primarily
+                controlled by standard <a href="https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/doc-files/net-properties.html">Java Networking Properties</a>.
+                The following properties are typically provided. Please refer to the documentation of your version of
+                Java for details.
```
Fishbowler, line 323: Should this be a bit more explicit, given the stuff below on what the launch script does?

Example:
```suggestion
                Java for details on available settings and how to control them (e.g. using <code>JAVA_TOOL_OPTS</code>.
```

Fishbowler (approved):
Some minor suggestions, but otherwise ace!

documentation/network-configuration-guide.html
guusdk: Do we even support that property? I'm not sure if we even have a generic way of applying these settings across all platforms that we support. To avoid confusion, I think we best leave this out for now.

PR #2568, opened 2024-10-26
OF-2896 & OF-2897: Fix memory leak when dealing with pre-authenticated Sessions

guusdk:
A collection of pre-authenticated sessions was maintained incorrectly (using a type-unsafe key that was both used as a stream-ID and a JID), which caused every session that was ever created to remain a part of that collection, until the server got restarted. This introduces a memory leak (and possibly functional issues, although those have never been reported).
    
The second commit resolves the problem. It hides access to the collection to enforce proper maintenance. A more type-safe key is used, which should help prevent future issues like these.
    
The changes in the second commit affect the API. I've opted to not include backwards-compatibility (with deprecation), as I believe the per-existing API to be simply broken. Any usage of that API should be revisited, which is enforced by removing the API.

The first commit introduces a periodic clean-up of pre-authenticated client sessions. This ensures that clients that are connected indefinitely before authenticating are eventually closed. Note that the XMPP specification mandates SASL, requiring every connection to be authenticated (which _may_ use the ANONYMOUS mechanism).


guusdk:
Even though the changes in this PR break Openfire's API, they're important enough in my book (fixing a memory leak) for this to be backported.

guusdk:
I've converted this to draft. The Aioxmpp tests seem to flag an issue. I'm not exactly sure what's going on. I wonder if Openfire processes stanzas 'addressed' to the pre-authenticated session after that session has been authenticated. With the cleanup now actually working, that might be failing.

guusdk:
Fixing OF-2896 (the memory leak) seems to have uncovered another bug: OF-2897. This PR fixes both.

PR #2570, opened 2024-10-30
Add demoboot documentation

guusdk:
The website needs an update after this gets released (to update the Openfire documentation page to point to the new guide introduced by this PR).

### DIFF
--- a/documentation/demoboot-guide.html
+++ b/documentation/demoboot-guide.html
@@ -1,0 +1,143 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Openfire: Demoboot Guide</title>
+    <link href="style.css" rel="stylesheet" type="text/css">
+</head>
+<body>
+
+<article id="top">
+
+    <header>
+        <img src="images/header_logo.gif" alt="Openfire Logo"/>
+        <h1>Demoboot Guide</h1>
+    </header>
+
+    <nav>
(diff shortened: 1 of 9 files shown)


documentation/demoboot-guide.html
```diff
@@ -0,0 +1,143 @@
...
+        <h2>Demoboot Overview</h2>
+
+        <p>
+            The 'demoboot' setup of Openfire allows one to start a fresh installation of Openfire into a certain
+            provisioned state, without running any of the setup steps. When running in 'demoboot' mode:
+        </p>
+        <ul>
+            <li>Openfire is configured to use an embedded database, which requires no setup</li>
+            <li>an administrative account is created using the username 'admin' and password 'admin'</li>
+            <li>two users are automatically created: 'jane' and 'john' (both using the value 'secret' as their password), which are on each-other's contact list</li>
```
Fishbowler, line 57:
```suggestion
            <li>two users are automatically created: 'jane' and 'john', both using the value 'secret' as their password, and who are on each other's contact list</li>
```

Fishbowler (approved):
One tweak, but otherwise fine.

There's some references to demoboot in `working-with-openfire.html` - does that want linking here too?

guusdk:
I've seen that. I thought adding more references there would make things a bit clunky. I'm not against it per-se, but... meh. I'll merge as-is, and would welcome a follow-up PR if you deem it useful enough. :)

PR #2571, opened 2024-10-30
OF-2899: Connect manager connect fail when upgrade openfire 4.9.0.

fjcondy:
because .connect manager session(LocalConnectionMultiplexerSession) have't SASLMechanisms. we change org.jivesoftware.openfire.net.StanzaHandler.java tlsNegotiated method when getSASLMechanisms() method return null. not add to features:


guusdk:
I have created https://igniterealtime.atlassian.net/browse/OF-2899 to track this issue. I'll rework the commit to introduce a reference to this ticket in the commit message.

guusdk:
I have reworked the commit, and applied the change to other places where the collection was used, but not explicitly checked for null.

PR #2577, opened 2024-10-30
OF-2898: Restore SessionDestroyed event

guusdk:
An earlier change prevented the SessionDestroyed event from being sent to event listeners, which is fixed in this commit.

I've not tested this change.


guusdk:
Community member Crabby73 reviewed and tested this fix: https://discourse.igniterealtime.org/t/session-destroyed-event-is-not-triggered/94842/7

PR #2585, opened 2024-11-04
OF-2900: Update Increase crypto config used by tests

guusdk:
Some stricter OS won't provide the 'lower' crypto configurations that our unit tests use. This makes tests fail.


akrherz:
I added the backport tag as we'd need this on the 4.9 branch for bamboo jobs to succeed...  Testing this change locally with `update-crypto-policies --set DEFAULT` configured and s2s tests passed.  I am boggling this some more though!

akrherz (approved):

PR #2587, opened 2024-11-06
OF-2901: Reduce verbosity of LDAP unencrypted connection warning

guusdk:
Introduces a new property that defines how long warnings are suppressed after they've been logged.


guusdk:
I've not tested this yet.

Fishbowler:
I've tested it in an insecure LDAP environment. Works for me.

> This warning will be suppressed for PT1H

Is there a nicer way to render durations?

guusdk:
> Is there a nicer way to render durations?

Not without dragging in additional code. I'm not sure if that's worth it?

Fishbowler (approved):

PR #2592, opened 2024-11-10
OF-2904: Fix parsing issue of OtherName

guusdk:
OtherName values are tagged, which should be taken into account by the parser.

Unit tests that generate a certificate should include an EXPLIT tagged OtherName.

After two days of ... pain ... I've resorted to converting the BouncyCastle-generated certificate to PEM and back again, which for some reason that's beyond me appears to be required to make the parsing be successfull on both Java 17 and 21.

Adds additional unit test that uses a certificate generated by OpenSSL.

Many thanks to @dwd for invaluable help on this.

### DIFF
--- a/xmppserver/src/test/java/org/jivesoftware/util/CertificateManagerTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/util/CertificateManagerTest.java
@@ -66,7 +66,6 @@ public class CertificateManagerTest
     public static final String KEY_ALGORITHM = "RSA";
     public static final String SIGNATURE_ALGORITHM = "SHA256withRSA";
 
-    private static KeyPairGenerator keyPairGenerator;
     private static KeyPair subjectKeyPair;
     private static KeyPair issuerKeyPair;
     private static ContentSigner contentSigner;
@@ -74,7 +73,7 @@ public class CertificateManagerTest
     @BeforeAll
     public static void initialize() throws Exception
     {
-        keyPairGenerator = KeyPairGenerator.getInstance( KEY_ALGORITHM );
+        final KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance(KEY_ALGORITHM);
         keyPairGenerator.initialize( KEY_SIZE );
 
         subjectKeyPair = keyPairGenerator.generateKeyPair();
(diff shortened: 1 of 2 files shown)


xmppserver/src/test/java/org/jivesoftware/util/CertificateManagerTest.java
```diff
@@ -272,23 +288,28 @@ public void testServerIdentitiesXmppAddrAndDNS() throws Exception
...
                 new GeneralName( GeneralName.otherName, otherName ),
                 new GeneralName( GeneralName.dNSName, subjectAltNameDNS )
         });
         builder.addExtension( Extension.subjectAlternativeName, true, subjectAltNames );
 
         final X509CertificateHolder certificateHolder = builder.build( contentSigner );
-        final X509Certificate cert = new JcaX509CertificateConverter().getCertificate( certificateHolder );
+        X509Certificate cert = new JcaX509CertificateConverter().getCertificate( certificateHolder );
+
+        // FIXME: Unsure why, but without this back-and-forth, tests will fail on Java 17.
```
dwd, line 309: I'm not sure this *is* a "FIXME", as such - if it is indeed some kind of error, it's one in the test code and not the code under test.

dwd (approved):
Looks good to me, and the "Prebuilt PEM" is one I generated independently so I'm confident it's handling all the cases I could throw at it.

xmppserver/src/test/java/org/jivesoftware/util/CertificateManagerTest.java
```diff
@@ -299,6 +320,105 @@ public void testServerIdentitiesXmppAddrAndDNS() throws Exception
...
+     *
+     * [ v3_ca ]
+     * subjectAltName = @alt_names
+     * basicConstraints = CA:TRUE
+     * keyUsage = digitalSignature, keyEncipherment
+     * extendedKeyUsage = serverAuth, clientAuth
+     *
+     * [ alt_names ]
+     * otherName.0 = 1.3.6.1.5.5.7.8.7;IA5:_xmpp-server.service.example.com
+     * otherName.1 = 1.3.6.1.5.5.7.8.7;IA5:_dns.service.example.net
```
Fishbowler, line 353: Why is this `assertFalse`d later?
I don't doubt that it should be, it's just not clear from the test.

dwd: Yeah, these assertions might benefit from some explanatory comments. In this case, it's a valid sRVName for the wrong service.

xmppserver/src/test/java/org/jivesoftware/util/CertificateManagerTest.java
guusdk: I agree that it's likely the test code. I'd like that to be fixed too, though.

dwd: Fair point.

xmppserver/src/test/java/org/jivesoftware/util/CertificateManagerTest.java
guusdk: That's a good suggestion. I've now added failure messages to the assertions that describe the (un)desired behavior.

guusdk:
Merging this in the assumption that @Fishbowler is happy with the newly added words to clarify why certain certificate SANs are expected to _not_ be in a test result.

I'm in a bit of a rush to get this PR merged, as it's a dependency of other PRs. We can always improve wording in follow-up commits.

Fishbowler:
Wholeheartedly.

PR #2593, opened 2024-11-11
OF-2905: Update Maven Wrapper to 3.3.2

guusdk:
These changes are the result of executing `./mvnw wrapper:wrapper`


guusdk:
Weird that it removes the JAR file.

Fishbowler:
> Weird that it removes the JAR file.

This is the `only-script` way, and is apparently the default now. https://maven.apache.org/wrapper/

Fishbowler:
There's a second maven wrapper in build/ci/updater.
Ideally, I'd find a way to integrate this more neatly. In the meantime, I'm struggling to get `./mvnw wrapper:wrapper` to generate something similar to the root version, even down to the version of the wrapper in use :(

guusdk:
Oh, that's a good find. I've got a slight preference for not changing the type of wrapper that we used, so I'll re-generate with

```
./mvnw wrapper:wrapper ./mvnw wrapper:wrapper -Dtype=bin
```

Within `build/ci/updater/` I'll try to do the same thing. It appears that we need an explicit environment variable for that plan to work. I've tried with the following from within that directory:

```
export OPENFIREVSN=5.0.0-SNAPSHOT && ./mvnw wrapper:wrapper -Dtype=bin
```

Fishbowler (approved):

Fishbowler:
Approved but not merged, given all the activity currently happening in the project and its plugins.

Merge whenever you're ready for it, Guus.

PR #2594, opened 2024-11-11
OF-2906: Fix setup page that checks config files

guusdk:
Fixes a copy/paste error that prevented the security config file from being parsed properly.

Fixes an error where a non-existing variable is checked, thus never passing the check.

This is an untested fix for the problem that is apparent on https://discourse.igniterealtime.org/t/reinstalled-openfire-but-now-get-an-error-that-it-doesnt-meet-server-requirements/94891/1

### DIFF
--- a/xmppserver/src/main/webapp/setup/index.jsp
+++ b/xmppserver/src/main/webapp/setup/index.jsp
@@ -155,7 +155,7 @@
     </c:if>
 
     <c:choose>
-        <c:when test="${not jreVersionCompatible or not servlet22Installed or not jsp11Installed or not jiveJarsInstalled or not openfireHomeExists or not configFailedLoading}">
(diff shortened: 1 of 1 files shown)


akrherz:
4.9 branch test, this appears:
Screenshot from 2024-11-11 14-19-27

trying this patch

akrherz:
With patch applied, error is gone and one can select the language.  I am unsure about OF-2868 association of this patch though.

guusdk:
I think we need a new ticket for this problem (but I do think it's introduced by changes for OF-2868).

Was there a work-around possible to get past this page?

akrherz:
> Was there a work-around possible to get past this page?

If one can deal with English, you can click the side menu items to move to the next setup screen.

Fishbowler (approved):

akrherz:
Workaround is to click here, if you can work with English setup.

Screenshot from 2024-11-11 14-46-14

PR #2597, opened 2024-11-12
OF-2907 Use older gz compression for RPM installer

akrherz:
From testing, this allowed a RPM built on Rocky Linux 9 to install on the older Amazon Linux.


Fishbowler (approved):

guusdk:
I'm guessing this was quite the search. Thanks!

Unsure how to verify this in other ways than you've already done. I doubt we'll get more feedback on this. Let's merge as-is, and take it from there.

PR #2600, opened 2024-11-12
Jetty 12.0.14 ee8 upgrade

hqv126:
Upgrading to Jetty 12.0.14 using the Java EE 8 platform.

Additional notes on previous commits:
- AdminConsolePlugin.java: Resource.newResource method is no longer available. Utilising ResourceFactory to construct new resource as suggested by Joakim.

### DIFF
--- a/plugins/pom.xml
+++ b/plugins/pom.xml
@@ -122,7 +122,7 @@
         <!-- Versions -->
         <openfire.version>5.0.0-SNAPSHOT</openfire.version>
         <!-- Note; the following jetty.version should be identical to the jetty.version in xmppserver/pom.xml -->
-        <jetty.version>10.0.18</jetty.version>
+        <jetty.version>12.0.14</jetty.version>
     </properties>
 
     <profiles>
--- a/xmppserver/pom.xml
+++ b/xmppserver/pom.xml
@@ -149,8 +149,8 @@
 
             <!-- Compile the JSP pages -->
             <plugin>
-                <groupId>org.eclipse.jetty</groupId>
-                <artifactId>jetty-jspc-maven-plugin</artifactId>
+                <groupId>org.eclipse.jetty.ee8</groupId>
+                <artifactId>jetty-ee8-jspc-maven-plugin</artifactId>
                 <version>${jetty.version}</version>
                 <executions>
                     <execution>
@@ -257,13 +257,13 @@
             <version>${jetty.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-servlets</artifactId>
+            <groupId>org.eclipse.jetty.ee8</groupId>
+            <artifactId>jetty-ee8-servlet</artifactId>
             <version>${jetty.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-webapp</artifactId>
+            <groupId>org.eclipse.jetty.ee8</groupId>
+            <artifactId>jetty-ee8-webapp</artifactId>
             <version>${jetty.version}</version>
         </dependency>
         <dependency>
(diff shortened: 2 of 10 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindManager.java
```diff
@@ -93,7 +93,7 @@ public final class HttpBindManager implements CertificateEventListener {
     public static final SystemProperty<Integer> HTTP_BIND_SECURE_PORT = SystemProperty.Builder.ofType(Integer.class)
         .setKey("httpbind.port.secure")
         .setDynamic(true)
-        .setDefaultValue(7444)
+        .setDefaultValue(7443)
```
guusdk: I believe that the problem being corrected here is not in the upstream Openfire source, which must mean that it gets introduces somewhere in this PR.

Rather than creating problem in one commit and needing another commit to fix it, please squash this commit with the one that introduces the problem (or edit the commit that introduces the issue, and remove the commit that fixes it). That way, it is as if it never happened (and this won't be in our source history).

plugins/pom.xml
```diff
@@ -341,8 +341,8 @@
                                 <jspc>
                                     <package>org.jivesoftware.openfire.plugin.${plugin.name}</package>
                                 </jspc>
-                                <sourceVersion>11</sourceVersion>
-                                <targetVersion>11</targetVersion>
+                                <sourceVersion>17</sourceVersion>
+                                <targetVersion>17</targetVersion>
```
guusdk: This change shouldn't be necessary, as Openfire's HEAD of the Main branch already has this change.

Can you rebase the PR on the latest HEAD of the Main branch of igniterealtime/Openfire please? That should take away the need for this change (and possibly others).

xmppserver/pom.xml
```diff
@@ -159,8 +159,8 @@
                             <goal>jspc</goal>
                         </goals>
                         <configuration>
-                            <sourceVersion>11</sourceVersion>
-                            <targetVersion>11</targetVersion>
+                            <sourceVersion>17</sourceVersion>
+                            <targetVersion>17</targetVersion>
```
guusdk: This change shouldn't be necessary, as Openfire's HEAD of the Main branch already has this change.

Can you rebase the PR on the latest HEAD of the Main branch of igniterealtime/Openfire please? That should take away the need for this change (and possibly others).

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindManager.java
```diff
@@ -94,7 +93,7 @@ public final class HttpBindManager implements CertificateEventListener {
     public static final SystemProperty<Integer> HTTP_BIND_SECURE_PORT = SystemProperty.Builder.ofType(Integer.class)
         .setKey("httpbind.port.secure")
         .setDynamic(true)
-        .setDefaultValue(7443)
+        .setDefaultValue(7444)
```
guusdk: Ah, this is an accidental change that is removed in a later commit. Please squash that commit with this commit (or edit this commit, and drop the other commit).

plugins/pom.xml
```diff
@@ -341,8 +341,8 @@
                                 <jspc>
                                     <package>org.jivesoftware.openfire.plugin.${plugin.name}</package>
                                 </jspc>
-                                <sourceVersion>17</sourceVersion>
-                                <targetVersion>17</targetVersion>
+                                <sourceVersion>11</sourceVersion>
+                                <targetVersion>11</targetVersion>
```
guusdk, line 345: This shouldn't be needed at all?

xmppserver/pom.xml
```diff
@@ -159,8 +159,8 @@
                             <goal>jspc</goal>
                         </goals>
                         <configuration>
-                            <sourceVersion>17</sourceVersion>
-                            <targetVersion>17</targetVersion>
+                            <sourceVersion>11</sourceVersion>
+                            <targetVersion>11</targetVersion>
```
guusdk, line 163: This shouldn't be needed at all?

plugins/pom.xml
guusdk: I stand corrected. This was _not_ in the HEAD of the Main branch yet (we've missed it). I would still like to see that Java-upgrade change be isolated from the Jetty upgrade change. I've pulled out this change in https://github.com/igniterealtime/Openfire/pull/2601

xmppserver/pom.xml
guusdk: I stand corrected. This was _not_ in the HEAD of the Main branch yet (we've missed it). I would still like to see that Java-upgrade change be isolated from the Jetty upgrade change. I've pulled out this change in https://github.com/igniterealtime/Openfire/pull/2601

plugins/pom.xml
guusdk: I stand corrected. This was _not_ in the HEAD of the Main branch yet (we've missed it). I would still like to see that Java-upgrade change be isolated from the Jetty upgrade change. I've pulled out this change in https://github.com/igniterealtime/Openfire/pull/2601

xmppserver/pom.xml
guusdk: I stand corrected. This was _not_ in the HEAD of the Main branch yet (we've missed it). I would still like to see that Java-upgrade change be isolated from the Jetty upgrade change. I've pulled out this change in https://github.com/igniterealtime/Openfire/pull/2601

guusdk:
Thanks @hqv126 - this looks great.

Let's squash these commits together, as they're basically the same change (only applied to different files:
- fbb0cbcf6273b7484d612fc5845043e07815603f
- df07ea2cbfc5f06eee9ce4c005032e6bdd3f29c2
- 535ebd3bbb7879e5dd33d342905f8caeec80d409

And lets also squash this commit to that, as it reverses a small issue introduced by one of the three commits above:
- 6ae8bb67fbd8d65af2e3884d85297d226d112cba

I was kind of thinking of the new guide that you added to be a more generic "this is how you adjust your proprietary plugin to be compatible with the Jetty upgrade" documentation, but we can adjust that in follow-up work (I also believe that at least one of the errors that you describe is unrelated to the Jetty upgrade, but I'll remove that later, too).

guusdk:
I've now isolated the Java-upgrade changes into a new commit, which has been applied to the Main branch. Please re-base this PR, so that those changes can be removed here.

hqv126:
Thank you Guusdk. I've squashed the relevant git commits.

guusdk:
Looks good! I think yo now need to drop the last commit, as that reverts Java back from 17 to 11.

guusdk (approved):

PR #2604, opened 2024-11-14
OF-2717: Allow plugins to register/unregister web context

guusdk:
The embedded webserver (Jetty) that powers the HTTP-Bind (BOSH) and Websocket endpoints can be used by plugins, for them to add web content of their own.

To do so, a plugin can dynamically add a 'handler' when it gets loaded in Openfire (that handler should be removed again when the plugin gets unloaded).

Due to changes to the Jetty API (which got updated from 10 to 12, using EE8), this mechanism was broken.

Prior to the upgrade, a `org.eclipse.jetty.server.handler.HandlerList` was used to maintain the collection of plugin-provided contexts. After upgrading to Jetty 12, this was replaced with a `org.eclipse.jetty.server.handler.ContextHandlerCollection`. This was part of https://github.com/igniterealtime/Openfire/pull/2600/commits/99fe5f98554a547a57f59ed6c3b1c65f47c5f6c6#diff-55709739e79c9b3141bb106d4e4cf428d2bc54e46acf59450c3d9e8cf988385b

The methods used by plugins to add and remove 'handlers' were not modified in that change. The 'handler' passed to them remained of type `org.eclipse.jetty.server.Handler`. This compiles fine, but seems to be a problem.

When plugins (which are stand-alone projects with their own source code repositories) get updated to use Jetty 12 instead of 10, their representation of their web-context switches from `org.eclipse.jetty.webapp.WebAppContext` to the Jetty 12 EE8 `org.eclipse.jetty.ee8.webapp.WebAppContext`. The former is compatible with the old implementation in Openfire, but the latter is not.

This commit adds new methods to register and unregister (Jetty 12 EE8) web contexts.

I considered leaving the original methods (possibly marked as being deprecated) in hopes of providing backwards compatibility. However old plugins that are compiled against Jetty 10 won't find the Jetty classes that they need in Openfire anyway (as Jetty is a 'provided' dependency, which now got updated), causing ClassCastExceptions to be thrown. That makes retaining the original add/remove methods (for backwards compatibility) rather pointless.


joakime:
Does your use of `org.eclipse.jetty.ee8.nested.ContextHandler` use any ee8 specific features? (like servlets? or servlet apis?)

Can it be a non servlet implementation?
if so, use the [`org.eclipse.jetty.server.handler.ContextHandler`](https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/server/handler/ContextHandler.html) version instead.

If you must use servlet features, then you are bound by the servlet environment that is provided by the `ee8` environment.

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindManager.java
```diff
@@ -757,12 +766,12 @@ protected ServletContextHandler createStaticContentHandler()
...
     /**
-     * Adds a Jetty handler to be added to the embedded web server that is used to expose BOSH (HTTP-bind)
-     * functionality.
+     * Adds a Jetty handler to be added to the embedded web server that is used to expose Openfire's public
+     * web-bindings (eg: BOSH / HTTP-bind and websocket).
      *
      * @param handler The handler (cannot be null).
      */
-    public void addJettyHandler( Handler handler )
+    public void addJettyHandler(final ContextHandler handler)
```
joakime, line 774: Either leave this as a `org.eclipse.jetty.server.Handler` or provide a new `addJettyHandler(org.eclipse.jetty.server.Handler handler)` as a secondary implementation allowing your users to enable other jetty features (like GzipHandler, CompressionHandler, QoSHandler, RewriteHandler, etc)

joakime:
Just to expand on what I said before ...

The `org.eclipse.jetty.server.handler.ContextHandlerCollection` holds non-environment core `org.eclipse.jetty.server.handler.ContextHandler` instances.

The ee8 `org.eclipse.jetty.ee8.nested.ContextHandler` is not core context handler.
As it attempts to maintain the servlet-ness of the ContextHandler found in Jetty 10/11.

When you upgrade to ee10 or ee11 (with Jetty 12.1.x) then that "servlet-ness" goes away, as there's no old Jetty version to attempt to maintain behavior with.

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindManager.java
joakime: It there any reason you don't want your users to just add Servlets or Filters to your existing context instead?

guusdk: This has historically been done to somewhat isolate the functionality that is being provided by individual plugins, both from each-other, as well as from the functionality that's provided by Openfire itself on that web context.

guusdk:
I must admit that my lack of knowhow makes it a bit hard to follow. Maybe it's easier to illustrate things with an example.

One of the Openfire plugins that adds content to Jetty is a plugin that adds a web-based chat client to Openfire. This plugin (called the JSXC-plugin) uses Jetty in two distinct ways:
- the plugin contains JSPs (pre-compiled) that add to very similar JSPs provided by Openfire, that make up a web-based administrative console. The plugin-provided JSPs allow for configuration of the web-client that's being added by the plugin.
- the webclient itself is a Javascript plugin, that is configured through a bit of JSON. That JSON is dynamically generated by a servlet.

My approach to make the pre-existing plugin code compatible with Jetty 12 (as well as the change in this PR) can be found here: https://github.com/igniterealtime/openfire-jsxc-plugin/pull/25

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindManager.java
joakime: Do you prevent two plugins from using the same overlapping context selection rules?
A ContextHandler is selected from the ContextHandlerCollection based on ...

* The context-path declared on the ContextHandler
* The incoming path-in-context for request when it reaches the ContextHandlerCollection
* The virtual-host declarations (either direct, or named)

To oversimplify this ...
If more than one ContextHandler entry results from the selection list above, then the resulting list is sorted, and then each entry is given the request to process following Jetty Core Handler rules, stopping when one of the handlers declares itself to be the one handling the generation of the response.

Is that something you attempt to prevent? or let the plugins do what they want?

joakime:
> I must admit that my lack of knowhow makes it a bit hard to follow. Maybe it's easier to illustrate things with an example.
> 
> One of the Openfire plugins that adds content to Jetty is a plugin that adds a web-based chat client to Openfire. This plugin (called the JSXC-plugin) uses Jetty in two distinct ways:
> 
>     * the plugin contains JSPs (pre-compiled) that add to very similar JSPs provided by Openfire, that make up a web-based administrative console. The plugin-provided JSPs allow for configuration of the web-client that's being added by the plugin.

JSPs would mean you are either using a ServletContextHandler or a WebAppContext.
I say this, because JSP requires a (jakarta|javax).servlet.ServletContext to function.
A raw ContextHandler is insufficient for JSP to function.

guusdk:
Openfire uses two different mechanisms:

JSPs provided by a plugin that are intended to be part of the admin console are _not_ added through the context that's being added by the API that's the subject of this PR. Instead, the per-compilation generates a `web.xml` like the one below. They are then merged into a [custom servlet that has already been initialized](https://github.com/igniterealtime/Openfire/blob/72f42357d7100bb5369ab10bf6b2e1a7611393ca/xmppserver/src/main/java/org/jivesoftware/openfire/container/PluginServlet.java#L155).

```xml
<?xml version="1.0" encoding="UTF-8"?>
<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
         version="3.1">
<context-param><param-name>org.eclipse.jetty.ee8.jsp.precompiled</param-name><param-value>true</param-value></context-param>

<!--
Automatically created by Apache Tomcat JspC.
-->


    <servlet>
        <servlet-name>org.jivesoftware.openfire.plugin.jsxc.jsxc_002dconfig_jsp</servlet-name>
        <servlet-class>org.jivesoftware.openfire.plugin.jsxc.jsxc_002dconfig_jsp</servlet-class>
    </servlet>

    <servlet-mapping>
        <servlet-name>org.jivesoftware.openfire.plugin.jsxc.jsxc_002dconfig_jsp</servlet-name>
        <url-pattern>/jsxc-config.jsp</url-pattern>
    </servlet-mapping>

<!--
End of content automatically created by Apache Tomcat JspC.
-->

</web-app>
```

This all is separate from the optional extension point that's the subject of this PR, through which a plugin could add ... whatever they want ... to Jetty, in order to make use of its functionality. In practice, that's mostly going to be servlet-based functionality, like the bit in JSXC that I pointed to in my previous commit. It uses something [like this](https://github.com/igniterealtime/openfire-jsxc-plugin/blob/cbb681fd7b98b8de7e0d4da18e6cd7d60019eff1/src/java/org/igniterealtime/openfire/plugin/jsxc/JSXCPlugin.java#L60-L64):

```java
// Add the Webchat sources to the same context as the one that's providing the BOSH interface.
context = new WebAppContext( null, pluginDirectory.getPath() + File.separator + "classes/", "/" + CONTEXT_ROOT );
context.setClassLoader( this.getClass().getClassLoader() );

HttpBindManager.getInstance().addJettyHandler( context );
```

To make available a web-app (as [defined here](https://github.com/igniterealtime/openfire-jsxc-plugin/tree/cbb681fd7b98b8de7e0d4da18e6cd7d60019eff1/classes)). This happens on a different different Jetty embedded server instance than the one that serves the admin console (as the webapp serves user-facing, public content, while the admin console serves restricted, administrative content).

hqv126:
Is the following alternative approach, which may have been mentioned in part in previous comments on this PR, of keeping, the more generic addJettyHandler(Handler handler) and removeJettyHandler(Handler handler) methods something we can go with? Plugins that use WebAppContext can simply call get() to pass a Handler instance, like so:

`HttpBindManager.getInstance().addJettyHandler( context.get() );`

This approach works with our current Openfire implementation and will also remain compatible if the changes in this [PR](https://github.com/igniterealtime/Openfire/pull/2609) are accepted.

guusdk:
Is there benefit in providing both the 'old' method, as well as the new one as introduced in this PR?

guusdk:
As suggested by @joakime and @hqv126, this PR has been updated (force-push) to retain the old add/remove methods that are based on the generic Handler.

As I expect many/most plugins to use a WebAppContext - which each plugin will need to update to Jetty12 EE8 - I've added a convenience overload for those methods that take instances of that class directly.

xmppserver/src/main/java/org/jivesoftware/openfire/http/HttpBindManager.java
```diff
@@ -756,6 +766,33 @@ protected ServletContextHandler createStaticContentHandler()
         return null;
     }
 
+    /**
+     * Adds a Jetty handler to be added to the embedded web server that is used to expose Openfire's public
+     * web-bindings (eg: BOSH / HTTP-bind and websocket).
+     *
+     * @param handler The handler (cannot be null).
+     */
+    public void addJettyHandler(final ContextHandler handler)
```
hqv126, line 775: Do you not want to use a @Nonnull check here like the removeJettyHandler(@Nonnull final ContextHandler handler) method?

hqv126 (approved):

PR #2606, opened 2024-11-15
OF-2910: No longer use Atlassian's Maven repository

guusdk:
Atlassian's Maven repository serves us a dependency (`com.github.jgonian:commons-ip-math:1.32`) that appears to be different from the dependency served under the same Maven coordinates by Maven-Central. Atlassian's one makes Openfire fail at runtime.

Atlassian has announced that they'll stop serving third-party artifacts in early 2025.

This commit removes Atlassian's Maven repository from Openfire's Maven configuration. There was one artifact used by Openfire in Atlassian's repository that couldn't be obtained elsewhere (`com.cenqua.shaj:shaj:0.5`). This has now been added to Ignite's own Maven repository.

### DIFF
--- a/build/ci/updater/pom.xml
+++ b/build/ci/updater/pom.xml
@@ -92,15 +92,11 @@
     </dependency>
   </dependencies>
 
-<repositories>
-        <repository>
-            <id>atlassian-public</id>
-            <url>https://maven.atlassian.com/repository/public</url>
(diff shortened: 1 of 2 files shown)


akrherz:
I would suggest we backport this so that `v4.9` branch keeps going too...

PR #2607, opened 2024-11-15
OF-2717: Plugins should use JSP compiler from new Jetty

guusdk:
In this version of Openfire, Jetty is updated from 10 to 12 (using EE8). When a plugin compiles its JSPs against this version of Openfire, it should use the JSP-compiler of that Jetty version (the Maven artifacts of the compiler have changed).

### DIFF
--- a/build/ci/updater/pom.xml
+++ b/build/ci/updater/pom.xml
@@ -92,15 +92,11 @@
     </dependency>
   </dependencies>
 
-<repositories>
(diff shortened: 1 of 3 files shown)


guusdk:
This PR contains the commit that's also in https://github.com/igniterealtime/Openfire/pull/2606 which is a prerequisite for the build to succeed.

PR #2610, opened 2024-11-19
OF-2134: Add option to enable certificate revocation checks

viv:
When enabled, certificates will be verified against Certificate Revocation Lists (CRL) and through Online Certificate Status Protocol (OCSP) to ensure they have not been revoked.

Additional settings are required to support OCSP, I plan to add these shortly.


viv:
I am planning to add the documentation changes to this PR but won't get a chance to do that until tomorrow.

distribution/src/bin/openfire.sh
```diff
@@ -138,6 +138,11 @@ case $arguments in
 esac
 done
 
+# Java security config
+OPENFIRE_OPTS="${OPENFIRE_OPTS} -Djava.security.properties=${OPENFIRE_HOME}/resources/security/java.security"
```
guusdk, line 142: Does this _add_ to the JVM-provided security properties, or does it replace them (I'm not immediately finding heaps of documentation on this property)? In the latter case, I worry about loosing other, possibly important, configuration.

viv: This adds to the existing opts by including them at the start `${OPENFIRE_OPTS} `

guusdk:
We've discussed this in a video call, but I'd like to capture two points that may need additional attention:
- If we configure Openfire to do revocation checking, but when Java is configured to not support (eg missing java security settings that enable OCSP), we should possibly inform the user that their system is not optimally configured (by having a warning on the admin console)
- When 'mutual authentication' is used (SASL EXTERNAL, either in client-to-server or server-to-server), then the 'client' in provides a certificate to the server. This certificate should also be the subject of revocation checks.

PR #2622, opened 2024-11-26
OF-2914: Reduce verbosity of logged messages when S2S fails.

guusdk:
Moves most verbose (and likely confusing to most users) messages related to S2S / TLS failures from WARN to DEBUG level of logging (effectively hiding them). The messages are replaced by a more human-descriptive message informing the user what's going on.


guusdk:
Merging this without a review to get it tested on Ignite.

PR #2623, opened 2024-11-26
OF-2919: Prevent exception after S2S TLS fails

guusdk:
When TLS fails in a server-to-server connection, this frequently is followed up with another exception, mentioning that a `^W` character couldn't be parsed as valid XMPP.

This End-of-Transmission character is likely sent by the peer, as the end of the (failed) TLS handshake. Openfire should let the TLS handler process that character. This prevents the XMPP handler from having to process it.


guusdk:
Merging this without a review to get it tested on Ignite.

PR #2626, opened 2024-11-28
fix: Improve certificate revocation checking

viv:
Configure PKIXRevocationChecker to only check end-entity certificates and use soft-fail for unavailable revocation information.

This attempts to balance security and reliability:
- Focusing revocation checks on leaf certificates, where revocation is most critical, avoiding issues with missing root certificate CRL distribution points
- Preventing validation failures when OCSP/CRL servers are unreachable or when revocation information isn't available for some certificates
- Avoiding common issues with intermediate/root certificate CRL checking


guusdk:
Can we make both options configurable please? That way, if Openfire is operating in a closed, tightly controlled network, an administrator can reconfigure that balance between security and reliability.

xmppserver/src/main/java/org/jivesoftware/openfire/keystore/OpenfireX509TrustManager.java
```diff
@@ -274,6 +274,23 @@ protected CertPath checkChainTrusted( CertSelector selector, X509Certificate...
...
+                // This avoids issues with chains where the root certificate isn't included
+                // in the chain (e.g. Let's Encrypt) and its CRL distribution point isn't accessible
+                PKIXRevocationChecker.Option.ONLY_END_ENTITY,
+
+                // Continue validation if revocation information is unavailable
+                // This prevents failures when OCSP/CRL servers are unreachable or
+                // when revocation information isn't available for some certificates
+                PKIXRevocationChecker.Option.SOFT_FAIL
+            ));
+            parameters.addCertPathChecker(revChecker);
```
guusdk, line 291: This looks a bit funny, but is probably OK. Just in case: the `PKIXRevocationChecker` is obtained from the `PathBuilder` instance, but after modification is added to an instance of `PKIXBuilderParameters`. Is that right? Is that PathBuilder instance guaranteed to give some soft of default instance?

xmppserver/src/main/java/org/jivesoftware/openfire/keystore/OpenfireX509TrustManager.java
```diff
@@ -274,6 +277,30 @@ protected CertPath checkChainTrusted( CertSelector selector, X509Certificate...
...
 
+        if (checkRevocation) {
+            // Configure revocation checking - using default OCSP preference (OCSP before CRL)
+            PKIXRevocationChecker revChecker = (PKIXRevocationChecker)pathBuilder.getRevocationChecker();
+
+            EnumSet<PKIXRevocationChecker.Option> options = EnumSet.noneOf(PKIXRevocationChecker.Option.class);
+
+            // Only verify revocation status of end-entity (leaf) certificates if configured
+            // This avoids issues with chains where the root certificate isn't included
+            // in the chain (e.g. Let's Encrypt) and its CRL distribution point isn't accessible
```
dwd, line 288: This comment seems wrong.

The chain checked is presumably after path-building, so if the old Trust Anchor is not included in the peer chain, we'll have used our own copy anyway, surely?

(Exception being if DANE is in play, but then we possibly don't do CRL checking anyway, and Openfire doesn't yet do DANE).

Is there a better explanation of EE-only status checking?

viv: Thanks Dave, your comment led me to realise I hadn't understood what was going on.

guusdk (approved):
Changes look good to me. @viv did you leave this in 'draft' intentionally?

viv:
No, I think I've covered everything.

PR #2627, opened 2024-11-29
OF-2918 Clear MUC Chat History

hqv126:


xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -1308,6 +1308,48 @@ public void removeOccupant(@Nonnull final MUCOccupant occupant) {
         MUCEventDispatcher.occupantLeft(occupant.getOccupantJID(), occupant.getUserAddress(), occupant.getNickname());
     }
 
+    /**
+     * Clears the chat history of a room. Sends out message retraction stanzas to all occupants, for all the message in the chat history.
+     */
+    public void clearChatHistory() {
+        Iterator<Message> history = roomHistory.getMessageHistory();
```
guusdk, line 1315: This will work in principle, but I think it should be improved in a number of ways:
1. As the generation and distribution of revocation messages is potentially a (very) long running process, it should be async.
2. Lets iterate over the history backwards, sending a revocation for the most recently received message first. That message is a lot more likely to be currently displayed in a connected client, than the oldest message in a chatroom. By working backwards, we start with the most visible messages. I believe that this will lead to a better user-experience.
3. Please add a configuration option (in the form of a SystemProperty) that allows an administrator to skip sending the revocation messages.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -1308,6 +1308,48 @@ public void removeOccupant(@Nonnull final MUCOccupant occupant) {
...
+            retractionMsg.addChildElement("retract", "urn:xmpp:message-retract:1").addAttribute(
+                "id",
+                new JID(getName(), getMUCService().getServiceDomain(), null).toBareJID()
+            );
+            // A fallback element is added to provide a fallback message for clients that do not support message retraction
+            retractionMsg.addChildElement("fallback", "urn:xmpp:fallback:0").addAttribute(
+                "for",
+                "urn:xmpp:message-retract:1"
+            );
+            retractionMsg.setBody("A request was received to retract a previous message as part of clearing the chat room history, but this is not supported by your client.");
```
guusdk, line 1338: Please make this text a translatable text (using the i18n file). You can use `org.jivesoftware.util.LocaleUtils#getLocalizedString(java.lang.String)` to obtain the text from that file.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -1308,6 +1308,48 @@ public void removeOccupant(@Nonnull final MUCOccupant occupant) {
         MUCEventDispatcher.occupantLeft(occupant.getOccupantJID(), occupant.getUserAddress(), occupant.getNickname());
     }
 
+    /**
+     * Clears the chat history of a room. Sends out message retraction stanzas to all occupants, for all the message in the chat history.
```
guusdk, line 1312: Please describe in another term what 'clear' means, as it is easy to misinterpret this (eg: remove from memory vs revocation vs completely deleted from database).

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
```diff
@@ -559,6 +562,39 @@ public static void deleteFromDB(MUCRoom room) {
...
+        PreparedStatement pstmt = null;
+        boolean abortTransaction = false;
+        try {
+            con = DbConnectionManager.getTransactionConnection();
+            pstmt = con.prepareStatement(CLEAR_ROOM_CHAT_HISTORY);
+            pstmt.setLong(1, room.getID());
+            pstmt.executeUpdate();
+
+            // Update the room (in memory) to indicate the it's no longer in the database.
+            room.setSavedToDB(false);
```
guusdk, line 586: When the conversation log has been removed from the database, the room itself has not been deleted from the database, so I don't think that `room.setSavedToDB(false)` is correct. This would signal that the room itself no longer is in the database.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
```diff
@@ -147,6 +147,9 @@ public class MUCPersistenceManager {
         "INSERT INTO ofMucConversationLog (roomID,messageID,sender,nickname,logTime,subject,body,stanza) VALUES (?,?,?,?,?,?,?,?)";
     private static final String DELETE_ROOM_HISTORY =
         "DELETE FROM ofMucConversationLog WHERE roomID=?";
+    // Clear the chat history for a room but don't clear the messages that set the room's subject
```
guusdk, line 150: This is a good find. I'm wondering if there are other usages of `DELETE_ROOM_HISTORY` that instead should preserve the room subject.

xmppserver/src/main/webapp/muc-room-clear-chat.jsp
```diff
@@ -0,0 +1,96 @@
+<%@ page contentType="text/html; charset=UTF-8" %>
+<%--
+  -
+  - Copyright (C) 2004-2008 Jive Software, 2017-2024 Ignite Realtime Foundation. All rights reserved.
```
guusdk, line 4: The copyright needs not include Jive Software. Please replace it with:
> Copyright (C) 2024 Ignite Realtime Foundation.

If any of the files that you edit have a copyright notice that doesn't include the current year for the Ignite Realtime Foundation, please update that, too.

xmppserver/src/main/webapp/muc-room-clear-chat.jsp
```diff
@@ -0,0 +1,96 @@
...
+    // Load the room object
+    MUCRoom room = webManager.getMultiUserChatManager().getMultiUserChatService(roomJID).getChatRoom(roomName);
+
+    // Handle a room clear:
+    if (clear) {
+        // Clear the room
+        if (room != null) {
+            room.clearChatHistory();
+            // Log the event
+            webManager.logEvent("clear chat history of MUC room ", roomName);
```
guusdk, line 65: Please use either the full address of the room, or include the room service. Multiple services can have a room of the same name.

guusdk (changes requested):
Thanks for this! I've left a bunch of suggestions. Most of them are pretty minor, but I believe some others are more important.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
hqv126: I haven't come across a use case yet for keeping a chatroom's topic after its deletion.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
hqv126: Thank you Guus. I've updated the code as suggested.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
hqv126: Thanks, this is done now.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
hqv126: This is done now.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
hqv126: Good spot, I've fixed this error.

xmppserver/src/main/webapp/muc-room-clear-chat.jsp
hqv126: Done.

xmppserver/src/main/webapp/muc-room-clear-chat.jsp
hqv126: Done.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -65,6 +65,7 @@
  */
 @JiveID(JiveConstants.MUC_ROOM)
 public class MUCRoom implements GroupEventListener, UserEventListener, Externalizable, Result, Cacheable {
+    public static final boolean BULK_MSG_RETRACTION_ENABLED = JiveGlobals.getBooleanProperty("chatManagement.bulkMsgRetractionEnabled", true);
```
guusdk, line 68: Instead of `JiveGlobals` can we use `SystemProperty` please? That has various benefits (better type-safety, being self-descriptive, etc). An example of a boolean-based SystemProperty can be found in https://github.com/igniterealtime/Openfire/blob/a446aa55d936ebb7ed016ac889cb91917e3fd12a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java#L71-L75

xmppserver/src/main/webapp/muc-room-clear-chat.jsp
```diff
@@ -60,9 +60,13 @@
     if (clear) {
         // Clear the room
         if (room != null) {
-            room.clearChatHistory();
-            // Log the event
-            webManager.logEvent("clear chat history of MUC room ", roomName);
+            webManager.logEvent("Making a request to clear chat history of MUC room ", roomJID.toString());
```
guusdk, line 63: These events are logged in an audit log, not a 'regular' (debug) log. Please do not add two messages, just one will be fine.

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
```diff
@@ -77,6 +77,45 @@ public class SequenceManager {
...
+        // Otherwise, create a new SequenceManager for the unique name.
+        int type = 1;
+        while (true) {
+            if (!managers.containsKey(type)) {
+                uniqueNameToType.put(uniqueName, type);
+                return new SequenceManager(type, size);
+            }
+            type++;
+        }
+    }
```
guusdk, line 118: This basically replaces a unique number with a unique name, but that's still something that needs to be correctly correlated between implementations (granted, the risk of accidental duplicate value usage is a bit lower).

This is solving an issue that, to the best of my knowledge, hasn't existed in the past two decades. It adds a bit of complexity that we'll need to maintain, which I dislike.

I am, for example, wondering if this holds up in a cluster. The `uniqueNameToType` map doesn't seem to share its content across cluster nodes, which may be an issue.

Also, how does this behave after a restart, especially when various bits of code in Openfire get initialized in a potentially different order. Does this code guarantee that the same code will use the same sequence again? If that's not guaranteed, then things are likely to get broken (as these sequence values are typically used as identifiers for database content).

I'm not sure if the benefit of this change outweighs the added complexity. I prefer to stick with the previous approach, of simply using magic numbers. An alternative like the one proposed in this commit would at the very least require a fair bit of (automated) tests, before I'd be comfortable using this.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCEventDispatcher.java
```diff
@@ -122,6 +122,16 @@ public static void roomDestroyed(JID roomJID) {
         }
     }
 
+    public static void roomClearChatHistory(JID roomJID) {
+        for (MUCEventListener listener : listeners) {
+            try {
+                listener.roomClearChatHistory(roomJID);
+            } catch (Exception e) {
+                Log.warn("An exception occurred while dispatching a 'roomClearChatHistory' event!", e);
```
guusdk, line 130: Please add the room JID to the logged message, which is going to be very helpful when we get to analyze log files that contain this warning.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -1432,6 +1434,9 @@ public void destroyRoom(JID alternateJID, String password, String reason) {
         MUCPersistenceManager.deleteFromDB(this);
         // Remove the history of the room from memory (preventing it to pop up in a new room by the same name).
         roomHistory.purge();
+        // If we are not preserving room history on deletion, fire event to clear chat room history
+        if(!preserveHistOnRoomDeletion)
```
guusdk, line 1438: I don't care much about people using different coding styles, but not using curly braces on single-line if-statements is an accident waiting to happen. Please use something like this instead:

```
if(!preserveHistOnRoomDeletion) {
    MUCEventDispatcher.roomClearChatHistory(getSelfRepresentation().getOccupantJID());
}
```

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -1353,6 +1353,8 @@ public CompletableFuture<Void> clearChatHistory() {
             MUCPersistenceManager.clearRoomChatFromDB(this);
             // Remove the history of the room from memory (preventing it to pop up in a new room by the same name).
             roomHistory.purge();
+            // Fire event to clear chat room history
+            MUCEventDispatcher.roomClearChatHistory(getSelfRepresentation().getOccupantJID());
```
guusdk, line 1357: Why `getSelfRepresentation().getOccupantJID()`? Isn't a simple `getJID()` enough?

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
guusdk: Also here `getJID()` rather than the occupant JID from the self-representation is probably preferred.

distribution/src/conf/openfire.xml
```diff
@@ -24,6 +24,10 @@
 
     <locale>en</locale>
 
+    <chatManagement>
```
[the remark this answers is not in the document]
hqv126: Thanks, this is done now.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
hqv126: Done.

xmppserver/src/main/webapp/muc-room-clear-chat.jsp
hqv126: Done.

xmppserver/src/main/java/org/jivesoftware/database/SequenceManager.java
hqv126: Thanks Guus.
In the interest of time, no changes have now been made to SequenceManager.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCEventDispatcher.java
hqv126: Done.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
hqv126: Done.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
hqv126: MUCEventDispatcher calls often uses getSelfRepresentation().getOccupantJID() in the existing code, but I have changed MUCEventDispatcher.roomClearChatHistory and MUCEventDispatcher.roomDestroyed to use getJID() as suggested.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -88,6 +86,12 @@ public class MUCRoom implements GroupEventListener, UserEventListener, Externali
         .setDynamic(true)
         .build();
 
+    public static final SystemProperty<Boolean> BULK_MSG_RETRACTION_ENABLED = SystemProperty.Builder.ofType(Boolean.class)
+        .setKey("xmpp.muc.bulkretraction")
```
guusdk, line 90: Please add an entry for `system_property.xmpp.muc.bulkretraction` in `openfire_i18n.properties` that is a single line, describing the purpose of this property.

hqv126: Done.

xmppserver/src/main/java/org/jivesoftware/database/bugfix/Issue369and370.java
```diff
@@ -29,29 +40,34 @@ public static void executeMigration() {
...
+            pstmt.setFetchDirection(ResultSet.FETCH_FORWARD);
             while (rs.next()) {
                 String roomJID = rs.getString("room");
                 long conversationID = rs.getLong("conversationID");
                 long roomID = roomJIDToRoomIDMap.computeIfAbsent(roomJID, k -> roomIDSequenceManager.nextUniqueID());
                 conversationIDToRoomIDMap.put(conversationID, roomID);
             }
-            Log.info("1 of 7. Generated room IDs for unique rooms");
+
+            pstmt.setFetchSize(250);
```
guusdk, line 55: Why do this here? Is this intended or a a copy/paste bug?

hqv126: Yes, they are redundant and misplaced calls. Removed now.

guusdk:
I've added a couple of minor tweaks and rebased this PR.

PR #2629, opened 2024-12-03
feat: Add Mac OS support to truststore update script

viv:
I made these updates while testing the recent certificate revocation work and thought it _might_ be useful to share.

The script now works on Mac and Linux. I'm not sure how useful this will be so feel free to close if it's not of use.


akrherz (approved):

guusdk:
Thanks @viv! I'm not using a Mac myself, so I've not tested this. From a read, I can't see how this would break existing functionality, which I think is enough of a bar to merge this.

PR #2630, opened 2024-12-05
OF-2760: Adds option to retire a MUC room name on deletion

viv:
Retiring a room means that this room's name cannot be used again after the room is deleted. 

Where a MUC may have particular meaning to plugins or to other systems, assume that deletion is explicit and intentional. Do not allow the MUC name (and so JID) to be repurposed, preventing impersonation / corruption / access of MUC history, or other unforeseen side effects.

### DIFF
--- a/distribution/src/database/openfire_db2.sql
+++ b/distribution/src/database/openfire_db2.sql
@@ -211,7 +211,7 @@ CREATE TABLE ofMucRoom (
   canDiscoverJID      INTEGER       NOT NULL,
   logEnabled          INTEGER       NOT NULL,
   preserveHistOnDel   INTEGER       NOT NULL,
-  tombstone           INTEGER       NOT NULL,
+  retireOnDeletion    INTEGER       NOT NULL,
   subject             VARCHAR(100),
   rolesToBroadcast    INTEGER       NOT NULL,
   useReservedNick     INTEGER       NOT NULL,
@@ -235,10 +235,10 @@ CREATE TABLE ofMucRoomProp (
   CONSTRAINT ofMucRoomProp_pk PRIMARY KEY (roomID, name)
 );
 
-CREATE TABLE ofMucRoomTombstone (
+CREATE TABLE ofMucRoomRetiree (
   serviceID           INTEGER       NOT NULL,
   name                VARCHAR(50)   NOT NULL,
-  CONSTRAINT ofMucRoomTombstone_pk PRIMARY KEY (serviceID, name)
+  CONSTRAINT ofMucRoomRetiree_pk PRIMARY KEY (serviceID, name)
 );
 
 CREATE TABLE ofMucAffiliation (
--- a/distribution/src/database/openfire_db2.sql
+++ b/distribution/src/database/openfire_db2.sql
@@ -238,6 +238,7 @@ CREATE TABLE ofMucRoomProp (
 CREATE TABLE ofMucRoomRetiree (
   serviceID           INTEGER       NOT NULL,
   name                VARCHAR(50)   NOT NULL,
+  retiredAt           TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
   CONSTRAINT ofMucRoomRetiree_pk PRIMARY KEY (serviceID, name)
 );
 
--- a/distribution/src/database/openfire_db2.sql
+++ b/distribution/src/database/openfire_db2.sql
@@ -238,6 +238,8 @@ CREATE TABLE ofMucRoomProp (
 CREATE TABLE ofMucRoomRetiree (
   serviceID           INTEGER       NOT NULL,
   name                VARCHAR(50)   NOT NULL,
+  alternateJID        VARCHAR(2000),
+  reason              VARCHAR(1024),
   retiredAt           TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
   CONSTRAINT ofMucRoomRetiree_pk PRIMARY KEY (serviceID, name)
 );
--- a/distribution/src/database/openfire_db2.sql
+++ b/distribution/src/database/openfire_db2.sql
@@ -211,6 +211,7 @@ CREATE TABLE ofMucRoom (
   canDiscoverJID      INTEGER       NOT NULL,
   logEnabled          INTEGER       NOT NULL,
   preserveHistOnDel   INTEGER       NOT NULL,
+  tombstone           INTEGER       NOT NULL,
   subject             VARCHAR(100),
   rolesToBroadcast    INTEGER       NOT NULL,
   useReservedNick     INTEGER       NOT NULL,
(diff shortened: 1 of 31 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/muc/MultiUserChatManager.java
```diff
@@ -991,4 +991,114 @@ public List<Multimap<String, String>> clusteringStateConsistencyReportForMucRoom
...
+
+        try {
+            con = DbConnectionManager.getConnection();
+
+            if ((startIndex == 0) && (numResults == Integer.MAX_VALUE)) {
+                pstmt = con.prepareStatement(GET_RETIREES);
+                pstmt.setLong(1, serviceID);
+                // Set the fetch size. This will prevent some JDBC drivers from trying
+                // to load the entire result set into memory.
+                DbConnectionManager.setFetchSize(pstmt, 500);
```
guusdk, line 1048: I happened to run into this in an unrelated project: only setting the fetch size is not sufficient to prevent the entire result set from being loaded in memory (at least not for MSSQL). That requires a bit more configuration of the query/result set Please use the changes in https://github.com/igniterealtime/openfire-monitoring-plugin/pull/410 as a guide.

I'm assuming that you've copied this from another query somewhere in Openfire. If possible, could you modify that implementation with the same change please?

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
```diff
@@ -1627,5 +1644,32 @@ public static void deleteLocalProperty(String subdomain, String name) {
...
-    
+
+    /**
+     * Check if a room name is retired for a given service.
+     *
+     * @param roomName the name of the room to check.
+     * @param multiUserChatService the service to check the room name against.
+     * @return true if the room name is retired for the supplied service, false otherwise.
+     */
+    public static boolean isRoomRetired(String roomName, MultiUserChatService multiUserChatService) {
```
guusdk, line 1655: I'm perfectly happy for an answer say: "that's overkill", but: is it worth caching the retirement state of rooms? Is there a chance that this method is invoked in a loop, or repeatedly? Instead of doing a query per room, it may make sense to load all retirees in a (possibly short-lived) cache. Changes to the retirement status of a room will be rare, and occur only in this class, I think, which would give a good opportunity to keep that cache up-to-date. We do need to make sure such a strategy would work in a cluster (by either making this a clustered cache, or in another way ensure that a change on one cluster node is visible on the other nodes).

xmppserver/src/main/webapp/muc-room-retirees.jsp
```diff
@@ -0,0 +1,302 @@
+<%@ page contentType="text/html; charset=UTF-8" %>
```
guusdk, line 1: We're trying to get away from scriptlets, and use JSTL instead. Could you at the very least condense the Java code to the start of the page, place any variables that you want on the request context, and use those in JSTL in the remainder of the page please?

distribution/src/database/openfire_db2.sql
```diff
@@ -234,6 +235,11 @@ CREATE TABLE ofMucRoomProp (
   CONSTRAINT ofMucRoomProp_pk PRIMARY KEY (roomID, name)
 );
 
+CREATE TABLE ofMucRoomRetiree (
+  serviceID           INTEGER       NOT NULL,
+  name                VARCHAR(50)   NOT NULL,
```
guusdk, line 240: It's almost a shame to not add a bit of metadata here, like the timestamp of the retirement and the JID of the entity that did the retiring, the 'destroy' message and alternative JID if those were used. All of this would be optional (apart from the timestamp, which can just be a default).

viv:
> This data is not yet exposed on the admin console retirees page. 

I'll pick this up on Monday morning.

distribution/src/database/openfire_db2.sql
viv: I have added alternateJID, reason, and retired-at timestamp.

xmppserver/src/main/webapp/muc-room-retirees.jsp
viv: I think I've covered this now.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
viv: I think this will have to wait for now.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MultiUserChatManager.java
viv: I've updated all instances of this.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MultiUserChatManager.java
```diff
@@ -991,4 +991,114 @@ public List<Multimap<String, String>> clusteringStateConsistencyReportForMucRoom
...
+        try {
+            con = DbConnectionManager.getConnection();
+            pstmt = con.prepareStatement(COUNT_RETIREE);
+            pstmt.setLong(1, serviceID);
+            rs = pstmt.executeQuery();
+            if (rs.next()) {
+                count = rs.getInt(1);
+            }
+        } catch (SQLException e) {
+            Log.error("An unexpected exception occurred while trying to count the number of MUC room retirees.", e);
```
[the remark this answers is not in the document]
viv: I've included the service ID.

viv:
I think this is ready for another review.

guusdk:
Rooms that are non-persistent but do have 'retire-on-delete' set, do not retire. I think in principle, the setting should be honored.

It might be desirable to have a _different_ default setting for 'retire-on-delete' for persistent vs. non-persistent rooms (but I'm happy for that improvement to go in a new ticket).

guusdk:
The 'retire-on-delete' option should be changeable through the room configuration form. 

In the screenshot below a new config setting that we've recently added does pop up, but not 'retire-on-delete'.

image

xmppserver/src/main/java/org/jivesoftware/openfire/muc/spi/MUCPersistenceManager.java
guusdk: I've raised this as https://igniterealtime.atlassian.net/browse/OF-2934 for someone to pick up in the future.

guusdk:
> I think this is ready for another review.

I've re-reviewed it, and left some suggestions to be improved in this PR, but also raised a couple of tickets for us to tackle outside of the scope of this PR ([OF-2933](https://igniterealtime.atlassian.net/browse/OF-2933), [OF-2934](https://igniterealtime.atlassian.net/browse/OF-2934)).

Are you aware of the 're-request review' functionality in GitHub? That has the benefit of sending out all the required emails etc.

guusdk (approved):
I've tested both new changes, which seem to work as intended. Thanks!

PR #2635, opened 2024-12-17
OF-2921: Prevent deadlock by not broadcasting synchronously

guusdk:
When a client drops offline, a presence update is broadcast on behalf of that client. We've observed deadlocks occurring when this happens (as the broadcast itself, especially when it occurs in context in of MUC rooms, will acquire mutexes).

There appears to be little reason for this broadcast to happen synchronous to the process of session termination process. To reduce the likelihood of deadlocks, this commit makes the broadcast happen in an asynchronous process.


akrherz:
IIRC, this was done synchronously due to issues we observed with remote MUC users in open_chat not seeing the shutdown when Ignite's Openfire was restarted.

guusdk:
I don't quite recall that issue, but given that this introduces deadlocks (which causes the system to become unusable and throw out-of-memory exceptions, the cure might be worse than the ailment in this case.

akrherz:
Agreed, I think this was https://igniterealtime.atlassian.net/browse/OF-1688

akrherz (approved):

PR #2636, opened 2024-12-17
OF-2921: Do not acquire mutex on unused StreamManager

guusdk:
Historically, disabling the Stream Management functionality has been a work-around for various issues.

To be able to work around a deadlock issue as described in OF-2921, this commit prevents a Stream Management-related mutex to be acquired when the Stream Management feature is disabled.


Fishbowler (approved):
Seems safe and obvious.

PR #2643, opened 2025-01-03
OF-2941: Improve Version comparison that involve SNAPSHOTs

guusdk:
For certain versions of plugins, Openfire incorrectly identifies that an update is available.

When Openfire has installed `10.1.7.3-SNAPSHOT` it incorrectly identifies that version `10.1.7.2` is an update.

This seems to be caused by Openfire’s abuse of the 4th numeric value in a version. It was originally intended to be a classifier for the status (eg: ‘the second beta release’). When that 4th value is used as part of a four-digit version, things do not quite compare as expected.

To illustrate, compare these two scenarios. The numbers 3 and 2 are all stored in the same field, which gets interpreted differently.

- `10.1.7 beta 3` is older than `10.1.7 release 2`
- `10.1.7.3-SNAPSHOT` is expected to be newer than `10.1.7.2` (release).

Prior to this commit, Openfire did not flag that `10.1.7.3-SNAPSHOT` is newer than `10.1.7.2`, which was causing confusion.

In this commit, _only_ the `SNAPSHOT` release status is now compared differently (as if the 4th value is part of the version string).

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/update/Update.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/update/Update.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2005-2008 Jive Software, 2017-2018 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2005-2008 Jive Software, 2017-2025 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


guusdk:
This magic remains a bit of a hack, but it's arguably a more consistent hack with this change applied.

akrherz (approved):

akrherz:
@guusdk I wouldn't mind seeing this go to 4.9 branch, can it?

PR #2642, opened 2025-01-03
Publish Docker image to GitHub registry

guusdk:
When a PR gets merged to the Main branch, or when a tag is created, a Docker image should be published to the GitHub-provided registry.

PR #2649, opened 2025-01-06
test: Run integration tests in CI workflow

viv:


viv:
This appears to be working, I'd like to have better visualisation of test outcome. It would also be great to not need to build the Docker image for the integration tests, something https://github.com/igniterealtime/Openfire/pull/2642 might resolve.

Fishbowler (approved):
Sexy.

Scutua (approved):

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -60,6 +60,53 @@ jobs:
...
+      - uses: actions/checkout@v4
+
+      - name: Build Docker Image
+        run: |
+          docker build -t openfire:latest .
+
+      - name: Checkout Integration Tests
+        uses: actions/checkout@v4
+        with:
+          repository: surevine/openfire-integration-tests
```
viv, line 84: Just spotted that this needs changing now that we have migrated the repository to IR. I'll push a change.

PR #2650, opened 2025-01-06
Denote Openfire 5.0.0 Alpha

akrherz:
The intention here is to create build and maven artifacts that can be targeted as we rapidly approach 5.0.0 Beta.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>parent</artifactId>
         <groupId>org.igniterealtime.openfire</groupId>
-        <version>5.0.0-SNAPSHOT</version>
+        <version>5.0.0-alpha</version>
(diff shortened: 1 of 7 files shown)


guusdk (approved):

akrherz:
I think the docker issue is chicken/egg problem, but am unsure.

guusdk:
I had hoped that the build wouldn't need the repository for a dependency that it is generating itself. Maybe the naming of the version (with the `-alpha`) is messing this up. I'm happy to ignore this for now.

akrherz:
> I had hoped that the build wouldn't need the repository for a dependency that it is generating itself.

My naive understanding of what is happening here is that the docker build process is not looking at the local maven repo, which uniquely has the newly named alpha artifact, which does not exist yet at ignite's maven repo....

PR #2652, opened 2025-01-07
Make close event handling asynchronous and remote session locator improvement

guusdk:
This revisits the fix for OF-2921, which introduces a problem in which a broadcast of presence occurs while the session is removed from the session manager. This causes issues, as the broadcast needs data associated to the session (such as its roster, privacy lists, etc). 

This PR holds two changes that improve on the previous solution. Please refer to individual commit messages for details.


akrherz (approved):
Might as well dogfood this.

PR #2664, opened 2025-01-22
OF-2437: Improve CachingPubsubPersistenceProvider

guusdk:
This new unit test verifies behavior of CachingPubsubPersistenceProvider (which fails, as indicated by some of these test failing).

The tests introduced here only cover one (of three) types of operations that are being scheduled by the implementation:
- Node creation/changing/removal

These two other types of operation still need to be added:
- Affiliation creation/update/removal
- Subscription creation/update/removal

### DIFF
--- a/xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderNodeOperationsTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderNodeOperationsTest.java
@@ -206,7 +206,7 @@ public class CachingPubsubPersistenceProviderNodeOperationsTest
         assertEquals(1, pending.size());
 
         final CachingPubsubPersistenceProvider.NodeOperation nodeOperation = pending.get(0);
-        assertEquals(mockOrigNode.getUniqueIdentifier(), nodeOperation.node.getUniqueIdentifier());
+        assertEquals(mockReplaceNode.getUniqueIdentifier(), nodeOperation.node.getUniqueIdentifier());
         assertNotEquals(new Date(0), nodeOperation.node.getCreationDate());
         assertEquals(CachingPubsubPersistenceProvider.NodeOperation.Action.UPDATE, nodeOperation.action);
         assertNull(nodeOperation.subscription);
--- a/xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderNodeOperationsTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderNodeOperationsTest.java
@@ -1,0 +1,438 @@
+/*
+ * Copyright (C) 2025 Ignite Realtime Foundation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.openfire.pubsub;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Verifies the implementation of {@link CachingPubsubPersistenceProvider}
+ *
+ * The unit tests in this class are limited to the operations that create, update and remove nodes.
+ *
+ * @author Guus der Kinderen, guus@goodbytes.nl
+ */
+@ExtendWith(MockitoExtension.class)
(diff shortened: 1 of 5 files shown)


guusdk:
@Fishbowler do you think that the test approach is valid? Also, is the behavior (as tested, and documented in the test) consistent and correct?

guusdk:
The second commit in this PR fixes the issues that are exposed by the unit tests that are added in the first commit.

guusdk:
The added unit tests do not verify assert the result of _obtaining_ data from a provider. This is certainly something I'd like to (eventually) add, but this might lead to a refactoring of the interface (the current interface definition requires a caller to pass along an instance of the pubsub service on which the data is to be loaded - that's awkward design, making it hard to write tests for).

xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderNodeOperationsTest.java
```diff
@@ -0,0 +1,438 @@
...
+        // Execute system under test.
+        provider.updateNode(mockOrigNode);
+        provider.updateNode(mockReplaceNode);
+
+        // Verify results.
+        final List<CachingPubsubPersistenceProvider.NodeOperation> pending = provider.nodesToProcess.computeIfAbsent(mockOrigNode.getUniqueIdentifier(), id -> new ArrayList<>());
+        assertEquals(1, pending.size());
+
+        final CachingPubsubPersistenceProvider.NodeOperation nodeOperation = pending.get(0);
+        assertEquals(mockOrigNode.getUniqueIdentifier(), nodeOperation.node.getUniqueIdentifier());
```
Fishbowler, line 209:
```suggestion
        assertEquals(mockReplaceNode.getUniqueIdentifier(), nodeOperation.node.getUniqueIdentifier());
```
I know it's entirely semantic...

xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderNodeOperationsTest.java
```diff
@@ -0,0 +1,438 @@
...
+        final CachingPubsubPersistenceProvider.NodeOperation firstNodeOperation = pending.get(0);
+        assertEquals(mockOrigNode.getUniqueIdentifier(), firstNodeOperation.node.getUniqueIdentifier());
+        assertEquals(new Date(0), firstNodeOperation.node.getCreationDate());
+        assertEquals(CachingPubsubPersistenceProvider.NodeOperation.Action.CREATE, firstNodeOperation.action);
+        assertNull(firstNodeOperation.subscription);
+        assertNull(firstNodeOperation.affiliate);
+
+        final CachingPubsubPersistenceProvider.NodeOperation secondNodeOperation = pending.get(1);
+        assertEquals(mockOrigNode.getUniqueIdentifier(), secondNodeOperation.node.getUniqueIdentifier());
+        assertNotEquals(new Date(0), secondNodeOperation.node.getCreationDate());
```
Fishbowler, line 169: What's going on with these date assertions?
The first one is 0, the second one is 'sometime after'?

xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderSubscriptionOperationsTest.java
```diff
@@ -0,0 +1,740 @@
...
+        final Node mockNode = Mockito.mock(Node.class);
+        Mockito.lenient().when(mockNode.getUniqueIdentifier()).thenReturn(new Node.UniqueIdentifier("mock-service-1", "mock-node-a"));
+        final JID subscriber = new JID("test-user-1", "example.org", null);
+        final String subscriptionId = "test-subscription-id-" + StringUtils.randomString(7);
+        final NodeSubscription subscription = new NodeSubscription(mockNode, subscriber, subscriber, NodeSubscription.State.subscribed, subscriptionId);
+
+        // Execute system under test.
+        provider.createSubscription(mockNode, subscription);
+        provider.updateSubscription(mockNode, subscription);
+        provider.removeSubscription(subscription);
```
Fishbowler, line 460: What if the create is rejected because the subscription already exists?
The update could be optimised away, but the remove might still be needed, right?

Although... Can an update ever occur that prevents removing the subscription? Like updating an owner?

Fishbowler (changes requested):
I've added a few questions, and a bunch of little suggestions (mostly javadoc)

xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderNodeOperationsTest.java
guusdk: I'm using the creation date to verify that after the update, the _original_ date was updated. In mockReplaceNode that date is set to `1`. My reasoning was that I don't really care what value it is, as long as it's no longer the original `0`. I'm happy to make the check to more strictly check for `1`.

xmppserver/src/test/java/org/jivesoftware/openfire/pubsub/CachingPubsubPersistenceProviderSubscriptionOperationsTest.java
guusdk: I've been debating things like this while I was coding this. I tried to make decisions about a very basic aspect of the system-under-test: it is strictly a data-access-object, in the sense that its sole responsibility is to interact with persistent storage. In that sense, it has less responsibility to verify that the invocation is correct (eg: flag that a creation should fail because it already exists). I think what you describe is something that's on the edge of that balance. I'm not really convinced of either approach to be honest.

Fishbowler (approved):
Discussed queue optimisations, which we've decided could only go wrong by another bug elsewhere in the business logic or database layer. Adding defensive code, or a flag to reduce optimisation, would only serve to hide (or exacerbate) that problem, which feels like smell.

PR #2667, opened 2025-02-06
OF-2722 OF-2902: Remove commons-fileupload

guusdk:
The commons-fileupload dependency that is used by Openfire has a CVE reported against it (rather: against one of its dependencies).

Java/Jetty can nowadays natively process file uploads. There no longer is a need for commons-fileupload.

This commit replaces commons-fileupload, which resolves the outstanding CVE report.

Additionally some plugin-upload properties were slightly refactored.


xmppserver/src/main/java/org/jivesoftware/openfire/update/PluginDownloadManager.java
```diff
@@ -55,16 +56,23 @@ public Update downloadPlugin(String url) {
...
+    public DownloadStatus installPlugin(String username, String url, String version, int hashCode) {
+
         UpdateManager updateManager = XMPPServer.getInstance().getUpdateManager();
 
         boolean worked = updateManager.downloadPlugin(url);
 
+        if (worked) {
+            final String name = url.substring(url.lastIndexOf('/') + 1);
+            SecurityAuditManager.getInstance().logEvent(username, "downloaded plugin " + name, "Downloaded from: " + url);
+        }
```
Fishbowler, line 74: Shouldn't we be logging an attempt (and the result), rather than only logging if it was successful?

xmppserver/src/main/webapp/available-plugins.jsp
```diff
@@ -144,7 +129,7 @@
     }
 
     function startDownload(url, version, id) {
-        downloader.installPlugin(url, version, id, downloadComplete);
+        downloader.installPlugin('${admin:escapeHTMLTags(webManager.user.username).replaceAll("'", "&quot;")}', url, version, id, downloadComplete);
```
Fishbowler, line 132: This is rather defensive of the username.
I can imagine an LDAP username being pretty complex.
What else?

Also, shouldn't this be sanitised at the point of use, rather than at the callsite, in case its called elsewhere in the future? (Not a strong opinion - just an idea)

xmppserver/src/main/webapp/plugin-admin-jsp.jsp
Fishbowler, line 1: Why the rename?

Fishbowler (commented):
A few comments / questions.
I don't pretend to understand (remember?) how servlets work, but it all appears rational.

xmppserver/src/main/java/org/jivesoftware/openfire/update/PluginDownloadManager.java
guusdk: Not necessarily I think. This is an _audit_ log, which I think should primarily register events that change the system. A failed attempt to install something doesn't change the system. If we start adding failures, then the audit log quickly becomes a regular log.

xmppserver/src/main/webapp/available-plugins.jsp
guusdk: I'm escaping this here because the username is going to be a _hard-coded_ value in the generated Javascript code. In that sense, I think it _is_ the point of use (the username isn't passed as a referenced value, it's quite literally in the Javascript code. I'm not actually sure if we allow a `'` character in the username, so escaping that (to not mess up the Javascript call may be overly defensive - but I can live with that.

xmppserver/src/main/webapp/plugin-admin-jsp.jsp
guusdk: Prior to this change, the request would be handled directly by the JSP page. We now need it to be handled by the new servlet. The servlet will forward the request to the new JSP page. I didn't want to change the address by the end-user, so I used the original address for the Servlet. The JSP page then has to have a distinct address.

xmppserver/src/main/java/org/jivesoftware/openfire/update/PluginDownloadManager.java
Fishbowler: I've got opinions on this that I'll take to the MUC, but it's not really the topic of this PR. If I convince you, I'll raise a new ticket in Jira 😉

Fishbowler (approved):

PR #2669, opened 2025-02-06
OF-2754: Deprecate FastDateFormat

guusdk:
This marks the third-party FastDateFormat class as deprecated, and replaces its usage with Java-native APIs.


xmppserver/src/main/java/org/jivesoftware/openfire/audit/spi/AuditorImpl.java
```diff
@@ -445,8 +440,8 @@ public Element getElement() {
          *
          * @return the date when the packet was audited.
          */
-        public Date getCreationDate() {
```
akrherz, line 448: Concerns of API break here?

guusdk: Although the _method_ is public, the _class_ is private. I think that makes it safe to change this API, as all consumers are limited to those that have visibility to this class (which is only the 'outer' class in the same file).

Fishbowler (approved):

PR #2670, opened 2025-02-06
OF-2831: Modify tests to cover offending data

guusdk:
This modifies existing unit tests to explicitly use data that appears to have triggered OF-2831. The resulting tests do not fail.


akrherz:
> Maybe not bother with the backport if that conflicts?

I was curious if this issue would reproduce on the 4.9 branch.  It is a simple patch, I'll manually effort now.

PR #2671, opened 2025-02-06
OF-2956: Drop database table ofSASLAuthorized

guusdk:
The database table ofSASLAuthorized is unused, while its definition causes database installation/upgrade errors for some users.

This commit removes the database table completely, both for fresh installs of Openfire, as for updates.

### DIFF
--- a/distribution/src/database/openfire_db2.sql
+++ b/distribution/src/database/openfire_db2.sql
@@ -154,13 +154,6 @@ CREATE TABLE ofPrivacyList (
 );
 CREATE INDEX ofPrivacyList_default_idx ON ofPrivacyList (username, isDefault);
 
-
-CREATE TABLE ofSASLAuthorized (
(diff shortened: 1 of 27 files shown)


akrherz:
> causes database installation/upgrade errors for some users.

Lets then remove the schema upgrades done in script 19 for this table.  Otherwise, those users are still stuck.

guusdk:
I know of at least one person that had an issue with the table definition (notably, the index that includes the 4000-size text column). https://discourse.igniterealtime.org/t/this-rootcollectionnode-is-null/95133/7?u=guus

guusdk:
I don't remember any (but that proves very, very little). In any case, I've now modified the older upgrade scripts in such a way that if there are still Openfire instances out there that haven't executed them, they run less risk of running into the issue that was reported.

akrherz (approved):

PR #2673, opened 2025-02-06
OF-2793: Bundle an Oracle database driver

guusdk:
Given changes in the Oracle license, we can now bundle their database drivers with Openfire. The changes in this commit do just that.

I've opted to use ojdbc11-production rather than ojdbc17-production as the latter is documented to be compatible with Jakarta. We do not use Jakarta. ojdbc11 has been documented to be compatible with Java 17 and 21, which should be good enough for us for now.

Jetty's JSP compiler plugin is not happy with the standard Oracle dependencies. It fails on expecting but not finding oracle.i18n.util.LocaleMapper. The internet suggests that this is caused by Oracle's xmlparserv2 dependency. Excluding that 'fixes' the problem. I'm not sure what functionality this affects.

As a side-effect, the `lib` folder of a distribution now contains an unneeded POM file. I can't find an easy way to exclude that.

### DIFF
--- a/build/ci/updater/pom.xml
+++ b/build/ci/updater/pom.xml
@@ -90,6 +90,19 @@
       <artifactId>mssql-jdbc</artifactId>
       <version>9.4.1.jre11</version>
     </dependency>
+    <dependency>
+      <groupId>com.oracle.database.jdbc</groupId>
+      <artifactId>ojdbc11-production</artifactId>
+      <version>23.7.0.25.01</version>
+      <type>pom</type>
+        <exclusions>
+            <exclusion>
+                <!-- This dependency causes the JSPC compilation to fail.-->
+                <groupId>com.oracle.database.xml</groupId>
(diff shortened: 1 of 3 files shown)


build/ci/updater/pom.xml
Fishbowler, line 1: Does this need the corresponding change in .github/workflows/continuous-integration-workflow.yml?

guusdk: In the sense of adding a new integration test for the Oracle database? I'd love that, but lets do that in a distinct PR.

Fishbowler: Have you been able to otherwise test this?

guusdk:
No

Op vr 7 feb 2025 10:37 schreef Dan Caseley ***@***.***>:

build/ci/updater/pom.xml
guusdk: No, I haven't.

Fishbowler (approved):

PR #2675, opened 2025-02-07
OF-2960: Replace Oracle database driver's BOM project with individual dependencies

guusdk:
This works around an issue that creeps up when using the BOM project: it causes a POM file to be included in Openfire's distirbution. That file in turn causes the RPM build to fail.

As the distribution cannot be easily changed, this commit prevents usage of the project that introduces the POM file.

More context is provided in https://igniterealtime.atlassian.net/browse/OF-2960

### DIFF
--- a/build/ci/updater/pom.xml
+++ b/build/ci/updater/pom.xml
@@ -16,6 +16,7 @@
     <application.entrypoint>com.igniterealtime.openfire.updaterunner.Main</application.entrypoint>
     <maven.compiler.source>17</maven.compiler.source>
     <maven.compiler.target>17</maven.compiler.target>
+    <ojdbc.version>23.7.0.25.01</ojdbc.version>
   </properties>
 
   <build>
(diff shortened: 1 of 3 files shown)


Fishbowler (approved):

build/ci/updater/pom.xml
```diff
@@ -90,18 +90,52 @@
       <artifactId>mssql-jdbc</artifactId>
       <version>9.4.1.jre11</version>
     </dependency>
-    <dependency>
+    <dependency> <!-- OF-2960: Including dependencies of ojdbc11-production rather than ojdbc11-production itself. -->
+      <groupId>com.oracle.database.jdbc</groupId>
+      <artifactId>ojdbc11</artifactId>
+      <version>${ojdbc.version}</version>
+    </dependency>
```
Fishbowler, line 97: This fails in the tests because my hacky CI Updater test rig isn't a child project of the parent POM. Should we make it be a child? No clue if there'd be side effects.

PR #2677, opened 2025-02-11
OF-2821: Admin Console Client Sessions page with configurable columns

guusdk:
In OF-2706, the 'resource' column disappeared from the admin console page that showed the overview of client sessions. This proved to be problematic for some users.

In this commit, the columns that are shown on that page can be configured by each admin user. To do so, a new configuration page is added (accessible through a small link in the top-right corner of the session summary page). Settings are saved per-user, to allow for personal preferences.

image

image


Fishbowler (approved):
Looks sensible, but not tested.
What's the worst that could happen? 😬

guusdk:
> What's the worst that could happen? 😬

Mushroom-clouds on the horizon, probably. That's unlikely to be caused by this change though.

PR #2678, opened 2025-02-11
OF-2883: Add fallback to older style base64 encoding for security.xml

guusdk:
In Openfire 4.9.0, the base64 encoder was changed. This causes a compatibility issue with security.xml files that were generated by older values, that use the 'MIME' type of decoding (with linebreaks) that's not used in later versions. While persisting data in 'security.xml', linebreaks are replaced by white space.

With the changes in this commit, the decoder checks for white space, and uses the older-style type of decoding when found.


Fishbowler (approved):
I've read the issue, the implementation, the tests. All looks sensible.
I've not run it, but I'm confident this will solve the problem.

PR #2679, opened 2025-02-11
OF-2940: After database install, run database upgrade check

guusdk:
With this change, any database upgrade scripts that update the database content beyond what is in the database install script will be executed directly after the install scripts are executed.

One side-effect of this approach is that under certain conditions, the 'success' message can be logged twice. I'm not bothered by that to much:

```
2025.02.11 20:56:20.914 INFO  [PluginMonitorExec-2]: org.jivesoftware.database.SchemaManager - Database update successful.
2025.02.11 20:56:20.914 INFO  [PluginMonitorExec-2]: org.jivesoftware.database.SchemaManager - Database update successful.
```

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/database/SchemaManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/database/SchemaManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2005-2008 Jive Software, 2016-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2005-2008 Jive Software, 2016-2025 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


akrherz:
Is there a scenario where this is possible?

guusdk:
Yes, quite easily. You can have a plugin that has an _install_ script that creates a database structure for, let's say, version 5, while update scripts exist for versions 1 through 10.

We typically don't do this (we always make sure that our install script immediately creates 'the latest version' of the database structure, but I've seen third-party plugins that don't: those only have an initial install script, followed by multiple upgrade scripts. For those plugins, the plugin currently needs to be reloaded for the entire database upgrade to complete (until that's done, the installation reports success, but isn't actually using the latest version, which can get very confusing). With this change, that is no longer needed.

Fishbowler:
Have you been able to test this?

Fishbowler (approved):

guusdk:
> Have you been able to test this?

Not exhaustive, but I tested with a regular plugin and a plugin that had its install script install less then the upgrade scripts.

PR #2680, opened 2025-02-12
OF-2952: Warn admins of certificate expiry

guusdk:
This commit adds a new feature that periodically checks if TLS certificates are expired, or are about to expire.

When such an expired is detected, an XMPP message is sent out to all admins (similar to the notifications sent out on available plugin updates).

A new admin console page has been added that can be used to configure this functionality.

image

### DIFF
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -1482,6 +1482,11 @@ system_property.xmpp.server.resolution-delay=The time to wait for a response for
 system_property.xmpp.server.resolution-timeout=The maximum amount of time to wait for successful resolution of a host of a target domain.
 system_property.xmpp.server.connection-max-workers=The maximum amount of worker threads attempting to set up a socket connection to a target remote XMPP domain. A value of '1' will effectively make 'Happy Eyeballs' impossible (as that requires concurrent connection attempts).
 
+system_property.ssl.certificates.expirycheck.service-enabled=Enables or disabled the period check for expiry of Openfire's TLS certificates.
+system_property.ssl.certificates.expirycheck.notify-admins=Determines if the administrators receive a message when an (almost) expired TLS certificate is detected.
+system_property.ssl.certificates.expirycheck.frequency=How often to check for (nearly) expired TLS certificates.
+system_property.ssl.certificates.expirycheck.warning-period=How long before a TLS certificate will expire a warning is to be sent out to admins.
+
 # Server properties Page
 
 server.properties.title=System Properties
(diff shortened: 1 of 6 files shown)


i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -1482,6 +1482,11 @@ system_property.xmpp.server.resolution-delay=The time to wait for a response for
 system_property.xmpp.server.resolution-timeout=The maximum amount of time to wait for successful resolution of a host of a target domain.
 system_property.xmpp.server.connection-max-workers=The maximum amount of worker threads attempting to set up a socket connection to a target remote XMPP domain. A value of '1' will effectively make 'Happy Eyeballs' impossible (as that requires concurrent connection attempts).
 
+system_property.ssl.certificates.expirycheck.service-enabled=Enables or disabled the period check for expiry of Openfire's TLS certificates.
+system_property.ssl.certificates.expirycheck.notify-admins=Determines if the administrators receive a message when an (almost) expired TLS certificate is detected.
+system_property.ssl.certificates.expirycheck.frequency=How often to check for (nearly) expired TLS certificates.
+system_property.ssl.certificates.expirycheck.warning-period=How long before a TLS certificate will expire a warning is to be sent out to admins.
```
Fishbowler, line 1488:
```suggestion
system_property.ssl.certificates.expirycheck.warning-period=How long (in hours) before a TLS certificate will expire for a warning is to be sent out to admins.
```

i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -2376,6 +2381,34 @@ ssl.certificates.trust-store=Trust store
 ssl.certificates.trust-store.s2s=Trust store used for connections from other servers
 ssl.certificates.trust-store.c2s=Trust store used for connections from clients
 
+# Expiry check
+sidebar.ssl.certificate.expirycheck=Expiry Check
```
Fishbowler, line 2385: No NL strings? 😲

Fishbowler (approved):
Once clarification change in the description of a system property, but looks good.
Would love to dogfood this with a nice high value to see that it works.

i18n/src/main/resources/openfire_i18n.properties
guusdk: I hoped you'd not notice. :)

guusdk:
> Would love to dogfood this with a nice high value to see that it works.

That'd be good to do. I did test this by having Openfire generate a self-signed certificate, adjusted to be valid for a negative amount of days.

PR #2682, opened 2025-02-17
OF-2967: Remove newlines in toString implementations

guusdk:
Where a stanza is used in a toString(), use to `toXML` rather than `toString` of the stanza, as the latter uses 'pretty print' (which introduces newlines).


akrherz:
Gemini review

dwd (approved):

PR #2683, opened 2025-02-19
update xpp3 from xpp3/xpp3/1.1.4c to org.codelibs/xpp3/1.1.4c.0

mihaimai:
The issue was described here:
https://discourse.igniterealtime.org/t/java-11-module-issue-xpp3-dependency-in-xmppserver-provides-duplicate-qname-class/95185

In short, xpp3/xpp3/1.1.4c contains javax.xml.namespace.QName class, which it should't
This started to cause issues since java11

### DIFF
--- a/xmppserver/pom.xml
+++ b/xmppserver/pom.xml
@@ -262,9 +262,9 @@
         </dependency>
 
         <dependency>
-            <groupId>xpp3</groupId>
+            <groupId>org.codelibs</groupId>
             <artifactId>xpp3</artifactId>
-            <version>1.1.4c</version>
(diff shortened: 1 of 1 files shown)


guusdk:
A concern is that the owner of that dependency is not well-known, and isn't very active. The dependency also doesn't seem to be used a lot, which doesn't do a whole lot for its reputation.

If we do need to change out the library, then perhaps it's better to create a fork ourselves.

guusdk:
I've created https://igniterealtime.atlassian.net/browse/OF-3071 to track this issue.

PR #2685, opened 2025-02-24
OF-2972: Replace (some) of the inline-script declarations

guusdk:
This PR resolves a couple of the subtasks of issue OF-2972 (isolated in individual commits).

These changes work towards being able to define stricter CSP directives by default (although a lot of additional changes are needed before that can happen).


guusdk:
gemini review

guusdk:
Gemini, what are your suggestions to improve readability and maintainability?

guusdk:
None of these are suggestions, are they? They seem to be observations.

dwd (approved):
I'm trusting that the extracted script files match, but I think that's a safe assumption.

PR #2687, opened 2025-02-26
OF-3026: Splitting up LocalRoutingTable

guusdk:
Instead of having one 'local' routing table to track routes from all types of connections (component, server, client) a distinct table for each of these types should be used.

This allows for more granular control of data access. It will become easier to move invocations under a guard of a lock form a corresponding cache.

In this commit, only minor functional changes are applied. This affects some exception handling (a more generic `Throwable` is caught when processing of a stanza fails) and the registration of a task that cleans up server sessions (this task is now cancelled when the RoutingTable is stopped).

Changes are applied to classes and methods that all have limited visibility outside of the RoutingTableImpl class (or its package). It's therefor unlikely that this change introduces compilation issues in other projects (such as Openfire plugins). The code changed here isn't expected to be re-used directly by other projects.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/LocalRoutingTable.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/LocalRoutingTable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2005-2008 Jive Software, 2017-2022 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2005-2008 Jive Software, 2017-2025 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


guusdk:
This change does not intent to _fix_ issue OF-3026. Instead, it refactors the code in a way that it becomes easier to apply a fix for that issue.

guusdk:
As this is the first PR in this project that uses the Gemini bot, some observations: its summary is ... correct, in the sense that it doesn't list things that aren't true. However, its full of things that are either a more long-winded version of the commit message, or things that are technically correct, but not relevant for reviewing purposes (it lists changes, but not the intention/motivation of the change). I'm worried that this will lead to people discussing the wrong thing, focusing on what is listed in the comment rather than the actual changes. It's more noise than value-adding content, in my opinion. I'm interested to see how this works out in the long(er) run, and what the actual reviewing function generates.

akrherz:
Gemini review

guusdk:
(force-pushed minor changes amended to the original commit)

guusdk:
Gemini review

dwd (approved):

PR #2688, opened 2025-02-26
OF-3026: State changes of LocalRoutingTable should happen under guard of a lock

guusdk:
This builds on the change in https://github.com/igniterealtime/Openfire/pull/2687 and introduces guards, when modifying state of LocalRoutingTable instances. The first commit in this PR is equal to that in #2687. The second commit in this PR is the relevant bit.

When modifications to a LocalRoutingTable is made, this should always happen under guard of a lock that also guards related changes to the corresponding cache.
    
This commit should reduce the possibility of state inconsistencies. It does so by re-using the Cache-sourced locks that were originally introduced by the changes for issue ENT-425.
    
In the past, we introduced deadlock potential by putting more operations under guard of a lock. From history, I don't think we ever attempted this change. Additionally, as the code that's now moved under guard of the pre-existing lock is not (intended to be) used by code outside of this package (the class has a corresponding access modified), this change should have minimal risks of introducing such a deadlock.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/spi/RoutingTableImpl.java
@@ -202,9 +202,16 @@ public class RoutingTableImpl extends BasicModule implements RoutingTable, Clust
 
     private String serverName;
     private XMPPServer server;
+
+    // Modifications should be made only after obtained a lock from #usersSessionsCache
     private final LocalRoutingTable<LocalClientSession> localClientRoutingTable = new LocalRoutingTable<>();
+
+    // Modifications should be made only after obtained a lock from #serversCache
     private final LocalRoutingTable<LocalOutgoingServerSession> localServerRoutingTable = new LocalRoutingTable<>();
+
+    // Modifications should be made only after obtained a lock from #componentsCache
     private final LocalRoutingTable<RoutableChannelHandler> localComponentRoutingTable = new LocalRoutingTable<>(); // TODO see if the type can be made more specific.
+
     private RemotePacketRouter remotePacketRouter;
     private IQRouter iqRouter;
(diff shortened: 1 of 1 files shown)


guusdk:
Gemini review only the second commit in this PR please.

guusdk:
I've now merged #2687 and rebased this PR on top of that.

guusdk:
Gemini review

guusdk:
From the last two Gemini reviews, the 'code duplication' issue seems false. I've addressed the second issue, although it won't affect functionality.

I disagree with the Clarity of Lock usage: the existing comments explicitly state that the scope for the locks is a modification of the data structure.

The last two suggestions seem not important. The (very small) risk of applying the suggested change introducing unexpected side effects doesn't outweigh the (smaller) benefit that the changes would bring.

guusdk:
In the interest of reviewing the application of Gemini in our PRs, I'm leaving some (overly) details feedback on its operation in this PR.

I'm not sure how to weight the "I understand" comment from an AI. I doubt if it actually can evaluate my reasoning, and agree or disagree. Without that, a comment like "I understand" seems meaningless. Is this just window dressing?

The "I'm okay with moving forward" comment bugs me. It implies that Gemini is a gatekeeper. Although I know it's not, a comment like this may easily be misunderstood by third parties that it is. As a result, I fear that people will be putting in pointless hours pleasing an AI ("must avoid computer-says-no scenario!), while today, Gemini certainly has proven to be fallible.

I've not taken actions to please Gemini. Instead, I kept a record for other humans to reflect on. In that sense, the "I appreciate you taking the time..." has a strong uncanny valley characteristic. Why not say something along the lines of "Thanks for explaining your decisions"?

guusdk:
So it said that it agrees that "I appreciate you taking the time" is best replaced by something else, yet it started that agreement with that exact same sentence. :)

dwd (approved):

PR #2695, opened 2025-02-27
OF-3028: Netty threads from 'child' EventLoop should use Netty-default settings

guusdk:
As we're overriding the thread factories used for Netty EventLoops to change the names of threads, we've also changed the default configuration of threads used by Netty.

Netty's default configuration can be expected to be optimized for performance. Openfire should use a similar configuration.

In this commit, the configuration is reverted back to the default configuration that's used by Netty (based on `io.netty.util.concurrent.DefaultThreadFactory`).


guusdk:
@gemini-code-assist what aspects that could be improved for clarity and maintainability are you referring to?

akrherz (approved):

PR #2696, opened 2025-02-27
OF-2811: Must not block on eventloop

guusdk:
OF-2808 describes an issue in which a session that is resumed gets closed. This is caused by the asynchronous 'close' operation to execute 'connection close' listeners on a session that, in another thread, has already been resumed.

The fix for OF-2808 introduces a Latch, that causes a thread to block. If that thread is one of a Netty EventLoop, all processing stops (possibly leading to the latch never be released, as the connection close listeners cannot operate).

This commit reverts the change for OF-2808, and applies a new fix for that issue: disallow a session that's already in the process of being closed from resumed.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
@@ -350,6 +350,8 @@ public class StreamManager {
         // OF-2811: Cannot resume a session that's already closed. That session is likely busy firing its 'closeListeners'.
         if (route.isClosed()) {
             Log.debug("Not allowing a client of '{}' to resume a session, as the preexisting session is already in process of being closed.", fullJid);
+            sendError(new PacketError(PacketError.Condition.unexpected_request));
+            return;
         }
 
         // Previd identifies proper session. Now check SM status
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2024 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2017-2025 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


guusdk:
Converting this to 'draft', as I've shot off this PR without even compiling, let alone testing this fix. Time constraints prevented me from doing so immediately. I did want to solicit feedback, which is why I raised the PR anyway.

@dwd as you are the author of Stream Management, can you please review the approach taken in this PR?

xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
```diff
@@ -346,6 +346,12 @@ private void startResume(String namespace, String previd, long h) {
             return;
         }
         Log.debug("Found existing session for '{}', checking status", fullJid);
+
+        // OF-2811: Cannot resume a session that's already closed. That session is likely busy firing its 'closeListeners'.
+        if (route.isClosed()) {
+            Log.debug("Not allowing a client of '{}' to resume a session, as the preexisting session is already in process of being closed.", fullJid);
```
dwd, line 352: Did you want to send an error and return here?

dwd (changes requested):

dwd:
Oh, you meant the XEP, right.

xmppserver/src/main/java/org/jivesoftware/openfire/streammanagement/StreamManager.java
guusdk: :facepalm:

guusdk:
I've thrown a bit of a load test against this. That occasionally gives me log warnings like this:

> 2025.02.27 20:20:42.980 WARN  [ForkJoinPool.commonPool-worker-8]: org.jivesoftware.openfire.spi.RoutingTableImpl - Client route not found for route test3@example.org, while user session still exists, Current content of users cache: {test3@example.org/res15=ClientRoute{nodeID=71ccbc28-1be9-44ab-9cfe-98c7a8ba626c, available=true}, test10@example.org/res17=ClientRoute{nodeID=71ccbc28-1be9-44ab-9cfe-98c7a8ba626c, available=true}, test1@example.org/res10=ClientRoute{nodeID=71ccbc28-1be9-44ab-9cfe-98c7a8ba626c, available=true}}. Current content of anonymous users cache: {}

guusdk:
I believe that those warnings that I found in the log are false positives. I've addressed this in OF-3029 / https://github.com/igniterealtime/Openfire/pull/2697

guusdk:
I've now marked this PR as 'ready for review', given that I've now had some time to test it myself.

PR #2699, opened 2025-03-01
OF-3031: Session's 'close' state no longer based on its backing connection

guusdk:
In the early days, a Session was always backed by a (socket) connection. When the connection was closed, so was the session, and vice versa.

With Stream Management, this no longer is the case. A session can remain unclosed (detached) while its connection is closed.

This commit changes the behavior of `LocalSession#isClosed()` to be based on the session's `status` field, and no longer on the connection's connectivity state.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004-2009 Jive Software, 2017-2024 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2004-2009 Jive Software, 2017-2025 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -500,13 +500,6 @@ public abstract class LocalSession implements Session {
         return Optional.ofNullable(conn)
                 .map(Connection::getPeerCertificates)
                 .orElse(new Certificate[0]);
-    }
-
-    @Override
-    public boolean isClosed() {
-        return Optional.ofNullable(conn)
-                .map(Connection::isClosed)
-                .orElse(Boolean.TRUE);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


guusdk:
Here be dragons. I fear that, although probably 'more correct', this change may upset a delicate balance of how sessions are being disconnected/reconnected.

akrherz (approved):

Fishbowler (approved):
This looks like a rational change, given SM, but understand your concerns about reach.

I think you're only gonna get this through dogfooding.

Minor nit: Should this update the javadocs in Session.java to remove the connection references there?

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
```diff
@@ -502,11 +503,16 @@ public Certificate[] getPeerCertificates() {
...
+    // TODO: Remove this override. The override (and the boolean property) serves as a emergency fallback for the change in OF-3031 and _should not_ be needed. It should be desirable to use #getStatus() only.
     @Override
     public boolean isClosed() {
-        return Optional.ofNullable(conn)
+        if (JiveGlobals.getBooleanProperty("xmpp.session.isclose.connectionbased", false)) {
+            return Optional.ofNullable(conn)
                 .map(Connection::isClosed)
                 .orElse(Boolean.TRUE);
+        } else {
+            return getStatus() == Status.CLOSED;
```
dwd, line 514: super.isClosed(), perhaps?

dwd (approved):
I agree this is correct and will need manual testing and deployment experience.

My concerns were that there might be code relying on the old behaviour in the '198 resume case, but that doesn't appear to be the case.

However, SessionManager::isActiveRoute does seem to rely on isClosed to be false in order to then call session.validate(), which "verifies the connection is still live". Can you check this one before merging?

The other callers of isClosed() all seem to be OK - in fact in some cases they look distinctly improved by this change. (For example, PEPService line 342 in sendNotification).

xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalSession.java
guusdk: I don't think that works with default methods on interfaces.

guusdk:
> Should this update the javadocs in Session.java to remove the connection references there?

I've now pushed a commit that updates the Javadoc to better distinguish between 'session' and 'connection'.

guusdk:
> However, SessionManager::isActiveRoute does seem to rely on isClosed to be false in order to then call session.validate(), which "verifies the connection is still live". Can you check this one before merging?

For context: nothing in Openfire itself calls `SessionManager::isActiveRoute` (but plugins may).

It's hard to reason about this. This arguably is an improvement? If the _session_ is closed, then doing a validate/keepalive on the connection is undesirable (as the session's state has already been marked as being closed, which is not recoverable afaik). Hand-wavy: this method could now (but currently doesn't) check if _any_ underlying connection mechanism is still viable.

PR #2702, opened 2025-03-02
OF-3036: Improve log message when remote domain is unreachable

guusdk:
This drops the stack trace, and changes the log message in such a way that it's more descriptive to an admin.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2005-2008 Jive Software, 2016-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2005-2008 Jive Software, 2016-2025 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalOutgoingServerSession.java
```diff
@@ -261,6 +262,11 @@ static LocalOutgoingServerSession createOutgoingSession(@Nonnull final DomainPai
             // Wait for the future to give us a session...
             // Set a read timeout so that we don't keep waiting forever
             return (LocalOutgoingServerSession) sessionInitialiser.init(listener).get(INITIALISE_TIMEOUT_SECONDS.getValue().getSeconds(), TimeUnit.SECONDS);
+        } catch (NetworkEntityUnreachableException e) {
+            Log.warn("Cannot connect to XMPP domain '{}' (from '{}'): Unable to create a socket connection to a host that associated " +
+                "to the domain. This is typically caused by an outage (the remote server may be turned off) or a network configuration " +
+                "issue (eg: missing or invalid DNS records, restrictive firewalls, etc.).", domainPair.getRemote(), domainPair.getLocal());
```
akrherz, line 268: Hmmm, if the problem is actually a "missing or invalid DNS records", then that should be reported as the problem.  Why is this being speculated?  From Openfire's side, it is or it isn't, right?

guusdk: At this point in the code, we lost the exact cause of the problem. Also, I expect that there's a lot of grey area here. What if there _are_ DNS records, but not one for the actual server that's used? What if when there _is_ an A or AAAA record, but not a SRV record. If connection is failing, it's hard to deduce if that SRV record would've made a difference.

guusdk:
I've raised https://igniterealtime.atlassian.net/browse/OF-3037 for more improvements around S2S unreachability issues.

PR #2716, opened 2025-03-12
CI: add DB2 support to database update test

guusdk:
Based on a Docker image that's documented in https://www.ibm.com/docs/en/db2/12.1?topic=system-linux

This is unlikely to be functional, as I don't think it is properly loading the 'old' database scripts. Creating a PR in the hope that someone smarter than me can fix this.

I suspect that this won't even run, as Openfire might not ship with a DB2 JDBC driver (see https://www.ibm.com/docs/en/db2-big-sql/6.0?topic=drivers-jdbc-driver )


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -450,6 +450,89 @@ jobs:
...
+      - name: Start database server with an empty database
+        run: |
+          mkdir olddb
+          echo 'CREATE TABLE imready AS (SELECT 1);' > $GITHUB_WORKSPACE/olddb/ZZZ-signal-ready.sql
+          docker compose -f ./build/ci/compose/db2.yml up --detach
+      - name: Execute Openfire's database install script
+        run: |
+          pushd ./build/ci/updater
+          ./mvnw package
+          java -jar ./target/updaterunner-1.0.0-jar-with-dependencies.jar
```
Fishbowler, line 490: The -with-dependencies suffix is failing the build right now

guusdk: Ah, that was replaced by the Maven exec plugin for other databases. I forgot to change that for the DB2 branch, which was developed in parallel with the changes for other databases.

I've applied the same change to DB2 now.

guusdk:
I suspect that at least two problems still exist:
- The DB2 docker image takes many minutes to become available, yet this test fails in approximately 90 seconds. This may suggest that the docker health check isn't working as intended
- I suspect that the provisioning of the DB2 database (with the `isready` table and/or the older Openfire database structure) isn't happening.

PR #2720, opened 2025-03-15
OF-2786: Openfire to use system-defined preferences for IPv6 over IPv4

stokito:
Split from https://github.com/igniterealtime/Openfire/pull/2563 to simplify of review.

This change is the openfire.init.d is same as was made in 30b4a176 for openfire.sh

> When establishing an outbound connection, Openfire now prefers the IP family as defined by the order in which the operating system returns addresses.

The Debian init script makes almost the same as the `openfire.sh` so they should be synced until the SystemD migration.

CC @guusdk


Fishbowler (approved):
Not tested anything, but ... Yep. This matches :)

guusdk:
Thanks!

PR #2721, opened 2025-03-15
OF-2542 fix: admin console setup page: check JRE version

stokito:
Split from #2563 to simplify of review.

Fixed a bug that I had on `23-ea` OpenJDK version.
Use [Runtime.version()](https://docs.oracle.com/javase/9/docs/api/java/lang/Runtime.Version.html) to compare JRE version. It works correctly now.

### DIFF
--- a/xmppserver/src/main/webapp/setup/index.jsp
+++ b/xmppserver/src/main/webapp/setup/index.jsp
@@ -43,12 +43,13 @@
     boolean openfireHomeExists = false;
     Path openfireHome = null;
 
-    // Check for JRE 1.8
+    // Check for min JRE requirement
+    int MIN_JAVA_VERSION = 11;
     try {
         String version = System.getProperty("java.version");
-        jreVersionCompatible = Integer.parseInt(version.split("\\.")[0]) >= 11;
+        jreVersionCompatible = Integer.parseInt(version.split("[-.]+")[0]) >= MIN_JAVA_VERSION;
     }
-    catch (Throwable t) {}
+    catch (Exception ignored) {}
     // Check for Servlet 2.3:
     try {
         Class c = ClassUtils.forName("javax.servlet.http.HttpSession");


xmppserver/src/main/webapp/setup/index.jsp
```diff
@@ -43,12 +43,9 @@
...
 
-    // Check for JRE 1.8
-    try {
-        String version = System.getProperty("java.version");
-        jreVersionCompatible = Integer.parseInt(version.split("\\.")[0]) >= 11;
-    }
-    catch (Throwable t) {}
+    // Check for min JRE requirement
+    Runtime.Version MIN_JAVA_VERSION = Runtime.Version.parse("11");
+    jreVersionCompatible = MIN_JAVA_VERSION.compareTo(Runtime.version()) <= 0;
```
Fishbowler, line 48: When a user runs Java 8, this will fail, since the Runtime API was introduced in Java 10.

I suspect other stuff will also fail in Java 8 too though... @guusdk ?

Fishbowler:
By that measure, why have the check at all?

Fishbowler:
So we should make sure it errors gracefully on Java 8?

guusdk:
That would be interesting: can we get to this page using JRE8? If so, then I believe that we should generate a 'proper' error message, rather than having an UnsupportedClassVersionError or something similar. As Openfire has been using Java 8 for years, I do suspect that a lot of people will still use Java 8 when installing Openfire. Let's give them a clear error condition, rather than a weird stack trace.

What was the original problem with `23-ea`? Maybe we can resolve that in a way that doesn't involve using API not available in Java 8?

akrherz:
This all seems very puzzling to me.

1. I would be surprised that you could to get to the referenced code when invoked from Java 8
2. I sort of agree with Fishbowler and wonder why this check exists at all?
3. Moreover, this check needs to reference something more global in scope than a magic number buried in a JSP

PR #2723, opened 2025-03-17
OF-2607: Add support for XEP-0421: Occupant identifiers for semi-anonymous MUCs

guusdk:
This commit adds an occupant-id property to MUCOccupant, ensures that each MUCOccupant instance receives an approprate occupant-id value, and ensures that this value is included in the appropriate mesasge and presence stanzas that are sent from the MUC room.

This PR uses that approach suggested by @dwd in https://github.com/igniterealtime/Openfire/pull/2714


dwd (approved):

Fishbowler:
Do your new interop tests pass against this implementation?

guusdk:
> Do your new interop tests pass against this implementation?

Yes, they do.

PR #2726, opened 2025-03-21
OF-3045: SchemaManager explict check for success

guusdk:
After SchemaManager runs database upgrade scripts, it should not assume success, but explicitly check for it.


akrherz (approved):

Fishbowler:
Definitely something up with this: https://github.com/igniterealtime/Openfire/actions/runs/14007796593/job/39223852281?pr=2726#step:4:440

Fishbowler:
With the false positive resolved, I think I've made database upgrades now _actually_ work :tada:

PR #2750, opened 2025-03-24
OF-3044: When resource binding, ensure that old detached sessions are removed

guusdk:
After resource binding succeeds, clean-up of older, detached sessions will be problematic (as that needs to update state, but typically also sends out presence unavailable on behalf of the user).

In this commit, any detached sessions matching the full JID that will be the result of a successful resource binding is terminated, prior to resource binding being allowed to complete.


guusdk:
I think this wasn't an issue before, until the fix for https://igniterealtime.atlassian.net/browse/OF-3039 was applied.

PR #2753, opened 2025-03-28
OF-2784: Fix admin console pages

guusdk:
The previous commit for this issue introduced nested HTML forms, which won't work. Additionally, the form shouldn't differ between null values, and empty string values for provided settings.


deleolajide (approved):
Looks good to me. Going to merge and test immediately

PR #2759, opened 2025-04-14
OF-2607: Remove _all_ client-provided `occupant-id` values

guusdk:
To prevent spoofing, not one, but _all_ provided `occupant-id` values should be removed. Thanks Travis!

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCOccupant.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCOccupant.java
@@ -40,6 +40,7 @@ import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


akrherz (approved):

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCOccupant.java
```diff
@@ -40,6 +40,7 @@
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
```
moparisthebest, line 46: What about the other 2 remove()'s in this file?

```
newPresence.getElement().remove(element);
// and
            if (mucUser != null) {
                // Remove any previous extendedInformation, then re-add it.
                presence.getElement().remove(mucUser);
            }
```
Seems like maybe those should remove all and not just 1 too ?

moparisthebest: also the same in `MultiUserChatServiceImpl.java`

guusdk: Those should be tackled too, I agree. I've created a separate PR for those: https://github.com/igniterealtime/Openfire/pull/2761

PR #2760, opened 2025-04-14
DRAFT: Get client sessions irrespectively of their presence state

guusdk:
`org.jivesoftware.openfire.SessionManager#getSessions(java.lang.String)` returned sessions only when they are 'available'. This is a (side?)effect of its implementation depending on `org.jivesoftware.openfire.RoutingTable#getRoutes`

I don't know if there are cases where it is desirable to get only 'available' sessions.

I do know that being able to retrieve only 'available' sessions introduces issues. For one, it prevents most code to evaluate non-available session. For example, code can fail to terminate a non-available session (or a session where the availability state hasn't been updated yet due to a race condition).

This commit introduces a new implementation that allows ClientSessions to be obtained irrespectively of their availability status.

I'm not at all certain that this is the right approach. I am offering this up as an illustration, for discussion purposes.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/RoutingTable.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/RoutingTable.java
@@ -23,6 +23,7 @@ import org.xmpp.packet.Packet;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 
 /**
  * <p>Maintains server-wide knowledge of routes to any node.</p>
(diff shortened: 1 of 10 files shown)


GregDThomas (commented):
Code changes are fine; but sorry no longer feel capable of providing any useful feedback on the value of "all" vs "available" sessions!

akrherz:
rebased to see what the rabbit thinks

akrherz:
@coderabbitai review

PR #2766, opened 2025-04-17
OF-3059 Remove anonymous route

guusdk:
The commits in this PR remove:
- the concept of having a _route_ that is anonymous. Instead, a _session_ can be anonymous
- the clustered cache that stores anonymous sessions (apart from 'regular' session). Instead, all sessions are stored in one cache

Please refer to the individual commit messages for more details.


akrherz:
I restarted the `aioxmpp` CI tests hoping the failure was transient, but it failed a second time

guusdk:
I did the same for the Smack test, that also consistently fails. There's a bug! There's a bug! Let's fire the author!

PR #2768, opened 2025-04-17
OF-3059: Remove 'anonymous route' concept

guusdk:
Openfire used the definition of an ‘anonymous route’ in various places (notably the routing table and session manager).

An anonymous route should be defined as “a route-able session of which the users has an anonymous authentication token.”

A ‘route’ is not inherently anonymous or non-anonymous, a session is.

Implementation that previously used the 'anonymous route' definition is interested only in the fact if the corresponding session is anonymous, not if the session is route-able.

In this commit, the implementations of methods named `isAnonymousRoute` are deprecated. Their implementation is replaced by checks that evaluate the authentication mechanism used in a session.

### DIFF
--- a/.github/actions/startserver-action/action.yml
+++ b/.github/actions/startserver-action/action.yml
@@ -13,6 +13,10 @@ inputs:
     description: 'Set a hosts file for the given IP and host (or for example.com if running locally)'
     required: false
     default: '127.0.0.1'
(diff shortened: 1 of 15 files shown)


guusdk:
This debugging branch has served its purpose. Closing this in favor of #2766

PR #2769, opened 2025-04-17
CI: Allow Openfire to log at different levels

guusdk:
This modifies the GitHub action that's used to start Openfire to take an optional log level. When set, the log4j configuration used by Openfire is modified accordingly.

The continuous integration jobs that use this action are now logging on level 'debug'. That should provide more analytics, in case a step fails.

### DIFF
--- a/.github/actions/startserver-action/startCIServer.sh
+++ b/.github/actions/startserver-action/startCIServer.sh
@@ -6,14 +6,15 @@ HOST='example.org'
 
 usage()
 {
-	echo "Usage: $0 [-i IPADDRESS] [-h HOST] [-b BASEDIR]"
+	echo "Usage: $0 [-i IPADDRESS] [-h HOST] [-b BASEDIR] [-l LOGLEVEL]"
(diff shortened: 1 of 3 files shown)


Fishbowler (approved):

.github/actions/startserver-action/startCIServer.sh
```diff
@@ -56,6 +60,11 @@ function launchOpenfire {
 	# Replace the default XMPP domain name ('example.org') that's in the demoboot config with the configured domain name.
 	sed -i -e 's/example.org/'"${HOST}"'/g' ${BASEDIR}/conf/openfire.xml
 
+	# Replace the default log level ('info') that's in the log4j config with the configured level.
+	if [[ -n "${LOGLEVEL}" ]]; then
+	    sed -i -e 's/info/'"${LOGLEVEL}"'/g' ${BASEDIR}/lib/log4j2.xml
```
akrherz, line 65: This feels rather brittle to me and the `/g` should definitely not be included.

guusdk: How about now?

PR #2789, opened 2025-05-24
Install4j update

guusdk:
- Now requires Install4j 11 (this appears to prevent an issue with running Openfire with Java 21 on a Mac, as tested by Dele)
- Mininimum version of Java set to 17
- Now generates a MacOS artifact
- All old images replaced various resolutions of the Openfire logo
- Updated the startup VM parameters to use the same values as used by the non-Install4j startup scripts

Removed the old Mac configuration pane project


guusdk:
I have deleted that workflow.

We'll probably need Bamboo changes when this gets merged;
- upgrade of install4j that's available in that environment
- publication of additional artifact to website

Anything else?

PR #2794, opened 2025-05-27
OF-3076: Refactor XML parsing in Launcher for unit test

guusdk:
A small refactoring is applied to the Launcher module, so that its XML parsing can be subject of a unit test.

Two small unit tests are added that assert that no XXE vulnerability is presenct in the XML parsing of the Launcher.

### DIFF
--- a/starter/pom.xml
+++ b/starter/pom.xml
@@ -61,6 +61,13 @@
             <version>10.0.3</version>
             <scope>provided</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.9.3</version>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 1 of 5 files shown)


guusdk:
Note, there aren't any functional changes in this PR yet. I first wanted to see if I could reproduce the issue in a unit test.

guusdk:
It turns out that the bug report was incorrect: Openfire is not susceptible to that particular issue (the preexisting guards put in place by @Alameyo are sufficient). This has been confirmed by the original reporter.

It is still a good thing to have an additional unit test that verifies this behavior, which is now provided in this PR. Let's merge this, but otherwise close the issue as 'not-a-bug'.

guusdk:
I honestly don't know for sure, as I'm not quite sure how that doctype declaration works exactly (I might have stolen this unit test from somewhere else). ROS noted that they were happy with this.

PR #2797, opened 2025-06-04
OF-3079: Bump Jetty from 12.0.21 to 12.0.22 (restore plugin compat)

guusdk:
This should fix a stacktrace that's shown when opening admin console pages of plugins that are compiled against older versions of Openfire (this issue appears only in unreleased versions of Openfire).

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -27,7 +27,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
-                <version>3.1.1</version>
+                <version>3.7.1</version>
                 <configuration>
                     <descriptors>
                         <descriptor>src/assembly/basic-distribution.xml</descriptor>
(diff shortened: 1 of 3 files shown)


guusdk:
The CI build failure is consistently happening when applying this upgrade, although I've not been able to reproduce it on my local laptop yet.

joakime:
I don't see any org/eclipse/jetty dependencies being downloaded in the logs of your build.

How are you using Jetty in your build?

guusdk:
I do see the downloads in the logs. I'm not sure if you can see the full build logs, as a non-project member. I'm attaching the logs for the (failing) Maven build step to this comment. If you grep for 'jetty' in those, you'll find them all over.

We're using Jetty in two ways:
- As a pretty standard library, to, in code, instantiate an embedded web server
- To precompile JSP pages

I believe that the build failure happens later. Our project has a 'distribution' module that generates an archive that contains our installable software (with start scripts, and libraries). I think it's this step that fails. That's odd, as it doesn't explicitly use Jetty as all, as far as I know.

[4_Build with Maven.txt](https://github.com/user-attachments/files/20596758/4_Build.with.Maven.txt)

joakime:
Looks like your bump of the maven-assembly-plugin did the trick.

PR #2808, opened 2025-06-11
OF-3084 OF-3085: Remove deprecated Base64 implementation

guusdk:
Java now provides a Base64 implementation in `java.util.Base64`. Use that (or a third-party implementation) instead of the implementation that is removed in this commit.


akrherz (approved):
:champagne:

PR #2810, opened 2025-06-11
OF-3084 OF-3087: Remove deprecated `PluginManager#getPlugin()`

guusdk:
The method removed by this commit depends on a name which is not under control of the plugin's author: the file name of the JAR that is used to deploy the plugin. The `getPluginByName` method provides more predictable results.

In some instances, the filename-based lookup is still required (for example, the Admin Console, which is loaded as a plugin, does not have a 'name'). For those instances, a new method is introduces (`getPluginByCanonicalName`) that is equivalent to the removed method (this effectively is a rename of that method, wrapping the result in an `Optional`). Now that the old method name is no longer available, developers are forced to explicitly choose between the two available alternatives. Generally speaking, the `getPluginByName` method should be used.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/cache/CacheFactory.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/cache/CacheFactory.java
@@ -859,13 +859,10 @@ public class CacheFactory {
 
     private static ClassLoader getClusteredCacheStrategyClassLoader() {
         PluginManager pluginManager = XMPPServer.getInstance().getPluginManager();
-        Plugin plugin = pluginManager.getPlugin("hazelcast");
-        if (plugin == null) {
-            plugin = pluginManager.getPlugin("clustering");
-            if (plugin == null) {
-                plugin = pluginManager.getPlugin("enterprise");
-            }
-        }
+        Plugin plugin = pluginManager.getPluginByCanonicalName("hazelcast")
+            .or(() -> pluginManager.getPluginByCanonicalName("clustering"))
+            .or(() -> pluginManager.getPluginByCanonicalName("enterprise"))
+            .orElse(null);
         PluginClassLoader pluginLoader = pluginManager.getPluginClassloader(plugin);
         if (pluginLoader != null) {
             if (log.isDebugEnabled()) {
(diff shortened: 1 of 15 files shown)


xmppserver/src/main/java/org/jivesoftware/util/cache/CacheFactory.java
```diff
@@ -859,13 +859,10 @@ public static synchronized void initialize() throws InitializationException {
...
-        Plugin plugin = pluginManager.getPlugin("hazelcast");
-        if (plugin == null) {
-            plugin = pluginManager.getPlugin("clustering");
-            if (plugin == null) {
-                plugin = pluginManager.getPlugin("enterprise");
-            }
-        }
+        Plugin plugin = pluginManager.getPluginByCanonicalName("hazelcast")
+            .or(() -> pluginManager.getPluginByCanonicalName("clustering"))
+            .or(() -> pluginManager.getPluginByCanonicalName("enterprise"))
```
GregDThomas, line 864: I dob't think either the clustering or the enterprise plugins work any more - they've long since been deprecated. Perhaps remove those from the list?

xmppserver/src/main/webapp/security-truststore.jsp
```diff
@@ -97,7 +97,8 @@
                     // When updating certificates through the admin console, do not cause changes to restart the website, as
                     // that is very likely to log out the administrator that is performing the changes. As the keystore change
                     // event is async, this line disables restarting the plugin for a few minutes.
-                    ((AdminConsolePlugin) XMPPServer.getInstance().getPluginManager().getPlugin("admin")).pauseAutoRestartEnabled(Duration.ofMinutes(5));
+                    XMPPServer.getInstance().getPluginManager().getPluginByCanonicalName("admin")
```
GregDThomas, line 100: I've seen the pattern

`((AdminConsolePlugin) XMPPServer.getInstance().getPluginManager().getPlugin("admin"))`

all over the place. Given you've got to (?) have an admin plugin, is there a case fpr

`public AdminConsolePlugin getAdminConsolePlugin() { ... }`

in the `XMPPServer` class itself?

GregDThomas (commented):
Two comments you might feel helpful. Or not :)

xmppserver/src/main/webapp/security-truststore.jsp
guusdk: ... maybe? I'm not sure if there's much added value over this approach though. In any case, I don't think we can remove the `getPluginByCanonicalName` method (as we'd probably need that for edge cases anyway).

PR #2811, opened 2025-06-11
OF-3084 OF-3088: Refactor Cache Locking implementation

guusdk:
Based on the single responsibility principle, the CacheFactory should not be managing Locks. Instead, Caches should do that themselves.

Back in 2019, the method in CacheFactory that generates locks was marked as deprecated because of that, with a removal note in the distant future: 5.0.0.

As 5.0.0 is upon us, the deprecated code has now been removed. The cache locking implementation has been moved to a dedicated class (`LocalCache`) that is used by caches directly.

This commit attempts to achieve the desired removal of deprecated methods with a minimum of refactoring.

Further improvements may be desired (e.g. not storing locks for all caches in one shared data structure, but rather in a per-cache data structure).

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/cache/lock/CacheKey.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/cache/lock/CacheKey.java
@@ -1,0 +1,35 @@
+package org.jivesoftware.util.cache.lock;
+
+import org.jivesoftware.util.cache.Cache;
+
+import java.util.Objects;
+
+/**
+ * A key of a cache, namespaced by the cache that it belongs to.
+ */
+class CacheKey
+{
+    final String cacheName;
+    final Object key;
+
+    CacheKey(Cache cache, Object key)
+    {
+        this.cacheName = cache.getName();
+        this.key = key;
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        CacheKey cacheKey = (CacheKey) o;
+        return cacheName.equals(cacheKey.cacheName) && key.equals(cacheKey.key);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(cacheName, key);
+    }
+}
--- a/xmppserver/src/main/java/org/jivesoftware/util/cache/lock/LocalLock.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/cache/lock/LocalLock.java
@@ -1,0 +1,122 @@
+package org.jivesoftware.util.cache.lock;
+
+import com.google.common.collect.Interner;
+import com.google.common.collect.Interners;
+import org.jivesoftware.util.cache.Cache;
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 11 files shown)


guusdk:
I've prepared a corresponding change in the Hazelcast plugin here: https://github.com/igniterealtime/openfire-hazelcast-plugin/pull/115

guusdk:
As I'd like to dogfood this test, I'm merging this without waiting for a review. Don't let that stop anyone from reviewing after-the-fact. :)

xmppserver/src/main/java/org/jivesoftware/util/cache/lock/LocalLock.java
```diff
@@ -0,0 +1,122 @@
...
+        lock.lock();
+    }
+
+    private void releaseLock(CacheKey key) {
+        ReentrantLock lock = lookupLockForRelease(key);
+        lock.unlock();
+    }
+
+    private ReentrantLock lookupLockForAcquire(CacheKey cacheKey) {
+        CacheKey mutex = interner.intern(cacheKey); // Ensure that the mutex used in the next line is the same for objects that are equal.
```
GregDThomas, line 88: I might be tempted to put this behind a `getRepresentativeInstance(cacheKey)` method, or similar, as you've done it twice already. Possibly as a method of CacheKey.

xmppserver/src/main/java/org/jivesoftware/util/cache/lock/CacheKey.java
```diff
@@ -0,0 +1,35 @@
...
+import org.jivesoftware.util.cache.Cache;
+
+import java.util.Objects;
+
+/**
+ * A key of a cache, namespaced by the cache that it belongs to.
+ */
+class CacheKey
+{
+    final String cacheName;
```
GregDThomas, line 12: `private` ?

guusdk: That'd probably lead to introducing a getter and setter for that field As this is package protected, having direct access to the field is an acceptable compromise maybe? It's slightly less code (for something that I don't expect to be used much anyway).

xmppserver/src/main/java/org/jivesoftware/util/cache/lock/LocalLock.java
guusdk: I've read this three times but I still don't get what you mean :)

GregDThomas: Sorry - let me be more explicit. 
I think this logic belongs in `CacheKey` and not `LocalLock`. So .. add a method to `CacheKey`
```java
public CacheKey getRepresentativeInstance() {
  return interner.intern(this);
}
```
and invoke that from here (and the other place). 

Taking that further, I might also be tempted to make the constructor private, and add a factory method to CacheKey ...
```java
public static CacheKey createCacheKey(Cache cache, Object key) {
  final CacheKey instance = new CacheKey(cache, key);
  return  interner.intern(CacheKey);
}
```
So it's impossible to obtain anything other than the representative instance of the class.

guusdk: Ah, now I see. I've opened a new PR for this: https://github.com/igniterealtime/Openfire/pull/2821

PR #2816, opened 2025-06-11
Update Install4j Maven artifacts

guusdk:
Switch to version 11.0.4, which now seems to be available on Maven Central.


akrherz (approved):

guusdk:
I have verified that locally, issuing:

```
mvn -U clean install -Dmaven.test.skip -Pinstall4j
```

caused installers to be generated in `distribution/target/media` as expected.

PR #2813, opened 2025-06-11
Denote Openfire 5.0.0 Beta Release

akrherz:

PR #2817, opened 2025-06-11
Update Maven plugin versions

guusdk:
This updates versions of the Maven plugins used by this project to the latest version that is available for them (without requiring Maven 4).

A plugin that was added for support of MINA OSGI was dropped (as we're no longer using MINA)

The enforcer plugin's configuration has been updated to require Maven 3.6.3 (which is the same version as the shipped Maven wrapper `mvnw` is).

All of these suggestions where found by running `./mvnw versions:display-plugin-updates`


akrherz:
cherry picked onto #2813

PR #2818, opened 2025-06-13
OF-3089: Consistently use the same property when configuring S2S idle time

guusdk:
This consistently uses the same property name when server-to-server maximum idle time is being configured. The property name used is the one that was the oldest one in use (`xmpp.server.session.idle` rather than `xmpp.server.idle`).

~To-do: migrate properties that are now no longer used.~

Ideally, I'd like to use the `java.time` API for this implementation, but given the timing of this fix (just prior to the 5.0.0 release), I'd like to prevent API changes).


akrherz:
@guusdk a 4.9 backport candidate?

akrherz (approved):

PR #2821, opened 2025-06-19
OF-3084 OF-3088: Refactor Cache Locking implementation

guusdk:
Apply review feedback from Greg.


Fishbowler (approved):
This looks sufficiently equivalent to me

PR #2880, opened 2025-07-04
OF-2034: Add support for XEP-0398: User Avatar to vCard-Based Avatars Conversion

guusdk:
This adds basic support for XEP-0398, that:
- converts vCard avatars into PEP avatars
- converts PEP avatars into vCard avatars
- adds an avatar hash to presence available stanzas sent by users (per section 4)

Additionally, the code checks if vCard avatars are to be migrated when a vCard is loaded from a vCard provider. This aims to pick up changes that are applied to external data stores (such as LDAP / AD).

The existing code does not _remove_ avatars, when a removal is initiated in one of both systems.

### DIFF
--- a/documentation/openfire.doap
+++ b/documentation/openfire.doap
@@ -422,6 +422,13 @@
         </implements>
         <implements>
             <xmpp:SupportedXep>
+                <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0398.html"/>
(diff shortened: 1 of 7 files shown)


guusdk:
This PR depends on changes in #2877 and #2878. I expect this PR to fail, until it is rebased onto a branch that contains changes from those other PRs.

PR #2885, opened 2025-07-07
OF-3102 Improve error handling in the embedded db

stokito:
While making a Debian package I made an error when the /var/lib/openfire/embedded-db wasn't created during setup. And the Embedded DB folder  /usr/share/openfire/embedded-db is a link to the non esisting /var/lib/openfire/embedded-db.

I fixed it but before that I had a problem on the http://localhost:9090/setup/setup-datasource-settings.jsp page. It saw that the folder is not exists and tried to create it. But there was the link, it just was broken. So the PR just improved error handling for this situation to simplify debugging.
Additionally it contains some changes:
Remove deprecated finialize() method since it won't be called anyway anymore.
Commits on Jul 7, 2025
[EmbeddedConnectionProvider.destroy(): add dataSource.close()](https://github.com/igniterealtime/Openfire/commit/0f6b0cf3c7bdc51955edbb27f81e5fb636272c1f) to make it similar to the `destroy()` method of the `DefaultConnectionProvider`.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
@@ -72,6 +72,7 @@ public class EmbeddedConnectionProvider implements ConnectionProvider {
             }
         } catch (IOException e) {
             Log.error("Unable to create 'embedded-db' directory", e);
+            throw new RuntimeException("Unable to create 'embedded-db' directory", e);
         }
 
         serverURL = "jdbc:hsqldb:" + databaseDir.resolve("openfire");
--- a/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
@@ -65,14 +65,14 @@ public class EmbeddedConnectionProvider implements ConnectionProvider {
     @Override
     public void start() {
         final Path databaseDir = JiveGlobals.getHomePath().resolve("embedded-db");
-        try {
-            // If the database doesn't exist, create it.
-            if (!Files.exists(databaseDir)) {
+        // If the database doesn't exist, create it.
+        if (!Files.exists(databaseDir)) {
+            try {
                 Files.createDirectory(databaseDir);
+            } catch (IOException e) {
+                Log.error("Unable to create 'embedded-db' directory", e);
+                throw new RuntimeException("Unable to create 'embedded-db' directory", e);
             }
-        } catch (IOException e) {
-            Log.error("Unable to create 'embedded-db' directory", e);
-            throw new RuntimeException("Unable to create 'embedded-db' directory", e);
         }
 
         serverURL = "jdbc:hsqldb:" + databaseDir.resolve("openfire");
--- a/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
@@ -117,7 +117,9 @@ public class EmbeddedConnectionProvider implements ConnectionProvider {
             DbConnectionManager.closeConnection(pstmt, con);
         }
         try {
-            dataSource.close();
+            if (dataSource != null) {
+                dataSource.close();
+            }
         } catch (final Exception e) {
             Log.error("Unable to close the data source", e);
         }
--- a/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
@@ -111,6 +111,11 @@ public class EmbeddedConnectionProvider implements ConnectionProvider {
         finally {
             DbConnectionManager.closeConnection(pstmt, con);
         }
+        try {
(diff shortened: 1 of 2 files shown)


guusdk (changes requested):
Thanks! I have one question and two minor suggestions:
- I have created ticket https://igniterealtime.atlassian.net/browse/OF-3102 to track this change. Can you please rewrite the commit messages to include 'OF-3102' so that JIRA/GitHub magic can automatically create release notes?
- Whenever a log message is written about some issue with a file or directory location, it would be helpful for debugging purposes to include the location itself (consider adding the absolute path value to the logged message).

Why did you remove the `finalize` method. I know that Finalization is being deprecated in Java, but as far as I can see, it doesn't really hurt to keep it around, for as long as it's still supported?

guusdk:
I agree that it has been _deprecated_, but it has not (yet) been _removed_ from Java. By removing this method from the Openfire class, we run the risk of introducing a behavioral change.

I'm all for improving the code / replacing that Finalization code, but perhaps it's better to do that in a separate PR. That will allow us to merge the other changes while we work on the close/cleanup/finalization code.

guusdk:
> It's simply not used by anyone.

This is untrue. It is correct that usage of the `finalize` method is discouraged (and I agree that we should look into a better approach) but currently (at least with Java 17, the minimum Openfire requirement), the Finalization mechanism is only deprecated (marked for deletion), but still available and in use by Java.

With the Finalization mechanism still being in use, the JVM will invoke the `finalize()` method when an instance is up for garbage collection. There does not need to be any code explicitly calling this method. The JVM will do this 'automatically'.

It can be easily demonstrated with the following code:

```java
public class Demo {
    public static void main(String[] args) throws InterruptedException
    {
        Demo demo = new Demo();
        System.out.println("New object was constructed.");
        demo = null;
        System.out.println("Object was dereferenced.");
        System.gc(); // No guarantees, but this usually does the trick
        Thread.sleep(1000);
    }

    @Override protected void finalize() {
        System.out.println("Instance's 'finalize' method was invoked.");
    }
}
```

With OpenJDK 17, this is the output that is generated:
```
/usr/lib/jvm/java-17-openjdk-amd64/bin/java -javaagent:/data/JetBrains/Toolbox/intellij-idea-ultimate/lib/idea_rt.jar=43453:/data/JetBrains/Toolbox/intellij-idea-ultimate/bin -Dfile.encoding=UTF-8 -classpath /home/guus/SourceCode/personal/finalize/target/classes Demo
New object was constructed.
Object was dereferenced.
Instance's 'finalize' method was invoked.

Process finished with exit code 0
```

xmppserver/src/main/java/org/jivesoftware/database/ConnectionProvider.java
```diff
@@ -74,5 +74,6 @@ public interface ConnectionProvider {
      * from one connection provider to another to ensure that there are no
      * dangling database connections.
      */
-    void destroy();
+    @Override
+    void close();
```
guusdk: This interface is likely used by third parties. This update would break those implementations. Where it is reasonably doable, we should try to prevent this.

I believe we can do that here with a few simple modifications:

1. keep the `destroy()` method, but annotate it as being deprecated:
  ```java
  /**
   * @deprecated Replaced by #close
   */
  @Deprecated(forRemoval=true); // Remove in or after Openfire 5.2.0
  void destroy();
  ```
2. Introduce the new `close()` method with a default implementation that calls the deprecated 'destroy' method:
  ```java
  default void close()
  {
    destroy();
  }
```

(I've created this code by hand, perhaps it has syntax errors)

This will have the benefit of warning developers that they need to change their implementation, without immediately breaking their implementation, allowing them to gracefully migrate.

**Review by guusdk — CHANGES_REQUESTED** [review]
Can you please update the copyright headers of the files that you changed? The copyright should 'end' in the current year (e.g. change "Copyright 2021-2023" into "Copyright 2021-2025").

The more that I think of this, the more that I wonder if making this `AutoCloseable` is a good idea. A `ConnectionProvider` is intended as a long-living object (similar to a factory). It's not intended to be used in a try-with-resources kind of structure. In that sense, I wonder if it makes sense to refactor it: We risk breaking APIs, but for what benefit?

In any case, let's move this out of this particular PR, too. The change would be important enough to have its own JIRA issue and release note entry (as it is likely to cause third-party implementors to have to do work to adjust their implementations).

File: xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java [hunk 2]
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004-2008 Jive Software, 2017-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2004-2008 Jive Software, 2017-20253 Ignite Realtime Foundation. All rights reserved.
```
guusdk: Maybe subtract 18228 from this value. :)

stokito: The Openfire will have a bright feature! Sorry, I'm on a sick leave now and have a bad concentration.

guusdk (approved):
Thanks!

xmppserver/src/main/java/org/jivesoftware/database/EmbeddedConnectionProvider.java
guusdk: Get well soon!

PR #2942, opened 2025-07-09
Makefile test

guusdk:


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: all clean dbkg eclipse
+.PHONY: all clean dbkg
```
Fishbowler: Should delete dbkg too. Git history suggests it's a typo of dpkg from the days of ant, that was fixed in one of two places in 2014, then removed in 1 of 2 places later that year.

PR #2943, opened 2025-07-09
OF-3106 Fix: when JAVA_HOME is empty it fails to detect the best java version

stokito:
The Dash which is the default shell in the Debian/Ubuntu the command returns nothing:

    grep -oe \/.*\/bin\/java

To solve this we need to quote the regexp argument.

https://igniterealtime.atlassian.net/jira/software/c/projects/OF/issues/OF-3106


guusdk (approved):
Thanks, I've verified that this works on Ubuntu, under both Bash and Dash.

Even if this is only used by scripts that are old and should be replaced, it doesn't hurt to have the old script be properly functional. Let's merge this.

PR #2949, opened 2025-07-14
OF-3107 RPM spec cleanup

stokito:


build/rpm/openfire.spec
```diff
@@ -4,11 +4,9 @@ Version: %{OPENFIRE_VERSION}
 Release: %{OPENFIRE_RELEASE}
 BuildRoot: %{_builddir}/%{name}-root
 Source0: %{OPENFIRE_SOURCE}
-%ifarch noarch
 # Note that epoch is set here to 1, this appears to be consistent with non-Redhat
```
stokito, line 7: > epoch is set here to 1

Could you explain what does the comment mean? It looks like very outdated.

akrherz: I don't recall if we once had to mess with the RPM epoch due to some release issue or not.  I also am unsure if it is needed anymore or not, but I thought once you bumped the epoch, you can't go back? shrug

PR #2982, opened 2025-08-18
OF-3116 & OF-3117: Iqhandler error and Testing on Windows

annovanvliet:
The openfire logs are filled with this Error.

It is not serious, but annoying. It is triggered by a server plugin asking the last activity to a server.

```
2024.11.06 09:55:49.573 [1;31mERROR[m [socket_s2s-thread-3]: org.jivesoftware.openfire.handler.IQHandler - Internal server error
java.lang.IllegalArgumentException: IQ must be of type 'set' or 'get'. Original IQ: <iq from="jchat.mlbl.example" to="jchat.oftst.example" type="result" id="544-830"><query xmlns="jabber:iq:last" seconds="12690"></query></iq>
    at org.xmpp.packet.IQ.createResultIQ(IQ.java:378) ~[tinder-2.0.0.jar:?]
    at org.jivesoftware.openfire.handler.IQLastActivityHandler.handleIQ(IQLastActivityHandler.java:56) ~[xmppserver-4.8.3.jar:4.8.3]
```

Second fix is for testing DB unit test on a windows environment.

### DIFF
--- a/xmppserver/src/test/java/org/jivesoftware/openfire/group/AbstractGroupProviderTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/openfire/group/AbstractGroupProviderTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022-2023 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2022-2025 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


guusdk:
I've tried to add WIndows and a whitespace-in-the-working-dir to expose this issue in our CI, in this PR: https://github.com/igniterealtime/Openfire/pull/2986

Turns out that I'm not smart enough to include that whitespace in a GitHub workflow working directory in a way that is maintainable. I've given up.

guusdk:
When running all unit tests from a directory with a space in it, I'm seeing another failure (on Linux) in the LauncherTest class. That test is new, but the construct in which it passes along a File (and it's path), is very old, and it's being used in the Launcher itself.

I wonder if this means that the Launcher is broken, when running from a directory that contains a space.

Could this be a Linux-only problem? The launcher is only used on Windows (as I believe it's only bundled with the Windows distribution). I had imagined people would have complained earlier if that fails when there's a space in the path.

annovanvliet:
You are correct. When i set up my Windows machine i used in my eternal wisdom my full name including spaces. An absolute showstopper for building typical Linux projects which use getResource.

PR #3020, opened 2025-09-14
Refactoring: simplify JSP page usage of escapeHTMLTags and escapeXML

stokito:
Removed null check and the escapeHTMLTags() will just return an empty string instead.

### DIFF
--- a/xmppserver/src/main/webapp/group-edit.jsp
+++ b/xmppserver/src/main/webapp/group-edit.jsp
@@ -308,10 +308,10 @@
         return;
     }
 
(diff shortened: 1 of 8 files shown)


guusdk:
Thanks! For boolean attribute parsing, we should ideally use `org.jivesoftware.util.ParamUtils#getBooleanParameter(javax.servlet.http.HttpServletRequest, java.lang.String)`. That would be consistent with most other JSP pages.

Instead of using scriptlets, can you switch some of the HTML escaping to make use of JSTL where possible? Something like `<c:out value="${foobar}"/>` negates the need for scriptlets.

PR #3021, opened 2025-09-14
i18n: remove preamble in properties files

stokito:
Keep only important note about { and ''.
Move important notes to the Translations documentation.

The dead link to language codes replaced with Wikipedia article.

This saves 20kb of space and makes it faster to parse the props.

### DIFF
--- a/documentation/translator-guide.html
+++ b/documentation/translator-guide.html
@@ -66,25 +66,26 @@ skin.message=<i>Beitrag</i></pre>
         <h2>Construct a Resource Bundle</h2>
 
         <p>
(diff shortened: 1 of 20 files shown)


guusdk (changes requested):
I'm not sure if it matters all that much, but sure, why not.

The automatic integration with Transifex may blow up over this (it's very unpredictable), which, in the worst case scenario, would cause a gazillion PR requests that we need to manually close. That's not a big deal.

My only issue is that this change seems to have modified texts in the `_sv` file. That's probably unintentional. Can you check that please?

PR #3029, opened 2025-09-18
OF-3127: Make MUCRoom's 'occupants' field more thread-safe

guusdk:
By switching the implementation of the List used to represent occupants of a MUC room, the chance of a concurrency issue occurring is reduced (notably, it is no longer possible that a ConcurrentModificationException is thrown when the occupants are iterated over while a change to the list is applied).

The new implementation is less efficient when the collection changes (or `size` is invoked). The number of these events should be small as compared to the amount of times that the collection is iterated over.


guusdk:
This was tested in an external demo environment. Seems to do the trick.

PR #3031, opened 2025-09-20
OF-2309 Fix NPE when plugin was removed manually

stokito:
I manually removed the search plugin from `plugins` folder and then the Openfire didn't started because of null pointer exception in the line:

    final String pluginName = getMetadata(canonicalName).getName();

This happened because the getMetadata(canonicalName) returned null.
So I added the null check to avoid the problem.

Additionally I noticed the isLocalizedTextRTL() method that can be simplified but also it should support more RTL languages.


guusdk (changes requested):
Can you please create a JIRA ticket for the plugin-related NullPointerException, that explains how to reproduce the problem. Then modify the commit message to reference the JIRA issue number please. This way, it becomes part of the release notes.

The RTL change is completely unrelated. Please move that to a different PR. I don't think we have translations for any of the languages that you added (so I'm not really seeing the point)? It shouldn't hurt, though, so I'm not against adding more language definitions.

PR #3032, opened 2025-09-24
OF-3121: Unit tests for plugin (un/re)loading

guusdk:
This adds a new set of unit tests that verify the behavior of PluginManager's loading, unloading and deleting of plugins.

The commit introduces a set of five test plugins. The source code of these plugins is added in a TAR archive (for future maintenance).

The test plugins live in a hierarchy. Child plugins reference classes that are provided by their parents, which allows for testing of class loading that occurs when plugins are initialized and destroyed.

This is a diagram of the hierarchy:

  Node_A
  ├── Node_A_B
  └── Node_A_C
      ├── Node_A_C_D
      └── Node_A_C_E

A README.txt file has been provided in the directory holding the test plugins to explain what they are and how they're to be maintained.


xmppserver/src/test/java/org/jivesoftware/openfire/container/PluginLoadingSimpleHierarchyTest.java
```diff
@@ -0,0 +1,409 @@
...
+            throw new AssertionFailedException("Unable to execute test: the plugin (that is to be reloaded as part of this test) could not be installed during the test fixture setup.");
+        }
+
+        // Execute system under test.
+        pluginManager.reloadPlugin("node_a");
+
+        // Verify results.
+        assertTrue(pluginManager.isInstalled("node_a"), "Expected the 'plugins' directory to contain the plugin jar file (but it does not).");
+        assertTrue(pluginManager.isExtracted("node_a"), "Expected the 'plugins' directory to contain a subdirectory that matches the canonical name of the plugin (but it does not).");
+        assertTrue(pluginManager.isLoaded("node_a"), "An instance of the plugin's class file is expected to have been created and registered with the plugin manager (but it has not).");
```
Fishbowler, line 159: Are there tests elsewhere that prove that reloadPlugin actually does a thing?

Fishbowler (approved):
Those are some _nice_ tests :heart: 

I tried to contrive more test cases, like "what's the order of plugin loading on restart, and could that cause problems", but I've convinced myself that it wouldn't matter given the tests and changes here.

xmppserver/src/test/java/org/jivesoftware/openfire/container/PluginLoadingSimpleHierarchyTest.java
guusdk: Heh, no. I replaced the implementation for `reloadPlugin()` with `return true` and all tests are happy. :) I'll see if I can improve on that.

guusdk: I have now added a change that tests for the 'last installation date' of a plugin, to assure that it actually got changed.

xmppserver/src/test/java/org/jivesoftware/openfire/container/PluginLoadingSimpleHierarchyTest.java
```diff
@@ -406,4 +423,16 @@ private boolean installPlugin(final String canonicalPluginName) throws IOExcepti
...
+    /**
+     * Returns the time that the plugin was installed (reloaded) last.
+     *
+     * @param canonicalPluginName The name of the plugin that is expected to be installed.
+     * @return the time when the plugin was installed.
+     */
+    private Instant getLastInstallTime(final String canonicalPluginName) throws IOException
+    {
+        final Path explodedPluginPath = tempPluginDirectory.resolve(canonicalPluginName);
+        return Files.getLastModifiedTime(explodedPluginPath).toInstant();
```
Fishbowler, line 434: I had _no idea_ that this what was happening on plugin reload :joy:

guusdk: We're not just killing it with _fire_, we're making it go boom!

guusdk:
Rebased.

PR #3035, opened 2025-10-01
OF-3131: Move 'subject' from history to room

guusdk:
This commit shifts responsibility for maintaining the last stanza that changed a MUC room's subject from its HistoryStrategy to the MUCRoom implementation itself.

The purpose for this change is to ensure that a MUCRoom has all data required to broadcast its latest subject.

The changes in this commit include:
- datatype change for `MUCRoom#subject` from String to Stanza
- deprecation of various subject-related methods in `HistoryStrategy`
- database migration to copy the latest subject stanza from table `ofMucConversationLog` to `ofMucRoom`

The database migration is not perfect: it does not migrate the stanza's timestamp, which means that the subject that is broadcast no longer mentions _when_ the subject was changed (as the stanzas in `ofMucConversationLog` do have an occupantJID in its 'from' attribute, the 'who' is correctly retained).

In rare occasions (see OF-3131) the room can have a subject, while the history does not. In those cases, the database migration scripts leave the (non-stanza) subject in the ofMucRoom table intact (this means that the column holds a mixture of plain text and XMPP data). The code that reads from the database will generate a stanza on the fly from such a plain text subject.


akrherz (approved):
Seems sensible

guusdk:
I've reverted the commit that introduces the test data. This way, the data remains part of the commit history, without continuing to be part of the migration scripts.

guusdk:
Rebased

Fishbowler:
Based on the dummy data tests you did, that covers that it upgrades without error on 4 systems (MySQL, MSSQL, Postgres, Oracle). You've also had a play with HSQLDB. From a risk POV, I think we're mostly there.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
```diff
@@ -2629,47 +2628,105 @@ public void nicknameChanged(MUCOccupant occupant, Presence newPresence, String o
...
+            // when empty), as MUC joins require 'subject' element to be present.
+            if (subject != null && !subject.isEmpty()) {
+                roomSubject.setSubject(subject);
+            } else {
+                roomSubject.getElement().addElement("subject");
+            }
+
+            changeSubject(roomSubject, this.selfOccupantData);
+        } catch (ForbiddenException e) {
+            Log.warn("Unable to change the subject of room {}", this.getJID(), e);
```
Fishbowler, line 2728: Why is this the only exception that could happen? Surely there could (should?) be database and other things that could occur and be trapped and bubbled up?

distribution/src/database/upgrade/38/openfire_hsqldb.sql
```diff
@@ -0,0 +1,25 @@
...
+    FROM ofMucConversationLog l
+    WHERE l.roomID = r.roomID
+      AND l.subject IS NOT NULL
+      AND l.logTime = (
+          SELECT MAX(l2.logTime)
+          FROM ofMucConversationLog l2
+          WHERE l2.roomID = r.roomID
+            AND l2.subject IS NOT NULL
+      )
+);
```
Fishbowler, line 29: LLM complained that this might have many full table scans, and suggested using ROW_NUMBER like you did for Oracle. I've no experience here, but [docs](https://hsqldb.org/doc/guide/builtinfunctions-chapt.html) suggest that it's possible.

xmppserver/src/main/java/org/jivesoftware/openfire/muc/MUCRoom.java
guusdk: This is the only _checked_ exception that is thrown. Things like database exceptions are handled within these calls.

There's a lot to be said about handling of checked vs unchecked exceptions, but also on the handling of database errors in Openfire. That, however, is probably best addressed in a distinct issue/topic/pull request. What we do here is at least roughly consistent with the rest of the code.

distribution/src/database/upgrade/38/openfire_hsqldb.sql
guusdk: ROW_NUMBER in HSQLDB appears to have quirks. LLM suggests that we can use it only for this particular purpose after upgrading from 2.7.1 to 2.7.4 (which we may want to do anyways).

I do believe that the query can be improved for a roughly comparable performance improvement, but with something that's usable under 2.7.1. I have added a commit that has that improvement.

PR #3036, opened 2025-10-01
CI: Help future-self with DATABASE_VERSION xcheck

akrherz:


guusdk:
This is going to be handy. Does it make sense to only run this when there are database-related changed? I think we can do that by depending on `should-do-database-tests`

akrherz:
rebased, this better be green or else.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -294,6 +294,33 @@ jobs:
               - 'xmppserver/src/main/java/org/jivesoftware/database/**'
 
 
+  xcheck-database-version:
+    name: Cross check that any new database upgrades match the code version
+    needs: [build, should-do-database-tests, check_branch]
+    runs-on: ubuntu-latest
+    if: ${{ needs.should-do-database-tests.outputs.check == 'true' || needs.check_branch.outputs.is_publishable_branch == 'true'}}
```
guusdk, line 301: What is the purpose of this `if`? Doesn't it duplicate the `needs`?

akrherz: I don't think so, needs just wires up the job order so that it runs after the needed step, so it can use the output variable.

PR #3037, opened 2025-10-02
OF-3132: Improve iteration over ClientSessions of a user

guusdk:
Previously, when iterating over all ClientSessions of a particular user, only those sessions that are 'avialable' were returned. This caused sessions that were not (yet) available to be missed.

This in turn is the source of various race conditions, where a session would be excluded because it wasn't done sending initial presence yet.

In this commit, an additional method in SessionManager is implemented, that allows for _all_ sessions of a particular client to be returned.

Additionally, the ClientSession-count, that had a 'todo' for this, now correctly returns the count for _all_ (rather than just available) sessions.

Finally, some usage of the old method (that only returned 'available' sessions for a client) have been replaced with the new method.


akrherz (approved):
seems sensible

PR #2204, opened 2023-06-21
Unit test improvements

guusdk:
This mostly migrates unit tests in Openfire's code base from JUnit 4 to 5. The PR includes some other minor unit test improvements. Please refer to the commit messages for more details.

PR #3038, opened 2025-10-03
Standardize test suite to use JUnit Jupiter

mureinik:
This PR follows up on the work previously done in #2204.
That PR migrated most of the tests to JUnit Jupiter, but left some JUnit 4-style tests. This PR finishes the job and migrates whatever leftovers that PR had to JUnit Jupiter. The goal is to have a standardized approach to tests throughout the codebase so it's easy to contributors to know how to write new tests.

This PR includes the following changes:

* Dependencies (pom.xml):
  * The `org.junit:junit-bom` dependency was upgraded to 5.13.4 to gain the functionality of `org.junit.jupiter.params.ParameterizedClass` (see below). Technically, it was introduced in 5.13.0, but we might as well get a few bugfixes added in newer patch versions.
  * The `org.dbunit:dbuit` dependency was upgraded to 3.0.0, which now longer depends on JUnit 4.
 * The `org.junit.vintage:junit-vintage-engine` was removed.
  * The `junit:junit:4.13.1` dependency was replaced with `org.junit.jupiter:junit-jupiter` in the `connectivtytests-action`
*  Mockito handling:
   * The `org.junit.runner.RunWith` annotations used to call `org.mockito.junit.MockitoJUnitRunner` in classes with no Mockito annotations was simply removed.
   * The `org.junit.runner.RunWith` annotations used to call `org.mockito.junit.MockitoJUnitRunner.Silent` were replaced with `org.junit.jupiter.api.extension.ExtendWith` calling `org.mockito.junit.jupiter.MockitoExtension`.
* Parameterization handling:
  * The `org.junit.runner.RunWith` annotation used to call `org.junit.runners.Parameterized` was replaced with `org.junit.jupiter.params.ParameterizedClass`
  * The `org.junit.runners.Parameterized.Parameters` annotation used to define the test's parameters was replaced with a `org.junit.jupiter.params.provider.MethodSource` annotation on the class level. The method signature and the names of the method and the local variables inside it was amended accordingly.
  * The class members initialized by the parameterized test were annotated with `org.junit.jupiter.params.Parameter`. Since the `ParameterizedClass` injects values directly to the members and not to the constructor, the constructor was removed and the `final` modifier was removed from the members
* Other JUnit Annotations:
  * `org.junit.jupiter.api.BeforeAll` was used as a drop in replacement for `org.junit.BeforeClass`.
  * `org.junit.jupiter.api.BeforeEach` was used as a drop in replacement for `org.junit.Before`.
  * `org.junit.jupiter.api.AfterEach` was used as a drop in replacement for `org.junit.After`.
  * `org.junit.jupiter.api.Test` was used as a drop in replacement for `org.junit.Test` for the simple case where the annotation had no arguments.
    * For the cases where the annotation had an `expected` argument, this replacement was still used, but the assertion was handled manually - see below
    * For the cases where the annotation had a `timeout` parameter, the `org.junit.jupiter.api.Timeout` annotation was added to provide this functionality.
  * `org.junit.Rule` annotations were removed and replaced with other mechanisms, see below.
  * Since the `org.dbunit.DBTestCase` is no longer executed with the JUnit 3 runner, the test lifecycle of the tests that extend it needed to be managed by the test itself. To that end:
    * All the test cases were annotated with `org.junit.jupiter.api.Test`
    * `setUp` methods were annotated with `org.junit.jupiter.api.BeforeEach`
    * `tearDown` methods were annotated with `org.junit.jupiter.api.AfterEach`
* Assertions
  * The `assertXYZ` methods from `org.junit.jupiter.api.Assertions` were used as drop in replacements for their counterparts from `org.junit.Assert` when no message is added.
  * The `assertXYZ` methods from `org.junit.jupiter.api.Assertions` were used as replacements for their counterparts from `org.junit.Assert` when a message is used, but the argument order is changed - `org.junit.jupiter.api.Assertions`' methods take the optional message argument as the last argument, not the first.
  * `org.junit.jupiter.api.Assertions#assertThrows` was used to assert that an expected exception was thrown instead of using `org.junit.Test#expected`. Note that this change has a theoretical advantage of making the test a bit stricter. With `org.junit.Test#expected`, the test would pass if any statement in the test threw the exception, including the code used to set up the pre-conditions. Using `org.junit.jupiter.api.Assertions#assertThrows` allows us to localize the expectation to a single statement that should throw the exception.
  * `org.junit.jupiter.api.Assertions#assertThrows` was used to assert that an expected exception was thrown instead of using `org.junit.org.junit.rules.ExpectedException`.
  *  Since `org.dbunit.DBTestCase` no longer extends `framework.junit.TestCase`, it no longer provides any `assertXYZ` methods to classes that extend it. Instead, these methods' counterparts were imported from `org.junit.jupiter.api.Assertions`.
* Other cleanups:
  * Unsued imports were removed

### DIFF
--- a/.github/actions/connectivitytests-action/conntest/pom.xml
+++ b/.github/actions/connectivitytests-action/conntest/pom.xml
@@ -28,9 +28,9 @@
             <version>4.3.4</version>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.1</version>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.13.4</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
+++ b/.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
@@ -4,8 +4,8 @@ import org.jivesoftware.smack.bosh.BOSHConfiguration;
 import org.jivesoftware.smack.bosh.XMPPBOSHConnection;
 import org.jivesoftware.smack.tcp.XMPPTCPConnection;
 import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
-import org.junit.Assert;
-import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
 public class ConnectivityTest
 {
@@ -57,7 +57,7 @@ public class ConnectivityTest
         }
 
         // Verify result.
-        Assert.assertEquals("Expected all BOSH connection attempts to result in an authenticated session, but not all did.", iterations, result);
+        Assertions.assertEquals(iterations, result, "Expected all BOSH connection attempts to result in an authenticated session, but not all did.");
     }
 
     /**
@@ -106,6 +106,6 @@ public class ConnectivityTest
         }
 
         // Verify result.
-        Assert.assertTrue("Expected all TCP connection attempts to result in an authenticated session, but not all did.", iterations >= result * .9); // FIXME make this an _exact_ comparison after stability of the BOSH connectivity is achieved.
+        Assertions.assertTrue(iterations >= result * .9, "Expected all TCP connection attempts to result in an authenticated session, but not all did."); // FIXME make this an _exact_ comparison after stability of the BOSH connectivity is achieved.
     }
 }


guusdk:
That's an impressive change, thanks!

guusdk (approved):

guusdk:
The test failure is not related to your changes. It affects many PRs.

We're currently trying to get that test suite stable. It annoyingly doesn't fail locally, but fails in CI. We're addressing various race conditions related to that failure.

guusdk:
We've addressed a major source of these race conditions (although I don't think we've addressed them all. Still, rebasing to the latest HEAD of the Main branch may give you a better chance of 'going green'.

mureinik:
> We've addressed a major source of these race conditions (although I don't think we've addressed them all. Still, rebasing to the latest HEAD of the Main branch may give you a better chance of 'going green'.

Thanks!
I've rebased the PR, let's see how it goes this time

mureinik:
> Nice work. Even the hairy ones like DBTestCase are working. This is awesome ❤️

Thanks!

PR #3042, opened 2025-10-07
Offer dialback when domain is not valid.

pelle-c:
This commit adds a configuration option, "xmpp.server.dialback.offer", that can be used to control if dialback stream feature should be offered.

One usecase is where openfire is used as a gateway between XMPP domains ("trunking") - https://download.igniterealtime.org/openfire/docs/latest/documentation/trunking-guide.html

In situations where dialback has to be enabled in the gateway, it also forces the server(s) behind the gateway to use dialback. Before this commit, the code did not add dialback unless the certificates are selfsigned. Which means that a session to a remote domain via the gateway will not be established both ways due to lack of dialback, when the certificates are signed by CAs and trust exists.

With "xmpp.server.dialback.offer" set to "true", dialback will be added as a stream feature and sessions can be established in both directions.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/session/LocalIncomingServerSession.java
@@ -27,6 +27,7 @@ import org.jivesoftware.openfire.server.ServerDialback;
 import org.jivesoftware.openfire.server.ServerDialbackErrorException;
 import org.jivesoftware.openfire.server.ServerDialbackKeyInvalidException;
 import org.jivesoftware.util.CertificateManager;
+import org.jivesoftware.util.JiveGlobals;
 import org.jivesoftware.util.StreamErrorException;
 import org.jivesoftware.util.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


guusdk:
Hi! Thanks for your contribution. I'll need to find some time to dig into it in more detail (I wonder if this is the most correct place to control that behavior), but one thing that does stand out to me is that we'd want an option like this ("enable Server Dialback Authenticaton") to be added to the admin console pages.

The admin console pages now only allow for "Mutual Authentication" (SASL EXTERNAL) to be required, which _implies_ that Server Dialback is to be used as a fallback. It would be far better to have the Server Dialback option be there explicitly (with maybe a warning if both SASL EXTERNAL _and_ Dialback are disabled).

Can you add that to `connection-settings-advanced.jsp` and friends?

image

pelle-c:
Sounds a bit tricky indeed. Too much flexibility often leads to increased complexity. 
Would it be possible to collect all settings that are common for each type (s2s, cs2s...) on each "startpage", and then have the settings that differ on a page like "advanced settings". Then have boxes like today, but with more clear distinction - encryption and certificate validation, ciphers and protocols, authentication etc. If a box is not directly applicable for the type, just an information note could point in a better direction. Not a big change page I assume, mostly a question of rephrasing and moving things around.

PR #3138, opened 2026-02-12
OF-3188 + OF-3189: Add generic support for XEP-0128: Service Discovery Extensions

Fishbowler:
I want to implement XEP-0504: Data Policy (https://xmpp.org/extensions/xep-0504.html) which relies on [XEP-0128: Service Discovery Extensions](https://xmpp.org/extensions/xep-0128.html).

We've already got 2 specific XEP implementations (Contact Addresses, Software Info) that depend on XEP-0128, and we've got the MUC Extended Info plugin that extends disco responses too.

This change creates a pluggable ExtendedDiscoInfoProviders capability for plugins to add support for XEPs (or other functionality) that rely on XEP-0128.

It also moves the Contact Addresses and the Software Info implementations to use the new thing, and updates the DOAP.

Lots of AI in the JavaDoc and unit tests, which I think I've fixed up, but use sufficient scepticism.

### DIFF
--- a/documentation/openfire.doap
+++ b/documentation/openfire.doap
@@ -206,7 +206,6 @@
         <implements>
             <xmpp:SupportedXep>
                 <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0128.html"/>
-                <xmpp:note xml:lang='en'>Provided by the 'MUC Service Discovery Extensions' plugin</xmpp:note>
             </xmpp:SupportedXep>
         </implements>
         <implements>
@@ -245,6 +244,11 @@
         <implements>
             <xmpp:SupportedXep>
                 <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0156.html"/>
+            </xmpp:SupportedXep>
+        </implements>
+        <implements>
+            <xmpp:SupportedXep>
+                <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0157.html"/>
             </xmpp:SupportedXep>
         </implements>
         <implements>
@@ -322,6 +326,11 @@
         <implements>
             <xmpp:SupportedXep>
                 <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0223.html"/>
+            </xmpp:SupportedXep>
+        </implements>
+        <implements>
+            <xmpp:SupportedXep>
+                <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0232.html"/>
             </xmpp:SupportedXep>
         </implements>
         <implements>
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright (C) 2026 Ignite Realtime Foundation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jivesoftware.openfire.disco;
+
+import org.jivesoftware.openfire.XMPPServer;
+import org.jivesoftware.openfire.user.UserManager;
+import org.jivesoftware.util.JiveGlobals;
+import org.xmpp.forms.DataForm;
+import org.xmpp.forms.FormField;
+import org.xmpp.packet.JID;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Provides XEP-0157 Contact Addresses for XMPP Services via service discovery.
+ * <p>
+ * This provider returns a data form containing administrative contact addresses for the server,
+ * including XMPP addresses for administrators and their email addresses when available.
+ * </p>
+ * <p>
+ * The contact information is only returned when:
+ * <ul>
+ *   <li>The {@code admin.disable-exposure} property is not set to true</li>
+ *   <li>The request is for the server domain (name is null or matches the server domain)</li>
+ *   <li>No specific node is requested (node is null)</li>
+ *   <li>At least one administrator is configured</li>
+ * </ul>
+ * </p>
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0157.html">XEP-0157: Contact Addresses for XMPP Services</a>
+ */
+public class ContactAddressesExtendedDiscoInfoProvider implements ExtendedDiscoInfoProvider {
+
+    @Override
+    public Set<DataForm> getExtendedInfos(String name, String node, JID senderJID) {
+        // Return empty set if admin exposure is disabled
+        if (JiveGlobals.getBooleanProperty("admin.disable-exposure")) {
+            return Collections.emptySet();
+        }
+
+        // Only respond for server-level requests (no node)
+        if (node != null) {
+            return Collections.emptySet();
+        }
+
+        // Only respond for server domain or null name
+        if (name != null && !name.equals(XMPPServer.getInstance().getServerInfo().getXMPPDomain())) {
+            return Collections.emptySet();
+        }
+
+        // Get admins and return empty if none
+        final Collection<JID> admins = XMPPServer.getInstance().getAdmins();
+        if (admins == null || admins.isEmpty()) {
+            return Collections.emptySet();
+        }
+
+        // Build XEP-0157 form
+        final DataForm dataForm = new DataForm(DataForm.Type.result);
+
+        final FormField fieldType = dataForm.addField();
+        fieldType.setVariable("FORM_TYPE");
+        fieldType.setType(FormField.Type.hidden);
+        fieldType.addValue("http://jabber.org/network/serverinfo");
+
+        final FormField fieldAdminAddresses = dataForm.addField();
+        fieldAdminAddresses.setVariable("admin-addresses");
+        fieldAdminAddresses.setType(FormField.Type.list_multi);
+
+        final UserManager userManager = UserManager.getInstance();
+        for (final JID admin : admins) {
+            fieldAdminAddresses.addValue("xmpp:" + admin.asBareJID());
+            if (admin.getDomain().equals(XMPPServer.getInstance().getServerInfo().getXMPPDomain())) {
+                try {
+                    final String email = userManager.getUser(admin.getNode()).getEmail();
+                    if (email != null && !email.trim().isEmpty()) {
+                        fieldAdminAddresses.addValue("mailto:" + email);
+                    }
+                } catch (Exception e) {
+                    // User not found or other error - skip email for this admin
+                    continue;
+                }
+            }
+        }
+
+        final Set<DataForm> dataForms = new HashSet<>();
+        dataForms.add(dataForm);
+        return dataForms;
+    }
+}
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
@@ -18,7 +18,6 @@ package org.jivesoftware.openfire.disco;
 
 import org.jivesoftware.openfire.XMPPServer;
 import org.jivesoftware.openfire.user.UserManager;
-import org.jivesoftware.util.JiveGlobals;
 import org.xmpp.forms.DataForm;
 import org.xmpp.forms.FormField;
 import org.xmpp.packet.JID;
@@ -51,7 +50,7 @@ public class ContactAddressesExtendedDiscoInfoProvider implements ExtendedDiscoI
     @Override
     public Set<DataForm> getExtendedInfos(String domain, String name, String node, JID senderJID) {
         // Return empty set if admin exposure is disabled
-        if (JiveGlobals.getBooleanProperty("admin.disable-exposure")) {
+        if (IQDiscoInfoHandler.DISABLE_EXPOSURE.getValue()) {
             return Collections.emptySet();
         }
 
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
@@ -49,7 +49,7 @@ import java.util.Set;
 public class ContactAddressesExtendedDiscoInfoProvider implements ExtendedDiscoInfoProvider {
 
     @Override
-    public Set<DataForm> getExtendedInfos(String name, String node, JID senderJID) {
+    public Set<DataForm> getExtendedInfos(String domain, String name, String node, JID senderJID) {
         // Return empty set if admin exposure is disabled
         if (JiveGlobals.getBooleanProperty("admin.disable-exposure")) {
             return Collections.emptySet();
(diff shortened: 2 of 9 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/disco/ExtendedDiscoInfoProvider.java
```diff
@@ -0,0 +1,65 @@
...
+ * </ul>
+ * </p>
+ *
+ * When multiple providers return forms with the same FORM_TYPE, the forms are merged by combining
+ * their fields. However, <b>each field must be unique</b> within a form type. If two providers contribute
+ * fields with the same field name to the same FORM_TYPE, this is treated as a configuration error. The duplicate
+ * provider's entire contribution is skipped and a warning is logged. Other providers continue to be processed
+ * normally.
+ *
+ * Implementations should use 'JiveGlobals.getBooleanProperty( "admin.disable-exposure" )' and consider appropriate
```
guusdk: I don't think this property is used at all? If it is:
- this description should be updated (as it's not clear what the purpose of this property is)
- it should use SystemProperty rather than JiveGlobals

xmppserver/src/test/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandlerTest.java
```diff
@@ -0,0 +1,547 @@
...
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+class IQDiscoInfoHandlerTest {
```
guusdk, line 62: AI will easily generate some appropriate javadoc for this class and its methods. :)

xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
```diff
@@ -0,0 +1,106 @@
...
+ * </p>
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0157.html">XEP-0157: Contact Addresses for XMPP Services</a>
+ */
+public class ContactAddressesExtendedDiscoInfoProvider implements ExtendedDiscoInfoProvider {
+
+    @Override
+    public Set<DataForm> getExtendedInfos(String name, String node, JID senderJID) {
+        // Return empty set if admin exposure is disabled
+        if (JiveGlobals.getBooleanProperty("admin.disable-exposure")) {
```
guusdk: Please use `org.jivesoftware.util.SystemProperty` instead of `JiveGlobals`. The former provides type-safety, is centrally defined, and self-documented , whereas `JiveGlobals` is legacy and less robust.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
```diff
@@ -0,0 +1,106 @@
...
+            return Collections.emptySet();
```
guusdk, line 79: I’m not sure whether this is an actual issue in this implementation, but it’s something I’ve run into a few times: Collections.emptySet() returns an immutable set. If a caller later assumes the returned collection is mutable and attempts to add elements, this will result in a runtime exception.

If immutability is intentional here, it might be worth making that contract explicit (e.g., via documentation or naming). Otherwise, returning a mutable empty collection could help avoid subtle bugs for callers down the line.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/SoftwareInfoExtendedDiscoInfoProvider.java
```diff
@@ -0,0 +1,115 @@
...
+        .setKey("xmpp.iqdiscoinfo.xformsoftwareversion")
+        .setDefaultValue(Boolean.TRUE)
+        .setDynamic(Boolean.TRUE)
+        .build();
+
+    @Override
+    public Set<DataForm> getExtendedInfos(String name, String node, JID senderJID) {
+        // Check if feature is enabled
+        if (!ENABLED.getValue()) {
+            return Collections.emptySet();
```
guusdk, line 89: More usage of immutable collections here. See earlier comment.

xmppserver/src/test/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProviderTest.java
```diff
@@ -0,0 +1,223 @@
...
+import org.xmpp.forms.DataForm;
+import org.xmpp.forms.FormField;
+import org.xmpp.packet.JID;
+
+import java.util.*;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+class ContactAddressesExtendedDiscoInfoProviderTest {
```
guusdk, line 55: AI can do wonders with generating a bit of nice javadoc. That'll be helpful for future us!

xmppserver/src/test/java/org/jivesoftware/openfire/disco/SoftwareInfoExtendedDiscoInfoProviderTest.java
```diff
@@ -0,0 +1,223 @@
...
+import org.xmpp.forms.DataForm;
+import org.xmpp.forms.FormField;
+import org.xmpp.packet.JID;
+
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+class SoftwareInfoExtendedDiscoInfoProviderTest {
```
guusdk, line 58: Old man complaining about missing javadoc again.

documentation/openfire.doap
```diff
@@ -206,7 +206,6 @@
         <implements>
             <xmpp:SupportedXep>
                 <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0128.html"/>
-                <xmpp:note xml:lang='en'>Provided by the 'MUC Service Discovery Extensions' plugin</xmpp:note>
```
guusdk, line 209: Does this mean we need to sunset the old plugin?

guusdk (changes requested):
I've left some minor feedback inline. There's an overarching concern: conceptually, this change feels odd to me. It doesn’t introduce new functionality, but rather an alternative way to achieve the same result. Openfire already provides an extension point for obtaining “extended infos” via org.jivesoftware.openfire.disco.DiscoInfoProvider#getExtendedInfos. This commit introduces a second mechanism for essentially the same purpose. In fact, the getExtendedInfos method in the new ExtendedDiscoInfoProvider has the exact same signature as the existing one, which feels like a code smell.

The main difference seems to be how the data from DiscoInfoProvider and ExtendedDiscoInfoProvider is combined into the final result, with the new provider adding more complex merging logic. I wonder whether this PR could be improved by eliminating the second provider and instead applying the merging logic to the existing extension point. Is there any functionality enabled by the current approach that wouldn’t be possible that way?

xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
Fishbowler: This was just me moving the existing JiveGlobals impl from it's old home. I can absolutely update it as part of this PR.

documentation/openfire.doap
Fishbowler: Not yet, it's still useful. But I'll update it at some point. I figured it was better to list Openfire as the implementation rather than a plugin (which was only partial anyway)

xmppserver/src/main/java/org/jivesoftware/openfire/disco/ExtendedDiscoInfoProvider.java
Fishbowler: This comment is for implementers of this interface, to let them know it's there and may want to be considered. It's used in both implementations inside Openfire (e.g. ContactAddressesExtendedDiscoInfoProvider).

When I update it to a SystemProperty, I'll update this doc, and make it clearer.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/ContactAddressesExtendedDiscoInfoProvider.java
Fishbowler: I dealt with this. The Set is iterated, and its contents used, but the set itself can be immutable - it's gonna get thrown away shortly after creation.

Fishbowler:
@guusdk 

> this change feels odd to me. It doesn’t introduce new functionality, but rather an alternative way to achieve the same result. Openfire already provides an extension point for obtaining “extended infos”

The existing functionality doesn't expose an ability to extend disco results _via a plugin_, except via the [proxy approach in mucextinfo](https://github.com/igniterealtime/openfire-mucextinfo-plugin/blob/8547833a3d1d940c5d465c1f0668e4247299ca93/src/main/java/org/igniterealtime/openfire/plugin/mucextinfo/DiscoInfoProviderProxy.java#L78-L96).

This felt like a good way to offer a simple hook to plugins. There's a much simpler version that deals with stuff like Data Policy, but I also wanted to cater for mucextinfo and similar, and bring the merge stuff in-house at the same time.

Right now, the testing has shown that this implementation has gaps (like not covering MultiUserChatServiceImpl or PubSubModule's implementations of getExtendedInfos), so if you've got thoughts on how to better encapsulate this functionality for a plugin to extend a disco#info, am happy to adapt.

I'm thinking that I might pull some stuff out of the anonymous DiscoInfoProvider at L616 and put it into the IQDiscoInfoHandler and call it from handleIQ instead (where it calls out to the other implementations). But at the point where its being hoisted to the root of the class, it could also all be moved to some sensible place in a file nearby, probably as mostly static methods... WDYT?

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -91,10 +90,20 @@ public class IQDiscoInfoHandler extends IQHandler implements ClusterEventListene
...
+    private final ContactAddressesExtendedDiscoInfoProvider contactAddressesProvider = new ContactAddressesExtendedDiscoInfoProvider();
+    private final SoftwareInfoExtendedDiscoInfoProvider softwareInfoProvider = new SoftwareInfoExtendedDiscoInfoProvider();
+
+    /**
+     * Controls whether extended service discovery information that may expose
+     * administrative details (e.g. contact addresses, software versions) should be
+     * included in disco#info responses. When set to true, such information is
+     * suppressed for security/privacy reasons.
+     */
+    public static final SystemProperty<Boolean> DISABLE_EXPOSURE = SystemProperty.Builder.ofType(Boolean.class)
```
guusdk, line 105: The _name_ of this is somewhat ambiguous to me. Am I also seeing distinct ENABLE properties for each specific provider that adds a detail? Do we need two settings that partially overlap? On first thought, I'd prefer to remove this particular one (as it's not very clear to what it does or does not apply).

Fishbowler: I thought the description in the i18n was pretty good? Its purpose seems more concrete in the old impl [here](https://github.com/igniterealtime/Openfire/blob/993f467533405b7ec6fd769884833e5144137ed0/xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java#L750) where it's fetch wraps a bunch of extra stuff that would get added to the disco info if this "privacy mode" isn't enabled.

Fishbowler: Presently, this is just tidying what was there before to fit The New Thing, intentionally not altering logic or behaviour of these properties or of the Contact Addresses / Software Info implementation. Happy to, though, if you think it's warranted.

guusdk: I think I was confused by this being in the generic class, and not in some specialized class that exposes contact information of administrative entities for the domain. I'm happy to leave it as-is.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -91,10 +90,20 @@ public class IQDiscoInfoHandler extends IQHandler implements ClusterEventListene
...
     private List<UserFeaturesProvider> anonymousUserFeatureProviders = new ArrayList<>();
     private List<UserFeaturesProvider> registeredUserFeatureProviders = new ArrayList<>();
+    private List<ExtendedDiscoInfoProvider> extendedDiscoInfoProviders = new ArrayList<>();
 
-    public static final SystemProperty<Boolean> ENABLED = SystemProperty.Builder.ofType(Boolean.class)
-        .setKey("xmpp.iqdiscoinfo.xformsoftwareversion")
-        .setDefaultValue(Boolean.TRUE)
+    private final ContactAddressesExtendedDiscoInfoProvider contactAddressesProvider = new ContactAddressesExtendedDiscoInfoProvider();
+    private final SoftwareInfoExtendedDiscoInfoProvider softwareInfoProvider = new SoftwareInfoExtendedDiscoInfoProvider();
+
```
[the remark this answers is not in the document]
Fishbowler: Thanks! Fixed.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -577,6 +599,151 @@ else if (field.element("media") != null && field.element("media").element("uri")
...
+    public void removeExtendedDiscoInfoProvider(ExtendedDiscoInfoProvider provider) {
+        extendedDiscoInfoProviders.remove(provider);
+    }
+
+    /**
+     * Merges incoming forms into the result set, combining forms with matching FORM_TYPE.
+     * Forms with the same FORM_TYPE have their fields merged. If duplicate field names
+     * are detected, the offending provider's contribution is skipped with a warning.
+     */
+    private void merge(Set<DataForm> result, Set<DataForm> incoming) {
```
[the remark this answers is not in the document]
Fishbowler: Thanks! Fixed.

xmppserver/src/test/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandlerTest.java
```diff
@@ -0,0 +1,590 @@
...
+     * @param handler the IQDiscoInfoHandler instance to test
+     * @param from the JID of the requesting entity
+     * @param to the JID of the target entity (null for server domain)
+     * @return the set of data forms included in the disco#info response
+     */
+    private Set<DataForm> getExtendedInfosViaHandleIQ(IQDiscoInfoHandler handler, JID from, JID to) {
+        try {
+            // Initialize the handler (this registers the server disco info provider)
+            handler.initialize(XMPPServer.getInstance());
+            handler.start();
```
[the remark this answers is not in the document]
Fishbowler: Thanks! Fixed.

xmppserver/src/main/java/org/jivesoftware/openfire/disco/IQDiscoInfoHandler.java
```diff
@@ -577,6 +599,151 @@ else if (field.element("media") != null && field.element("media").element("uri")
...
+     * Adds a provider that contributes extended information (as per XEP-0128) to disco#info responses.
+     * This allows plugins to add custom data forms or additional fields to existing forms.
+     * <p>
+     * Providers are called for all disco#info requests and their forms are merged with existing forms.
+     * Forms with the same FORM_TYPE are merged, assuming field names are unique.
+     *
+     * @param provider The provider of extended disco info forms.
+     * @see ExtendedDiscoInfoProvider
+     */
+    public void addExtendedDiscoInfoProvider(ExtendedDiscoInfoProvider provider){
```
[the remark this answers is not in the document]
Fishbowler: Thanks! Fixed.

guusdk (approved):
This seems functional. I've tested with the new ContactAddresses provider, and combined that with a change in the PubSub Server Info plugin: https://github.com/igniterealtime/openfire-pubsubserverinfo-plugin/pull/9

That resulted in both extended forms to be added:

```xml
<iq type="result" id="H6GG6-1" from="example.org" to="john@example.org/test">
  <query xmlns="http://jabber.org/protocol/disco#info" node="https://www.igniterealtime.org/projects/openfire/#2/o098yXw9DTZUBG9pKun2BTdW8=">
    <identity category="server" name="Openfire Server" type="im"/>
    <identity category="pubsub" type="pep"/>
<!-- snip -->
    <feature var="urn:xmpp:serverinfo:0"/>
    <feature var="http://jabber.org/protocol/disco#info"/>
    <x xmlns="jabber:x:data" type="result">
      <field var="FORM_TYPE" type="hidden">
        <value>
          http://jabber.org/network/serverinfo
        </value>
      </field>
      <field var="admin-addresses" type="list-multi">
        <value>
          xmpp:admin@example.org
        </value>
        <value>
          mailto:admin@example.com
        </value>
      </field>
      <field var="serverinfo-pubsub-node">
        <value>
          xmpp:pubsub.example.org?;node=serverinfo
        </value>
      </field>
    </x>
  </query>
</iq>
```
My PubSub Server Info test tool correctly picked up the new form, too.

PR #3154, opened 2026-03-04
OF-3187: Make AuthCheckFilter's destroy a no-op

Fishbowler:
This is called when the AdminConsolePlugin restarts (by PluginServlet). It wipes all exclusions that have been registered, requiring plugins to re-register them. This change retains the static list of excludes. Plugins remain responsible for unregistering their excludes, which they should have been doing anyway, rather than relying on an eventual restart of the AdminConsole.

Fixes the generic case of https://github.com/igniterealtime/openfire-restAPI-plugin/issues/205 (which has a fix in https://github.com/igniterealtime/openfire-restAPI-plugin/pull/232, but feels like it should be unnecessary)


guusdk (approved):
Looks good. Can you tempt you to create a JIRA issue for this and reword the commit message to reference the new JIRA number? That way, things get tracked for future-us.

Fishbowler:
Jira: :heavy_check_mark:

PR #3220, opened 2026-03-22
OF-2554: Link the docker-related documentation together

Fishbowler:
We've got 2 disparate bits of docker documentation. I considered unifying them, but they're addressing 2 different audiences, so I've added links in both directions.


guusdk (approved):
That's sensible, thanks!

PR #3234, opened 2026-03-27
OF-3222: Improve XEP-0359 stanza-id handling for one-on-one routing and archiving

guusdk:
Add stanza-id generation for direct messages delivered to local bare and full JIDs using recipient bare JID ownership.

Add sender-owned stanza-id for local-to-remote messages only after routing completes so outbound wire stanzas are not modified while post-processing interceptors can archive IDs.

Exclude local component subdomains from sender-owned stanza-id injection so existing MUC stanza-id handling in MultiUserChatServiceImpl remains authoritative and is not duplicated.


akrherz:
rebased to see what the rabbit thinks...

PR #3237, opened 2026-03-28
OF-3025: Refactor System Properties UI to remove inline scripts and improve table layout

MilanTyagi2004:
This PR refactors the System Properties administrative page to ensure better Content Security Policy (CSP) compliance and fixes several UI rendering issues.

**Changes included:**
* **Removed Inline JavaScript:** Extracted inline JS from `system-properties.jsp` into external files (`system-properties.js` and `list-pager.js`) to comply with strict CSP directives without requiring `'unsafe-inline'`.
* **Fixed Data Layout Bugs:** Corrected an issue where invisible whitespaces and line-breaks broke the edit/delete buttons, and moved dialog confirmation properties away from the `<body>` tag so they are no longer stripped out by SiteMesh.
* **Improved Table UX & CSS:** Added `overflow-wrap` to property names to prevent horizontal scrolling on small screens. For excessively long property values (like complex CSP headers), added `text-overflow: ellipsis` to restrict the height to a single line, combined with `title` attributes so users can view the full un-truncated values on hover.


guusdk (approved):
.

guusdk (changes requested):
Thank you for the effort, but I am still going to ask you to split this in separate PRs. The PR description now says that this is fixing three different issues. Please make those three different PRs, so that they can be evaluated one-by-one.

This PR is making changes to ListPager. That is used on many different pages, not just the one that is being changed in this PR. Have you verified that these changes do not break other pages?

I'm not sure why the change in LocalOutgoingServerSessionTest is included. Does that relate to this PR?

The PR contains a couple of commits that need to be fixed - they are not very concise. Perhaps you should squash some of them (but also split things up over different PR's, as suggested above).

The current PR can no longer be merged, probably because other PRs got merged that change the base of Openfire that this was created on. Please re-base on the latest HEAD of the Main branch and address any merge conflicts!

MilanTyagi2004:
@coderabbitai
check this pr

PR #3239, opened 2026-04-02
OF-2549: Fix JavaScript error and improve UX in MUC Create Permission page

MilanTyagi2004:
This PR resolves a JavaScript error (Uncaught TypeError: Cannot read properties of undefined (reading 'focus')) on the MUC room creation permissions page and enhances the user interface to be more responsive.

Root Cause
The error was caused by a scope conflict where a radio button tried to focus an input field (userJID) located in a separate form. Additionally, the target field was often missing from the DOM because it was only rendered by the server when the room creation restriction was already enabled.

Changes
 - Dynamic Visibility: Replaced server-side JSP conditional rendering with client-side CSS toggling (display: block/none) for the             "Allowed Users" section  .
 
 - Robust Event Handling: Introduced a toggleAllowedUsers() JavaScript function to manage visibility and safely apply focus only when the element is available.
 - Improved UX: The "Allowed Users" section now appears or disappears immediately upon selecting a policy, without requiring a page reload.
 - Code Cleanup: Removed redundant and broken inline onclick handlers from JID and Group selection fields that were using incorrect this.form references.

Verification
 - Verified that the "Allowed Users" section shows/hides correctly on toggle.
 - Confirmed that the JID input field is automatically focused when "Only specific..." is selected.
 - Manually tested that no JavaScript errors are thrown in the browser console.
 - Ensured that "Add" and "Delete" actions still persist correctly after the UI changes.

### DIFF
--- a/xmppserver/src/main/webapp/muc-create-permission.jsp
+++ b/xmppserver/src/main/webapp/muc-create-permission.jsp
@@ -1,7 +1,7 @@
 <%@ page contentType="text/html; charset=UTF-8" %>
 <%--
   -
-  - Copyright (C) 2004-2008 Jive Software, 2017-2025 Ignite Realtime Foundation. All rights reserved.
+  - Copyright (C) 2004-2008 Jive Software, 2017-2026 Ignite Realtime Foundation. All rights reserved.
   -
   - Licensed under the Apache License, Version 2.0 (the "License");
   - you may not use this file except in compliance with the License.
@@ -126,6 +126,7 @@
         if (add) {
             mucService.addUsersAllowedToCreate(allowedJIDs);
             mucService.setAllRegisteredUsersAllowedToCreate(allowAllRegisteredUsers);
+            mucService.setRoomCreationRestricted(true);
             // Log the event
             webManager.logEvent("updated MUC room creation permissions for service "+mucname, null);
             response.sendRedirect("muc-create-permission.jsp?addsuccess=true&mucname="+URLEncoder.encode(mucname, StandardCharsets.UTF_8));
@@ -151,6 +152,22 @@
 <meta name="subPageID" content="muc-perms"/>
 <meta name="extraParams" content="<%= "mucname="+URLEncoder.encode(mucname, StandardCharsets.UTF_8) %>"/>
 <meta name="helpPage" content="set_group_chat_room_creation_permissions.html"/>
+<script>
+    function toggleAllowedUsers() {
+        const isRestricted = document.getElementById('rb02').checked;
+        const container = document.getElementById('allowedUsersContainer');
+        const input = document.getElementById('userJIDtf');
+
+        if (isRestricted) {
+            container.style.display = 'block';
+            if (input) {
+                input.focus();
+            }
+        } else {
+            container.style.display = 'none';
+        }
+    }
+</script>
 </head>
 <body>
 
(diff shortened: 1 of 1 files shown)


xmppserver/src/main/webapp/muc-create-permission.jsp
```diff
@@ -1,7 +1,7 @@
 <%@ page contentType="text/html; charset=UTF-8" %>
 <%--
   -
-  - Copyright (C) 2004-2008 Jive Software, 2017-2025 Ignite Realtime Foundation. All rights reserved.
+  - Copyright (C) 2004-2026 Jive Software, 2017-2026 Ignite Realtime Foundation. All rights reserved.
```
guusdk: To be super pedantic: we cannot change the copyright years of Jive Software (or anyone else), only those for the Ignite Realtime Foundation!

xmppserver/src/main/webapp/muc-create-permission.jsp
```diff
@@ -255,7 +271,7 @@
         <p>
         <label for="userJIDtf"><fmt:message key="muc.create.permission.add_jid" /></label>
         <input type="text" name="userJID" size="30" maxlength="100" value="<%= (userJID != null ? userJID : "") %>"
```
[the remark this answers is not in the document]
guusdk: @MilanTyagi2004 can you please address this issue?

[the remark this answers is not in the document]
MilanTyagi2004: sure.

guusdk (changes requested):
(as mentioned in my previous commit: the setting no longer seems to be saved correctly).

MilanTyagi2004:
sure i will do this and squash commits into one commit.

MilanTyagi2004:
image


i fix this, now it is working fine.

MilanTyagi2004:
hi @guusdk,
once you review these changes,  then  i  will   squash  the commits into one.

MilanTyagi2004:
I squashes all the commits into one with all description.

MilanTyagi2004:
@Fishbowler if you are not busy could you please review this.

Fishbowler:
When I pick All Registered Users... 

image

...it's cleared on save

image

xmppserver/src/main/webapp/muc-create-permission.jsp
```diff
@@ -151,6 +152,22 @@
 <meta name="subPageID" content="muc-perms"/>
 <meta name="extraParams" content="<%= "mucname="+URLEncoder.encode(mucname, StandardCharsets.UTF_8) %>"/>
 <meta name="helpPage" content="set_group_chat_room_creation_permissions.html"/>
+<script>
+    function toggleAllowedUsers() {
+        const isRestricted = document.getElementById('rb02').checked;
```
Fishbowler, line 157: If we're gonna use the ID in JS, would renaming the ID to something more friendly make sense?

Fishbowler (changes requested):
Note the issue above. I suspect this might be pre-existing, but I haven't checked.
Added one nitpick below.

xmppserver/src/main/webapp/muc-create-permission.jsp
MilanTyagi2004: Thanks for review, 
i will look into this.

MilanTyagi2004:
@Fishbowler 
image


request changes are fixed, 
please review them when you have time.

Fishbowler (approved):
LGTM! I've pushed one commit to do a label rename to match an id rename.

guusdk:
Thanks!

PR #3240, opened 2026-04-02
OF-3237: Add exploratory Firebird database support

guusdk:
This is an exploratory change to assess the feasibility of adding Firebird (https://firebirdsql.org) as a supported database back-end for Openfire.

Firebird is an open-source relational database with standard SQL support, backed by the mature Jaybird JDBC driver. The change bundles that driver, adds Firebird to the database-type detection logic, provides an installation schema (adapted from the PostgreSQL schema, with Firebird-compatible types), a schema upgrade script for version 38 (to be able to run a database upgrade job in CI), a Docker Compose service definition for CI, install and upgrade CI jobs, a preset in the setup wizard, and a section in the database documentation.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -706,6 +706,117 @@ jobs:
           ./mvnw -B clean package exec:java -Dexec.args="$GITHUB_WORKSPACE"
 
 
+  firebird-install:
+    name: Test Firebird install script
+    needs: [build, should-do-database-tests, check_branch]
+    runs-on: ubuntu-latest
+    if: ${{ needs.should-do-database-tests.outputs.check == 'true' || needs.check_branch.outputs.is_publishable_branch == 'true'}}
+    steps:
+      - name: Checkout Openfire
+        uses: actions/checkout@v6
+      - name: Set up JDK 17 Zulu
+        uses: actions/setup-java@v5
+        with:
+          java-version: 17
+          distribution: zulu
+          cache: maven
+      - name: Restore mvn repo artifacts from build job
+        uses: actions/download-artifact@v8
+        with:
+          name: Maven Repository
+          path: ~/.m2/repository/org/igniterealtime/openfire/
+      - name: Set environment variables
+        run: |
+          echo "CONNECTION_STRING=jdbc:firebirdsql://localhost:3050//var/lib/firebird/data/openfire.fdb" >> $GITHUB_ENV
+          echo "CONNECTION_DRIVER=org.firebirdsql.jdbc.FBDriver" >> $GITHUB_ENV
+          echo "CONNECTION_USERNAME=sysdba" >> $GITHUB_ENV
+          echo "CONNECTION_PASSWORD=SecurePa55w0rd" >> $GITHUB_ENV
+          OPENFIREVSN=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
+          echo "OPENFIREVSN=$OPENFIREVSN" >> $GITHUB_ENV
+          echo "JAVA_HOME=$(echo $JAVA_HOME_17_X64)" >> $GITHUB_ENV
+      - name: Start database server with an empty database
+        run: |
+          mkdir olddb
+          docker compose -f ./build/ci/compose/firebird.yml up --wait
+      - name: Execute Openfire's database install script
+        run: |
+          pushd ./build/ci/updater
+          ./mvnw -B clean package exec:java -Dexec.args="$GITHUB_WORKSPACE"
+
+  # Firebird support was added recently and upgrade scripts are not available yet by virtue of support being new.
+  # Skip Firebird upgrade execution until the target upgrade directory exists. This step can be removed completely
+  # after things have matured.
+  firebird-upgrade-precheck:
+    name: Check Firebird upgrade prerequisites
+    needs: [should-do-database-tests, check_branch]
+    runs-on: ubuntu-latest
+    if: ${{ needs.should-do-database-tests.outputs.check == 'true' || needs.check_branch.outputs.is_publishable_branch == 'true'}}
+    outputs:
+      upgrade-dir-exists: ${{ steps.check-firebird-upgrade-dir.outputs.exists }}
+    steps:
+      - name: Checkout Openfire
+        uses: actions/checkout@v6
+      - name: Check if Firebird upgrade directory exists
+        id: check-firebird-upgrade-dir
+        run: |
+          if [ -d distribution/src/database/upgrade/39 ]; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "Skipping firebird-upgrade: Firebird support is new and upgrade scripts are not available yet."
+          fi
+
+
+  firebird-upgrade:
+    name: Test Firebird upgrade scripts
+    needs: [build, should-do-database-tests, check_branch, firebird-upgrade-precheck]
+    runs-on: ubuntu-latest
+    if: ${{ needs.should-do-database-tests.outputs.check == 'true' || needs.check_branch.outputs.is_publishable_branch == 'true' }}
+    steps:
+      - name: Skip Firebird upgrade test when scripts are not available yet
+        if: ${{ needs.firebird-upgrade-precheck.outputs.upgrade-dir-exists != 'true' }}
+        run: 'echo "Skipping firebird-upgrade: Firebird support is new and upgrade scripts are not available yet."'
+      - name: Checkout Openfire
+        if: ${{ needs.firebird-upgrade-precheck.outputs.upgrade-dir-exists == 'true' }}
+        uses: actions/checkout@v6
+      - name: Set up JDK 17 Zulu
+        if: ${{ needs.firebird-upgrade-precheck.outputs.upgrade-dir-exists == 'true' }}
+        uses: actions/setup-java@v5
+        with:
+          java-version: 17
+          distribution: zulu
+          cache: maven
+      - name: Restore mvn repo artifacts from build job
+        if: ${{ needs.firebird-upgrade-precheck.outputs.upgrade-dir-exists == 'true' }}
+        uses: actions/download-artifact@v8
+        with:
+          name: Maven Repository
+          path: ~/.m2/repository/org/igniterealtime/openfire/
+      - name: Set environment variables
+        if: ${{ needs.firebird-upgrade-precheck.outputs.upgrade-dir-exists == 'true' }}
+        run: |
+          echo "CONNECTION_STRING=jdbc:firebirdsql://localhost:3050//var/lib/firebird/data/openfire.fdb" >> $GITHUB_ENV
+          echo "CONNECTION_DRIVER=org.firebirdsql.jdbc.FBDriver" >> $GITHUB_ENV
+          echo "CONNECTION_USERNAME=sysdba" >> $GITHUB_ENV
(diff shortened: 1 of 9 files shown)


xmppserver/src/main/webapp/setup/setup-datasource-standard.jsp
```diff
@@ -222,6 +222,7 @@
     presets.add(new String[]{"Oracle","oracle.jdbc.driver.OracleDriver","jdbc:oracle:thin:@HOSTNAME:1521:SID"});
     presets.add(new String[]{"Microsoft SQL Server (legacy)","net.sourceforge.jtds.jdbc.Driver","jdbc:jtds:sqlserver://HOSTNAME/DATABASENAME;appName=Openfire"});
     presets.add(new String[]{"PostgreSQL","org.postgresql.Driver","jdbc:postgresql://HOSTNAME:5432/DATABASENAME"});
+    presets.add(new String[]{"Firebird","org.firebirdsql.jdbc.FBDriver","jdbc:firebirdsql://HOSTNAME:3050/ABSOLUTE_PATH_TO_DATABASE_FILE"});
```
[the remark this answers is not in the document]
guusdk: I'll amend this change to the original commit.

documentation/database.html
```diff
@@ -333,6 +334,65 @@ <h3>Further Help</h3>
...
+            <p>
+                The JDBC driver for Firebird is bundled with Openfire, so you do not need to download and install a
+                separate driver.
+            </p>
+            <p>
+                In the Openfire web-based setup tool, use the following values:
+            </p>
+            <ul style="list-style: none">
+                <li><span class="short-tab">driver:</span> <code>org.firebirdsql.jdbc.FBDriver</code></li>
+                <li><span class="short-tab">server:</span> <code>jdbc:firebirdsql://HOSTNAME:PORT/ABSOLUTE_PATH_TO_DATABASE_FILE</code></li>
```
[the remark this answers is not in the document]
guusdk: I'll amend this change to the original commit.

build/ci/updater/src/main/java/com/igniterealtime/openfire/updaterunner/Main.java
```diff
@@ -49,6 +49,7 @@ public static void main(String[] args) throws Exception
         "Oracle","oracle.jdbc.driver.OracleDriver","jdbc:oracle:thin:@HOSTNAME:1521:SID"
         "Microsoft SQL Server (legacy)","net.sourceforge.jtds.jdbc.Driver","jdbc:jtds:sqlserver://HOSTNAME/DATABASENAME;appName=Openfire"
         "PostgreSQL","org.postgresql.Driver","jdbc:postgresql://HOSTNAME:5432/DATABASENAME"
+        "Firebird","org.firebirdsql.jdbc.FBDriver","jdbc:firebirdsql://HOSTNAME:3050/ABSOLUTE_PATH_TO_DATABASE_FILE"
```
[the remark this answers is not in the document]
guusdk: I'll amend this change to the original commit.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -603,6 +603,117 @@ jobs:
...
+      - name: Checkout Openfire
+        uses: actions/checkout@v6
+      - name: Check if Firebird upgrade directory exists
+        id: check-firebird-upgrade-dir
+        run: |
+          if [ -d distribution/src/database/upgrade/39 ]; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "Skipping firebird-upgrade: Firebird support is new and upgrade scripts are not available yet."
```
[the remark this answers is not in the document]
guusdk: This is a job that's added to make sure that upgrade scripts that are added in the future will automatically be tested by CI (this attempts to make these CI jobs consistent with those of other databases).

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -603,6 +603,117 @@ jobs:
...
+          distribution: zulu
+          cache: maven
+      - name: Restore mvn repo artifacts from build job
+        uses: actions/download-artifact@v8
+        with:
+          name: Maven Repository
+          path: ~/.m2/repository/org/igniterealtime/openfire/
+      - name: Set environment variables
+        run: |
+          echo "CONNECTION_STRING=jdbc:firebirdsql://localhost:3050//var/lib/firebird/data/openfire.fdb" >> $GITHUB_ENV
```
mrotteveel: For ease of use, if you add `FIREBIRD_CONF_DatabaseAccess: Restrict /var/lib/firebird/data` in your docker compose file, you can also use only the filename instead of the full path, i.e. `jdbc:firebirdsql://localhost:3050/openfire.fdb`

(NB: prefix `jdbc:firebird:` is also supported, and port (`:3050`) is already the default)

distribution/src/database/openfire_firebird.sql
```diff
@@ -0,0 +1,441 @@
+-- In Firebird, TEXT is represented as BLOB SUB_TYPE TEXT CHARACTER SET UTF8.
```
mrotteveel, line 1: If it makes it easier, you could also define a [domain](https://firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-datatypes-custom.html#fblangref50-datatypes-custom) called TEXT:

```sql
create domain TEXT as blob sub_type text character set UTF8;
```

distribution/src/database/openfire_firebird.sql
```diff
@@ -0,0 +1,441 @@
...
+
+CREATE TABLE ofUser
+(
+    username          VARCHAR(64) NOT NULL,
```
mrotteveel, line 5: Defining VARCHAR or CHAR without specifying a character set will mean its actual character set depends on the database default.

I recommend that you specify the actual desired character set so you don't get surprises later on, or when supporting customers with a different character set (or even NONE) as the default.

.github/workflows/continuous-integration-workflow.yml
guusdk: Ah, thanks! Not hardcoding a container filesystem path in the JDBC string would be a nice improvement - I'll add that!

I don't mind about explicitly using default values. It provides some clarity.

distribution/src/database/openfire_firebird.sql
guusdk: I think it's fine to keep using Firebird's native terminology. Unless there's a benefit that I'm missing, I'd prefer to keep it as-is.

distribution/src/database/openfire_firebird.sql
guusdk: This, sadly, is a historical omission in our application. I'm not arguing against this, but we should probably take this on as a larger effort. I've created a new ticket in our bug tracker for this: https://igniterealtime.atlassian.net/browse/OF-3298

guusdk:
I have rebased this PR to the current HEAD of the Main branch, and added a new commit that addresses some of the suggestions made by @mrotteveel (thanks!)

distribution/src/database/openfire_firebird.sql
mrotteveel: > I think it's fine to keep using Firebird's native terminology. Unless there's a benefit that I'm missing, I'd prefer to keep it as-is.

For Firebird itself, it doesn't matter; I was more suggesting it in case if you want to use common terminology across different engines.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -706,6 +706,117 @@ jobs:
...
+        uses: actions/checkout@v6
+      - name: Check if Firebird upgrade directory exists
+        id: check-firebird-upgrade-dir
+        run: |
+          if [ -d distribution/src/database/upgrade/39 ]; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "Skipping firebird-upgrade: Firebird support is new and upgrade scripts are not available yet."
+          fi
```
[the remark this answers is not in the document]
guusdk: I think we need this, to skip executing the 'upgrade' CI while we have no upgrades available? Firebird support gets introduced when the rest of the databases are on version 38. The earliest possible version of Firebird support will thus be script version 38, too. Running the CI database upgrade checks only makes sense after a 39 (or later) has been added to the repository. Until then, the update step should be skipped.

guusdk:
> **Actionable comments posted: 1**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (1)
> 
> > xmppserver/src/main/java/org/jivesoftware/database/DbConnectionManager.java (1)> `1215-1249`: _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> > > **Fix Firebird paging: don’t default to `LIMIT` SQL**
> > > `DbConnectionManager.DatabaseType.firebird` falls through `getResultSetLimitKeyword()` to `ResultSetLimitKeyword.LIMIT`, and `DefaultPubSubPersistenceProvider.getPublishedItems(...)` then prepares `LOAD_LAST_ITEMS_LIMIT`, which hardcodes `... LIMIT ?`. Firebird doesn’t support MySQL/PostgreSQL-style `LIMIT`; it uses `FETCH FIRST ... ROWS ONLY` / `OFFSET ... FETCH` instead, so this will generate invalid SQL for Firebird.
> > > 
> > > Suggested fix
> > > ```diff
> > >          public ResultSetLimitKeyword getResultSetLimitKeyword()
> > >          {
> > >              return switch (this) {
> > >                  case sqlserver -> ResultSetLimitKeyword.TOP;
> > > -                case oracle, db2 -> ResultSetLimitKeyword.FETCH_FIRST;
> > > +                case oracle, db2, firebird -> ResultSetLimitKeyword.FETCH_FIRST;
> > >                  default -> ResultSetLimitKeyword.LIMIT;
> > >              };
> > >          }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@xmppserver/src/main/java/org/jivesoftware/database/DbConnectionManager.java`
> > > around lines 1215 - 1249, DatabaseType.firebird currently falls through to
> > > ResultSetLimitKeyword.LIMIT which produces MySQL-style "LIMIT" SQL; update
> > > getResultSetLimitKeyword() so firebird returns the appropriate keyword (e.g.,
> > > ResultSetLimitKeyword.FETCH_FIRST) instead of DEFAULT/LIMIT so callers like
> > > DefaultPubSubPersistenceProvider.getPublishedItems(...) and the
> > > LOAD_LAST_ITEMS_LIMIT path generate "FETCH FIRST ... ROWS ONLY"/"OFFSET ...
> > > FETCH" SQL valid for Firebird; change the switch in
> > > DatabaseType.getResultSetLimitKeyword() to include case firebird ->
> > > ResultSetLimitKeyword.FETCH_FIRST.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> 🪄 Autofix (Beta)
> 
> 
> ℹ️ Review info

That's valuable feedback, thanks! This mechanism is new (it was added after this PR was originally created). I've now added a commit to this PR to add Firebird support.

distribution/src/database/openfire_firebird.sql
guusdk: I appreciate the sentiment, but I'm afraid that ship has sailed. Back in 2004. :)

akrherz (approved):

PR #3277, opened 2026-04-15
OF-3252 & OF-3253: Upgrade various libraries

guusdk:
This PR updates the following libraries:
- Apache's commons-dbcp2 from 2.9.0 to 2.14.0
- Apache's commons-pool2 from 2.9.0 to 2.13.1
- org.jsmpp:jsmpp from 2.3.10 to 3.0.2

All of these are maintenance releases, each of them bringing about half a decade of changes - mostly minor ones.

There was enough overlap between the three for me to include them all in one PR.

This also replaces usage of the old `JiveGlobals` with `SystemProperty` where applied in the SMS service.

### DIFF
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -1509,6 +1509,21 @@ system_property.ssl.certificates.expirycheck.notify-admins=Determines if the adm
 system_property.ssl.certificates.expirycheck.frequency=How often to check for (nearly) expired TLS certificates.
 system_property.ssl.certificates.expirycheck.warning-period=How long (in hours) before a TLS certificate will expire for a warning is to be sent out to admins.
 
+system_property.sms.smpp.connections.maxAmount=The maximum amount of SMPP connections that the SMS Service will maintain.
+system_property.sms.smpp.connections.idleMillis=Time after which idle MPPP connections used by the SMS Service are allowed to be evicted.
+system_property.sms.smpp.source.ton=The type-of-number value for the source of SMS messages.
+system_property.sms.smpp.source.npi=The number-plan-indicator value for the source of SMS messages.
+system_property.sms.smpp.source.address=The source address of SMS messages.
+system_property.sms.smpp.destination.ton=The type-of-number value for the destination of SMS messages.
+system_property.sms.smpp.destination.npi=The number-plan-indicator value for the destination of SMS messages.
+system_property.sms.smpp.host=The host name of your SMPP Server or SMSC, i.e. smsc.example.org.
+system_property.sms.smpp.port=The port on which the SMSC is listening.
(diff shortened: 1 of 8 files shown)


i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -1509,6 +1509,21 @@ system_property.ssl.certificates.expirycheck.notify-admins=Determines if the adm
 system_property.ssl.certificates.expirycheck.frequency=How often to check for (nearly) expired TLS certificates.
 system_property.ssl.certificates.expirycheck.warning-period=How long (in hours) before a TLS certificate will expire for a warning is to be sent out to admins.
 
+system_property.sms.smpp.connections.maxAmount=The maximum amount of SMPP connections that the SMS Service will maintain.
+system_property.sms.smpp.connections.idleMillis=Time after which idle MPPP connections used by the SMS Service are allowed to be evicted.
```
akrherz: `MPPP` ?

xmppserver/src/main/java/org/jivesoftware/util/SmsService.java
```diff
@@ -147,10 +288,10 @@ public void sendImmediately( String message, String recipient ) throws Exception
      * Checks if an exception in the chain of the provided throwable contains a 'command status' that can be
      * translated in a somewhat more helpful error message.
      *
-     * The list of error messages was taken from http://www.smssolutions.net/tutorials/smpp/smpperrorcodes/
+     * The list of error messages was taken from <a href="http://www.smssolutions.net/tutorials/smpp/smpperrorcodes/">http://www.smssolutions.net/tutorials/smpp/smpperrorcodes/</a>
```
akrherz: Can this be https instead?

akrherz (approved):
This seems fine to me, but aren't there some cleanups/fixes in the Sms logic that perhaps should be covered by a Jira?

i18n/src/main/resources/openfire_i18n.properties
guusdk: Oops. Fixed.

xmppserver/src/main/java/org/jivesoftware/util/SmsService.java
guusdk: Yes. I've now done that.

guusdk:
To be honest, I don't really see any that warrant a ticket. There should not be functional changes, and apart from the move from JiveGlobals to SystemProperty (which possibly warrants a ticket, but meh), the changes are either related to the library upgrade (for which I had to change a few calls to methods in those libraries that were deprecated in the newer versions), or are minor documentation or typo fixes. Did I miss something?

akrherz:
> I don't really see any that warrant a ticket.

My bad, there are not such changes that I thought I spotted.

PR #3291, opened 2026-04-24
Fix S2S dialback not offered in post-TLS features when self-signed certs accepted

milne1282:
ServerDialback.isEnabledForSelfSigned() reads the legacy property xmpp.server.certificate.accept-selfsigned (TLS_ACCEPT_SELFSIGNED_CERTS), but ConnectionListener.setAcceptSelfSignedCertificates() — which backs the Admin Console UI checkbox — writes to the per-connection-type prefixed property xmpp.socket.ssl.certificate.accept-selfsigned. These diverged when per-type prefixed properties were introduced but the readers in ServerDialback were not updated.

The consequence is that after a successful TLS handshake with a self-signed peer certificate, LocalIncomingServerSession.getAvailableStreamFeatures() calls isEnabledForSelfSigned(), gets false, and omits <db:dialback> from the post-TLS <stream:features>. The connecting server then finds no usable authentication mechanism and closes the connection. This causes all encrypted S2S connections to fail silently when the only option is dialback, even if the operator has enabled self-signed certificate acceptance via the UI.

Fix: check both property names in isEnabledForSelfSigned(), preserving full backwards compatibility with installations that have the legacy property set.

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/ClientTrustManager.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/ClientTrustManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004-2008 Jive Software, 2017-2025 Ignite Realtime Foundation. All rights reserved.
+ * Copyright (C) 2004-2008 Jive Software, 2017-2026 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -173,6 +173,8 @@ public class ClientTrustManager implements X509TrustManager {
 
         loadCRL();
 
+        final ConnectionListener connectionListener = XMPPServer.getInstance().getConnectionManager().getListener(ConnectionType.SOCKET_C2S, false);
+
         boolean verify = JiveGlobals.getBooleanProperty("xmpp.client.certificate.verify", true);
         if (verify) {
             int nSize = x509Certificates.length;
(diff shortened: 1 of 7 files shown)


guusdk:
Oh, interesting. Thanks for this @milne1282 ! Did you happen to check if _other_ connection types suffer from a similar issue, triggered by the same type of divergence after the prefixed properties were introduced?

milne1282:
Hey @guusdk I have not checked other properties as I was focused on an implementation topology that this affected and have not made a more general search

guusdk:
@milne1282 I've raised a ticket for this change in https://igniterealtime.atlassian.net/browse/OF-3259

I've also added a couple of commits to this PR, that addresses similar issues with certificate-verification related properties. I'd love your feedback on those. Please refer to the individual commit messages for details.

guusdk:
> ⚠️ Outside diff range comments (1)
> 
> > xmppserver/src/main/java/org/jivesoftware/openfire/spi/ConnectionListener.java (1)> `747-762`: _⚠️ Potential issue_ | _🔴 Critical_ | _⚡ Quick win_
> > > **Bug: `oldValue` retrieves from wrong method.**
> > > Line 749 calls `verifyCertificateValidity()` but should call `acceptSelfSignedCertificates()`. This causes the comparison on line 754 to be incorrect, and the log messages will report wrong old/new values for the self-signed certificate acceptance setting.

This code review finding is something that is already being fixed in a different PR: https://github.com/igniterealtime/Openfire/pull/3308

guusdk:
I've rebased this to the latest HEAD of the Main branch, which has some improvements to the CI workflow. Hopefully, tests are more stable now.

PR #3299, opened 2026-05-04
Fix CI test flakes

Fishbowler:
Herein lie attempts to fix the recent test flakes in CI. I believe them all to be reasonable changes or reasonable compromises.

### DIFF
--- a/.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
+++ b/.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
@@ -57,7 +57,7 @@ public class ConnectivityTest
         }
 
         // Verify result.
-        Assertions.assertEquals(iterations, result, "Expected all BOSH connection attempts to result in an authenticated session, but not all did.");
+        Assertions.assertTrue(iterations >= result * .9, "Expected all BOSH connection attempts to result in an authenticated session, but not all did."); // FIXME make this an _exact_ comparison after stability of the BOSH connectivity is achieved.
     }
 
     /**
--- a/.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
+++ b/.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
@@ -106,6 +106,6 @@ public class ConnectivityTest
         }
 
         // Verify result.
-        Assertions.assertTrue(iterations >= result * .9, "Expected all TCP connection attempts to result in an authenticated session, but not all did."); // FIXME make this an _exact_ comparison after stability of the BOSH connectivity is achieved.
+        Assertions.assertEquals(iterations, result, "Expected all TCP connection attempts to result in an authenticated session, but not all did.");
     }
 }
--- a/xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalOutgoingServerSessionTest.java
+++ b/xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalOutgoingServerSessionTest.java
@@ -119,7 +119,6 @@ public class LocalOutgoingServerSessionTest
     @BeforeEach
     public void setUpEach() throws Exception {
         final XMPPServer xmppServer = Fixtures.mockXMPPServer();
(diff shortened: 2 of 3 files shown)


.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
```diff
@@ -57,7 +57,7 @@ public void testBoshConnections() throws Exception
         }
 
         // Verify result.
-        Assertions.assertEquals(iterations, result, "Expected all BOSH connection attempts to result in an authenticated session, but not all did.");
+        Assertions.assertTrue(iterations >= result * .9, "Expected all BOSH connection attempts to result in an authenticated session, but not all did."); // FIXME make this an _exact_ comparison after stability of the BOSH connectivity is achieved.
```
guusdk, line 60: I think the original commit simply applied the fix to the wrong place. I don't think that the TCP tests need tolerance. Maybe _remove_ it there?

guusdk, line 60: I think the original commit simply applied the fix to the wrong place. I don't think that the TCP tests need tolerance. Maybe _remove_ it there?

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
```diff
@@ -81,6 +81,10 @@
 @ExtendWith(MockitoExtension.class)
 public class LocalIncomingServerSessionTest
 {
+    // Increased from 2s to avoid false failures on slow CI runners (e.g. Windows), where session state propagation takes longer.
+    // This mirrors the `xmpp.server.session.initialise-timeout` in LocalOutgoingServerSessionTest
+    private static final int ASYNC_ASSERT_TIMEOUT_SECONDS = 10;
```
guusdk: I don't expect this to be effective, other than causing the tests that fail to fail more slowly. Only last month, I've introduced the 2 second delay (prior to that, there wasn't any delay). There wasn't a notable improvement between 'no delay' and '2 second delay' - I'm not expecting bumping that from 2 to 10 to make a difference.

guusdk (commented):
I don't really like the blunt retries - it hides a problem to the point where having the test may not reliably prove much any more.

I was considering to simply disable these tests, until we can get them more stable.

Happy to test these changes to see if they make a difference, but meh - I don' have high hopes.

guusdk (commented):
I don't really like the blunt retries - it hides a problem to the point where having the test may not reliably prove much any more.

I was considering to simply disable these tests, until we can get them more stable.

Happy to test these changes to see if they make a difference, but meh - I don' have high hopes.

xmppserver/src/test/java/org/jivesoftware/openfire/session/LocalIncomingServerSessionTest.java
Fishbowler: I agree. It was my least favourite of the changes. Will drop this commit.

Fishbowler:
> I don't really like the blunt retries - it hides a problem to the point where having the test may not reliably prove much any more.

Me neither, but in this case, it feels like a necessary evil, and didn't feel altogether all that evil....

Retries are added to 2 places:
- on the SSL socket layer, where the retry is dealing with a test setup race condition. We've already got the same line in RemoteReceivingServerDummy
- for a single retry on the outgoing test, where the expectation is testing combinations of connection logic for an outcome. If something is broken, it's likely always broken. Whilst it might fail for other reasons, it's unlikely to "accidentally pass". It's likely only hiding some test-only warm-up bug.

.github/actions/connectivitytests-action/conntest/src/test/java/ConnectivityTest.java
Fishbowler: Done

Fishbowler: Duplicated comment for a duplicated fix? :)

guusdk (approved):

PR #3320, opened 2026-05-08
OF-3274: Bump org.glassfish.jaxb:jaxb-runtime from 2.3.3 to 2.3.9

guusdk:
jaxb-runtime 4 is Jakarta JAXB (jakarta.xml.bind.), which is a package-rename break from javax.xml.bind.. so a direct upgrade to 4.0.x is not safe for the current code.

We should keep JAXB on the 2.3.x line for now (same javax namespace)


Fishbowler (approved):
Changelog looks safe

PR #3321, opened 2026-05-08
Cache conversations.apk in tests

Fishbowler:
Sometimes f-droid has comms flakes. We also shouldn't burn their bandwidth unnecessarily.


guusdk (approved):
I think this is OK as-is. The retry stuff suggested by rabbit would probably make it even more robust?

PR #3326, opened 2026-05-09
OF-3025: Replace inline <style> element in system-properties.jsp

MilanTyagi2004:
- Add inlineJsDisabled flag to ListPager to conditionally suppress inline onclick/onchange attributes and pageFunctions output (defaults to false, so all existing pages are unaffected)
- Extract page-specific JavaScript to js/system-properties.js using event delegation instead of inline onclick/onchange handlers
- Extract ListPager client-side pagination logic to js/list-pager.js
- Replace inline <script> block with external script includes
- Use data-* attributes to pass property metadata to JavaScript
- Use c:set/c:out for safely passing localized confirm messages
- Replace scriptlet getHiddenFields() with EL expression
- Improve table layout: use word-wrap for property names, add title tooltips for truncated values, reduce search input sizes


MilanTyagi2004:
Hi @guusdk,
this PR is the replace of old PR for the same issue OF-3025.
when this is reviewed and approved i will close the previous PR.

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
```diff
@@ -228,12 +229,30 @@ public boolean isSortDescending() {
...
+        this.inlineJsDisabled = inlineJsDisabled;
+    }
+
     /**
      * @return a string that contains HTML for selecting the page size
      */
     public String getPageSizeSelection() {
         final StringBuilder sb = new StringBuilder();
-        sb.append(String.format("<select name='%s' onchange='return setPageSize(this.value);'>", REQUEST_PARAMETER_KEY_PAGE_SIZE));
+        sb.append(String.format("<select name='%s'%s>", REQUEST_PARAMETER_KEY_PAGE_SIZE, inlineJsDisabled ? "" : " onchange='return setPageSize(this.value);'"));
```
Fishbowler: Could this be made more readable?
Something like this?

```suggestion
        sb.append(String.format("<select name='%s'", REQUEST_PARAMETER_KEY_PAGE_SIZE));
        if( !inlineJsDisabled ){
            sb.append(" onchange='return setPageSize(this.value);'");
        }
        sb.append(">");
```

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
```diff
@@ -281,9 +300,9 @@ private void appendPageLink(final StringBuilder sb, final int pageToLink) {
             cssClass = "";
         }
 
-        sb.append(String.format("\n<a href='%s' onclick='return jumpToPage(%d)'%s>%s</a>",
+        sb.append(String.format("\n<a href='%s'%s%s>%s</a>",
             String.format("?%s=%d", REQUEST_PARAMETER_KEY_CURRENT_PAGE, pageToLink),
-            pageToLink,
+            inlineJsDisabled ? "" : String.format(" onclick='return jumpToPage(%d)'", pageToLink),
```
Fishbowler: Same comment about readability and maintainability here.
We're already using a StringBuilder - we can split it up into simple logical steps.

xmppserver/src/main/webapp/system-properties.jsp
```diff
@@ -30,19 +31,20 @@
...
         }
 
         .valueColumn {
-            text-overflow: ellipsis;
-            overflow: hidden;
-            white-space: nowrap;
             max-width: 250px;
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
```
Fishbowler, line 47: What's this change for? Does the reordering do things in CSS?

xmppserver/src/main/webapp/js/system-properties.js
```diff
@@ -0,0 +1,148 @@
+/*
+ * Copyright (C) 2025 Ignite Realtime Foundation. All rights reserved.
```
Fishbowler:
```suggestion
 * Copyright (C) 2026 Ignite Realtime Foundation. All rights reserved.
```

xmppserver/src/main/webapp/js/list-pager.js
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright (C) 2025 Ignite Realtime Foundation. All rights reserved.
```
Fishbowler:
```suggestion
 * Copyright (C) 2026 Ignite Realtime Foundation. All rights reserved.
```

Fishbowler (commented):
Some initial comments. I haven't done any testing yet.

xmppserver/src/main/webapp/system-properties.jsp
MilanTyagi2004: No functional change here, I only reordered the CSS properties for formatting/readability. The behavior remains the same.

xmppserver/src/main/java/org/jivesoftware/util/ListPager.java
MilanTyagi2004: sure i will go through this and make it more readable.

MilanTyagi2004:
hi @Fishbowler
i have implemented the change could you please review them.

guusdk (changes requested):
To be frank: I don't think this is the way to go. I do not like the added complexity of the new `inlineJsDisabled` flag. This gives us a 'sometimes' fix. The solution should simply always work.

One feature that the ListPager has, is that it automatically generates forms and corresponding functionality, after it was configured by a number of fields. With this change, it appears that users need to use the same fields in different places. This introduces a need to create duplicated data, and takes away from the value of the ListPager.

Having hard-coded snippets of javascript per page that uses the ListPager goes directly against the purpose of ListPager, too.

I realize that this is far from an easy task, but I'd like for us to find a way to remove these hard-coded code templates. The purpose of OF-3025 is to prevent _inline_ elements. Can we replace them by a call to a servlet-backed URL that dynamically generates the javascript that is correct for the particular use case?

MilanTyagi2004:
Thanks for the feedback. I agree that introducing `inlineJsDisabled` adds complexity and does not fully solve the problem. I also understand your concern about requiring pages to duplicate configuration and provide hard-coded JavaScript, as that reduces the value of `ListPager`.

I'll take another look at the implementation and explore a solution where `ListPager` can always generate the required behavior without page-specific JavaScript. The servlet-backed approach you suggested sounds promising, and I'll investigate whether we can use that to generate the appropriate JavaScript dynamically for each use case while keeping the existing configuration model intact.

PR #3368, opened 2026-06-03
Group dependabot PRs

akrherz:


akrherz:
I am unsure this is right, we want to group by maven artifact name and not lump them all together, hmm...

Fishbowler (approved):
Blind approve. This doesn't look malicious.

PR #3376, opened 2026-06-04
OF-31: Preserve conference service context in room creation flow

MilanTyagi2004:
## Summary

Fixes OF-31 by preserving the current conference service context when creating a new MUC room.

Previously, when an administrator navigated to **Create New Room** from a service-specific MUC administration page, the room creation form defaulted to the alphabetically first conference service because the selected service context was lost during navigation.

This change derives the active conference service from the originating request and forwards it through the room creation flow, allowing the room creation form to preselect the correct service and load the appropriate service-specific defaults.

## Root Cause

The **Create New Room** sidebar entry is a top-level navigation item generated from a static URL. As a result, service-specific request parameters (`mucname`, `mucName`, or `roomJID`) were not carried to the room creation flow.

When `muc-room-edit-form.jsp` was opened without service context, it defaulted to the first available conference service.

## Changes

* Updated `muc-room-create.jsp` to derive the active conference service from the originating request and forward it to the room creation form.
* Updated `muc-room-edit-form.jsp` to:

  * Support both `mucName` and `mucname` parameters.
  * Resolve the target conference service from the provided context.
  * Fall back to the existing default behavior when no valid service can be determined.

## Backward Compatibility

No behavior changes occur when service context is unavailable.

The room creation form continues to default to the first available conference service in the following cases:

* No service context is provided.
* The originating request cannot be resolved.
* The resolved service is invalid or no longer exists.

## Testing

Verified with multiple conference services configured:

1. Open a service-specific MUC administration page (e.g. `conference2`).
2. Click **Create New Room**.
3. Confirm that:

   * The selected conference service defaults to `conference2`.
   * Service-specific room defaults are loaded correctly.
4. Verify that room creation succeeds under the selected service.
5. Verify that navigation from non-service-specific pages continues to use the existing default service selection behavior.

### DIFF
--- a/xmppserver/src/main/webapp/muc-room-create.jsp
+++ b/xmppserver/src/main/webapp/muc-room-create.jsp
@@ -16,7 +16,11 @@
   - limitations under the License.
 
 --%>
-
+<%@ page import="java.util.function.BiFunction" %>
+<%@ page import="java.util.regex.Pattern" %>
+<%@ page import="java.util.regex.Matcher" %>
+<%@ page import="static java.nio.charset.StandardCharsets.UTF_8" %>
+<%@ page import="java.net.URLEncoder,java.net.URLDecoder" %>
 <%
     // OF-31: To select the current conference service automatically on room creation,
     // we extract the active service context from the HTTP Referer header. This allows us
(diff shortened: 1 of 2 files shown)


xmppserver/src/main/webapp/muc-room-create.jsp
```diff
@@ -18,6 +18,96 @@
...
+            : referrer.substring(idx + 8);
+
+        try {
+            String decodedVal = java.net.URLDecoder.decode(
+                val,
+                java.nio.charset.StandardCharsets.UTF_8.name()
+            );
+            String encodedVal = java.net.URLEncoder.encode(
+                decodedVal,
+                java.nio.charset.StandardCharsets.UTF_8.name()
```
stokito: nitpick: shorter `URLEncoder.encode(decodedVal, StandardCharsets.UTF_8)
feel free to static import the constant `URLEncoder.encode(decodedVal, UTF_8)

PR #3381, opened 2026-06-05
OF-2010 Specify port protocol for Media Proxy

MilanTyagi2004:
### Description

The Media Proxy service in Openfire opens UDP listeners, but the corresponding port configuration fields in the Admin Console do not explicitly indicate that these ports use the UDP protocol. This can cause confusion when configuring firewall or network rules.

This change updates the labels of the Media Proxy port configuration fields to clearly indicate that they are UDP ports.

### Changes

Updated the default English localization resource file (`openfire_i18n.properties`):

* `mediaproxy.form.minport`: `Port Range Min` → `UDP Port Range Min`
* `mediaproxy.form.maxport`: `Port Range Max` → `UDP Port Range Max`
* `mediaproxy.form.echoport`: `Echo Test Port` → `UDP Echo Test Port`

### Testing

* Verified that the project compiles successfully using `mvn compile`.
* Verified that the updated labels are displayed correctly on the Media Proxy settings page in the Admin Console.

### Notes

This is a UI text-only change. No functional behavior has been modified.

OF-2010


MilanTyagi2004:
Hi @guusdk,
changes are unrelated to this CI/CD failure.
please trigger rerun.

PR #3382, opened 2026-06-05
OF-2710: Replace two group-member queries with one in DefaultGroupPro…

MilanTyagi2004:
## Description

`DefaultGroupProvider` currently performs two database queries when loading a group's users:

- One query to retrieve regular members.
- One query to retrieve administrators.

As both user types are stored in the same table (`ofGroupUser`) and differ only by the value of the `administrator` column, the data can be retrieved using a single query and categorized in memory.

This PR optimizes group loading by replacing the two queries with a single query that retrieves both members and administrators.

### Changes

- Added a new query constant, `LOAD_MEMBERS_AND_ADMINS`, that retrieves both `username` and `administrator` values from `ofGroupUser`.
- Removed the separate `LOAD_MEMBERS` and `LOAD_ADMINS` query constants.
- Replaced the `getMembers(String, boolean)` helper method with `loadMembersAndAdmins(String, Collection<JID>, Collection<JID>)`.
- Updated group-loading logic to populate member and administrator collections from a single result set.
- Updated `createGroup` and `getGroup` to use the new helper method.

### Benefits

- Reduces the number of database queries required to load group membership data from **two to one**.
- Lowers database workload and query overhead.
- Preserves existing behavior and APIs.
- Requires no schema or architectural changes.

### Compilation

```bash
mvn compile

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
@@ -57,10 +57,8 @@ public class DefaultGroupProvider extends AbstractGroupProvider {
     private static final String DELETE_GROUP =
         "DELETE FROM ofGroup WHERE groupName=?";
     private static final String GROUP_COUNT = "SELECT count(*) FROM ofGroup";
-     private static final String LOAD_ADMINS =
-        "SELECT username FROM ofGroupUser WHERE administrator=1 AND groupName=? ORDER BY username";
-    private static final String LOAD_MEMBERS =
-        "SELECT username FROM ofGroupUser WHERE administrator=0 AND groupName=? ORDER BY username";
+    private static final String LOAD_MEMBERS_AND_ADMINS =
+        "SELECT username, administrator FROM ofGroupUser WHERE groupName=? ORDER BY username";
     private static final String LOAD_GROUP =
         "SELECT description FROM ofGroup WHERE groupName=?";
     private static final String REMOVE_USER =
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java [hunk 1]
```diff
@@ -105,9 +103,9 @@ public Group createGroup(String name) throws GroupNameInvalidException, GroupAlr
...
         }
 
-        // TODO Replace these two database queries with one. OF-2710
-        Collection<JID> members = getMembers(name, false);
-        Collection<JID> administrators = getMembers(name, true);
+        List<JID> members = new ArrayList<>();
+        List<JID> administrators = new ArrayList<>();
+        loadMembersAndAdmins(name, members, administrators);
 
         return new Group(name, "", members, administrators);
```
Fishbowler, line 110: Under what circumstances would a group already have members?

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
```diff
@@ -541,7 +534,14 @@ private Collection<JID> getMembers(String groupName, boolean adminsOnly) {
...
                 }
-                members.add(userJID);
+
+                if (userJID != null) {
+                    if (isAdmin) {
+                        administrators.add(userJID);
+                    } else {
+                        members.add(userJID);
+                    }
+                }
```
Fishbowler, line 544: Should these populate temporary lists until the db iteration is complete and error free, then mutate the original?

Fishbowler: e.g.
```
List<JID> tmpMembers = new ArrayList<>();
List<JID> tmpAdmins = new ArrayList<>();
// fill tmp lists from ResultSet...
// on success:
members.addAll(tmpMembers);
administrators.addAll(tmpAdmins);
```

MilanTyagi2004:
@Fishbowler 
Apology for the delayed response.

This PR is under resolving state i am working on it.

xmppserver/src/main/java/org/jivesoftware/openfire/group/DefaultGroupProvider.java
MilanTyagi2004: Yes. the main idea was to populate the list until db iteration is complete then mutate the original.

MilanTyagi2004:
you can review  my other PRs
Thanks and DOAP file is also under resolving state.

thanks for the point issue.

PR #3385, opened 2026-06-07
OF-3305: Abort Blowfish PBKDF2 migration when security.xml is not persistable

viv:
## Problem

The Blowfish SHA1 to PBKDF2 migration (OF-3075) re-encrypts the database properties
with a PBKDF2 key derived from a freshly generated salt, and stores that salt and the
`kdf=pbkdf2` flag in `conf/security.xml`.

When `conf/security.xml` cannot be loaded or written (it is missing, empty, corrupt,
or not writable), `JiveGlobals` falls back to a non-persisting, in-memory-only
`XMLProperties` instance. In that state:

- `getBlowfishSalt()` generates a salt and calls `setProperty()`, which cannot save
  (`IllegalStateException: No file specified`). The error is logged and the boolean
  result is ignored.
- The migration re-encrypts and commits the database properties regardless.
- `setBlowfishKdf()` likewise fails to persist, and the tool reports success.

After a restart, `security.xml` is still unusable, so the server defaults to the
`sha1` KDF with no salt and can no longer decrypt the PBKDF2-encrypted values. Among
those values are the certificate-store passwords (`*.keypass` / `*.trustpass`), which
Openfire stores as encrypted database properties. With them unreadable, the keystores
fail to load (`UnrecoverableKeyException: Password verification failed`), including the
`WEBADMIN` store used by the admin console, so the admin console cannot start its HTTPS
listener. On a server that reaches the admin console over HTTPS, this locks the
administrator out (in the original report there was no admin listener on either port,
and the server logged "admin console not started due to configuration settings"). The
randomly generated salt is unrecoverable, so the re-encrypted data can be recovered
only from a pre-migration database backup.

See OF-3305, ADR-004, ADR-005.

Follow-up to #3067 (OF-3075), which added this migration. Reported on the forums: https://discourse.igniterealtime.org/t/admin-gui-not-starting-after-migration-to-pbkdf2/96500

## Fix

Fail fast, before any irreversible change, if the security configuration cannot be
persisted:

- Add `XMLProperties.isPersistable()` (true only when backed by a file).
- Add `JiveGlobals.isSecurityPropertiesPersistable()`.
- `JiveGlobals.migrateXMLPropertiesFromSHA1ToPBKDF2()` throws `IllegalStateException`
  if security properties are not persistable, before deriving any key or touching the
  database. The servlet runs this before any database work and reports that no
  database changes were made.
- `BlowfishMigrationServlet.doPost()` adds a pre-flight check (after the CSRF, backup,
  and clustering gates) that blocks migration and shows a clear admin message
  (`security.blowfish.migration.error.security-xml-not-writable`).
- Correct the load-failure log message in `JiveGlobals.loadSecurityProperties()` to
  name `security.xml` rather than `openfire.xml`.

The check runs at the very start of the migration, before any key derivation or
database write, so an unwritable `security.xml` is reported to the administrator as a
clear, actionable error and the database is never modified.

## Testing

- `XMLPropertiesTest`: `isPersistable()` returns false for a non-persisted instance and
  true for a file-backed one.
- `BlowfishMigrationServletTest`: the migration throws with a `security.xml` message
  when not persistable; `doPost()` sets the new error key and redirects without
  attempting migration. Existing migration and clustering tests pass.
- Full `xmppserver` suite green via `mvn clean verify`: 1911 tests, 0 failures, 0 errors.

### DIFF
--- a/doc/adr/004-manual-blowfish-pbkdf2-migration.md
+++ b/doc/adr/004-manual-blowfish-pbkdf2-migration.md
@@ -255,6 +255,7 @@ The migration tool uses multiple `ClusterManager` methods to detect unsafe clust
 **Migration blocked when**:
 1. `clusteringEnabled && !clusteringStarted` — clustering configured but not yet running (race condition risk: other nodes might be starting simultaneously)
 2. `clusteringStarted && nodeCount > 1` — multiple nodes active in cluster
+3. `!isSecurityPropertiesPersistable()` — `conf/security.xml` cannot be loaded or written (missing, empty, corrupt, or not writable), so the freshly generated PBKDF2 salt and the `kdf=pbkdf2` flag could not be persisted and would be lost on the next restart, leaving the re-encrypted data unreadable (OF-3305)
 
 **Migration allowed when**:
 1. `!clusteringAvailable` — Hazelcast plugin not installed, clustering impossible
--- a/i18n/src/main/resources/openfire_i18n.properties
+++ b/i18n/src/main/resources/openfire_i18n.properties
@@ -2490,6 +2490,9 @@ security.blowfish.migration.error=Migration failed
 security.blowfish.migration.error.detail=Migration failed: {0}
 security.blowfish.migration.error.backups-required=You must confirm all backups before proceeding with migration.
 security.blowfish.migration.error.csrf=CSRF validation failed. Please refresh the page and try again.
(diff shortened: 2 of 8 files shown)


guusdk:
Before we merge this - this is a good opportunity to ensure that a log entry is written to the audit log, when this migration is performed. I think that's currently missing.

Fishbowler (approved):

guusdk (approved):

PR #3394, opened 2026-06-16
OF-3314, OF-3315 & OF-3316: Minor FaviconServlet fixes

guusdk:
The FaviconServlet is used to display an icon on the admin console, when displaying server-to-server connections. The changes in this PR harden the implementation. Please refer to individual commit messages for details.

The one notable change is that this replaces the caches named "Favicon Hits" and "Favicon Misses" with one new cache: "Favicon".

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
@@ -100,7 +100,7 @@ public class FaviconServlet extends HttpServlet {
             LOGGER.warn("Unable to retrieve default favicon", e);
         }
         // Initialize cache.
-        faviconCache = CacheFactory.createCache("Favicon Cache");
+        faviconCache = CacheFactory.createCache("Favicon");
     }
 
     @Override
--- a/xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
+++ b/xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
@@ -77,14 +77,12 @@ public class FaviconServlet extends HttpServlet {
      * Pool of HTTP connections to use to get the favicons
      */
     private CloseableHttpClient client;
+
     /**
-     * Cache the domains that a favicon was not found.
+     * Cache of favicons. A present value holds the favicon bytes; an absent value
+     * records that the host was checked and has no (usable) favicon.
      */
-    private Cache<String, Integer> missesCache;
-    /**
-     * Cache the favicons that we've found.
-     */
-    private Cache<String, byte[]> hitsCache;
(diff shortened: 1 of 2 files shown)


akrherz:
@coderabbitai please review

xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
```diff
@@ -102,9 +99,8 @@ public void init(ServletConfig config) throws ServletException {
         catch (final IOException e) {
             LOGGER.warn("Unable to retrieve default favicon", e);
         }
-        // Initialize caches.
-        missesCache = CacheFactory.createCache("Favicon Misses");
-        hitsCache = CacheFactory.createCache("Favicon Hits");
+        // Initialize cache.
+        faviconCache = CacheFactory.createCache("Favicon");
```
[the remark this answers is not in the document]
guusdk: Didn't I just apply that fix?

xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
```diff
@@ -226,24 +218,53 @@ private byte[] getImage(@Nonnull final String host) {
...
+                        }
+                        // Resolve relative redirects against the current URL.
+                        // Rather than using LaxRedirectStrategy, re-validate redirect targets to prevent SSRF (OF-3315)
+                        final URI target = url.resolve(location.getValue());
+                        if (!"http".equalsIgnoreCase(target.getScheme()) && !"https".equalsIgnoreCase(target.getScheme())) {
+                            break;
+                        }
+                        final String targetHost = target.getHost();
+                        if (targetHost == null || !isConnectedHost(targetHost)) {
+                            LOGGER.info("Ignoring redirect to non-connected or unparseable host: {}", location.getValue());
```
Fishbowler, line 255: Should this check be this strict? If we attempt to get the favicon of "xmpp.example.org", we wouldn't allow a redirect to example.org?

Fishbowler (approved):
Approved, with 1 small question

xmppserver/src/main/java/org/jivesoftware/util/FaviconServlet.java
guusdk: The registrable-domain relaxation alone is probably the right call for usability, but it does widen the trust boundary from "hosts I have S2S sessions with" to "anything under a connected peer's domain." Although I can see the value, it loosens a security constraint for something that's nothing more than window dressing: it's showing an icon on the admin console. I think I prefer to err on the side of caution.

Fishbowler: Fair

PR #3410, opened 2026-06-28
Enable SASL2 tests on Openfire

Fishbowler:
Enable SASL2 Conversations test.

This is on top of #3113


dwd:
Sorta-merged into #3113

PR #3423, opened 2026-07-06
OF-3331: Guard against double counting in new connection rate throttling

guusdk:
Openfire re-fires the 'channel active' event after successful TLS negotiation. This can cause non-idempotent channel handlers to misbehave (see OF-3330).

This commit is a low-risk approach to make the NewConnectionRateLimitHandler idempotent. This is intended to be a change suitable for a patch release.


Fishbowler (approved):

Fishbowler:
Looks safe (but I've not tested it)

PR #3427, opened 2026-07-07
ci: fix Docker layer cache never being read back

guusdk:
The Docker build's cache-to scope appended `${{ github.sha }}` while cache-from did not, so every run wrote its layer cache to a unique per-commit scope that no later run ever read from.

The GHA cache was effectively write-only, and even trivially reusable layers (base images, the poms stage) rebuilt from scratch on every run. The build logs show zero CACHED steps.

Drop the `-${{ github.sha }}` suffix so cache-to and cache-from target the same scope. GHA cache is already isolated per-branch by Actions' own scoping, and mode=max handles overwrites, so the per-SHA suffix served no purpose beyond defeating the cache.


Fishbowler (approved):
LGTM, consistent with docs at https://docs.docker.com/build/cache/backends/gha/#scope. I assume the original intent was to match GHA actions/cache rules where you write specifically and read generically, but that doesn't seem like how this one works.

PR #3437, opened 2026-07-08
OF-3335: Add dependency-check suppressions

Fishbowler:
Add suppressions for two dependencies

- Suppress 2 dependencies for httpcore4. They are HTTP/2 related, and only affect httpcore5, which was a complete rewrite - httpcore4 doesn't even have any HTTP/2 capabilities.
- Suppress all jetty related matches for Mortbay Jasper JSP engine, which isn't Jetty. We do have Jetty, but this confuses the numbers. Mortbay Jasper is v9 but brings in Jetty v9 issues. Openfire has Jetty v12. This breaks the relationship between the two.

This has a nice side effect of leaving the report completely clean :)

### DIFF
--- a/.dependency-check-suppressions.xml
+++ b/.dependency-check-suppressions.xml
@@ -100,4 +100,18 @@
         <packageUrl regex="true">^pkg:maven/org\.mortbay\.jasper/apache\-jsp@.*$</packageUrl>
         <cve>CVE-2023-41080</cve>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+        False positive: this CVE is against org.apache.httpcomponents.core5:httpcore5 (HttpCore 5.x), which was a ground-up rewrite. The CPE range is "up to 5.4.2", and so dependency-check matches. CVE-2026-54428 is in the HTTP/2 HPACK decoder, which does not exist in HttpCore 4.x at all.
+   ]]></notes>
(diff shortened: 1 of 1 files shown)


guusdk:
Oh, this is useful, thanks. Particularly the jetty/mortbay false positive keeps popping up.

For the HttpCore one, if we exclude it now, but decide to upgrade our dependency in the future, will this exclusion be problematic? If so, can we somehow guard against that (maybe just leave a comment near the dependency in `pom.xml` pointing at this exclusion, instructing a developer to remove the inclusion when the dependency is updated from 4.x to 5.x?

Fishbowler:
If the dependency goes from 4 to 5, then `httpcore@4` won't match in this file and it'll be inert and cause no harm whatsoever. 

I think if we want to start caring for the dependency-check as a first-class citizen, doing things like incorporating it into CI are part of that work, along with the comments etc that will help us achieve that ongoing maintenance.

guusdk:
Ah, I missed the `@4` scoping! I agree, that should do the trick.

Didn't we _have_ this in CI at some point? We might have removed it to cut back on execution time, assuming that dependabot serves a similar purpose? If we do want to bring it back (and I'm not objecting) then it might make sense to run this periodically rather than at every build (much like what dependabot does), as new vulnerabilities could have been registered/reported outside of our code changes. It would have the added benefit of not slowing down our PR-builds.

None of this needs to block this particular PR, which I think is fine as-is?

guusdk (approved):

PR #3438, opened 2026-07-08
OF-3334: Use monotonic timers to avoid system clock dependencies

Fishbowler:
Monotonic timers are immune to changes in the system clock. This PR prevents some areas of the codebase from misbehaving when the system clock changes. 

Each altered area of the code is its own separate commit.


dwd (approved):
LGTM!

PR #3453, opened 2026-07-24
OF-3342 Presence support adds hash to existing Element text

akrherz:
backport #3448


akrherz:
Will blame the github outage for this misfire :/

PR #3461, opened 2026-08-14
Bump the Maven dependency-check plugin to 13.0.0

guusdk:


guusdk:
I've backported this manually.

PR #3474, opened 2026-08-31
ci: Conversations e2e test improvements

Fishbowler:
- Fixes a typo in the device profile
- Normalises the names of generated artifacts
- Produces Maestro log artifacts always (not just on failure)
- Add a `name` to all of the tests, for easier maintenance later
- Small resilience tweak for the Bind2 test, where the keyboard might still be up, hiding important content
- Tighten the regex on the Channel Binding test, since SCRAM-SHA-1-PLUS should never be negotiated between Conversations and Openfire.

### DIFF
--- a/.github/actions/conversationstest-action/action.yml
+++ b/.github/actions/conversationstest-action/action.yml
@@ -61,8 +61,8 @@ runs:
         path: ${{ runner.temp }}/sidecar.log
         retention-days: 7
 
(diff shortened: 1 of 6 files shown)


Fishbowler:
@coderabbitai review

Fishbowler:
@coderabbitai full review

guusdk (approved):

PR #3476, opened 2026-09-02
OF-3358: Publish the SASL mechanism registry as an immutable snapshot

guusdk:
initMechanisms assigned an empty set and then filled it, so a concurrent reader could observe a partial mechanism list. The field was not volatile either, giving readers no happens-before edge to a writer at all.

The registry is now replaced wholesale rather than mutated in place, and the three writers are synchronized. Readers are unchanged.


Fishbowler (approved):
LGTM! Nice catch!

PR #3477, opened 2026-09-02
OF-3359: Do not let unofferable FAST mechanisms shape the stream features

guusdk:
The FAST mechanisms were carried into two calculations that cannot use them: the channel-binding types, which were derived from HT-*-UNIQ/ENDP/EXPR variants even when the FAST inline feature was not being advertised, and the emptiness check that suppresses a SASL1 mechanisms element, which counted mechanisms that a SASL1 element never renders.

appendSASLFeatures now passes on only what is actually offered, and the element builder distinguishes the two profiles rather than re-deciding whether FAST is enabled.


guusdk:
Rebased

guusdk:
Rebased.

Fishbowler (approved):

PR #3490, opened 2026-09-09
OF-2534: Expose Bind 2 usage in session details admin console

guusdk:
Records, on the client session, whether resource binding for that session happened via an inline XEP-0386 Bind 2 request (as opposed to legacy IQ-based binding), and which inline feature requests (SM, CSI, Carbons) were successfully negotiated as part of it.

Surfaces this on session-details.jsp: a new "Bind Method" row, a "Bind 2 Inline Features" row listing negotiated features by their handler's display name, and inline "(negotiated via Bind 2)" annotations next to the affected Stream Management, CSI, and Message Carbons status rows.

image

### DIFF
--- a/xmppserver/src/main/java/org/jivesoftware/openfire/net/Bind2Request.java
+++ b/xmppserver/src/main/java/org/jivesoftware/openfire/net/Bind2Request.java
@@ -45,6 +45,13 @@ public class Bind2Request {
     private static final Map<String, Bind2InlineHandler> elementHandlers = new ConcurrentHashMap<>();
 
     /**
+     * The namespaces of inline feature requests that were successfully processed
+     * during {@link #processFeatureRequests}, for observability purposes. Empty until that
+     * method has run.
+     */
+    private Set<String> negotiatedFeatureNamespaces = Collections.emptySet();
+
+    /**
      * Registers a handler for processing inline elements with a specific namespace.
      *
      * Only one handler can be registered for each namespace. Attempting to register a handler for an already registered
@@ -103,6 +110,7 @@ public class Bind2Request {
      */
     public Element processFeatureRequests(LocalClientSession clientSession, Element successElement) {
         Element bound = successElement.addElement(new QName("bound", new Namespace("", NAMESPACE)));
+        final Set<String> succeeded = new LinkedHashSet<>();
 
         for (Element element : featureRequests) {
             String namespace = element.getNamespaceURI();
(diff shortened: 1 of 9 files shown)


xmppserver/src/main/java/org/jivesoftware/openfire/net/Bind2Request.java
```diff
@@ -103,27 +110,32 @@ public static boolean unregisterElementHandler(@Nonnull final Bind2InlineHandler
...
             String namespace = element.getNamespaceURI();
             Bind2InlineHandler handler = elementHandlers.get(namespace);
 
             if (handler != null && handler.isEnabled()) {
                 try {
-                    if (!handler.handleElement(clientSession, bound, element)) {
-                        Log.info("Handler for namespace {} failed to process element", namespace);
+                    if (handler.handleElement(clientSession, bound, element)) {
+                        Log.trace("Handler for namespace {} successfully processed element for session: {}", namespace, clientSession);
+                        succeeded.add(namespace);
```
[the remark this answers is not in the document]
guusdk: You're right. Both handlers were falling through to return true regardless of element name. Fixed by rejecting unrecognized elements in each, matching the pattern already in Bind2StreamManagementHandler. Turned out Carbons had a small related behavior bug too: any non-"enable" name was being treated as "disable" rather than ignored, so that's tightened up as well. Pushed as a follow-up commit.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
```diff
@@ -237,6 +237,20 @@ public class SASLAuthentication {
      */
     private static final String SASL2_RESUME_REQUEST = "Sasl2.resume-request";
 
+    /**
+     * Session Data property name used to record that a session's resource was bound via an inline XEP-0386 Bind 2
+     * request during SASL2 authentication, as opposed to legacy IQ-based binding. Set only once binding has actually
+     * completed successfully; never set for sessions that bind via legacy IQ, and never set when an inline XEP-0198
+     * resume made the inlined bind2 request moot.
+     */
+    public static final String BIND2_USED = "Bind2.used";
```
Fishbowler, line 246: There's related stuff in #3496 where I've added a `SASL2_BIND2_PENDING_OR_SUCCEEDED`. I wonder if we need to dedupe?

Fishbowler (approved):
LGTM. One awkwardness that should be addressed after this lands.

xmppserver/src/main/java/org/jivesoftware/openfire/net/SASLAuthentication.java
guusdk: Possibly. Lets merge this, and I'll rebase the other PR on top.

PR #3496, opened 2026-09-13
OF-3361 OF-3362: Handle bad clients during SASL2 authentication

Fishbowler:
Whilst adding tests to SINT Server Extensions for SASL2, I found 2 issues

OF-3361 - you can't send weird things mid authentication
OF-3362 - you can't send another authenticate after success

This PR fixes both.


xmppserver/src/test/java/org/jivesoftware/openfire/net/StanzaHandlerTest.java
```diff
@@ -214,4 +217,102 @@ public void startedSASL_shouldBeResetAfterSasl2Bind2Authentication() throws Exce
...
+     * Verifies that traffic other than {@code <response/>} or {@code <abort/>} received while a SASL2 negotiation
+     * is in progress causes the connection to be disconnected with a stream error, per XEP-0388 § 2.4 ("Servers
+     * MUST disconnect Clients immediately if any other traffic is received"). Regression test for OF-3361.
+     */
+    @Test
+    public void strayStanzaDuringSasl2Negotiation_shouldDisconnect() throws Exception
+    {
+        // Setup test fixture.
+        final Connection connection = mock(Connection.class);
+        when(connection.isEncrypted()).thenReturn(false);
```
guusdk: I'm not sure if this is relevant, but explicitly mocking `isEncrypted` to `false` surprises me. Is this significant to the test? I'd assume that a 'regular' connection would be encrypted.

Fishbowler: That's just the agent being consistent with the other tests in the class. It's inert. The mock already returns false.

guusdk: Lets remove that to reduce ambiguity.

guusdk:
This feels like it's growing large. Especially the edits to the parser / classes that we generally never, ever touch, make me uncomfortable, being part of a largeish PR. Can we separate out fixes in distinct PRs?

xmppserver/src/main/java/org/jivesoftware/openfire/nio/XMLLightweightParser.java
```diff
@@ -361,6 +382,10 @@ else if (ch == '/' && !head.isEmpty()) {
                     depth = 1;
                 }
                 else {
+                    // A character received at the top level, between stanzas (not inside one - this state is
+                    // entered fresh after every completed stanza, see foundMsg()). Typically a whitespace
+                    // keep-alive ping (RFC 6120 § 4.6.1), but flagged regardless of content.
+                    nonStanzaDataReceived = true;
```
guusdk: I wonder if this is the only place where this needs to be captured. Isn't this equally applicable to the `status == OUTSIDE` block?

I'm not very keen on the approach where we introduce a polling mechanism in this class. I'm not sure if we can improve on it, but this feels forced, an not in line with the rest of the class' approach. Perhaps we should _emit_ an event when we receive this data, rather than record it for something to eventually poll for it?

Fishbowler:
I've dropped OF-3370 out of this PR. That fix was not like the others. I'll PR that separately.

guusdk (approved):

guusdk:
I have rebased this, and added a small refactoring (name change and documentation update) to remove any ambiguity between a constant that was introduced in parallel on Main, and one introduced by this PR.

PR #3498, opened 2026-09-13
OF-3368: Return credentials-expired for invalidated FAST tokens

guusdk:
Invalidating a FAST token (XEP-0484) hard-deleted its row, so a later authentication attempt against that token was indistinguishable from one that had never been issued. The server therefore returned the SASL 'not-authorized' condition instead of 'credentials-expired', in violation of §4.2, which requires 'credentials-expired' for any token the server no longer trusts so the client knows to fall back to another mechanism rather than treating it as a hard authentication failure.

Tokens the server no longer trusts are now expired in place instead of deleted:

- invalidateToken() backdates the token's expiry rather than deleting the row, so validateTokenHt2() still finds it, recognises it as expired, and reports 'credentials-expired' as required.
- The token superseded during 'new' slot promotion is expired the same way instead of being deleted outright, since a client that replays it is in the same "no longer trusted" situation.
- purgeExpiredTokens() now retains rows for a configurable grace period after expiry (xmpp.fast.token.purge-grace-period) rather than deleting them as soon as they are expired, so the credentials-expired window does not depend on how soon the periodic cleanup task happens to run. The invalidation backdate is accounted for in the purge cutoff so the two settings don't interact.
- Backdating uses a fixed offset rather than the exact current instant, so a token still reads as expired despite modest clock skew between cluster nodes.

Account-level invalidation (on user deletion) and clearing an unacknowledged 'new' token before issuing a replacement are left as hard deletes, since neither has a corresponding "tell the client to fall back" scenario to support.

I have dropped one unit test that I simply could not wrap my head around.


i18n/src/main/resources/openfire_i18n.properties
```diff
@@ -1302,6 +1302,7 @@ system_property.xmpp.auth.ssl.enforce_sni=Controls if the server enforces the us
 system_property.xmpp.fast.enabled=Enables FAST (XEP-0484), which lets clients re-authenticate with a long-lived token instead of their password.
 system_property.xmpp.fast.token.expiry=How long an issued FAST token remains valid. This is the window in which a leaked token, or a token belonging to a password changed outside Openfire (as with LDAP), can still be used to authenticate - lower values shorten that exposure at the cost of more frequent full re-authentication.
 system_property.xmpp.fast.token.rotation-threshold=How close to expiry a token must be before a successful authentication issues a replacement. It must be comfortably longer than the longest interval at which a client is expected to reconnect, or clients will find their token expired and have to re-authenticate with a password.
+system_property.xmpp.fast.token.purge-grace-period=The duration for which a purged FAST token remains in the database before being permanently deleted. This allows the server to distinguish between expired and invalid tokens.
```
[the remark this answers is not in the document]
guusdk: Fair. I amended this change to the original commit, and force-pushed.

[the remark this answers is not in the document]
guusdk: Fair. I've amended again.

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
```diff
@@ -157,6 +157,33 @@ public class FastTokenManager {
...
+    /**
+     * How long an invalidated or naturally-expired token's row is retained past its expiry before being purged. While
+     * the row exists, a client redeeming that token is refused with SASL 'credentials-expired' (XEP-0484 §4.2), telling
+     * it to fall back to another mechanism. Once purged, the same attempt is indistinguishable from a token that was
+     * never issued and gets the generic 'not-authorized' instead. This is a minimum retention guarantee, independent of
+     * how often the purge task itself runs ({@link FastTokenLifecycle#CLEANUP_INTERVAL}).
+     */
+    public static final SystemProperty<Duration> TOKEN_PURGE_GRACE_PERIOD = SystemProperty.Builder.ofType(Duration.class)
+        .setKey("xmpp.fast.token.purge-grace-period")
+        .setDefaultValue(Duration.ofHours(1))
```
Fishbowler: I think the use case should cover someone being away for a long weekend and getting the expected message back on their return to work. WDYT?

```suggestion
    public static final SystemProperty<Duration> TOKEN_PURGE_GRACE_PERIOD = SystemProperty.Builder.ofType(Duration.class)
        .setKey("xmpp.fast.token.purge-grace-period")
        .setDefaultValue(Duration.ofDays(4))
```

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
```diff
@@ -157,6 +157,33 @@ public class FastTokenManager {
...
+        .setMinValue(Duration.ZERO)
+        .build();
+
+    /**
+     * How far in the past to backdate a token's expiry when it is explicitly invalidated, rather than left to expire
+     * naturally. A backdate (instead of {@code Instant.now()} exactly) guarantees the token reads as already-expired
+     * even across a modest clock skew between the node that invalidates it and a different node (in a clustered
+     * deployment) that later validates a redemption attempt against it.
+     */
+    private static final Duration INVALIDATION_BACKDATE = Duration.ofMinutes(5);
```
Fishbowler, line 181: Why not kick this back further to make it easier to identify tokens at a point in time? I know it's only a heuristic, but it costs nothing.

```suggestion
    private static final Duration INVALIDATION_BACKDATE = Duration.ofYears(1);
```

Fishbowler (changes requested):
This passes the test that previous failed, but fails a test that previous passed.

> Upon successful use of any token, the server MUST invalidate all tokens issued to the same client with an earlier expiry than the current token (even if those tokens have not yet reached their expiry time)

The test:
- gets a FAST token by password auth (getting token T1 in the 'new' slot)
- uses the FAST token and requests a new one (promoting T1 to 'current' and getting T2 in 'new'
- uses the second token (promoting T2 to 'current', which should invalidate T1)
- attempts to use the first token again, expecting failure

Openfire is reporting SQL problems:
```
2026.09.14 10:13:44.866 ERROR [socket_c2s-handler-11]: org.jivesoftware.openfire.fast.FastTokenManager - Failed to fetch HT2 FAST token for user 'sinttest-qke2k-9' mechanism 'HT-SHA-256-NONE'
java.sql.SQLIntegrityConstraintViolationException: integrity constraint violation: unique constraint or index violation ; OFFASTTOKEN_PK table: OFFASTTOKEN
    at org.hsqldb.jdbc.JDBCUtil.sqlException(Unknown Source)
    at org.hsqldb.jdbc.JDBCUtil.sqlException(Unknown Source)
    at org.hsqldb.jdbc.JDBCPreparedStatement.fetchResult(Unknown Source)
    at org.hsqldb.jdbc.JDBCPreparedStatement.executeUpdate(Unknown Source)
    at org.apache.commons.dbcp2.DelegatingPreparedStatement.executeUpdate(DelegatingPreparedStatement.java:137)
    at org.apache.commons.dbcp2.DelegatingPreparedStatement.executeUpdate(DelegatingPreparedStatement.java:137)
    at org.jivesoftware.openfire.fast.FastTokenManager.validateTokenHt2(FastTokenManager.java:453)
    at org.jivesoftware.openfire.sasl.AbstractHtSaslServer.lambda$new$0(AbstractHtSaslServer.java:107)
    at org.jivesoftware.openfire.sasl.HtSaslServer.doEvaluateResponse(HtSaslServer.java:106)
    at org.jivesoftware.openfire.sasl.AbstractHtSaslServer.evaluateResponse(AbstractHtSaslServer.java:155)
    at org.jivesoftware.openfire.net.SASLAuthentication.handle(SASLAuthentication.java:646)
    at org.jivesoftware.openfire.net.StanzaHandler.processStanza(StanzaHandler.java:267)
    at org.jivesoftware.openfire.net.StanzaHandler.process(StanzaHandler.java:150)
    at org.jivesoftware.openfire.nio.NettyConnectionHandler.channelRead0(NettyConnectionHandler.java:179)
    at org.jivesoftware.openfire.nio.NettyConnectionHandler.channelRead0(NettyConnectionHandler.java:51)
    at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
    at io.netty.channel.AbstractChannelHandlerContext.lambda$fireChannelRead$2(AbstractChannelHandlerContext.java:366)
    at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:148)
    at io.netty.util.concurrent.DefaultEventExecutor.run(DefaultEventExecutor.java:66)
    at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:1204)
    at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
    at java.base/java.lang.Thread.run(Unknown Source)
Caused by: org.hsqldb.HsqlException: integrity constraint violation: unique constraint or index violation ; OFFASTTOKEN_PK table: OFFASTTOKEN
    at org.hsqldb.error.Error.error(Unknown Source)
    at org.hsqldb.Constraint.getException(Unknown Source)
    at org.hsqldb.index.IndexAVL.insert(Unknown Source)
    at org.hsqldb.persist.RowStoreAVL.indexRow(Unknown Source)
    at org.hsqldb.TransactionManager2PL.addInsertAction(Unknown Source)
    at org.hsqldb.Session.addInsertAction(Unknown Source)
    at org.hsqldb.Table.insertSingleRow(Unknown Source)
    at org.hsqldb.StatementDML.update(Unknown Source)
    at org.hsqldb.StatementDML.executeUpdateStatement(Unknown Source)
    at org.hsqldb.StatementDML.getResult(Unknown Source)
    at org.hsqldb.StatementDMQL.execute(Unknown Source)
    at org.hsqldb.Session.executeCompiledStatement(Unknown Source)
    at org.hsqldb.Session.execute(Unknown Source)
    ... 21 more 
```

The promotion of T2 to the new current failed. T1 needs to be expired, so that it can fail in the test (with `credentials-expired`)

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
guusdk: I have no strong opinion on this. This trades seeing the 'correct' error message (credentials-expired vs not-authorized) against the potential size of the database.

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
guusdk: I disagree, I think there's value in keeping it as-is.

There's a clean-up that runs every hour. It deletes from the database every token that has an expiry date that is further than TOKEN_PURGE_GRACE_PERIOD in the past.

The INVALIDATION_BACKDATE is used to set the expiry date of a token. It is deliberately set to something that is barely in the past. This ensures that it does not get immediately deleted from the database, if the clean-up cycle happens to kick in only moments after the token expiry data was backdated.

In other words, I believe that INVALIDATION_BACKDATE must not cause the expiry-date of an invalidated token to be backdated more than TOKEN_PURGE_GRACE_PERIOD, because that runs the risk of the (expired, back-dated) token to be deleted from the database whenever the clean-up run executes next.

I'm using 5 minutes rather than one millisecond to allow for some clock skew (for example in a cluster), but otherwise, I think it's important for the backdating to happen as close to 'now' (the time of invalidation) as possible.

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
Fishbowler: Yes, that's exactly the trade I'd like to make. The number of tokens accrued in 4 days is likely very low. For abnormal scenarios, folks can tweak the parameter.

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
Fishbowler: Okay. We've got a debug log line for it too.

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
guusdk: Done!

xmppserver/src/main/java/org/jivesoftware/openfire/fast/FastTokenManager.java
```diff
@@ -411,11 +441,27 @@ public static Ht2ValidationResult validateTokenHt2(@Nonnull final String usernam
...
+                // 'C' slot entirely, not just have its expiry updated in place - otherwise promoting this 'N' row into
+                // 'C' collides with the still-'C'-labelled retired row. Only one retired row is kept per client; an
+                // older one still waiting on purgeExpiredTokens() is superseded rather than accumulated, since a second
+                // rotation within one grace period is a rare edge case and not one XEP-0484 requires unbounded history
+                // for.
+                pstmt = con.prepareStatement(DELETE_RETIRED_TOKEN);
                 pstmt.setString(1, username);
                 pstmt.setString(2, mechanism);
                 pstmt.setString(3, matchedClientId);
                 pstmt.executeUpdate();
```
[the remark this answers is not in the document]
guusdk: I think that this is the point of the change? This deletes _previous_ retired tokens, and the next query (a few lines down) updates the _current_ token to be retired. Due to the database primary key definition, we can't have more than one token in 'retired' state, which makes this a bit of a hack.

xmppserver/src/test/java/org/jivesoftware/openfire/fast/FastTokenPersistenceTest.java
```diff
@@ -542,39 +635,47 @@ void invalidatingAUserRemovesEveryTokenTheyHold() throws Exception
...
+
+            final String justWithinGracePeriod = StringUtils.zeroPadString(
+                String.valueOf(millis + 1), 15);
+            final String justBeyondGracePeriod = StringUtils.zeroPadString(
+                String.valueOf(millis - 1), 15);
+
+            assertTrue(justWithinGracePeriod.compareTo(cutoff.getValue()) > 0,
+                "A token just inside the retention window would be deleted too early (OF-3368).");
+            assertTrue(justBeyondGracePeriod.compareTo(cutoff.getValue()) < 0,
+                "A token just outside the retention window would survive indefinitely.");
```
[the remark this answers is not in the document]
guusdk: That'd be better, but makes the test implementation depend on SystemProperty, which adds undesired complexity.

Fishbowler (approved):
LGTM. Tested with a bunch of new SINT Server Extensions tests. This feels solid.

PR #3499, opened 2026-09-14
OF-3370: Handle non-stanza content during SASL2 authentication

Fishbowler:
XEP-0388 §2.4, "During Authentication":

Clients MUST NOT send whitespace [...] Servers MUST disconnect Clients immediately if any other traffic is received.


Fishbowler:
Superseded by #3502

PR #3500, opened 2026-09-14
OF-3371: Disable test whilst issue remains open

Fishbowler:
Right now, offline messages aren't removed. But we also don't have a MAM implementation for this. XEP-0386 removes the offline messages _because_ they'll come from MAM. So keeping offline messages makes sense for now.

This proactively disables the test that's not yet in the main branch of SINT Server Extensions.


guusdk (approved):
Looks good to me. Let's not forget that this _does not_ close OF-3371 upon merge.

PR #3503, opened 2026-09-15
Partially revert "OF-3313: Return previous value from SystemProperty#setValue"

guusdk:
The original change altered the signature of the existing public method SystemProperty#setValue from `void setValue(T)` to `T setValue(T)`, so that callers could see the previous value.

Although source-compatible, this broke binary compatibility: plugin jars compiled against the old `void setValue(T)` descriptor (setValue(Ljava/lang/Object;)V) fail at runtime with a NoSuchMethodError against the new jar, since the descriptor no longer exists, e.g.:

  java.lang.NoSuchMethodError: 'void org.jivesoftware.util.SystemProperty.setValue(java.lang.Object)'
      at org.jivesoftware.openfire.plugin.InterpretedXMLPrinter.setEnabled(...)

This would have forced every plugin calling setValue to be recompiled and republished against the new Openfire release.

This commit reverts setValue to its original `void` signature, restoring binary compatibility for existing plugins, and reintroduces the "return the previous value" behaviour under a new method, getAndSetValue(T), so callers who need the old value can opt in explicitly without breaking anyone else.

Tests updated accordingly: the setValue-return-value assertions from OF-3313 now target getAndSetValue instead.


Fishbowler (approved):
LGTM. I'd also say that this is more of a "different approach" than "partially revert" :)

PR #3507, opened 2026-09-17
Fix Statement leak in DbConnectionManager.testConnection

guusdk:
The Statement was only closed on the success path; if executeQuery threw (e.g. schema not yet installed), it was never closed. Use try-with-resources instead.


guusdk:
This is a result of me experimenting with Claude Code. The fix in itself is pretty minor: I fully expect there be no practical impact, as database drivers can be expected to close resources when the database connection is closed. This is a hygiene fix.

akrherz (approved):

Fishbowler (approved):

Fishbowler:
This is likely a no-op hygiene fix, so we'll forego the Jira ticket.

PR #3510, opened 2026-09-19
OF-3374: Preserve the key derivation method for user credentials when using the default auth provider

Fishbowler:
The PBKDF2 migration left Openfire with no way to know that the passwords were encrypted with SHA1. Instead, PBKDF2 was used, all comparisons would fail, and all users were locked out. Affected anyone using default credential storage.

The migration now records the KDF used for passwords in security.xml (`encrypt.blowfish.user-kdf`), and AuthFactory uses it, falling back to `encrypt.blowfish.kdf` when it's not set. That means that pre-migration, post-migration and new installations that don't need migration all get the correct KDF for authentication.

- Implementation
  - JiveGlobals: new encrypt.blowfish.user-kdf security property. getBlowfishKdfForPasswords() reads it and falls back to encrypt.blowfish.kdf.
  - BlowfishMigrationServlet: records user-kdf=sha1 before switching kdf to PBKDF2.
  - AuthFactory: the password cipher uses getBlowfishKdfForPasswords() instead of the shared KDF.
- Testing
  - Unit tests for the fallback and for passwords staying on SHA1 across the migration.
  - Upgrade tests in Docker:
    - 5.0.5 → 5.1.2 (or the tip of the 5.1 branch) then migrate and restart: the lockout reproduces.
    - Same, with this change applied to the tip of 5.1: logins keep working.
    - With this change and no migration (a fresh 5.1 install): password changes or new users all keep working too.


guusdk (changes requested):
Thanks for digging into this. I think this approach would work, but has significant downsides:

1. **The name won't survive the next KDF change.** `preserveLegacyPasswordKdf` and `encrypt.blowfish.user-kdf` encode "which of two historical KDFs". A future third generation needs either a new value with new comparison logic everywhere, or another flag. We would be carrying this along forever.
2. **It pins the data rather than migrating it.** `AuthFactory.encryptPassword()` uses the same cipher, so every password set *after* the migration is also written SHA1-derived, and nothing ever clears the flag. Possibly the most sensitive dataset on the server stays permanently on the KDF this migration was created for to retire. Granted, it didn't retire it at all initially, but still.
3. **It doesn't repair instances that already migrated.** Although it's likely that those instances roll back (as users may no longer be able to log in) not everyone will be able to roll back 

I'd like us to reconsider the approach before merging. I'll suggest an approach in a new comment on this PR.

guusdk:
As suggested in the review feedback: I'd like us to consider another approach. Instead of flagging this data as 'not migrated', we should migrate this data. If the migration of data was considered 'worth it' for properties, it should _certainly_ be worth it for user credentials.

One complication is installations that already ran the migration: On those, passwords set _after_ the migration may have been encrypted with PBKDF2, while older ones still use SHA1. There's a real chance of mixed data, so we can't decrypt all rows with a single KDF. We need to be able to tell the two apart.

I believe that we can use the SCRAM credentials to do that. Since (I believe - please verify) Openfire 4.0.0 (also known as 'a long long time ago'), every time a password is set, SCRAM credentials are stored for the user as well (originally as columns in `ofUser`, now in `ofUserScram`). That data is derived from the plaintext password, independently of Blowfish, and the migration never touches it.

So for each `encryptedPassword` of which we are unsure which KDF was used, we can:
- decrypt it with SHA1,
- check whether the result matches the user's stored SCRAM credential,
- re-encrypt it with PBKDF2 if it does.

Every other row can be left alone.

The one gap is users who have no SCRAM credential: their rows can't be verified. I believe the best we can do is to report them to the administrator.

(When the property migration has _not_ executed yet, we could in theory short-circuit this by treating passwords just like the properties being migrated. That would give us two distinct migration paths that could diverge, so I'd prefer a single implementation, even if it's slower. One thing that we're missing out on is that before the migration, every row is known to be SHA1-encrypted, so rows without a SCRAM credential can still be re-encrypted rather than reported. In practice, I don't think this happens much)

